Issue #776, opened 2022-11-23
Add metric for memory footprint of Transactions


bfish713:
Should be a gauge which takes the sums the serialized size of the transactions still pending.

PR #794, opened 2022-12-01
Add Metric which Keeps Track of Transactions' Memory

bfish713:
#closes: https://github.com/EspressoSystems/HotShot/issues/776

### DIFF
--- a/consensus/src/lib.rs
+++ b/consensus/src/lib.rs
@@ -117,7 +117,7 @@ pub struct ConsensusMetrics {
     pub rejected_transactions: Box<dyn Counter>,
     /// Number of outstanding transactions
     pub outstanding_transactions: Box<dyn Gauge>,
-    /// Memory size of the serialized transactions still outstanding
+    /// Memory size in bytes of the serialized transactions still outstanding
     pub outstanding_transactions_memory_size: Box<dyn Gauge>,
     /// Number of views that timed out
     pub number_of_timeouts: Box<dyn Counter>,
--- a/consensus/src/lib.rs
+++ b/consensus/src/lib.rs
@@ -86,6 +86,9 @@ pub struct Consensus<TYPES: NodeTypes> {
     /// Amount of invalid QCs we've seen since the last commit
     /// Used for metrics.  This resets to 0 on every decide event.
     pub invalid_qc: usize,
+
+    /// Used for metrics, the size of the pending transactions in memory.
+    pub pending_transaction_mem: usize,
 }
 
 /// The metrics being collected for the consensus algorithm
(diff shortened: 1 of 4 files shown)


consensus/src/lib.rs
```diff
@@ -86,6 +86,9 @@ pub struct Consensus<TYPES: NodeTypes> {
     /// Amount of invalid QCs we've seen since the last commit
     /// Used for metrics.  This resets to 0 on every decide event.
     pub invalid_qc: usize,
+
+    /// Used for metrics, the size of the pending transactions in memory.
```
DieracDelta, line 90: Might be worth mentioning here that the size is measured in bytes

DieracDelta (approved):

consensus/src/replica.rs
```diff
@@ -373,6 +375,15 @@ impl<A: ConsensusApi<TYPES>, TYPES: NodeTypes> Replica<A, TYPES> {
                 .metrics
                 .invalid_qc_views
                 .add_point(consensus.invalid_qc as f64);
+            #[allow(clippy::cast_possible_truncation)]
+            let included_txn_size = included_txns_set.iter().fold(0, |total, txn| {
+                total + bincode_opts().serialized_size(&txn).unwrap_or(0) as usize
+            });
+            consensus.pending_transaction_mem -= included_txn_size;
```
jbearer, line 382: Should we add something like `update(delta: i64)` to the `Gauge` trait so we don't have to store the current value of the gauge separately for metrics like this?

bfish713: Yeah i've been thinking of this, I'm going to add it.  I thought it had to do with the Prometheus gauge type not having any kind of update but I checked and it expose, get, add and subtract.

consensus/src/replica.rs
```diff
@@ -375,15 +375,14 @@ impl<A: ConsensusApi<TYPES>, TYPES: NodeTypes> Replica<A, TYPES> {
...
-                total + bincode_opts().serialized_size(&txn).unwrap_or(0) as usize
-            });
-            consensus.pending_transaction_mem -= included_txn_size;
+                total + bincode_opts().serialized_size(&txn).unwrap_or(0)
+            }) as i64;
             consensus
                 .metrics
                 .outstanding_transactions_memory_size
-                .set(consensus.pending_transaction_mem);
+                .update(included_txn_size);
```
jbearer, line 385: Should this be `-included_txn_size`?

bfish713: 🤦Yes good catch

jbearer (approved):
LGTM

PR #796, opened 2022-12-01
Include chain height in Leaf.

### LINKED ISSUE #787: Add `height` to leaf
I previously suggested that a block height field be removed from the [block/leaf structure](https://www.notion.so/espressosys/HotShot-block-structure-604788fca01b4486b2370f5494a89b90) because I couldn't think of a use case for it. I just discovered a use case for it:

The query service uses the block height of new leaves to check if any leaves have been skipped (due to catchup, e.g.). If so it inserts placeholders for the missing leaves and starts trying to fetch the missing data asynchronously from elsewhere. The Espresso query service gets the block height from the CAP ValidatorState, but that's application-specific -- we should add a height field to Leaf so the generic query service can do it in an application-agnostic way

jbearer:
Closes #787

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2323,6 +2323,7 @@ version = "0.1.0"
 dependencies = [
  "ark-bls12-381",
  "async-compatibility-layer",
+ "async-lock",
  "async-std",
  "async-trait",
  "blake3",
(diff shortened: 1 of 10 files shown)


DieracDelta (approved):
LGTM! Nice usage of `u64_field` and ensure. Didn't realize those existed!

PR #797, opened 2022-12-02
Include QC for latest leaf in decide event

### LINKED ISSUE #795: Add QC for last leaf to Decide event
Currently, the Decide event only includes `leaf_chain: Vec<Leaf>`, and each Leaf includes the QC that signed the leaf _before_ it (`justify_qc`) in order to link the leaves into a chain. This means the Decide event does not include a QC with signatures for the most recent leaf in the chain, which means
* we cannot authenticate the last leaf
* we cannot get the block hash the leaf is committing to (qc.block_commitment) (technically we can use `leaf.deltas.commit()`, but this is not compatible with DA)

jbearer:
Closes #795

### DIFF
--- a/consensus/src/replica.rs
+++ b/consensus/src/replica.rs
@@ -281,6 +281,7 @@ impl<A: ConsensusApi<TYPES>, TYPES: NodeTypes> Replica<A, TYPES> {
         let mut last_view_number_visited = self.cur_view;
         let mut new_commit_reached: bool = false;
         let mut new_decide_reached = false;
+        let mut new_decide_qc = None;
         let mut leaf_views = Vec::new();
         let mut included_txns = HashSet::new();
         let old_anchor_view = consensus.last_decided_view;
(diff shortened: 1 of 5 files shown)


consensus/src/replica.rs
```diff
@@ -300,6 +301,9 @@ impl<A: ConsensusApi<TYPES>, TYPES: NodeTypes> Replica<A, TYPES> {
                             if current_chain_length == 2 {
                                 new_locked_view = leaf.view_number;
                                 new_commit_reached = true;
+                                // The next leaf in the chain, if there is one, is decided, so this
+                                // leaf's justify_qc would become the QC for the decided chain.
+                                new_decide_qc = Some(leaf.justify_qc.clone());
```
elliedavidson, line 306: This looks right.

consensus/src/replica.rs
```diff
@@ -281,6 +281,7 @@ impl<A: ConsensusApi<TYPES>, TYPES: NodeTypes> Replica<A, TYPES> {
         let mut last_view_number_visited = self.cur_view;
         let mut new_commit_reached: bool = false;
         let mut new_decide_reached = false;
+        let mut new_decide_qc = None;
```
elliedavidson, line 284: We could _technically_ initialize this to the genesis QC because that is the last decided view if there are no other decided views, but I don't think it is necessarily incorrect to do it this way either.  Just mentioning it.

testing/tests/consensus.rs
```diff
@@ -705,3 +710,114 @@ async fn test_chain_height() {
 
     test.execute().await.unwrap();
 }
+
```
elliedavidson, line 713: Good tests!

elliedavidson (approved):
LGTM.

PR #799, opened 2022-12-02
Bump jf + libp2p

### LINKED ISSUE #781: Update dependencies
Libp2p and dependabot deps should be updated and a CVE should be corrected.

DieracDelta:
Closes #781.

Upgrades:
- zeroize to ^1.5
- jellyfish to the 0.1.2-patch.1 tag
- libp2p to 0.50

The subtleties here are all related to libp2p. This was a large version bump.

- A bunch of functions are deprecated
- Some functions/modules were renamed/moved around
- Overall behaviour event types are different and DHT get record requests have been changed. Now there is a new event for each new record found instead of recv-ing all records at once in a single event. As a result, the records need to be tracked in the dht behaviour's wrapper struct.


libp2p-networking/src/network/behaviours/dht.rs
```diff
@@ -184,11 +184,11 @@ impl DHTBehaviour {
                 // failed try again later
                 query.progress = DHTProgress::NotStarted;
                 query.backoff.start_next(false);
-                warn!("Error publishing to DHT: {e:?} for peer {:?}", self.peer_id);
+                error!("Error publishing to DHT: {e:?} for peer {:?}", self.peer_id);
                 self.queued_put_record_queries.push_back(query);
             }
             Ok(qid) => {
-                info!("Success publishing {:?} to DHT", qid);
+                error!("Success publishing {:?} to DHT", qid);
```
nyospe, line 191: ????

nyospe (dismissed):
That's a lot of event model changes.

Is that one info! => error! intentional, and meant to be kept in?

libp2p-networking/src/network/behaviours/dht.rs
DieracDelta: asdf that is a typo

shenkeyao (approved):

PR #815, opened 2022-12-12
Nvim nix support

DieracDelta:
It doesn't look like there's an environment variable that can be fed into rust-analyzer to set feature flags. So instead, I set an env variable when entering the nix shell that I've modified my neovim config to feed through to the rust-analzyer setup on start.

Relevant code to stick in the `init.lua` is:

```lua
local rust_tools = require('rust-tools')
rust_tools.setup({server = {settings = {["rust-analyzer"] = {cargo = {features = os.getenv("NVIM_RUST_ENABLED_FEATURE_FLAGS") and { os.getenv("NVIM_RUST_ENABLED_FEATURE_FLAGS") } or "all"}}}}})
```

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -57,6 +57,7 @@
           ];
 
         CARGO_TARGET_DIR = "target_dirs/nix_rustc";
+        NVIM_RUST_ENABLED_FEATURE_FLAGS = "full-ci";
 
         rustOverlay = final: prev: {
           rustc = fenixStable;
@@ -186,7 +187,7 @@
           ];
       in {
         devShell = pkgs.mkShell {
-          inherit CARGO_TARGET_DIR;
+          inherit CARGO_TARGET_DIR NVIM_RUST_ENABLED_FEATURE_FLAGS;
           buildInputs = [ fenixStable ] ++ buildDeps;
         };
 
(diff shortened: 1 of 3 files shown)


flake.nix
```diff
@@ -57,6 +57,7 @@
           ];
 
         CARGO_TARGET_DIR = "target_dirs/nix_rustc";
+        NVIM_RUST_ENABLED_FEATURE_FLAGS = "full-ci";
```
nyospe, line 60: Do we actually want to push this globally? I would expect to have it set per-user and just inherit it during the nix initialization, but maybe I'm misunderstanding an aspect of the way the `flake.nix` file is used.

DieracDelta: Summarizing from meeting:
- example config for neovim in `.nvim`
- source `.nvim/config.sh` in a shellhook for nix shell
- readme for user about this

DieracDelta: Conceptually this can't work from the `flake.nix`, because nix shells only know about staged or committed files.

I'll need to get nvim to search from the root directory of the project for a `.nvim/env.sh`.

nyospe: 🙀

DieracDelta: Should be good to go 👍

nyospe (approved):

Issue #706, opened 2022-11-08
Define data availability receipt


jbearer:
Must include at least
* some kind of commitment to the data
* signature of the committee member
* proof of committee eligibility (associated type of the specific DA implementation)

Issue #705, opened 2022-11-08
Modify block interface to support DA


jbearer:
- Type for full block with data
- Type for block commitment
- Interface to add transactions to the full block
- Interface to get a commitment from full block
- Change all replica logic to work with the commitment, not the full block

PR #818, opened 2022-12-15
Add interfaces for DA

shenkeyao:
- Adds leaf, proposal and certificate types for validating and DA.
- Refactors existing code to be generic over new types.
- Fix tests.

Closes https://github.com/EspressoSystems/HotShot/issues/792, closes https://github.com/EspressoSystems/HotShot/issues/705, closes https://github.com/EspressoSystems/HotShot/issues/706, closes https://github.com/EspressoSystems/HotShot/issues/709, closes https://github.com/EspressoSystems/HotShot/issues/711.


types/src/message.rs
```diff
@@ -17,62 +18,89 @@ use serde::{Deserialize, Serialize};
...
+    TYPES: NodeType,
+    LEAF: LeafType<NodeType = TYPES>,
+    PROPOSAL: ProposalType<NodeType = TYPES>,
+> {
     /// The sender of this message
     pub sender: TYPES::SignatureKey,
 
     /// The message kind
-    pub kind: MessageKind<TYPES>,
+    pub kind: MessageKind<TYPES, LEAF, PROPOSAL>,
```
nyospe, line 30: Just a note to remember that this, in particular, is going to need to be refactored back out in the work we do next year. We want to have the MessageKind more customized to the consensus layer, which probably means taking the specific messages out of the struct, and entirely redoing this enum.

nyospe (dismissed):
It looks like this fully captures the intended pre-holiday plan for updates for DA. Nicely done.

types/src/message.rs
shenkeyao: Sounds good! I just added a TODO comment to the enum.

nyospe (approved):

nyospe (approved):

Issue #837, opened 2023-01-04
Associate the use of certificates with the election


shenkeyao:
We mixed up the use of structs and types for certificates (QC and DAC). E.g., the `Replica` struct is associated with `ELECTION`, but its `high_qc` field is a `QuorumCertificate` struct that's unrelated to `ELECTION`. We should change it to the `ELECTION::QuorumCertificate` type.

This will give us more flexibility for refactoring, since the certificates will become traits associated with a round of the election.

PR #839, opened 2023-01-05
Replace certificate structs with associated types

shenkeyao:
- Adds QC and DAC associated types to `LeafType`.
- Makes the certificate types consistent between `Election` and its associated `LeafType`.
- Replaces the use of certificate structs with associated types.

Closes https://github.com/EspressoSystems/HotShot/issues/837.

### DIFF
--- a/src/traits/election/vrf.rs
+++ b/src/traits/election/vrf.rs
@@ -6,12 +6,13 @@ use commit::{Commitment, Committable, RawCommitmentBuilder};
 use derivative::Derivative;
 use espresso_systems_common::hotshot::tag;
 use hotshot_types::{
-    certificate::{DACertificate, QuorumCertificate},
+    certificate::DACertificate,
     data::LeafType,
     traits::{
         election::{
             Checked::{self, Inval, Unchecked, Valid},
-            Election, ElectionConfig, ElectionError, TestableElection, VoteToken,
+            Election, ElectionConfig, ElectionError, SignedCertificate, TestableElection,
+            VoteToken,
         },
         node_implementation::NodeType,
         signature_key::{EncodedPublicKey, EncodedSignature, SignatureKey, TestableSignatureKey},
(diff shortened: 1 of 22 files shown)


bfish713 (approved):
LGTM

testing/tests/consensus.rs
```diff
@@ -299,7 +307,11 @@ where
 /// Submits proposals for 0..NUM_VIEWS rounds where `node_id` is the leader
 fn test_validating_proposal_queueing_round_setup<
     TYPES: NodeType<ConsensusType = ValidatingConsensus>,
-    ELECTION: Election<TYPES, LeafType = ValidatingLeaf<TYPES>>,
+    ELECTION: Election<
```
DieracDelta, line 310: This is a mouthful. Wish we could type alias the trait within an `impl` block.

src/traits/election/vrf.rs
```diff
@@ -282,17 +284,17 @@ where
...
-#[derivative(Eq, PartialEq)]
+#[derivative(Debug, Eq, PartialEq)]
 pub struct VrfImpl<TYPES, LEAF: LeafType<NodeType = TYPES>, SIGSCHEME, VRF, VRFHASHER, VRFPARAMS>
 where
     VRF: Vrf<VRFHASHER, VRFPARAMS> + Sync + Send,
     TYPES: NodeType,
 {
     /// the stake table
-    #[derivative(PartialEq = "ignore")]
+    #[derivative(Debug = "ignore", PartialEq = "ignore")]
```
DieracDelta, line 294: Printing the stake table is most likely useful for debugging. But, we no longer care about vrf based election (afaik?) so I don't think this matters.

types/src/certificate.rs
```diff
@@ -59,53 +58,81 @@ pub struct QuorumCertificate<TYPES: NodeType, LEAF: LeafType<NodeType = TYPES>>
...
-    pub fn genesis() -> Self {
-        Self {
-            // block_commitment: fake_commitment(),
-            leaf_commitment: fake_commitment::<LEAF>(),
-            view_number: <TYPES::Time as ConsensusTime>::genesis(),
-            signatures: BTreeMap::default(),
-            genesis: true,
-        }
-    }
+    pub is_genesis: bool,
```
DieracDelta, line 61: Nice!

DieracDelta (approved):
LGTM!

testing/tests/consensus.rs
shenkeyao: Me too! Maybe after the big refactor, we could try to make them look better. I think it's okay for now since we are going to change the `Election` trait anyway.

PR #851, opened 2023-01-10
Implement signing and voting interfaces

shenkeyao:
- Implements proposal signing interfaces.
- Adds vote types and implements vote signing for each type.
- Implements validation functions for proposals, votes and certificates.

Closes https://github.com/EspressoSystems/HotShot/issues/710. Closes https://github.com/EspressoSystems/HotShot/issues/712.


consensus/src/traits.rs
```diff
@@ -153,36 +152,85 @@ pub trait ConsensusApi<
...
+    }
+
+    /// Sign a positive vote on validating or commitment proposal.
+    fn sign_yes_vote(
+        &self,
+        leaf_commitment: Commitment<LEAF>,
+    ) -> (EncodedPublicKey, EncodedSignature) {
+        let signature = TYPES::SignatureKey::sign(
+            self.private_key(),
+            &VoteData::<TYPES, LEAF>::Yes(leaf_commitment).as_bytes(),
```
DieracDelta, line 192: It may be worth adding a comment here explaining why this is correct. IMO this is a bit subtle. Something like:
```
/// the leaf commitment and the type of the vote (yes) are signed
/// This is the minimum amount of information necessary. 
/// Any users of the subsequently constructed QC should be able to check 
/// that this node voted `Yes` on that leaf. 
/// The leaf is expected to be reconstructed based on other information in the yes vote.
```

types/src/traits/election.rs
```diff
@@ -133,30 +161,84 @@ pub trait Election<TYPES: NodeType>: Clone + Eq + PartialEq + Send + Sync + 'sta
...
+                )
+            })
+            .fold(0, |acc, x| (acc + u64::from(x.1 .1.vote_count())));
 
-    /// check that the data availability certificate is valid
-    fn is_valid_dac(&self, qc: <Self::LeafType as LeafType>::DACertificate) -> bool;
+        if stake >= u64::from(self.threshold()) {
+            return true;
+        }
+        false
```
DieracDelta, line 187:
```suggestion
        stake >= u64::from(self.threshold())
```

DieracDelta (approved):
LGTM

DieracDelta:
Merging this should close #418 . Note that we're not authenticating data messages as both enum variants are handled in other ways. Notably, we expect transaction authentication to be done by the application layer. And newest QC is unused.

consensus/src/traits.rs
shenkeyao: Sounds good! I combined your suggested changes for `sign_yes_vote` and `sign_timeout_vote` and added comments for all vote signing functions.

types/src/traits/election.rs
shenkeyao: I thought Clippy would catch this! 😅

consensus/src/traits.rs
DieracDelta: 🎉

types/src/traits/election.rs
DieracDelta: You'd think so!

DieracDelta (approved):

PR #853, opened 2023-01-12
DA leader implementation

bfish713:
Logic just for the DA Leader to send out proposal and to collect votes on them

Depends on: [PR: 851](https://github.com/EspressoSystems/HotShot/pull/851)

Closes: #[714](https://github.com/EspressoSystems/HotShot/issues/714)

### DIFF
--- a/consensus/src/da.rs
+++ b/consensus/src/da.rs
@@ -46,6 +46,15 @@ pub struct DALeader<
     /// Limited access to the consensus protocol
     pub api: A,
 
+    pub vote_collection_chan: Arc<
+    Mutex<
+        UnboundedReceiver<
+            ConsensusMessage<TYPES, DALeaf<TYPES>, DAProposal<TYPES, ELECTION>>,
+        >,
+    >,
+>,
+/// The 
+
     #[allow(missing_docs)]
     #[allow(clippy::missing_docs_in_private_items)]
     pub _pd: PhantomData<ELECTION>,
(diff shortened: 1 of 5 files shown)


consensus/src/next_leader.rs
```diff
@@ -87,63 +96,76 @@ where
...
+                            }
+
+                            // if the signature on the vote is invalid,
+                            // assume it's sent by byzantine node
+                            // and ignore
+
+                            if !self.api.is_valid_vote(
+                                &vote.signature.0,
+                                &vote.signature.1,
+                                VoteData::Yes(vote.leaf_commitment),
```
shenkeyao, line 119: Unlike `block.commit()` that we can directly use to validate the DA vote, here we don't have something like `leaf.commit()` that we know is true. But if we use `vote.leaf_commitment` for the validation, we need a way to ensure it's correct/consistent among honest nodes.

We can probably change `stake_casted` to a map from leaf commitment to count, and we construct the QC only when the count of a commitment meets the threshold.

consensus/src/next_leader.rs
```diff
@@ -87,63 +96,76 @@ where
...
+                                vote.current_view,
```
shenkeyao, line 120: Similarly, we should check `vote.current_view == self.cur_view`.

shenkeyao (commented):
I had some comments about checking the block/leaf commitment and view number. Those comments in `next_leader.rs` should perhaps be addressed in my PR #851 since you were just rebasing on it and didn't change anything.

Let me update that file in my PR.

shenkeyao:
> I had some comments about checking the block/leaf commitment and view number. Those comments in `next_leader.rs` should perhaps be addressed in my PR #851 since you were just rebasing on it and didn't change anything.
> 
> Let me update that file in my PR.

Actually, no, I'll open another PR to fix the commitment and view number in `next_leader.rs`. Don't want to make merging too complicated. 😄

consensus/src/da.rs
```diff
@@ -0,0 +1,273 @@
...
+                                != <TYPES::SignatureKey as SignatureKey>::to_bytes(&sender)
+                            {
+                                continue;
+                            }
+
+                            // Ignore invalid signature
+                            if !self.api.is_valid_vote(
+                                &vote.signature.0,
+                                &vote.signature.1,
+                                VoteData::DA(vote.block_commitment),
```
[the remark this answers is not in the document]
nyospe: 👍

consensus/src/da.rs
```diff
@@ -0,0 +1,273 @@
...
+                                vote.current_view,
```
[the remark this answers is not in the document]
nyospe: Yes on this as well.

types/src/message.rs
```diff
@@ -67,6 +67,79 @@ impl<
     }
 }
 
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+#[serde(bound(deserialize = ""))]
+/// Votes sent by consensus messages.
+pub enum Vote<TYPES: NodeType, LEAF: LeafType<NodeType = TYPES>> {
```
nyospe, line 73: 👍  We should have an issue noting that this is a stopgap until the traits refactoring is complete, and link it in a comment here.

nyospe (approved):
Solid progress. Seconding Keyao's comment about the checked conditions, but that really comes back to me getting a table of the constraint properties together.

types/src/message.rs
shenkeyao: Oh this is also something added by the voting interface PR that's already merged, so not part of this PR actually. I just created an issue: https://github.com/EspressoSystems/HotShot/issues/856 for this. I could link it in the next PR for fixing the next leader logic (https://github.com/EspressoSystems/HotShot/issues/854), so Brendon could merge this PR sooner.

shenkeyao (approved):

consensus/src/da.rs
```diff
@@ -0,0 +1,269 @@
...
+        }
+
+        // Wait for DA votes or Timeout
+        let lock = self.vote_collection_chan.lock().await;
+        let mut vote_outcomes = HashMap::new();
+        let threshold = self.api.threshold();
+        let mut stake_casted = 0;
+
+        while let Ok(msg) = lock.recv().await {
+            if Into::<ConsensusMessage<_, _, _>>::into(msg.clone()).view_number() != self.cur_view {
```
shenkeyao, line 207: Just noticed that we had this check before validating the vote, so it actually doesn't matter whether we use `self.cur_view` (as in the current version) or `vote.view_number()` (as in your previous implementation) for validation.

I'll add this check to `next_leader.rs` as well, but the DA logic looks good!

PR #855, opened 2023-01-13
Fix node mismatch warnings

DieracDelta:
Turns out the node mismatch warnings and failures on aarch64 are being caused by hashing/eq mismatches. This fixes that by invoking derivative. Not sure why this causes failures on aarch64 but not other archs... Related: we should devote time to improving our leaf comparison assertions in the testing framework. 

Those who have Apple silicon, could you double check that the tests pass for you on this branch?


nyospe (approved):

elliedavidson (approved):
The tests that failed before are passing for me now.

Issue #843, opened 2023-01-06
Implement `ConnectedNetwork` for existing network layers


bfish713:
- [ ] libp2p
- [ ] centralized server
- [ ] memory network

Issue #842, opened 2023-01-06
Define Networking Traits: `CommunicationChanne`l and `ConnectedNetwork`


bfish713:
ConnectedNetwork - is where we define how we connected to different participants.  There will be an impl for both LibP2P and nginx
CommunicationChannel - is where we define which group we are sending messages two.  There will be an impl for DA and for general consensus (everyone voting)

DieracDelta:
Merged and finished

PR #870, opened 2023-01-18
Networking DA changes

### LINKED ISSUE #872: Migrate NetworkingImplementation to ConnectedNetwork + CommunicationChannel

DieracDelta:
PR for splitting NetworkingImplementation into CommunicationChannel and ConnectedNetwork. Note: `{broadcast,direct}_msg` is still blocking and requestids are not tracked. These will be filled out further in a subsequent PR.
  
  Things that are missing:
  - A non-blocking API for ConnectedNetwork and CommunicationChannel
  - Cancellable requests
  - The `CommunicationChannel` implementation is copy pasta. I'd like to include a default implementation but this forces the requirement that the underlying struct implements connectednetwork, which is not necessarily true (only true in the trivial case).
  - A less hacky way to insert a global topic for libp2p

Things that were done:

- [x] Add ConnectedNetwork and CommunicationChannel traits
- [x] Implement ConnectedNetwork
  - [x] For Libp2p
  - [x] For CentralizedServer
  - [x] For MemoryNetwork
  - [x] Re-enable MemoryNetwork unit tests on top of ConnectedNetwork
- [x] Implement Identity CommunicationChannels (e.g. with only one underlying ConnectedNetwork)
  - [x] For Libp2p
  - [x] For CentralizedServer
  - [x] For MemoryNetwork (testing purposes)
- [x] s/NetworkingImplementation/{CommunicationChannel,ConnectedNetwork}
  - [x] Tie in with consensus
- [x] Re-enable TestableNetworkingImplementation on top of CommunicationChannel
- [x] Re-enable testing framework and tests
  - [x] Fix up macros
  - [x] Fix up any manual tests

Closes #872 , #843 , #842

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -303,7 +303,9 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> HotShot<TYPES::ConsensusType
         // Add the transaction to our own queue first
         trace!("Adding transaction to our own queue");
         // Wrap up a message
-        let message = DataMessage::SubmitTransaction(transaction);
+        // TODO place a view number here that makes sense
+        // we haven't worked out how this will work yet
+        let message = DataMessage::SubmitTransaction(transaction, TYPES::Time::new(0));
 
         let api = self.clone();
         async_spawn(async move {
@@ -394,7 +396,6 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> HotShot<TYPES::ConsensusType
                     Message { sender: pk, kind },
                     // TODO this is morally wrong
                     inner.election.clone().as_ref(),
-                    <TYPES::Time as ConsensusTime>::new(0),
                 )
                 .await
                 .is_err()
@@ -552,12 +553,12 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> HotShot<TYPES::ConsensusType
     /// Handle an incoming [`DataMessage`] that was broadcasted on the network
     async fn handle_broadcast_data_message(
         &self,
-        msg: DataMessage<TYPES, I::Leaf>,
+        msg: DataMessage<TYPES>,
         _sender: TYPES::SignatureKey,
     ) {
         // TODO validate incoming broadcast message based on sender signature key
         match msg {
-            DataMessage::SubmitTransaction(transaction) => {
+            DataMessage::SubmitTransaction(transaction, _view_number) => {
                 let size = bincode_opts().serialized_size(&transaction).unwrap_or(0);
 
                 // The API contract requires the hash to be unique
@@ -582,31 +583,18 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> HotShot<TYPES::ConsensusType
                         .update(size as i64);
                 }
             }
-            DataMessage::NewestQuorumCertificate { .. } => {
-                // Log the exceptional situation and proceed
-                warn!(
-                    ?msg,
-                    "Newest QC received over broadcast channel. This shouldn't be possible."
-                );
-            }
         }
     }
 
     /// Handle an incoming [`DataMessage`] that directed at this node
     async fn handle_direct_data_message(
         &self,
-        msg: DataMessage<TYPES, I::Leaf>,
+        msg: DataMessage<TYPES>,
         _sender: TYPES::SignatureKey,
     ) {
         debug!(?msg, "Incoming direct data message");
         match msg {
-            DataMessage::NewestQuorumCertificate { .. } => {
-                warn!(
-                    ?msg,
-                    "Newest QC received over direct channel. This shouldn't be possible."
-                );
-            }
-            DataMessage::SubmitTransaction(_) => {
+            DataMessage::SubmitTransaction(_, _) => {
                 // Log exceptional situation and proceed
                 warn!(?msg, "Broadcast message received over direct channel");
             }
@@ -1051,7 +1039,6 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>>
                 },
                 // TODO this is morally wrong
                 self.inner.election.clone().as_ref(),
-                <TYPES::Time as ConsensusTime>::new(0),
             )
             .await?;
         Ok(())
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -41,7 +41,7 @@ pub mod tasks;
 
 use crate::{
     certificate::QuorumCertificate,
-    traits::{NetworkingImplementation, NodeImplementation, Storage},
+    traits::{NodeImplementation, Storage},
     types::{Event, HotShotHandle},
 };
 use async_compatibility_layer::{
@@ -79,6 +79,10 @@ use hotshot_types::{
         State,
     },
     HotShotConfig,
+};
+use hotshot_types::{
+    message::{MessageKind, ProcessedConsensusMessage},
+    traits::network::CommunicationChannel,
 };
 use hotshot_utils::bincode::bincode_opts;
 #[allow(deprecated)]
@@ -390,7 +394,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> HotShot<TYPES::ConsensusType
         async_spawn_local(async move {
             if inner
                 .networking
-                .broadcast_message(Message { sender: pk, kind })
+                .broadcast_message_cc(Message { sender: pk, kind }, nll_todo(), nll_todo())
                 .await
                 .is_err()
             {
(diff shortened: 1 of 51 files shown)


src/traits/networking/libp2p_network.rs
```diff
@@ -572,21 +542,24 @@ impl<
...
     }
 
-    #[instrument(name = "Libp2pNetwork::message_node", skip_all)]
-    async fn message_node(
-        &self,
-        message: Message<TYPES, LEAF, PROPOSAL>,
-        recipient: TYPES::SignatureKey,
-    ) -> Result<(), NetworkError> {
+    // TODO make non-blocking
+    // TODO make requestid meaningful (e.g. track it)
```
shenkeyao, line 546: I'm curious about why we would want to track `RequestId`. Is it for testing purposes?

types/src/traits/network.rs
```diff
@@ -109,105 +106,147 @@ pub enum NetworkError {
...
+/// common traits we would like olur network messages to implement
+pub trait NetworkMsg:
+    Serialize + for<'a> Deserialize<'a> + Clone + std::fmt::Debug + Sync + Send + 'static
+{
+}
+
+/// represents a networking implmentration
+/// exposes low level API for interacting with a network
+/// intended to be implemented for libp2p, the centralized server,
+/// and memory network
```
shenkeyao, line 199: I'm a bit confused about why we have the `ConnectedNetwork` trait specifically implemented for libp2p, the centralized server, and the memory network, but the implementation of each function is just called by the corresponding function of `CommunicationChannel`. Isn't this the same as putting the implementations directly in `CommunicationChannel` and removing `ConnectedNetwork`?

I might have missed something. I thought we wanted these two traits to support more than one instance of the networking implementation, but I'm not sure how this implementation fits into this idea.

types/src/traits/network.rs
```diff
@@ -109,105 +106,147 @@ pub enum NetworkError {
...
+    async fn shut_down_cc(&self) -> ();
+
+    /// broadcast message to those listening on the communication channel
+    /// non-blocking
+    async fn broadcast_message_cc(
         &self,
         message: Message<TYPES, LEAF, PROPOSAL>,
-    ) -> Result<(), NetworkError>;
+        election: &ELECTION,
+        view_number: TYPES::Time,
```
shenkeyao, line 156: Do we need the view number here? It can be retrieved from `message` anyway.

src/lib.rs
```diff
@@ -390,7 +388,13 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> HotShot<TYPES::ConsensusType
         async_spawn_local(async move {
             if inner
                 .networking
-                .broadcast_message(Message { sender: pk, kind })
+                .broadcast_message_cc(
+                    Message { sender: pk, kind },
+                    // TODO this is morally wrong
+                    inner.election.clone().as_ref(),
+                    // TODO this is conceptually wrong
```
shenkeyao, line 395: Won't be a problem if we decide it's okay to remove the view number from here.

src/lib.rs
```diff
@@ -390,7 +388,13 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> HotShot<TYPES::ConsensusType
         async_spawn_local(async move {
             if inner
                 .networking
-                .broadcast_message(Message { sender: pk, kind })
+                .broadcast_message_cc(
+                    Message { sender: pk, kind },
+                    // TODO this is morally wrong
```
shenkeyao, line 393: Why? Because of the `clone()`? 😄

shenkeyao (changes requested):
Looks good overall! I have some comments and questions.

types/src/traits/network.rs
```diff
@@ -109,105 +106,147 @@ pub enum NetworkError {
...
     PROPOSAL: ProposalType<NodeType = TYPES>,
+    ELECTION: Election<TYPES>,
 >: Clone + Send + Sync + 'static
 {
     /// Returns true when node is successfully initialized
     /// into the network
-    ///
     /// Blocks until node is ready
-    async fn ready(&self) -> bool;
+    async fn ready_cc(&self) -> bool;
```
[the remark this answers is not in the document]
DieracDelta: If both `ConnectedNetwork` and `CommunicationChannel` are in scope then one must turbofish whenever calling `ready` to specify which trait to use. Adding `cc` to mean `communication_channel` seemed easier

src/traits/networking/libp2p_network.rs
DieracDelta: I was thinking to be able to report the status of the request. E.g. to answer the following sorts of questions: was the request completed? Is the request in the process of being cancelled? Has the request already been cancelled?

types/src/traits/network.rs
DieracDelta: Oooh yes 100%, I'll remove this 👍

src/lib.rs
DieracDelta: I think now this is assuming that Election is threadsafe but not putting it behind a rwlock or mutex (maybe I should do that?).

testing/tests/consensus.rs
```diff
@@ -547,6 +547,8 @@ async fn test_bad_proposal() {
 
     test.execute().await.unwrap();
 }
+// hotshot::traits::implementations::MemoryNetwork< ValidatingProposal<StaticCommitteeTestTypes, GeneralStaticCommittee<StaticCommitteeTestTypes, ValidatingLeaf<StaticCommitteeTestTypes>, JfPubKey<BLSSignatureScheme<ark_bls12_381::Parameters>>>>, JfPubKey<BLSSignatureScheme<ark_bls12_381::Parameters>>>:
+// TestableNetworkingImplementation<StaticCommitteeTestTypes, ValidatingLeaf<StaticCommitteeTestTypes>, ValidatingProposal<StaticCommitteeTestTypes, GeneralStaticCommittee<StaticCommitteeTestTypes, ValidatingLeaf<StaticCommitteeTestTypes>, JfPubKey<BLSSignatureScheme<ark_bls12_381::Parameters>>>>, GeneralStaticCommittee<StaticCommitteeTestTypes, ValidatingLeaf<StaticCommitteeTestTypes>, JfPubKey<BLSSignatureScheme<ark_bls12_381::Parameters>>>>
```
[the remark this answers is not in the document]
DieracDelta: Yes!

types/src/traits/network.rs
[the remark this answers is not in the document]
nyospe: I wasn't planning for the same type to implement both traits... it seemed like aggregation would be a better approach for this (e.g. `CommunicationChannel`(s) have handles to `ConnectedNetwork`(s))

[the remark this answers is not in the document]
DieracDelta: Maybe a wrapper type makes sense here? RIght now `CommunicationChannel` just passes through to `ConnectedNetwork`.

types/src/traits/network.rs
DieracDelta: Actually it looks like like some cases don't have a view number? Specifically the `DataMessage::Transaction`. We'll probably need a view number for this if we just want to send it to the DA committee.

src/traits/networking/libp2p_network.rs
shenkeyao: Yeah makes sense!

types/src/traits/network.rs
shenkeyao: If the view number is indeed needed when sending the transaction, maybe we can add `view_number` to the `DataMessage::SubmitTransaction` variant. So instead of `SubmitTransaction(TYPES::Transaction)`, we may have `SubmitTransaction(txn, view_number)`. I feel this is better than having a duplicate view number somewhere besides `Message`.

src/lib.rs
shenkeyao: Would it be simpler to make `braodcast_message` take the committee set (i.e., the `BTreeSet` retrieved from `get_committee(election, view_number)`) rather than the whole election as input to avoid the thread safety problem? That's the only info needed for message broadcasting.

shenkeyao (approved):
The update looks good! There are merge conflicts though.

types/src/traits/network.rs
DieracDelta: I'll do that! I'm a bit concerned about adding slightly extra bandwidth in the future when we send the data messages over the network. 8 extra bytes per message is probably fine/something we can consider in the future.

src/lib.rs
DieracDelta: This is a good point! I think this is simpler.

Long term, my understanding is we'll need more from election than just the keyset. Notably, the DA committee won't be fixed, so we'll probably want to query a read-only version of the Election every time we need a few more keys. Maybe this could be a stream? Or some sort iterator of keys? @nyospe are there other things that broadcast might need from election? If not, I'll switch this back to a keyset, and add Election as maybe an associated type? We still would need to use the Election generic somewhere.

types/src/traits/network.rs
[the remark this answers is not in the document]
DieracDelta: Added a wrapper type.

src/lib.rs
shenkeyao: Yeah sounds good to me. If we are going to query more items from `Election`, it makes sense to have a read-only version.

shenkeyao: Yeah, this makes sense!

shenkeyao (approved):
Looks good!

PR #875, opened 2023-01-19
Implement committee member logic

shenkeyao:
Implement the logic of collecting and voting on DA proposal for committee member.

Closes https://github.com/EspressoSystems/HotShot/issues/717.


shenkeyao:
Need to merge with lib.rs changes for next leader.

bfish713:
This looks good to me, I merged my PR so the conflicts can be resolved now and I can approve after that

nyospe (dismissed):

consensus/src/da_member.rs
```diff
@@ -0,0 +1,289 @@
...
+                            },
+                        );
+                        let safety_check = outcome.is_ok();
+                        if let Err(e) = outcome {
+                            self.api.send_view_error(self.cur_view, Arc::new(e)).await;
+                        }
+
+                        // Skip if both saftey and liveness checks fail.
+                        if !safety_check && !liveness_check {
+                            warn!("Failed safety check and liveness check");
```
DieracDelta, line 188: Are safety and liveness checks necessary here? My (potentially incorrect) understanding is that we check for structural integrity and then immediately vote yes. And then these liveness and safety checks happen later when we are a replica validating the CommitmentProposal.

DieracDelta (approved):

consensus/src/da_member.rs
bfish713: +1, At the very least replicas not on the committee will have to do this same check

bfish713 (dismissed):

consensus/src/da_member.rs
DieracDelta: I wonder if we need the liveness check to be +2 or +1. +2 was added to make the chernoff bounds work when we were doing VRF on consensus (not DA). Since we're not doing that anymore, when voting on the CommitmentProposal, I think we can change the liveness check back to +1.

src/lib.rs
```diff
@@ -895,6 +897,19 @@ where
             let txns = consensus.transactions.clone();
             (high_qc, txns)
         };
+        let mut send_to_member = hotshot.replica_channel_map.write().await;
```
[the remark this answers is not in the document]
shenkeyao: Yeah, there's a comment in this function:
```
// TODO probably cleaner to separate this into a function
// e.g. insert the view and remove the last view
```
I'll modify the comment a bit to make it clearer.

consensus/src/da_member.rs
```diff
@@ -0,0 +1,289 @@
...
+            match &parent_view.view_inner {
+                ViewInner::Leaf { leaf } => {
+                    if let Some(leaf) = consensus.saved_leaves.get(leaf) {
+                        leaf
+                    } else {
+                        warn!("Failed to find high QC parent.");
+                        return None;
+                    }
+                }
+                // can happen if future api is whacked
```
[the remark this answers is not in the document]
shenkeyao: This comment was copied from `leader.rs`, and also appeared in `da_leader.rs`. I'm going to just remove it in these files.

src/lib.rs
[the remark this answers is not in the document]
shenkeyao: Added an issue for this: https://github.com/EspressoSystems/HotShot/issues/897.

consensus/src/da_member.rs
shenkeyao: Thanks! I've removed the safety and liveness checks here. Also fixed the liveness check for the validating replica.

DieracDelta (approved):

PR #877, opened 2023-01-20
Implement Next Leader and the Rest of Leader for DA consensus

bfish713:
Adding logic for The leader and next leader in the DA/Sequencing consensus
- The `DAConsensusLeader` is the task which builds and broadcast the `CommitmentProposal`.  For now this tasks depends on the DAProposal succeeding and can only begin after that.  
- DANextLeader is the next leader task which collects votes on the previous leaders proposal and builds the QC

The vote collection logic for `DAProposals` and `CommitmentProposals` is quite similar and I intend to refactor to use the new election trait after this.  It should reduce the duplicated code to some degree

Also I stubbed out the network function for sending the DA proposal.  Currently, the networking traits all assume 1 type of proposal but we need to send DA and Commitment proposals.  I stubbed it out because this exactly what @DieracDelta is working on with the networking trait refactor.  

Closes: https://github.com/EspressoSystems/HotShot/issues/715


consensus/src/traits.rs
```diff
@@ -152,6 +152,12 @@ pub trait ConsensusApi<
         .await;
     }
 
+    /// Send a broadcast to the DA comitee, stub for now
+    async fn send_da_broadcast<DAPROPOSAL: ProposalType<NodeType = TYPES>>(
```
DieracDelta, line 156: I think this might need a view number to determine the DA committee for this view.

src/lib.rs
```diff
@@ -995,6 +1035,23 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>>
             .await
     }
 
+    async fn send_da_broadcast<DAPROPOSAL: ProposalType<NodeType = TYPES>>(
+        &self,
+        _message: ConsensusMessage<TYPES, I::Leaf, DAPROPOSAL>,
+    ) -> std::result::Result<(), NetworkError> {
+        // TODO: Should look like this code but it won't work due to only 1 Proposal type being
+        // associated with self.inner.networking.
```
DieracDelta, line 1043: Makes sense! Like you said, the DA networking changes should help with this.

DieracDelta (approved):

consensus/src/traits.rs
bfish713: I will add view number to the function signature as a placeholder for now.

consensus/src/da.rs
```diff
@@ -194,33 +267,202 @@ where
...
+            state
+        } else {
+            warn!("Don't have last state on parent leaf");
+            return None;
+        };
+        if let Ok(new_state) = starting_state.append(&self.block, &self.cur_view) {
+            let leaf = DALeaf {
+                view_number: self.cur_view,
+                // TODO: what is this height for?
+                height: 0,
```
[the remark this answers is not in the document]
bfish713: I also am not sure

PR #888, opened 2023-01-24
Remove the bound on Leaf in `HotShotHandle::start()`

bfish713:
from Jeb on Zulip
"
Also, `HotShotHandle::start` requires `I::Leaf: From<StoredView<Types, I::Leaf>>`. `ValidatingLeaf` implements this `From` but `DALeaf` doesn't. Since both leaf types implement `LeafType` which has `from_stored_view`, and `NodeImplementation::Leaf` already implements `LeafType`, should the bound on `start` be removed and `from` be replaced with `from_stored_view`?
"

### DIFF
--- a/src/types/handle.rs
+++ b/src/types/handle.rs
@@ -12,11 +12,8 @@ use hotshot_types::{
     error::{HotShotError, RoundTimedoutState},
     event::EventType,
     traits::{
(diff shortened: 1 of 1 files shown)


jbearer (approved):
LGTM

shenkeyao (approved):

Issue #883, opened 2023-01-23
Implement New Election Trait for DA


bfish713:
DA collects votes for 2 different reasons.  The vote collection logic is very similar, but not the same between the two.  For DA committee we collect votes from a subset of the network and form a DA certificate.  Then for the commitment proposal we collect votes from all nodes and form a Quorum Certificate.  We should implement both of these election/voting mechanisms to avoid having a lot of the duplicated code in the implementation.

PR #908, opened 2023-02-01
Refactor Vote Collection

bfish713:
We collect votes in 3 places and the logic was just duplicated everywhere.  This removes most of that duplication and puts it into an `Accumlator`.  There is still some funkiness in election because it has both DA and QC certificate types.  I also couldn't figure a way not have a `accumulate_da_vote` and `accumulate_qc_vote` functions despite the logic looking the same

Closes: https://github.com/EspressoSystems/HotShot/issues/883

### DIFF
--- a/src/traits/election/static_committee.rs
+++ b/src/traits/election/static_committee.rs
@@ -1,6 +1,8 @@
+use super::vrf::JfPubKey;
 use ark_bls12_381::Parameters as Param381;
 use commit::{Commitment, Committable, RawCommitmentBuilder};
 use espresso_systems_common::hotshot::tag;
+use hotshot_types::certificate::DACertificate;
 use hotshot_types::{
     data::LeafType,
     traits::{
@@ -14,8 +16,6 @@ use jf_primitives::signatures::BLSSignatureScheme;
 use serde::{Deserialize, Serialize};
 use std::marker::PhantomData;
 use std::num::NonZeroU64;
-use hotshot_types::certificate::DACertificate;
-use super::vrf::JfPubKey;
 
 /// Dummy implementation of [`Election`]
 
--- a/src/traits/election/static_committee.rs
+++ b/src/traits/election/static_committee.rs
@@ -2,7 +2,6 @@ use super::vrf::JfPubKey;
 use ark_bls12_381::Parameters as Param381;
 use commit::{Commitment, Committable, RawCommitmentBuilder};
 use espresso_systems_common::hotshot::tag;
-use hotshot_types::certificate::DACertificate;
 use hotshot_types::{
     data::LeafType,
     traits::{
@@ -95,7 +94,8 @@ where
 
     type QuorumCertificate = LEAF::QuorumCertificate;
 
-    type DACertificate = DACertificate<TYPES>;
+    type DACertificate = LEAF::DACertificate;
+    // type DACertificate = DACertificate<TYPES>;
 
     type LeafType = LEAF;
 
--- a/src/traits/election/static_committee.rs
+++ b/src/traits/election/static_committee.rs
@@ -14,7 +14,7 @@ use jf_primitives::signatures::BLSSignatureScheme;
 use serde::{Deserialize, Serialize};
 use std::marker::PhantomData;
 use std::num::NonZeroU64;
-
+use hotshot_types::certificate::DACertificate;
 use super::vrf::JfPubKey;
 
 /// Dummy implementation of [`Election`]
@@ -95,7 +95,7 @@ where
 
     type QuorumCertificate = LEAF::QuorumCertificate;
 
-    type DACertificate = LEAF::DACertificate;
+    type DACertificate = DACertificate<TYPES>;
 
     type LeafType = LEAF;
 
--- a/types/src/traits/election.rs
+++ b/types/src/traits/election.rs
@@ -14,7 +14,6 @@ use crate::{
     },
 };
 use bincode::Options;
-use blake3::traits::digest::crypto_common::rand_core::block;
 use commit::{Commitment, Committable};
 use either::Either;
 use hotshot_utils::bincode::bincode_opts;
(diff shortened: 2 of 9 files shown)


types/src/traits/election.rs
```diff
@@ -101,9 +102,33 @@ pub trait Accumulator<T, U>: Sized {
...
+//         valid_signatures.insert(key, (sig.clone(), token.clone()));
+
+//         stake_casted += u64::from(token.vote_count());
+
+//         if stake_casted >= u64::from(threshold) {
+//             return Either::Right((self.valid_signatures));
+//         }
+//         Either::Left(self)
+//     }
+// }
```
DieracDelta, line 131: nit: Can this be deleted?

types/src/traits/election.rs
```diff
@@ -113,7 +138,17 @@ where
     Self: Send + Sync + Clone + Serialize + for<'a> Deserialize<'a>,
     LEAF: Committable,
 {
-    type Accumulator: Accumulator<(EncodedSignature, SIGNATURE), Self>;
+    // type Accumulator: Accumulator<
+    //     (Commitment<LEAF>, (EncodedPublicKey, (EncodedSignature, TOKEN))),
+    //     BTreeMap<EncodedPublicKey, (EncodedSignature, TOKEN)>,
+    // >;
```
DieracDelta, line 144: nit: Can this be deleted?

types/src/traits/election.rs
```diff
@@ -234,6 +269,83 @@ pub trait Election<TYPES: NodeType>: Clone + Eq + PartialEq + Send + Sync + 'sta
...
+    ) -> Either<
+        CertificateAccumulator<TYPES::VoteTokenType, Self::LeafType>,
+        QuorumCertificate<TYPES, Self::LeafType>,
+    > {
+        if !self.is_valid_vote(
+            encoded_key,
+            encoded_signature,
+            VoteData::Yes(leaf_commitment),
+            view_number,
+            // Ignoring deserialization errors below since we are getting rid of it soon
```
DieracDelta, line 291: I think this was copied over from before, but I'm not sure what exactly we're getting rid of. Serialiazation? Does anyone know? If not we can probably delete the comment and the one below

types/src/traits/election.rs
```diff
@@ -234,6 +269,83 @@ pub trait Election<TYPES: NodeType>: Clone + Eq + PartialEq + Send + Sync + 'sta
...
+    /// Accumlate the vote, return the QC if the threshold was reached, if not return the updated accumlator
+    #[allow(clippy::type_complexity)]
+    fn accumulate_qc_vote(
+        &self,
+        encoded_key: &EncodedPublicKey,
+        encoded_signature: &EncodedSignature,
+        leaf_commitment: Commitment<Self::LeafType>,
+        vote_token: TYPES::VoteTokenType,
+        view_number: TYPES::Time,
+        accumlator: CertificateAccumulator<TYPES::VoteTokenType, Self::LeafType>,
```
DieracDelta, line 281: nit: `accumulator`

types/src/traits/election.rs
```diff
@@ -234,6 +269,83 @@ pub trait Election<TYPES: NodeType>: Clone + Eq + PartialEq + Send + Sync + 'sta
         is_valid_signature && is_valid_vote_token
     }
 
+    /// Accumlate the vote, return the QC if the threshold was reached, if not return the updated accumlator
+    #[allow(clippy::type_complexity)]
+    fn accumulate_qc_vote(
+        &self,
+        encoded_key: &EncodedPublicKey,
+        encoded_signature: &EncodedSignature,
+        leaf_commitment: Commitment<Self::LeafType>,
```
DieracDelta, line 278: An idea to reduce code duplication:

Maybe we could stick all these arguments into a struct, and make that struct generic over the type of commitment? This will fix the clippy lint. Denote this struct `VoteMetaData<T>`. 

We could then type alias `pub type QcVote<WhateverSelfIs> = VoteMetaData<WhateverSelfIs::LeafType` and `pub type DACVote<WhateverSelfIs> = VoteMetaData<WhateverSelfIs::Block>`. Then this `accumulate_vote` could be made generic over the commitment type and certificate type, and we could have const function aliases as part of the trait that monomorphize over the struct type aliases and certificate types.

DieracDelta (dismissed):
LGTM! I had one thought on merging those two functions

src/traits/election/static_committee.rs
```diff
@@ -95,7 +95,7 @@ where
 
     type QuorumCertificate = LEAF::QuorumCertificate;
 
-    type DACertificate = LEAF::DACertificate;
+    type DACertificate = DACertificate<TYPES>;
```
shenkeyao, line 98: Nit: I'm curious why this change is needed for DAC but not for QC.

types/src/traits/election.rs
shenkeyao: Maybe the code below this comment used to do some deserialization, and we forgot to remove/update the comment? It doesn't seem to be related to serialization now. I guess we can delete it.

shenkeyao (approved):
LGTM

src/traits/election/static_committee.rs
bfish713: It's because the DAC is based of the blocks commitment rather than the leaf commitment and the LeafType::DAC bounds DAC with `SignedCertificate<..., Self>`.  So the bounds are mismatched with Election bound DAC with `SignedCertificate<..., TYPES::BlockType>`

I actually reverted this and updated the trait bounds in LeafType and that seems to work and make more sense

types/src/traits/election.rs
bfish713: Based of this suggestion I did something similar to this but wasn't sure about the type alias part.  LMK what you think

DieracDelta: The type aliases were meant as a convenience so we don't accidentally feed in the wrong commitment. But, it looks like Rust is inferring the type where needed, which is cleaner IMO. Nice, LGTM!

DieracDelta (approved):

shenkeyao (approved):

PR #911, opened 2023-02-01
Implement the replica logic for sequencing consensus

shenkeyao:
- Implements the replica logic.
- Adds message creation functions.

Closes https://github.com/EspressoSystems/HotShot/issues/864.

### DIFF
--- a/consensus/src/da_replica.rs
+++ b/consensus/src/da_replica.rs
@@ -7,16 +7,18 @@ use crate::{
 };
 use async_compatibility_layer::channel::UnboundedReceiver;
 use async_lock::{Mutex, RwLock, RwLockUpgradableReadGuard, RwLockWriteGuard};
+use bincode::Options;
 use commit::Committable;
-use either::Right;
+use either::{Left, Right};
 use hotshot_types::{
-    certificate::QuorumCertificate,
+    certificate::{DACertificate, QuorumCertificate},
     data::{CommitmentProposal, DALeaf},
     message::{ConsensusMessage, ProcessedConsensusMessage},
     traits::{
         election::{Election, SignedCertificate},
         node_implementation::NodeType,
         signature_key::SignatureKey,
+        Block,
     },
 };
 use hotshot_utils::bincode::bincode_opts;
@@ -33,6 +35,7 @@ pub struct DAReplica<
     ELECTION: Election<
         TYPES,
         LeafType = DALeaf<TYPES>,
+        DACertificate = DACertificate<TYPES>,
         QuorumCertificate = QuorumCertificate<TYPES, DALeaf<TYPES>>,
     >,
 > {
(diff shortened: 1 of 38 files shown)


consensus/src/da_replica.rs
```diff
@@ -0,0 +1,526 @@
...
+                                error!(
+                                    "Failed to generate vote token for {:?} {:?}",
+                                    self.cur_view, e
+                                );
+                            }
+                            Ok(None) => {
+                                info!("We were not chosen for committee on {:?}", self.cur_view);
+                            }
+                            Ok(Some(vote_token)) => {
+                                info!("We were chosen for committee on {:?}", self.cur_view);
```
bfish713, line 150: nit: maybe the logging should specify the committee type, since here the committe is bascially anyone with stake, but in DAMember it's really checking membership on the DA comittee

src/lib.rs
```diff
@@ -950,6 +950,9 @@ impl<
             api: c_api.clone(),
         };
         let _ = da_member.run_view().await;
+        // TODO (da) Replica task isn't added since the proposal it listens to is the commitment
+        // proposal but `I::Proposal` here is the DA proposal. We need to support the two proposal
+        // types for sequencing consensus in the refactored node implementation.
```
bfish713, line 955: Can we explicitly pass the Proposal type as a generic parameter to `create_or_obtain_chan_from_read` and `create_or_obtain_chan_from_write` and then use that to create one recv channel for each proposal type?

bfish713 (commented):
The general logic looks good, but I'm wondering if it's possible we can hook this into the main `run_view` function with the suggestion I made?

Also I wonder how much of the logic can be shared between this and what's in replica.rs but we can save that for the next big refactor step.

consensus/src/da_replica.rs
```diff
@@ -0,0 +1,526 @@
...
+                                let parent_commitment = match self.parent_leaf().await {
+                                    Some(parent) => parent.commit(),
+                                    None => {
+                                        break None;
+                                    }
+                                };
+                                let block_commitment = p.data.block_commitment;
+                                let leaf = DALeaf {
+                                    view_number: self.cur_view,
+                                    height: 0,
```
DieracDelta, line 165: Should this depend on the parent's height?

src/lib.rs
bfish713: Hmm also the `channel_map` is typed with a signle proposal type so yeah it's not very simple to do

DieracDelta (commented):
LGTM

consensus/src/da_replica.rs
bfish713: Oh we should add height to `CommitmentProposal` it's on the leaf and also `ValidatingProposal` and the leader already depends on parent height to set this value

shenkeyao: I was thinking about using `parent.height` but it's not guaranteed to be the same as the height the leader used, so yeah, I think it's a good idea to add `height` to `CommitmentProposal`.

shenkeyao:
I accidentally pushed changes from another branch. I've reverted the commit, so the last two commits are irrelevant. You could just review this commit: https://github.com/EspressoSystems/HotShot/pull/911/commits/bf9f891c5bd2d29149bcad5d180624eaf26e1004. @bfish713 @DieracDelta

bfish713 (approved):

DieracDelta (approved):

Issue #884, opened 2023-01-23
Tie together DA Tasks for Leader, Replica, Next leader


bfish713:
When we complete the individual implementations we have to make sure that all the tasks work together and we can complete views.

Things we need to do:
- Properly send signals from task to task.  Next Leader task must get internal timeout message from replica task.  DA Task and leader task must get signal to stop when next round is started (also timeout)
- Add tests and examples to prove it works
- Join tasks at the end of a view

As a note: we eventually want to get rid of the per view functions and simply have each task spin up and shut down based on signals and messages from the network and not be purely based on running 1 view at a time.

PR #915, opened 2023-02-03
Spawn `run_view` tasks async and tie together them at the end

bfish713:
This does a few things each is pretty small/simple:

- Collect all the run views in tasks_handles and join them
- Update HighQC and return properly at the end of run view
- Add a new timeout function for DA with generic proposal type to be used later
- GC old leader channels

I also notice a few things still left todo and made comments on them.  The biggest one is the same issue Keyao ran into where the channel map is generic over just one proposal type.  So right now the next leader has the wrong type of channel and is collecting votes for DAProposal (which it shouldn't get) .  Fortunately the logic doesn't really change and we just need to update the type.

Closes: https://github.com/EspressoSystems/HotShot/issues/884

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -987,7 +987,9 @@ impl<
             high_qc: high_qc.clone(),
             api: c_api.clone(),
         };
-        async_spawn(async move { da_member.run_view() });
+        let member_handle = async_spawn(async move { da_member.run_view().await });
+        task_handles.push(member_handle);
+
         let _da_replica = {};
         // TODO: ADD DA replica task handle and push it to `task_handles`.
 
@@ -995,7 +997,6 @@ impl<
 
         async_spawn({
             let next_view_timeout = hotshot.inner.config.next_view_timeout;
-            let next_view_timeout = next_view_timeout;
             let hotshot: HotShot<TYPES::ConsensusType, TYPES, I> = hotshot.clone();
             async move {
                 async_sleep(Duration::from_millis(next_view_timeout)).await;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -292,6 +292,29 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> HotShot<TYPES::ConsensusType
         };
     }
 
+    /// Marks a given view number as timed out. This should be called a fixed period after a round is started.
+    ///
+    /// If the round has already ended then this function will essentially be a no-op. Otherwise `run_round` will return shortly after this function is called.
+    /// # Panics
+    /// Panics if the current view is not in the channel map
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -949,11 +987,45 @@ impl<
...
-        // TODO tie all the tasks together and do correct book keeping.
-        #[allow(deprecated)]
-        nll_todo()
+        // TODO: ADD DA replica task handle and push it to `task_handles`.
+
+        let children_finished = futures::future::join_all(task_handles);
+
+        async_spawn({
+            let next_view_timeout = hotshot.inner.config.next_view_timeout;
+            let next_view_timeout = next_view_timeout;
```
shenkeyao, line 998: Nit: remove this?

src/lib.rs
```diff
@@ -949,11 +987,45 @@ impl<
             high_qc: high_qc.clone(),
             api: c_api.clone(),
         };
-        let _ = da_member.run_view().await;
+        async_spawn(async move { da_member.run_view() });
         let _da_replica = {};
-        // TODO tie all the tasks together and do correct book keeping.
-        #[allow(deprecated)]
-        nll_todo()
+        // TODO: ADD DA replica task handle and push it to `task_handles`.
```
shenkeyao, line 992: Looks like `task_handles` only includes handles of the DA leader and the next leader. Besides the replica handle which we can't add now, I think we could add the handles of the consensus leader and DA member to `task_handles`.

shenkeyao (changes requested):
Looks good! I think we need to add two more task handles in `run_view`. Also left a few nit comments.

types/src/traits/election.rs
```diff
@@ -53,6 +53,7 @@ pub enum Checked<T> {
 
 /// Data to vote on for different types of votes.
 #[derive(Serialize)]
+// TODO (da) : LEAF type should just be changed to a commitment and bounded by Commitable
```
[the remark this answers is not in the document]
bfish713: Yeah you are right it doesn't make sense.  I was going through some stuff that would have to change for having 2 proposal types and for some reason got to this and thought it might need to change, but looking at it now I realize it does not.

src/lib.rs
bfish713: The consensus leader is together with the DA leader task since they are sequential.  So `leader_handle` includes both the DA leader and consensus leader.  I did forget the DA memeber

shenkeyao (approved):
LGTM

shenkeyao:
Ahh sorry I think my renaming PR caused the conflicts.

shenkeyao (approved):

Issue #856, opened 2023-01-13
Modify the Vote enum after the consensus trait refactoring


shenkeyao:
The `Vote` enum that we have for consensus messages is a stopgap. We should make it a trait after refactoring the consensus traits.

We may keep the enum for the yes and no votes since they are both signing on the leaf commitment and we need a way to distinguish the vote type.

PR #951, opened 2023-02-14
Refactor vote

shenkeyao:
- Splits vote type into DA and quorum votes.
- Converts vote enum to a trait.
- Adds vote generic to API, message and network types.
- Removes unused leaf generic.

There should be two vote types for sequencing consensus, but it's currently using `DAVote` since `NodeImplementation` only has one vote type. This will be fixed after adding `ConsensusExchange` and `SequencingNodeImplementation`. 

Closes https://github.com/EspressoSystems/HotShot/issues/932. Closes https://github.com/EspressoSystems/HotShot/issues/856.


testing/src/lib.rs
```diff
@@ -561,31 +562,34 @@ pub enum ConsensusTestError {
...
     ELECTION,
 > {
     _pd_0: PhantomData<TYPES>,
     _pd_1: PhantomData<LEAF>,
     _pd_2: PhantomData<PROPOSAL>,
-    _pd_3: PhantomData<NETWORK>,
-    _pd_4: PhantomData<STORAGE>,
-    _pd_5: PhantomData<ELECTION>,
+    _pd_3: PhantomData<VOTE>,
+    _pd_4: PhantomData<NETWORK>,
```
DieracDelta, line 579: nit:
```rust
  pd: PhantomData<(TYPES, LEAF, PROPOSAL, VOTE, STORAGE, ELECTION)>
```
might be simpler to parse

DieracDelta (approved):
LGTM!!

testing/src/lib.rs
shenkeyao: Yes! I updated all places where multiple `PhantomData` were used.

DieracDelta (approved):

PR #952, opened 2023-02-15
Improve docs

DieracDelta:
I noticed a bunch of warnings when generating docs. This PR fixes those warnings and adds a stricter check (checks on library items).

Related: we should do a documentation pass at some point. Some of these docs are super out of date (there's still stuff about BlockContents which was renamed ~8 months ago!)


elliedavidson (approved):
Great PR!

bfish713 (approved):

PR #955, opened 2023-02-17
Generic orchestrator

DieracDelta:
Rewrites the orchestrator to be generic over the various types in the same style of the validators. Renames and moves around files to match each validator binary with a orchestrator binary.

Nukes `src/bin/*` because really each orchestrator binary corresponds directly to a demo. Should work now for both election types so we no longer need separate vrf and static comittee files.


shenkeyao (approved):
LGTM!

PR #960, opened 2023-02-20
Refactor vote accumulator

shenkeyao:
* Moves vote-related implementations from `election.rs` and `message.rs` to a separate file.
* Renames `CertificateAccumulator` to `VoteAccumulator`.
* Adds the accumulator to `NodeImplementation`.

Closes https://github.com/EspressoSystems/HotShot/issues/933.


DieracDelta (dismissed):
LGTM!

types/src/vote.rs
```diff
@@ -0,0 +1,166 @@
...
+/// A vote on DA proposal.
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+#[serde(bound(deserialize = ""))]
+pub struct DAVote<TYPES: NodeType, LEAF: LeafType<NodeType = TYPES>> {
+    /// TODO we should remove this
+    /// this is correct, but highly inefficient
+    /// we should check a cache, and if that fails request the qc
+    pub justify_qc_commitment: Commitment<LEAF::QuorumCertificate>,
+    /// The signature share associated with this vote
+    /// TODO ct/vrf make ConsensusMessage generic over I instead of serializing to a [`Vec<u8>`]
```
nyospe, line 38: Yes, I'm looking forward to the end of the `Vec<u8>` type erasures.

types/src/vote.rs
```diff
@@ -0,0 +1,166 @@
...
+    pub signature: (EncodedPublicKey, EncodedSignature),
+    /// The block commitment being voted on.
+    pub block_commitment: Commitment<TYPES::BlockType>,
+    /// The view this vote was cast for
+    pub current_view: TYPES::Time,
+    /// The vote token generated by this replica
+    pub vote_token: TYPES::VoteTokenType,
+}
+
+/// A positive or negative vote on valiadting or commitment proposal.
```
nyospe, line 48: typo: s/valiadting/validating

types/src/vote.rs
```diff
@@ -0,0 +1,166 @@
...
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+#[serde(bound(deserialize = ""))]
+pub struct YesOrNoVote<TYPES: NodeType, LEAF: LeafType<NodeType = TYPES>> {
```
nyospe, line 51: Do we really want to have the type for `QuorumVote::Yes` and `QuorumVote::No` be the same? I've got a slightly uneasy feeling about the type safety aspect... and the fact that the signatures end up being the same means a leader might be able to cheat? At the very least, there should be a YES or NO constant added to the signature...

nyospe (approved):
LG, with a caveat about the `YesOrNoVote` and the possible vulnerability of the signature.

types/src/vote.rs
shenkeyao: The YES or NO constant is indeed added to the signature, by the enum `VoteData`. For example, if the replica wants to vote yes, they will create a signature by the following steps.
1. Constructing `VoteData::Yes(leaf_commitment)`
2. Signing on the above `VoteData`
3. Using the above signature to construct a `YesOrNoVote`
4. Sending a `QuorumVote::Yes` with the above `YesOrNoVote`

If a leader wants to cheat, they will have to redo step 2 to create a signature on `VoteData::No`, which isn't possible, so I think this is fine. Let me know if I miss anything!

nyospe (approved):
LGTM

PR #970, opened 2023-02-22
Rearrange testing infrastructure

DieracDelta:
Moves a bunch of `testing/tests/common/*` general infrastructure to `testing/src` since we're using it to generate test specifications. This also is needed to add a procedural macro (these can only live in a library with a cargo.toml flag enabled).

This also gets a documentation pass on clippy that isn't run on the testing/tests/* files (not sure why).

NOTE: this also nukes the `cross_all_types` and `cross_all_types_proptest` macros. They're being rewritten, so this shouldn't change much in the short term.


shenkeyao (approved):
LGTM!

PR #973, opened 2023-02-23
Rewrite spec generating declarative macro as procedural macro

### LINKED ISSUE #965: Add macro tests for sequencing consensus
Add a macro similar to `cross_all_types` but use sequencing consensus types.

DieracDelta:
This PR rewrites the `cross_tests` and `cross_all_types` macros from `hotshot-testing` and adds them in a separate crate containing a `cross_tests` procedural macro. This massively uncurses the code.

Note: I'm dropping the proptest tests for now. Copy-pasta-ing from zulip:

- We never run these tests (they're hidden under the --slow-tests feature flag)
- They don't build.
- I'm having issues porting the old tests over. They introduce custom tokens which are tricky to parse. And the quote macro doesn't play nicely with proptest!. I can't get their example readme of an empty test to compile from within the quote macro.

At some point in the future we should reintegrate proptest back in and add a GHA workflow to actually run the slow tests one every few weeks. I've opened a new issue for this: #1008 

Closes #965 

NOTE: The tests time out because we're cross-product-ing over more than just `MemoryNetwork`. As a result, significantly more tests are generated. Should I move the majority of the cross product tests to be guarded behind `--slow-tests`?


testing/Cargo.toml
```diff
@@ -7,8 +7,8 @@ authors = ["Victor Koenders <vkoenders@espressosys.com>"]
 
 [features]
 default = []
-full-ci = ["async-std-executor", "demo"]
-tokio-ci = ["tokio-executor", "demo"]
+full-ci = ["async-std-executor", "demo", "channel-async-std"]
```
nyospe, line 10: Just confirming, this is the crate that refers to, rather than being included in, the top level crate, correct?

DieracDelta: Yes!

nyospe (approved):
Looks good, and verified locally.

PR #989, opened 2023-02-27
Libp2p: 0.50.0 -> 0.51.0

DieracDelta:
There aren't any major changes between releases. There were some type renamings and deprecated functions, but that's all AFAICT.


elliedavidson (dismissed):
Looks good to me!

centralized_web_server/Cargo.toml
```diff
@@ -33,7 +33,7 @@ async-std = { version = "1.12", optional = true }
 async-trait = "0.1.59"
 bincode = "1.3.3"
 clap = { version = "4.0", features = ["derive", "env"], optional = false }
-commit = { git = "ssh://github.com/EspressoSystems/commit", tag = "0.2.0" }
+commit = { git = "https://github.com/EspressoSystems/commit", tag = "0.2.0" }
```
elliedavidson, line 36: Why do we need to change this?  Isn't doing this using `ssh` better?

orchestrator/Cargo.toml
```diff
@@ -32,7 +32,7 @@ async-std = { version = "1.12", optional = true }
 async-trait = "0.1.59"
 bincode = "1.3.3"
 clap = { version = "4.0", features = ["derive", "env"], optional = false }
-commit = { git = "ssh://github.com/EspressoSystems/commit", tag = "0.2.0" }
+commit = { git = "https://github.com/EspressoSystems/commit", tag = "0.2.0" }
```
elliedavidson, line 35: And here?

centralized_web_server/Cargo.toml
DieracDelta: I changed it b/c I was having errors fetching from ssh on my computer D: . Not sure why, I'll need to investigate.

We can definitely switch to `ssh`, but I think we should be consistent. E.g. I think we should either use ssh everywhere or https everywhere. The rest of the repo uses https for git links, so maybe we should switch everything over?

Even for commit it looks like there are occurrences using [https](https://github.com/EspressoSystems/HotShot/blob/main/centralized_server/Cargo.toml#L33) too so I think we're effectively fetching it twice?

(Why is using ssh better? Is it faster and/or more secure? I'm not familiar with the differences btwn the protocols.)

nyospe: If the repository is OSS, it should be https...

elliedavidson: Makes sense, @nyospe! We can use http

elliedavidson (approved):

centralized_web_server/Cargo.toml
nyospe: FYI, the recommendation is: if you're contributing to a git repo, clone via ssh, for both performance and convenience (use of your key pair with ssh)

For dependencies (e.g. what cargo does) it should always be https.

nyospe: If the dependency is private, (implying that the project is also private) you can use ssh, but this will need to be updated once the dependency is made public, and that can lead to some surprises, as above. It's probably better to use a url substitution in each dev's .gitconfig, with
```
[url "ssh://git@github.com/EspressoSystems/"]
    insteadOf = https://github.com/EspressoSystems/
```
... and then use the https version in all the Cargo.toml files.

Mind you, `commit` is already public, so that needs to be https already.

PR #990, opened 2023-02-27
Nuke State::ConsensusType

DieracDelta:
I think we established this isn't need in zulip, so here's a PR nuking it.


bfish713 (approved):

nyospe (approved):
Good. I feel like we created this relationship, in this (incorrect) direction, when we were trying to work out how to make the State morphology adapt to the needs of each consensus.

Removing the associated type is long overdue.

PR #1013, opened 2023-03-06
Ed/integrate orchestrator v2

elliedavidson:
This PR: 
* Removes the old centralized server in favor of the web server.  We always have the git copy if we need to pull it back in.
* Modifies `examples/infra` to work with the new orchestrator
  * Most of the Libp2p code stayed the same, but was moved around.  It would be great if @DieracDelta could check that it works.  
  * Renames `ClientConfig` to `Run`
  * Separates out the orchestrator functionality from the web server; all networking implementations use the same orchestrator interface
  * Fixes a bug with election initialization (it is always initialized to default for now since we do not pass the election config into the network config; before it would panic on a bad unwrap)
* Updates the example executable names in `Cargo.toml`
* Cleans up `init_state_and_hotshot`; it had redundant logging and unused variables
* Moves code around so that the `mod.rs` file is structured as follows: Orchestrator, Generic Validator, Libp2p, Web Server

NOTE: This PR does not fully support libp2p bootstrap nodes.  That will come in a later PR for this issue: https://github.com/EspressoSystems/HotShot/issues/999.  This PR does fully support running with the web server.


nyospe (commented):
LGTM, but I'll let @DieracDelta check that moving to the common impl didn't break anything for libp2p

Cargo.toml
```diff
@@ -80,34 +80,30 @@ channel-async-std = [
...
 required-features = ["demo", "libp2p/rsa"]
-path = "examples/libp2p/orchestrator.rs"
+path = "examples/web-server/web-server.rs"
 
 [[example]]
-name = "centralized-orchestrator"
+name = "web-server-validator"
 required-features = ["demo", "libp2p/rsa"]
-path = "examples/centralized/orchestrator.rs"
+path = "examples/web-server/validator.rs"
```
DieracDelta, line 106: Big fan of s/centralized-web-server/web-server

examples/infra/mod.rs
```diff
@@ -54,8 +58,281 @@ use libp2p::{
...
+    HotShot<TYPES::ConsensusType, TYPES, NODE>: ViewRunner<TYPES, NODE>,
+    Self: Sync,
+{
+    /// Initializes networking, returns self
+    async fn initialize_networking(
+        config: NetworkConfig<TYPES::SignatureKey, TYPES::ElectionConfigType>,
+    ) -> Self;
+
+    /// Initializes the genesis state and HotShot instance; does not start HotShot consensus
+    async fn initialize_state_and_hotshot(&self) -> (TYPES::StateType, HotShotHandle<TYPES, NODE>) {
```
DieracDelta, line 188: nit: might make sense to have a `# Panics` section in here. Looks like this only panics if we can't generate a genesis block, fail to init, or fail to get the anchored view?

examples/infra/mod.rs
```diff
@@ -54,8 +58,281 @@ use libp2p::{
...
+pub struct ValidatorArgs {
+    /// The address the orchestrator runs on
+    host: IpAddr,
+    /// The port the orchestrator runs on
+    port: u16,
+}
+
+/// Defines the behavior of a "run" of the network with a given configuration
+#[async_trait]
+pub trait Run<
```
DieracDelta, line 158: Fantastic! I'm a big fan of how this is the same for all implementations.

examples/infra/mod.rs
```diff
@@ -54,8 +58,281 @@ use libp2p::{
...
+    ) -> Self;
+
+    /// Initializes the genesis state and HotShot instance; does not start HotShot consensus
+    async fn initialize_state_and_hotshot(&self) -> (TYPES::StateType, HotShotHandle<TYPES, NODE>) {
+        let genesis_block = TYPES::BlockType::genesis();
+        let initializer =
+            hotshot::HotShotInitializer::<TYPES, ValidatingLeaf<TYPES>>::from_genesis(
+                genesis_block,
+            )
+            .unwrap();
```
DieracDelta, line 194: nit: 
```suggestion
            .expect("Couldn't generate genesis block");
```

examples/infra/mod.rs
```diff
@@ -54,8 +58,281 @@ use libp2p::{
...
+            network,
+            MemoryStorage::new(),
+            MEMBERSHIP::create_election(known_nodes, election_config),
+            initializer,
+            NoMetrics::new(),
+        )
+        .await
+        .expect("Could not initialize hotshot");
+
+        let state = hotshot.storage().get_anchored_view().await.unwrap().state;
```
DieracDelta, line 220: nit:
```suggestion
        let state = hotshot.storage().get_anchored_view().await.expect("Couldn't get HotShot's anchored view").state;
```

examples/infra/mod.rs
```diff
@@ -54,8 +58,281 @@ use libp2p::{
...
+            ..
+        } = self.get_config();
+
+        let size = mem::size_of::<TYPES::Transaction>();
+        let adjusted_padding = if padding < size { 0 } else { padding - size };
+        let mut txns: VecDeque<TYPES::Transaction> = VecDeque::new();
+        let state = hotshot.get_state().await;
+
+        // This assumes that no node will be a leader more than 5x the expected number of times they should be the leader
+        // FIXME  is this a reasonable assumption when we start doing DA?
```
DieracDelta, line 241: I'm not clear about this. For now: super reasonable assumption. Long term: I think we're sampling from a stake distribution, which might result in a leader owning much of the stake. 

How is this assumption being used?

examples/infra/mod.rs
```diff
@@ -54,8 +58,281 @@ use libp2p::{
...
+        // TODO ED: In the future we should have each node generate transactions every round to simulate a more realistic network
```
DieracDelta, line 242: +1! I think this is vital for future benchmarking.

examples/infra/mod.rs
```diff
@@ -342,296 +581,152 @@ impl<
...
-
-    async fn wait_for_ready(&self) {
-        CommunicationChannel::wait_for_ready(&self.get_network()).await;
+impl OrchestratorClient {
+    /// Creates the client that connects to the orchestrator
+    async fn connect_to_orchestrator(args: ValidatorArgs) -> Self {
+        let base_url = format!("{0}:{1}", args.host, args.port);
+        let base_url = format!("http://{base_url}").parse().unwrap();
+        let client = surf_disco::Client::<ClientError>::new(base_url);
+        // TODO ED: Add healthcheck wait here
```
DieracDelta, line 661: I'm curious, what would this entail?

examples/infra/mod.rs
```diff
@@ -342,296 +581,152 @@ impl<
...
-                0
-            };
-            error!("Submitting {} transactions", num_submitted);
-
-            // Start consensus
-            error!("  - Waiting for consensus to occur");
-            debug!("Waiting for consensus to occur");
+            .boxed()
+        };
+        let _result: Result<(), ClientError> =
```
DieracDelta, line 698: Is it okay for this to return an error?

examples/infra/mod.rs
```diff
@@ -649,519 +744,32 @@ pub async fn main_entry_point<
...
-        fn tag() -> String {
-            tag::ORCHESTRATOR_BLOCK.to_string()
-        }
-    }
 
-    impl Block for TestBlock {
-        type Error = TestError;
-        type Transaction = TestTransaction;
+    let run = RUN::initialize_networking(run_config.clone()).await;
+    let (_state, hotshot) = run.initialize_state_and_hotshot().await;
```
DieracDelta, line 768: Clean!

examples/web-server/types.rs
```diff
@@ -10,13 +12,15 @@ use hotshot_types::{
     vote::QuorumVote,
 };
 
-use crate::infra::CentralizedConfig;
+// use crate::infra::CentralizedConfig;
```
DieracDelta, line 15: I think this can be deleted

examples/web-server/types.rs
```diff
@@ -10,13 +12,15 @@ use hotshot_types::{
...
 pub type ThisMembership =
     GeneralStaticCommittee<VDemoTypes, ThisLeaf, <VDemoTypes as NodeType>::SignatureKey>;
-pub type ThisNetwork = CentralizedCommChannel<VDemoTypes, ThisProposal, ThisVote, ThisMembership>;
+pub type ThisNetwork =
+    CentralizedWebCommChannel<VDemoTypes, ThisProposal, ThisVote, ThisMembership>;
 pub type ThisProposal = ValidatingProposal<VDemoTypes, ThisLeaf>;
 pub type ThisVote = QuorumVote<VDemoTypes, ThisLeaf>;
 pub type ThisNode = VDemoNode<ThisNetwork, ThisMembership>;
-pub type ThisConfig = CentralizedConfig<VDemoTypes, ThisMembership>;
+// TODO ED Run makes more sense to me than Config
```
DieracDelta, line 25: Agreed, it more represents a complete `Run` rather than just the config.

DieracDelta (approved):
LGTM!

examples/infra/mod.rs
elliedavidson: Good catch!  This actually cannot return an error; I've updated the type to reflect this!

elliedavidson:
@DieracDelta I've addressed your comments!

DieracDelta (approved):

PR #1017, opened 2023-03-07
Add errors to `run_background_receive`

kmchicoine:
Added new errors and error handling for unexpected exits in `run_background_receive` in the web server


src/traits/networking/centralized_web_server_network.rs
```diff
@@ -285,6 +289,9 @@ impl<
                 vote_index = 0;
             }
         }
+        Err(NetworkError::CentralizedWebServer {
+            source: CentralizedWebServerNetworkError::ClientDisconnected,
+        })
```
DieracDelta, line 294: If this loop stops running, does that mean the Client is disconnected?  Wondering what the Client is in this case and what the difference is between this and the replica being shutdown (e.g. `NetworkError::ShutDown`)

DieracDelta (approved):
I have one conceptual question, but other than that LGTM!

src/traits/networking/centralized_web_server_network.rs
kmchicoine: Hmm good question- the only way for the polling loop to exit is for the `inner.running` boolean to become false, which happens in the ConnectedNetwork `shutdown()` function, so you're right: `NetworkError::ShutDown` would probably make more sense here

DieracDelta: Ah, gotcha! 👍

DieracDelta (approved):

elliedavidson (approved):

PR #1018, opened 2023-03-08
Example debug configs

DieracDelta:
Augments `gen_launch.py` to also generate examples debug configurations. Adds command to list examples to `justfile`. Adds generated examples debug configurations to `.vscode/launch.json`.


elliedavidson (approved):
LGTM!

PR #1025, opened 2023-03-10
Fix ViewRunner

shenkeyao:
- Adds replica task to `run_view`.
- Separate channel maps for the DA member and replica.
- To be merged into https://github.com/EspressoSystems/HotShot/pull/996.


nyospe (approved):
LGTM

PR #1041, opened 2023-03-16
Ed/orchestrator libp2p integration

elliedavidson:
This PR allows the new orchestrator to work with libp2p: 
* Nodes now must identity with the orchestrator before receiving the network config.  This identity is the IP address nodes will use for libp2p
* The orchestrator will automatically choose the first _n_ nodes to be bootstrap nodes.  The number of bootstrap nodes can now be specified in the libp2p configuration file
* There are now two options for port mapping, which can be specified in the configuration file: 
  * Option 1 will map ports to the libp2p base_port + node_index.  This is useful for local deployments where all nodes are running on the same IP address
  * Option 2 will not index libp2p ports by node_index, and will instead designate the libp2p base_port as the port for all nodes


examples/infra/mod.rs
```diff
@@ -450,15 +444,19 @@ where
...
+        let bound_addr: Multiaddr = format!(
             "/{}/{}/tcp/{}",
             if libp2p_config.public_ip.is_ipv4() {
                 "ip4"
             } else {
                 "ip6"
             },
             libp2p_config.public_ip,
-            libp2p_config.base_port + node_index as u16
+            libp2p_config.base_port as u64 + port_index
```
nyospe, line 459: Do we really want to allow an invalid port range like this? I'm assuming this is because of the lack of u16 support in the config specification?

nyospe (approved):
Looks good enough for now. A few loose threads that eventually need to be cleaned up, but none that are an issue when properly configured/used.

examples/infra/mod.rs
elliedavidson: I'll update this in the PR for the sequencing examples

Issue #1056, opened 2023-03-21
Audit allow lints


DieracDelta:
enable stricter lints everywhere we can: tests, examples, etc. Figure out which lints we care/don't care about (e.g. review the `allow` lints everywhere).

elliedavidson:
Audit all `allow` lints in the code

PR #1100, opened 2023-04-05
Remove a lot of #[allow] attributes and fix lints

jbearer:
Progress on #1056. There’s more to do but this is a good start. This does not apply @dieracdelta’s commit 4cac2885e38eebf0bdfa71c1b234c552438c7eae from #1055, which causes over 400 new errors. I'll work on that next.

Thoughts on the remaining #[allow]ed lints:
* The cast precision loss ones are probably ok, especially since they are mostly used for metrics. I think it is fine to continue allowing these, with localized #[allow] attributes as warning flags.  I got rid of all the cast-possible-wrap lints as wrapping is (1) easier to prevent, by clamping, and (2) potentially worse behavior as it can give wildly erroneous results instead of just rounded results.
* The deprecated lints we should try to get rid of, but it might be more involved than the scope of this PR
* too_many_arguments and too_many_lines are good heuristics for refactoring, but not urgent, so I think it’s fine to continue #[allow]ing those with localized exceptions.
* Panic in test code is fine
* type_complexity is in a similar category as too_many_arguments, in that it’s a light code smell but not an urgent problem. I’m leaving those alone because many if not all of them seem like they will be addressed after the current round of refactoring

### DIFF
--- a/examples/infra/mod.rs
+++ b/examples/infra/mod.rs
@@ -261,7 +261,7 @@ pub trait Run<
             sk,
             config.node_index,
             config.config,
-            MemoryStorage::new(),
+            MemoryStorage::empty(),
             quorum_exchange,
             committee_exchange,
             initializer,
--- a/examples/infra/mod.rs
+++ b/examples/infra/mod.rs
@@ -265,7 +265,7 @@ pub trait Run<
             quorum_exchange,
             committee_exchange,
             initializer,
-            NoMetrics::new(),
+            NoMetrics::boxed(),
         )
         .await
         .expect("Could not init hotshot");
@@ -566,7 +566,7 @@ where
 
         let node_config = config_builder.build().unwrap();
         let network = Libp2pNetwork::new(
-            NoMetrics::new(),
+            NoMetrics::boxed(),
             node_config,
             pubkey.clone(),
             Arc::new(RwLock::new(
--- a/examples/infra/mod.rs
+++ b/examples/infra/mod.rs
@@ -819,7 +819,6 @@ impl OrchestratorClient {
 
     /// Tells the orchestrator this validator is ready to start
     /// Blocks until the orchestrator indicates all nodes are ready to start
-    #[allow(clippy::let_unit_value)]
     async fn wait_for_all_nodes_ready(&self, node_index: u64) -> bool {
         let send_ready_f = |client: Client<ClientError>| {
             async move {
@@ -833,7 +832,8 @@ impl OrchestratorClient {
             }
             .boxed()
         };
-        () = self.wait_for_fn_from_orchestrator(send_ready_f).await;
+        self.wait_for_fn_from_orchestrator::<_, _, ()>(send_ready_f)
+            .await;
 
         let wait_for_all_nodes_ready_f = |client: Client<ClientError>| {
             async move { client.get("api/start").send().await }.boxed()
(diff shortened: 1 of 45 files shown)


examples/infra/mod.rs
```diff
@@ -833,7 +832,8 @@ impl OrchestratorClient {
             }
             .boxed()
         };
-        () = self.wait_for_fn_from_orchestrator(send_ready_f).await;
+        self.wait_for_fn_from_orchestrator::<_, _, ()>(send_ready_f)
+            .await;
```
elliedavidson, line 836: Nice!  I couldn't figure out to do this when I originally wrote it!

src/lib.rs
```diff
@@ -257,7 +250,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> HotShot<TYPES::ConsensusType
             high_qc: anchored_leaf.get_justify_qc(),
 
             metrics: Arc::new(ConsensusMetrics::new(
-                inner.metrics.subgroup("consensus".to_string()),
+                &*inner.metrics.subgroup("consensus".to_string()),
```
elliedavidson, line 253: For my own knowledge, why is this change better than the old version?

elliedavidson (approved):
LGTM!

shenkeyao (approved):
This is great!

src/lib.rs
jbearer: The old version consumed the `Box` in the callee, which prevents the caller from ever using it again. This is unnecessary because the callee doesn't actually take ownership of the `Box`, it just borrows and then drops it.

In this particular case it doesn't matter because the caller is simply creating a temporary `Box` and then dropping it when the callee is done. But it's good practice for APIs to take objects by reference if they don't require ownership of the object.

src/traits/election/vrf.rs
```diff
@@ -250,9 +250,7 @@ where
         let new_seed = *hasher.finalize().as_bytes();
         let mut prng = rand::rngs::StdRng::from_seed(new_seed);
 
-        #[allow(clippy::let_unit_value)]
-        let parameters = SIGSCHEME::param_gen(Some(&mut prng)).unwrap();
```
nyospe, line 254: So, uh... 1) this param_gen() apparently actually needed to be called for some `SignatureScheme` implementations (but not this one?) before calling `key_gen()`, in at least the version of jellyfish we were using when we wrote this, according to the documentation, but 2) this is basically dead code now. It was used for the sortition based VRF election scheme that turned out to be unsafe, and nothing uses this implementation of the election trait now. Even though we're likely to use a VRF based selection mechanism for DA committee membership in the future, it won't be this one.

src/traits/election/vrf.rs
```diff
@@ -1146,10 +1147,7 @@ impl ElectionConfig for VRFStakeTableConfig {}
 //         let stake_per_node = NonZeroU64::new(100).unwrap();
 //         let genesis_seed = [0u8; 32];
 //         for _i in 0..num_nodes {
-//             // TODO we should make this more general/use different parameters
-//             #[allow(clippy::let_unit_value)]
-//             let parameters = BLSSignatureScheme::<Param381>::param_gen(Some(rng)).unwrap();
-//             let (sk, pk) = BLSSignatureScheme::<Param381>::key_gen(&parameters, rng).unwrap();
+//             let (sk, pk) = BLSSignatureScheme::<Param381>::key_gen(&(), rng).unwrap();
```
nyospe, line 1150: Same as above.

nyospe (commented):
LGTM

PR #1101, opened 2023-04-05
Add stake table to webserver

kmchicoine:
With static committee, the orchestrator pushes a node's key to the webserver as nodes join the network.

### DIFF
--- a/web_server/src/lib.rs
+++ b/web_server/src/lib.rs
@@ -6,6 +6,8 @@ use clap::Args;
 use config::DEFAULT_WEB_SERVER_PORT;
 use futures::FutureExt;
 
+use hotshot_types::traits::signature_key::EncodedPublicKey;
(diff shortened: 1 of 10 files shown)


web_server/src/lib.rs
```diff
@@ -172,6 +181,20 @@ impl WebServerDataSource for WebServerState {
 
         Ok(())
     }
+
+    fn post_staketable(&mut self, key: Vec<u8>) -> Result<(), Error> {
+        // KALEY TODO: need security checks here
```
DieracDelta, line 186: Should we have an issue for this?

DieracDelta (approved):

elliedavidson (approved):

web_server/src/lib.rs
kmchicoine: Good call- https://github.com/EspressoSystems/HotShot/issues/1108

PR #1112, opened 2023-04-07
Fix bug where next leader channel is created with wrong view number.

jbearer:
The channel to send/receive votes is index by the current view number (the view number of the vote) not the next view number.

With this change, the sequencer tests get farther (previously the leader was never receiving votes, now it receives them but the replicas eventually get out of sync).

Having problems running the tests locally so I don't know if this actually works yet.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -991,7 +991,7 @@ where
         send_to_next_leader.cur_view += 1;
         let (send_commitment_vote_chan, recv_commitment_vote_chan) = {
             let vq = HotShot::<SequencingConsensus, TYPES, I>::create_or_obtain_chan_from_write(
-                cur_view + 1,
+                cur_view,
                 send_to_next_leader,
             )
             .await;


jbearer:
Tests pass, would like to know how the tests pass both with and without this change

jbearer:
I realized this is not obvious, so here's why I think this fix is correct:

[Here](https://github.com/EspressoSystems/HotShot/blob/main/src/lib.rs#L601) is the place where we assume the channel is indexed by current view number (note we use `msg_time` to get the channel compared to [`msg_time + 1`](https://github.com/EspressoSystems/HotShot/blob/main/src/lib.rs#L592) to check if we're the next leader). Also in `ValidatingConsensus` we [create the channel](https://github.com/EspressoSystems/HotShot/blob/main/src/lib.rs#L801-L804) indexed by the current view after using [the next view](https://github.com/EspressoSystems/HotShot/blob/main/src/lib.rs#L799) for the leader check. So I think this change just makes `SequencingConsensus` consistent with `ValidatingConsensus` and consistent with what `handle_direct_consensus_message` expects.

DieracDelta (approved):

shenkeyao (approved):

PR #1120, opened 2023-04-12
Use updated view number for decide events

jbearer:


bfish713:
I'm closing this because it only touches dead code

PR #1128, opened 2023-04-14
Ed/view sync 1

elliedavidson:
Here is the first chunk of view sync work: the basic update to use Yes/No votes and form Yes/No QCs.  This PR: 
* Actually uses the No votes instead of throwing them away
* Does not include updates to form Timeout QCs
* Makes `VoteData` generic over something that is Committable instead of over LEAF and TYPES.  
* Does not implement the Yes/No voting for DA yet. 
* Does not have testing (will wait until the updates to the testing harness are done) 
* Labels outstanding items with `TODO ED`
* Adds `VoteData` to signatures in QCs - there is likely a more optimized way to do this, but I want to keep it simple for now.  

This PR only contains a subset of the view sync work to date, but I want to get the work in piece by piece and get feedback on the design.  Based on this PR's feedback I may go back and update the stuff I am currently working on.  Although the view sync work isn't complete in this PR, I think it is best to merge it in incrementally instead of all at the end.  The tests should all pass.  

Note: I'd really like feedback on the implementation of the vote accumulator logic as well as the extra trait bounds I added.  I feel like there is probably a more optimized way to do both of those things.


types/src/traits/election.rs
```diff
@@ -235,8 +243,13 @@ pub trait ConsensusExchange<TYPES: NodeType, LEAF: LeafType<NodeType = TYPES>, M
     }
 
     /// Threshold required to approve a [`Proposal`](Self::Proposal).
-    fn threshold(&self) -> NonZeroU64 {
-        self.membership().threshold()
+    fn success_threshold(&self) -> NonZeroU64 {
```
nyospe, line 246: This is an interesting change. Does that mean we assume all exchanges support yes/no votes?

elliedavidson: I figured we'd want to in the future, like how we want to allow Yes No votes for DA.  What do you think?

nyospe: I believe we have cases where we won't want to use No votes - including the DA certificate - but a) I don't think it does any harm, and b) in the best case, it'll allow speeding up the next round of DA committee size expansions.

nyospe (approved):
Excellent, I think this is ready to go.

Issue #1071, opened 2023-03-22
Refactor `NodeImplementation`


shenkeyao:
Refactor `NodeImplementation` with `Exchanges` type, which is either a `ValidatingExchanges` or a `SequencingExchanges`. Then remove the `QuorumExchange` and `CommitteeExchange` from the trait.

PR #1133, opened 2023-04-18
Refactor `NodeImplementation` and `Message` related types

shenkeyao:
- Refactors `NodeImplementation` with `Exchanges` type and remove `CommitteeExchange` and `QuorumExchange` from it.
- Refactors `Message` related types for validating and sequencing consensus.
- Refactors `HotShot` to have different implementations depending on the consensus type.
- Updates leader, DA member, and replica logic.
- Updates tests and examples.

Closes https://github.com/EspressoSystems/HotShot/issues/1071.


consensus/src/traits.rs
```diff
@@ -146,10 +126,80 @@ pub trait ConsensusApi<
...
+        VOTE: VoteType<TYPES>,
+    >(
+        &self,
+        message: ValidatingMessage<TYPES, I>,
+    ) -> std::result::Result<(), NetworkError>;
+}
+
+/// The API that [`HotStuff`] needs to talk to the system for sequencing applications. This should be implemented in the `hotshot` crate and passed to all functions on `HotStuff`.
+///
+/// [`HotStuff`]: struct.HotStuff.html
```
DieracDelta, line 165: What does `struct.HotStuff.html` refer to?

src/lib.rs
```diff
@@ -150,6 +157,12 @@ pub struct HotShot<CONSENSUS: ConsensusType, TYPES: NodeType, I: NodeImplementat
     /// The hotstuff implementation
     hotstuff: Arc<RwLock<Consensus<TYPES, I::Leaf>>>,
 
+    // TODO: use associated traits to pair these maps with respective exchanges
+    // e.g. per exchange
+    // struct ChannelMaps {
+    //    proposal_channel:Arc<RwLock<SendToTasks<TYPES, I>>>,
+    //    votes_channel<:Arc<RwLock<SendToTasks<TYPES, I>>>,
+    // }
```
DieracDelta, line 165: Should we open an issue for this?

src/lib.rs
```diff
@@ -463,41 +452,289 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> HotShot<TYPES::ConsensusType
...
+                    .await;
+
+                if new {
+                    // If this is a new transaction, update metrics.
+                    let consensus = self.hotstuff().read().await;
+                    consensus.metrics.outstanding_transactions.update(1);
+                    consensus
+                        .metrics
+                        .outstanding_transactions_memory_size
+                        .update(i64::try_from(size).unwrap_or(i64::MAX));
```
DieracDelta, line 595: Should this warn if the conversion fails? Might be useful in the case of overflow.

src/lib.rs
```diff
@@ -463,41 +452,289 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> HotShot<TYPES::ConsensusType
...
+                shut_down.clone(),
+                TransmitType::Direct,
+                exchange.clone().into(),
+            )
+            .instrument(info_span!("HotShot Direct Task",)),
+        );
+
+        async_spawn(
+            tasks::network_lookup_task(self.clone(), shut_down.clone())
+                .instrument(info_span!("HotShot network lookup task",)),
```
DieracDelta, line 674: nit
```suggestion
                .instrument(info_span!("HotShot Network Lookup Task",)),
```

src/lib.rs
```diff
@@ -608,138 +808,328 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> HotShot<TYPES::ConsensusType
...
-                // Log exceptional situation and proceed
-                warn!(?msg, "Broadcast message received over direct channel");
-            }
-        }
-    }
+        msg: SequencingMessage<TYPES, I>,
+        sender: TYPES::SignatureKey,
+    ) {
+        // TODO validate incoming data message based on sender signature key
+        // <github.com/ExpressoSystems/HotShot/issues/418>
```
DieracDelta, line 963: The link is broken (`EspressoSystems`). This got closed, I think we're doing this? Maybe?

src/lib.rs
```diff
@@ -608,138 +808,328 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> HotShot<TYPES::ConsensusType
...
+                            .committee_exchange()
+                            .is_leader(msg_time);
+                        if !is_leader || msg_time < channel_map.cur_view {
+                            warn!(
+                                "Throwing away VoteType<TYPES>message for view number: {:?}, Channel cur view: {:?}",
+                                msg_time,
+                                channel_map.cur_view,
+                            );
+                            return;
+                        }
```
DieracDelta, line 1114: This is very similar to line 1066-1078. I wonder if we can separate this out into a function to avoid duplication.

DieracDelta, line 1114: It might be worth using [this](https://doc.rust-lang.org/std/any/fn.type_name.html) to print the type name instead of `VoteType<TYPES>`. Maybe post monomorphized types would be more informative? Or maybe printing both?

src/lib.rs
```diff
@@ -1154,14 +1555,154 @@ where
...
+                        _phantom: PhantomData,
+                    },
+                    recipient,
+                )
+                .await
+        });
+        Ok(())
+    }
+
+    // TODO remove and use exchange directly.
```
DieracDelta, line 1668: Do we have an issue for this?

src/lib.rs
```diff
@@ -1247,75 +1834,52 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>>
...
+            .quorum_exchange()
             .network()
             .broadcast_message(
                 Message {
                     sender: self.inner.public_key.clone(),
-                    kind: message.into(),
+                    kind: MessageKind::from_consensus_message(message),
+                    _phantom: PhantomData,
                 },
                 // TODO this is morally wrong!
```
DieracDelta, line 1850: I think I wrote this comment. No idea why I thought this was morally wrong, but it's probably fine and worth removing the comment?

src/traits/networking/centralized_server_network.rs
```diff
@@ -1021,6 +1023,32 @@ where
...
+where
+    TYPES::SignatureKey: TestableSignatureKey,
+    MessageKind<TYPES::ConsensusType, TYPES, I>: ViewMessage<TYPES>,
+{
+    fn generator(
+        expected_node_count: usize,
+        num_bootstrap: usize,
+        network_id: usize,
+    ) -> Box<dyn Fn(u64) -> Self + 'static> {
+        let generator=<CentralizedServerNetwork<TYPES::SignatureKey, TYPES::ElectionConfigType> as TestableNetworkingImplementation<TYPES, Message<TYPES, I>>>::generator(expected_node_count, num_bootstrap, network_id);
```
DieracDelta, line 1043: I'm surprised cargo fmt was okay with no spaces between `generator` and `<`. Might be worth trying to format only this file.

DieracDelta (dismissed):
This is FANTASTIC!! Much cleaner types by the looks of things. Big fan, excited to see this merged! I only have a couple of small nits.

kmchicoine (approved):
Agree with Justin- much easier to understand/follow with the new types!

consensus/src/traits.rs
shenkeyao: This was copied from an outdated comment. I'll update it!

src/lib.rs
shenkeyao: Yes! I just created an [issue](https://github.com/EspressoSystems/HotShot/issues/1193) and will add a link here.

src/lib.rs
shenkeyao: Also found some other places that needed capitalization. I'll update them all.

src/lib.rs
shenkeyao: Nope, and the description here isn't very accurate. I created an [issue](https://github.com/EspressoSystems/HotShot/issues/1194).

src/lib.rs
shenkeyao: Yeah I think we could remove it!

src/traits/networking/centralized_server_network.rs
shenkeyao: Good catch! Auto-formating doesn't seem to work for complicated types like this. I'll manually fix this one. I've manually formatted some other places but missed this one. 😅

src/lib.rs
shenkeyao: They are slightly different in the exchange type. We are calling `committee_exchange()` here but `quorum_exchange()` in the other place, which I think makes it a bit harder to extract the common parts. I'll leave it like this for now, but we could refactor it if we have too many duplications like this in the future!

DieracDelta (approved):

Issue #1062, opened 2023-03-22
Share resources between `CommunicationChannels`


bfish713:
- ConnectedNetwork is always created when we create a `CommunicationChannel`
- Most times we only want one connected network to be shared between all channels that want to use it.

PR #1140, opened 2023-04-20
Share `ConnectedNetwork` b/t `CommunicationChannel`s

bfish713:
This PR does 3 related things

1. Makes the `CommunicationChannel` accept and `Arc` for the network instead of the object allowing us to share one instance
2. Moves `TestableNetworkImplementation` to be on the `ConnectedNetwork` allowing us to generate a `ConnectedNetwork` Directly from the test launcher
3. Removes the second Message generic from `ConnectedNetwork` and makes it so `WebSeverNetwork` doesn't need two message type generics

The test parts are a bit awkward without having `ConsensusExchanges` completed yet.  Basically the tests have to create one ConnectedNetwork type but that type isn't readily exposed at the `TestableNodeImplementation` level and only gotten through the ConsensusExchange's communication channel type.  This is why we have the super complex type def called `NetworkTypes` in node_implementation.rs.  Also it means we have to introduce the restriction in test that the `QuorumExchange`'s associated `CommunicationChannel`'s associated `ConnectedNetwork` equals the  `CommitteeExchange`'s associated `CommunicationChannel`'s associated `ConnectedNetwork`.  In other words if the Quorum's underlying network is LibP2p the Committee's must also be LibP2p for the test.  This is fine because that's how all the tests were working anyhow and it doesn't matter for the Validating tests (b/c Committee doesn't get used)

This can get fixed once we have `ConsensusExchanges` implemented because we can delegate the `ConnectedNetwork` and `CommunicationChannel` creation to a `TestableConsensusExchanges` trait.

Closes: #1062

### DIFF
--- a/types/src/traits/node_implementation.rs
+++ b/types/src/traits/node_implementation.rs
@@ -10,8 +10,7 @@ use super::{
     block_contents::Transaction,
     election::{ConsensusExchange, ElectionConfig, VoteToken},
     network::{
-        CommunicationChannel, ConnectedNetwork, TestableChannelImplementation,
-        TestableNetworkingImplementation,
+        CommunicationChannel, TestableChannelImplementation, TestableNetworkingImplementation,
     },
     signature_key::TestableSignatureKey,
     state::{ConsensusTime, ConsensusType, TestableBlock, TestableState},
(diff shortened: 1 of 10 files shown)


elliedavidson (approved):
LGTM!  The message purpose a nice addition.

types/src/traits/node_implementation.rs
```diff
@@ -301,7 +303,32 @@ pub type CommitteeMembership<TYPES, I> =
         <I as NodeImplementation<TYPES>>::Leaf,
         Message<TYPES, I>,
     >>::Membership;
-
+// TODO remove this really ugly type once we have `ConsensusExchanges` impl and can use that to do all generation
+/// Type for the underlying `ConnectedNetwork` that will be shared (for now) b/t Communication Channels
```
shenkeyao, line 307: With the `ConsensusExcahnges`, I added types that are similarly ugly, just shorter than this. 😅

shenkeyao (approved):

Issue #1066, opened 2023-03-22
Create `CommunicationChannel` for Centralized Server with Libp2p fallback


bfish713:
This channel should have a `ConnectedNetwork` handle for both the Centralized Server and Libp2p and send/receive messages on both. 

If a message is received on the faster channel (Centralized) we should cancel the libp2p gossip or ignore that message if we receive it again on the slower channel

PR #1157, opened 2023-04-28
Create CommunicationChannel for WebServer + Libp2p

bfish713:
Very basic implementation with the webserver being the primary and the libp2p being the backup.

Most of the work was getting this all to play nicely with the testing stuff.  Because I couldn't implement it generically means I can't use MemoryNetworks to force certain failure modes.  This means the test is more or less testing that things work with the webserver as there should not be any errors/fallback happening.

There are for sure improvements to the logic that we can make, specifically around cancelling broadcasts, and handling duplicated messages.  I want to try to get some more failure tests so I can really make sure the logic works

Closes: #1066

### DIFF
--- a/src/traits/networking/web_sever_libp2p_fallback.rs
+++ b/src/traits/networking/web_sever_libp2p_fallback.rs
@@ -1,7 +1,6 @@
-use super::{
-    memory_network::MemoryNetwork, FailedToSerializeSnafu, NetworkError, NetworkReliability,
-    NetworkingMetrics,
-};
+use super::NetworkError;
+use crate::traits::implementations::Libp2pNetwork;
+use crate::traits::implementations::WebServerNetwork;
 use crate::NodeImplementation;
 use async_compatibility_layer::{
     art::{async_sleep, async_spawn},
@@ -11,6 +10,7 @@ use async_lock::{Mutex, RwLock};
 use async_trait::async_trait;
 use bincode::Options;
 use dashmap::DashMap;
+use futures::join;
 use futures::StreamExt;
 use hotshot_types::traits::network::TestableChannelImplementation;
 use hotshot_types::traits::network::ViewMessage;
@@ -41,10 +41,7 @@ use std::{
         Arc,
     },
 };
-use crate::traits::implementations::WebServerNetwork;
-use crate::traits::implementations::Libp2pNetwork;
 use tracing::{debug, error, info, info_span, instrument, trace, warn, Instrument};
-
 /// A communication channel with 2 networks, where we can fall back to the slower network if the
 /// primary fails
 #[derive(Clone)]
@@ -99,7 +96,7 @@ impl<
 
     pub fn network(
         &self,
-    ) -> WebServerNetwork<
+    ) -> &WebServerNetwork<
         Message<TYPES, I>,
         TYPES::SignatureKey,
         TYPES::ElectionConfigType,
@@ -107,10 +104,10 @@ impl<
         PROPOSAL,
         VOTE,
     > {
-        self.networks.0
+        &self.networks.0
     }
-    pub fn fallback(&self) -> Libp2pNetwork<Message<TYPES, I>, TYPES::SignatureKey> {
-        self.networks.1
+    pub fn fallback(&self) -> &Libp2pNetwork<Message<TYPES, I>, TYPES::SignatureKey> {
+        &self.networks.1
     }
 }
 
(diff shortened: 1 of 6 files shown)


src/traits/networking/web_sever_libp2p_fallback.rs
```diff
@@ -0,0 +1,254 @@
...
+    ) -> Result<(), NetworkError> {
+        let recipients =
+            <MEMBERSHIP as Membership<TYPES>>::get_committee(election, message.get_view_number());
+        let fallback = self
+            .fallback()
+            .broadcast_message(message.clone(), recipients.clone());
+        let network = self.network().broadcast_message(message, recipients);
+        match join!(fallback, network) {
+            (Err(e), Err(_)) => Err(e),
+            _ => Ok(()),
```
DieracDelta, line 188: nit: It might be worthwhile to log if one of the networks is down

src/traits/networking/web_sever_libp2p_fallback.rs
```diff
@@ -0,0 +1,254 @@
...
+        &self,
+        message: Message<TYPES, I>,
+        election: &MEMBERSHIP,
+    ) -> Result<(), NetworkError> {
+        let recipients =
+            <MEMBERSHIP as Membership<TYPES>>::get_committee(election, message.get_view_number());
+        let fallback = self
+            .fallback()
+            .broadcast_message(message.clone(), recipients.clone());
+        let network = self.network().broadcast_message(message, recipients);
```
DieracDelta, line 185: Is the reason for broadcasting both to prevent censorship?

The alternative is to try the primary network, and if that fails to then broadcast to the secondary network. Not sure if there are soundness issues here though.

src/traits/networking/web_sever_libp2p_fallback.rs
```diff
@@ -0,0 +1,254 @@
...
+        match self.network().recv_msgs(transmit_type).await {
+            Ok(msgs) => Ok(msgs),
+            Err(_e) => self.fallback().recv_msgs(transmit_type).await,
+        }
+    }
+
+    async fn lookup_node(&self, pk: TYPES::SignatureKey) -> Result<(), NetworkError> {
+        match self.network().lookup_node(pk.clone()).await {
+            Ok(msgs) => Ok(msgs),
+            Err(_e) => self.fallback().lookup_node(pk).await,
```
DieracDelta, line 220: I think we may want to do this regardless for libp2p. It does use more bandwith but it also keeps our routing table up to date. This is useful if we ever store something on the DHT.

src/traits/networking/web_sever_libp2p_fallback.rs
```diff
@@ -0,0 +1,254 @@
...
+        message: Message<TYPES, I>,
+        recipient: TYPES::SignatureKey,
+    ) -> Result<(), NetworkError> {
+        match self
+            .network()
+            .direct_message(message.clone(), recipient.clone())
+            .await
+        {
+            Ok(_) => Ok(()),
+            Err(_e) => self.fallback().direct_message(message, recipient).await,
```
DieracDelta, line 203: Might also be worthwhile logging on the sad path here. `info` or `warn` level or something

src/traits/networking/web_sever_libp2p_fallback.rs
```diff
@@ -0,0 +1,254 @@
...
+        }
+    }
+
+    async fn recv_msgs(
+        &self,
+        transmit_type: TransmitType,
+    ) -> Result<Vec<Message<TYPES, I>>, NetworkError> {
+        match self.network().recv_msgs(transmit_type).await {
+            Ok(msgs) => Ok(msgs),
+            Err(_e) => self.fallback().recv_msgs(transmit_type).await,
```
DieracDelta, line 213: Probably something to push to a future sprint, but we may want to try listening on both channels with a timeout in case the centralized server is down for other nodes.

DieracDelta (approved):
LGTM, nice!

src/traits/networking/web_sever_libp2p_fallback.rs
bfish713: It's for the case where other nodes can't reach the server for some reason.  Also helps with censorship.

src/traits/networking/web_sever_libp2p_fallback.rs
bfish713: That's a good point, I think this one should just do both, especially since it doesn't return anything

src/traits/networking/web_sever_libp2p_fallback.rs
bfish713: I've been thinking a bunch on this one, I think we should probably should be clearing both queues and on each call here.  Otherwise the backup queue will just keep growing, and we won't really know if we missed something.  

I'm actually going to tackle this one this sprint, probably next.  This needs something a bit more sophisticated

DieracDelta (approved):

PR #1164, opened 2023-05-01
Upgrade libp2p

DieracDelta:


elliedavidson (approved):

kmchicoine (approved):
LGTM!

libp2p-networking/src/network/mod.rs
```diff
@@ -223,32 +222,72 @@ pub async fn gen_transport(
...
     Ok(transport
-        .upgrade(upgrade::Version::V1)
-        // authentication: messages are signed
-        .authenticate(noise::NoiseConfig::xx(noise_keys).into_authenticated())
-        // muxxing streams
-        // useful because only one connection opened
-        // with multiple substreams
-        // https://docs.libp2p.io/concepts/stream-multiplexing/
+        .upgrade(Version::V1)
+        .authenticate(libp2p_noise::Config::new(&identity).unwrap())
```
nyospe, line 233: This is interesting. So the authentication model for this changed?

libp2p-networking/src/network/mod.rs
```diff
@@ -223,32 +222,72 @@ pub async fn gen_transport(
...
         .multiplex(multiplexing_config)
-        .timeout(Duration::from_secs(20))
-        .map_err(|err| io::Error::new(io::ErrorKind::Other, err))
+        .timeout(std::time::Duration::from_secs(20))
         .boxed())
 }
+// pub async fn development_transport(
```
nyospe, line 238: Adding an entirely commented out block?

nyospe (commented):
What's the reason for adding the commented out code?

libp2p-networking/src/network/mod.rs
DieracDelta: This is a mistake! I'll update it in the test PR

libp2p-networking/src/network/mod.rs
DieracDelta: I don't think it necessarily changed. Rather, XX became the default and they deprecated the other handshakes. I was manually specifying XX because I was afraid the default would change and I wouldn't notice. https://github.com/libp2p/rust-libp2p/blob/master/transports/noise/CHANGELOG.md

PR #1166, opened 2023-05-02
Commit blocks besides genesis

DieracDelta:
This changes the signature on `next_block` such that it may be called without a previous state. This lets us commit blocks besides genesis. I also renamed `StateCommitmentType` since it either `State` or `()`, but never actually a commitment.

### DIFF
--- a/types/src/traits/storage.rs
+++ b/types/src/traits/storage.rs
@@ -134,7 +134,7 @@ pub struct StoredView<TYPES: NodeType, LEAF: LeafType<NodeType = TYPES>> {
     /// The justify QC of this view. See the hotstuff paper for more information on this.
     pub justify_qc: QuorumCertificate<TYPES, LEAF>,
     /// The state of this view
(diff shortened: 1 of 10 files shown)


types/src/traits/storage.rs
```diff
@@ -134,7 +134,7 @@ pub struct StoredView<TYPES: NodeType, LEAF: LeafType<NodeType = TYPES>> {
     /// The justify QC of this view. See the hotstuff paper for more information on this.
     pub justify_qc: QuorumCertificate<TYPES, LEAF>,
     /// The state of this view
-    pub state: LEAF::StateCommitmentType,
+    pub state: LEAF::MaybeState,
```
bfish713, line 137: 👍

bfish713 (approved):
LGTM! MaybeState is a good name for the associated type

shenkeyao (approved):
Looks good!

Issue #1193, opened 2023-05-15
Pair channel maps with respective exchanges


shenkeyao:
The `HotShot` struct currently has 4 channel maps, for two types of leaders, DA members, and replicas. However, not all maps are needed for validating consensus. We could use associated traits to pair these maps with respective exchanges.

E.g.. per exchange
```
struct ChannelMaps {
  proposal_channel:Arc<RwLock<SendToTasks<TYPES, I>>>,
  votes_channel<:Arc<RwLock<SendToTasks<TYPES, I>>>,
}
```

PR #1241, opened 2023-06-02
Pair channel maps with respective exchanges

shenkeyao:
- Creates `ChannelMaps` struct corresponding to an exchange.
- Adds a function to `NodeImplementation` to create `ChannelMaps`(s) depending on the consensus type.
- Replaces all `*_channel_map` fields of `HotShot` with the channels created by the associated `NodeImplementation`.
- Updates `NodeImlementation` impls and macros.

Closes https://github.com/EspressoSystems/HotShot/issues/1193.


bfish713 (approved):
This is a good change if we need the maps for run view.

PR #1245, opened 2023-06-06
Ed/small da committees v2

elliedavidson:
This PR: 
* Allows HotShot to work properly with small DA committees; it does this by allowing CommitteeElection types to have a completely separate election configuration from the QuorumElection type.  
* Adds an `is_da` flag to the test network generator and web server.  The web server needs to know if it is running in DA mode so non-DA committee nodes do waste bandwidth polling for DA proposals (this optimization is still a work in progress)
* Updates the StaticCommittee election type to actually verify if nodes are part of the committee, instead of assuming all nodes are always part of the committee.


kmchicoine (approved):
LGTM!

PR #1252, opened 2023-06-07
Add Debug trait bound

sveitser:
This avoids having to specify this bound in the sequencer repo. If we
specify the bounds manually we get a trait requirement overflow (E0275).

Pending some discussion if this is really required / the right approach.

### DIFF
--- a/libp2p-networking/src/network/mod.rs
+++ b/libp2p-networking/src/network/mod.rs
@@ -239,7 +239,7 @@ pub async fn gen_transport(
 /// a single node, connects them to each other
 /// and waits for connections to propagate to all nodes.
 #[instrument]
-pub async fn spin_up_swarm<S: std::fmt::Debug + Default>(
+pub async fn spin_up_swarm<S: Debug + Default>(
     timeout_len: Duration,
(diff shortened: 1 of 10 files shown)


DieracDelta:
LGTM this seems like the right approach. It's not that big of an imposition on the implementer, who can do a dummy implementation if there is no debug information.

elliedavidson (commented):
Small nit: `std::fmt::Debug` is already imported (or should be) at the top of the files, so we can just use `Debug` here.

sveitser:
> Small nit: `std::fmt::Debug` is already imported (or should be) at the top of the files, so we can just use `Debug` here.

In network.rs it's not imported and used as `std::fmt::Debug` a few times. Can happily change that.

elliedavidson:
> > Small nit: `std::fmt::Debug` is already imported (or should be) at the top of the files, so we can just use `Debug` here.
> 
> In network.rs it's not imported and used as `std::fmt::Debug` a few times. Can happily change that.

Sorry, I only checked election.rs.  It would be great if you updated it, but since it's not already imported in network.rs that's fine.

elliedavidson (approved):
Wonderful!

PR #1255, opened 2023-06-08
Simplify poll

### LINKED ISSUE #1211: Remove pin_project from task abstraction
The lack of this is making `HST::poll` very repetitive and long. It should be possible to do without, since there are only `Option<FUT>`, and no actual `FUT` held. [Option](https://doc.rust-lang.org/std/marker/trait.Unpin.html#impl-Unpin-for-Option%3CT%3E) implements `Unpin`.

DieracDelta:
Utilize pin project to simplify the Future::poll implementation of the hotshot task. It's still not *great* because the logic for the message stream vs event stream is identical, but it's definitely easier than before to grokk. I also caught a few bugs while doing this (there was a mismatched if-else statement in a couple places).

Closes #1211

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -48,6 +48,7 @@ use async_compatibility_layer::{
 };
 use async_lock::{Mutex, RwLock, RwLockUpgradableReadGuard, RwLockWriteGuard};
 use async_trait::async_trait;
+use nll::nll_todo::nll_todo;
 use bincode::Options;
 use commit::{Commitment, Committable};
 use either::{Left, Right};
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -708,7 +709,7 @@ where
             network_direct_task_handle,
             committee_network_broadcast_task_handle: None,
             committee_network_direct_task_handle: None,
```
shenkeyao, line 711: These four network task handles are also `nll_todo()`, similar to `consensus_task_handle`.

task/src/task.rs
shenkeyao, line 1: Good simplification in this file!

shenkeyao (commented):
Looks good! Only a minor comment.

Issue #1295, opened 2023-06-22
Integrate network task into mod.rs


shenkeyao:
`src/tasks/mod.rs` is currently using dummy networking task types. We should update it with new types in `task-impls/src/network.rs`, and replace `network_task` calls in `src/lib.rs`.

This is a follow-up task for https://github.com/EspressoSystems/HotShot/issues/1180.

PR #1304, opened 2023-06-26
Integrate network task implementations

shenkeyao:
- Modifies `add_network_task` function to construct `NetworkTaskState`, create message and event streams, and register all of them.
- Updates `view_runner` and `spawn_all`.
- Adds message handler to `NetworkTaskState` and removes the `run` function.
- Removes the `Success` variant from the `HotShotTaskCompleted` enum.

Closes https://github.com/EspressoSystems/HotShot/issues/1295.


task-impls/src/network.rs
```diff
@@ -37,11 +35,10 @@ pub struct NetworkTaskState<
...
 > {
-    channel: COMMCHANNEL,
-    events: UnboundedStream<SequencingHotShotEvent<TYPES, I>>,
-    event_stream: ChannelStream<SequencingHotShotEvent<TYPES, I>>,
-    view: ViewNumber,
-    phantom: PhantomData<(TYPES, Message<TYPES, I>, PROPOSAL, VOTE, MEMBERSHIP)>,
+    pub channel: COMMCHANNEL,
+    pub event_stream: ChannelStream<SequencingHotShotEvent<TYPES, I>>,
+    pub view: ViewNumber,
+    pub phantom: PhantomData<(TYPES, Message<TYPES, I>, PROPOSAL, VOTE, MEMBERSHIP)>,
```
DieracDelta, line 41: I wonder if this can be simplified slightly:
```suggestion
    pub phantom: PhantomData<(PROPOSAL, VOTE, MEMBERSHIP)>,
```

task-impls/src/network.rs
```diff
@@ -72,11 +69,44 @@ impl<
...
+                    CommitteeConsensusMessage::DAProposal(proposal) => {
+                        SequencingHotShotEvent::DAProposalRecv(proposal.clone(), proposal.signature)
+                    }
+                    CommitteeConsensusMessage::DAVote(vote) => {
+                        SequencingHotShotEvent::DAVoteRecv(vote.clone(), vote.signature.1)
+                    }
+                },
+            },
+            MessageKind::Data(_) => {
+                panic!("Got unexpected message type in network task!");
```
DieracDelta, line 98: It might be better to log an error here and return `None`. Otherwise the adversary could send a data message and cause the entire networking task to fail.

src/tasks/mod.rs
```diff
@@ -382,44 +332,100 @@ pub type ViewSyncTaskTypes =
...
+        },
+    ));
+    let network_event_handler = HandleEvent(Arc::new(
+        move |event, mut state: NetworkTaskState<_, _, _, _, MEMBERSHIP, _>| {
+            async move {
+                if let SequencingHotShotEvent::Shutdown = event {
+                    (Some(HotShotTaskCompleted::ShutDown), state)
+                } else {
+                    state.handle_event(event, exchange.membership()).await;
+                    (None, state)
```
DieracDelta, line 395: Would it make sense to check the return value in case it returns `true`? (Can that happen?)

task/src/task_impls.rs
```diff
@@ -208,9 +208,9 @@ impl<
         STATE: TS,
     > HotShotTaskTypes for HSTWithEventAndMessage<ERR, EVENT, ESTREAM, MSG, MSTREAM, STATE>
 {
-    type Event = ();
+    type Event = EVENT;
     type State = STATE;
-    type EventStream = DummyStream;
+    type EventStream = ESTREAM;
```
DieracDelta, line 213: Oops! Good catch, this is definitely a typo.

DieracDelta (approved):
LGTM! We'll still need to hook in the message stream, but I think that can be done in a separate PR.

task-impls/src/network.rs
shenkeyao: I changed it to `warn!` but didn't make it return `None`, since this function doesn't have a return value. Let me know if you think it's better to return something!

task-impls/src/network.rs
```diff
@@ -72,11 +69,44 @@ impl<
...
+            }
+            MessageKind::_Unreachable(_) => unimplemented!(),
```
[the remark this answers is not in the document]
shenkeyao: I think this is fine for the `Unreachable` case since the adversary won't be able to send an `Unreachable` message, right?

src/tasks/mod.rs
shenkeyao: Actually the return value of `handle_event` is redundant--it returns `true` only if the event is `Shutdown`, which is already checked by the caller of `handle_event`. I'll remove the return type.

task-impls/src/network.rs
DieracDelta: Oops! I confused this with the message handler type. Yes, this makes sense!

task-impls/src/network.rs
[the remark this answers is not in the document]
DieracDelta: yes this makes sense! It looks like we wrap the messages explicitly, so this should never show up.

PR #1309, opened 2023-06-28
Generator stream

DieracDelta:
Add a stream created from a generator. Useful in multiple places in the run view refactor


shenkeyao (approved):
LGTM!

Issue #1305, opened 2023-06-27
Enable networking implementations to return a stream


shenkeyao:
Neither `CommunicationChannel` nor `ConnectedNetwork` exposes a stream to listen on, so the message handler has nothing to handle for the network task. We could add a function to get a `Stream<Item = MSG>` (or `Vec<Msg>` if `Stream` doesn't work with the web server) to `CommunicationChannel` or `ConnectedNetwork`, then use it as the message stream for network tasks.

As a temporary solution, the stream may be of type `Vec<Message>` rather than `Message`. We can use `GeneratedStream` (added in https://github.com/EspressoSystems/HotShot/pull/1309) to get two streams with different `TransmitType`, then combine them using `Merge`.

PR #1315, opened 2023-06-29
Enable networking implementations to return a stream

shenkeyao:
- Creates two streams with different `TransmitType`.
- Combines them into one message stream for the networking task.

Closes https://github.com/EspressoSystems/HotShot/issues/1305.

### DIFF
--- a/src/traits/networking/centralized_server_network.rs
+++ b/src/traits/networking/centralized_server_network.rs
@@ -4,6 +4,8 @@
 //!
 #[cfg(feature = "async-std-executor")]
 use async_std::net::TcpStream;
+use hotshot_task::{BoxSyncFuture, boxed_sync};
+use nll::nll_todo::nll_todo;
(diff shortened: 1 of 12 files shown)


src/traits/networking/centralized_server_network.rs
```diff
@@ -825,6 +830,8 @@ impl<M: NetworkMsg, K: SignatureKey + 'static, E: ElectionConfig + 'static> Conn
                     .context(FailedToDeserializeSnafu),
             }
         };
+        let _ =
+            <CentralizedServerNetwork<K, E> as ConnectedNetwork<M, K>>::shut_down::<'_, '_>(self);
```
DieracDelta, line 834: I don't think this will have the intended effect. When `recv_msgs` is called, `shut_down` will be called before the closure is evaluated. Is this what we want?

DieracDelta (approved):

PR #1326, opened 2023-07-01
Make the commitment of all VoteData of fixed length

dailinsubjam:
This PR

- Implement Commitment<T> for all the kinds of [VoteData](https://github.com/EspressoSystems/HotShot/blob/paper_benchmarking/types/src/traits/election.rs#L76). So that the new commitment (the actual one we'll use) takes in the type and the sub-commitment (like leaf_commitment or block_commitment).
- Change the type of input data so that sign() takes in GenericArray.

This PR can (only?) pass the web server sequencing tests and will be merged to `[sishan/signature_aggregation_with_hsp](https://github.com/EspressoSystems/HotShot/tree/sishan/signature_aggregation_with_hsp)` once it's approved.

### DIFF
--- a/src/traits/election/vrf.rs
+++ b/src/traits/election/vrf.rs
@@ -228,14 +228,14 @@ where
         println!("Inside sign() of SignatureKey for JfPubKey.");
         // Sign it
         // Sishan NOTE: for QC Aggregation
-        // Sishan NOTE TODO: change 32 to a defined constant
-        let pointer = data.as_ptr() as *const [u8; 32];
-        let mut msg_test = unsafe { *pointer };
+        let mut generic_msg_test = GenericArray::from_slice(data);
+        println!("generic_msg_test = {:?}", generic_msg_test);
         println!("data = {:?}", data);
         let agg_signature_test =
             BitvectorQuorumCertificate::<BLSOverBN254CurveSignatureScheme>::partial_sign(
                 &(),
-                &msg_test.into(),
+                &generic_msg_test,
+                // &msg_test.into(),
                 key_pair_test.sign_key_ref(),
                 &mut rand::thread_rng());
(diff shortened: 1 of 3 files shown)


src/traits/election/vrf.rs
```diff
@@ -226,13 +227,10 @@ where
...
-         now the msg is a test message, partial_sign only support msg with [u8] in length 32, cannot support general `data`.*/
-        //<BLSOverBN254CurveSignatureScheme as SignatureScheme>::Signature
-        // let msg_test: [u8; data.len()] = data.clone();
-        let msg_test = [72u8; 32]; // Sishan TODO: change to `data` after hotshot-primitives is updated
-        // msg_test[..data.len()].clone_from_slice(&data); 
-        println!("msg_test = {:?}", msg_test);
+        // Sishan NOTE: for QC Aggregation
+        // Sishan NOTE TODO: change 32 to a defined constant
+        let pointer = data.as_ptr() as *const [u8; 32];
+        let mut msg_test = unsafe { *pointer };
```
nyospe, line 233: Ummm. Can't you use `GenericArray` here to avoid the unsafe?

types/src/traits/election.rs
```diff
@@ -91,6 +91,35 @@ pub enum VoteData<COMMITTABLE: Committable + Serialize + Clone> {
...
+            VoteData::ViewSync(commitment) =>
+                unimplemented!(),
+        }
+    }
+
+    fn tag() -> String {
+            ("VOTE_DATA_COMMIT").to_string()
+    }
+}
+
```
nyospe, line 122: This looks good.

dailinsubjam (commented):
Not sure what to do for the input message with length 8 (their form is like a view-number following with 7 zeros), now I simply extend its length from 8 to 32, but I feel there may be better options on this.

Issue #1179, opened 2023-05-08
Create Long running DA Task


bfish713:

shenkeyao:
DA task implementations added to [keyao/da-task](https://github.com/EspressoSystems/HotShot/tree/keyao/da-task) branch. Not integrated into the main task yet.

PR #1337, opened 2023-07-05
Implement DA task for run view

shenkeyao:
- Implements `DATaskState` and its event handling functions.
- Integrates the DA task into the main task in mod.rs.
- Fixes the shutdown handling for the networking task. 

Closes https://github.com/EspressoSystems/HotShot/issues/1179.


task-impls/src/da.rs
```diff
@@ -0,0 +1,334 @@
...
+                    let state = DAVoteCollectionTaskState {
+                        committee_exchange: self.committee_exchange.clone(),
+                        accumulator,
+                        cur_view: vote.current_view,
+                        event_stream: self.event_stream.clone(),
+                    };
+                    let name = "DA Vote Collection";
+                    let filter = FilterEvent::default();
+                    // TODO (run_view refactor) `TaskBuilder` is created but the task isn't added to the task runner.
+                    let _builder =
```
bfish713, line 273: I updated the consensus task vote collection task logic to include the registering and running it in my pr if you want to have a look.  We should also add in our tech debt clean up to reuse the code for vote collection.  Looks like this is done in all 3 of the main tasks (excluding the network one)

task-impls/src/da.rs
```diff
@@ -0,0 +1,334 @@
...
+                    }
+                    Ok(None) => {
+                        info!("We were not chosen for DA committee on {:?}", view);
+                    }
+                    Ok(Some(vote_token)) => {
+                        info!("We were chosen for DA committee on {:?}", view);
+
+                        // Generate and send vote
+                        let message = self.committee_exchange.create_da_message::<I>(
+                            self.high_qc.commit(),
```
bfish713, line 219: Looks like this never actually gets updated.  Maybe the consensus task can emit an event when it updates it's high_qc, or maybe this qc should come from the DA proposal?  TBH I'm not really sure if we need a QC on the DA message now that it's separated from the consensus logic.

task-impls/src/da.rs
```diff
@@ -0,0 +1,334 @@
...
+                            info!("Sending vote to the DA leader {:?}", vote);
+                            self.event_stream
+                                .publish(SequencingHotShotEvent::DAVoteSend(vote))
+                                .await;
+                        }
+                    }
+                }
+            }
+            SequencingHotShotEvent::DAVoteRecv(vote, sender) => {
+                if vote.signature.0 != <TYPES::SignatureKey as SignatureKey>::to_bytes(&sender) {
```
bfish713, line 237: Just like in consensus, we should probably check first that we are the leader for the view of this vote

src/tasks/mod.rs
```diff
@@ -514,27 +494,47 @@ pub async fn add_consensus_task(
...
+        Commitment = TYPES::BlockType,
+    >,
+{
     // build the da task
-    let da_state = DATaskState {};
+    let da_state = DATaskState {
+        cur_view: TYPES::Time::new(0),
+        high_qc: QuorumCertificate::<TYPES, I::Leaf>::genesis(),
+        committee_exchange: committee_exchange.into(),
+        vote_collector: (TYPES::Time::new(0), async_spawn(async move {})),
```
bfish713, line 523: I ended up storing the task_id rather than a future/JoinHandle here.  The GlobalRegistry allows you to manually shut down, so that's what I use.

PR #1338, opened 2023-07-05
Update Consensus Task with DA split

bfish713:
This updates all the logic for the DA split to work with the consensus task.  Basically the task waits for both a valid proposal and a valid DAC from the last leader before sending a yes vote. It integrates the consensus task into the main runner too.

This also includes the basis for creating some unit tests.  I'll follow up by writing some this week but now there is a skeleton for creating and running any one task, and for testing a certain stream of input events yields and expected output.

Finally it adds the endpoints to the web_server for posting/getting the DAC


consensus/src/sequencing_replica.rs
```diff
@@ -208,6 +209,16 @@ where
                                         };
                                         let justify_qc_commitment = justify_qc.commit();
                                         let leaf_commitment = leaf.commit();
+                                        let Some(dac) = p.data.dac else {
```
DieracDelta, line 212: nice use of let-else!

src/lib.rs
```diff
@@ -176,7 +173,7 @@ pub struct SystemContextInner<TYPES: NodeType, I: NodeImplementation<TYPES>> {
 #[derive(Clone)]
 pub struct SystemContext<CONSENSUS: ConsensusType, TYPES: NodeType, I: NodeImplementation<TYPES>> {
     /// Handle to internal hotshot implementation
-    inner: Arc<SystemContextInner<TYPES, I>>,
+    pub inner: Arc<SystemContextInner<TYPES, I>>,
```
DieracDelta, line 176: hmm what's the reason for this? If we need this long term, maybe `pub crate` or a getter would suffice?

src/tasks/mod.rs
```diff
@@ -472,33 +467,84 @@ where
...
+    >,
+    CommitteeEx<TYPES, I>: ConsensusExchange<
+        TYPES,
+        Message<TYPES, I>,
+        Certificate = DACertificate<TYPES>,
+        Commitment = TYPES::BlockType,
+    >,
+{
+    let consensus = hotshot.get_consensus();
+    let c_api: HotShotSequencingConsensusApi<TYPES, I> = HotShotSequencingConsensusApi {
```
DieracDelta, line 499: Do we have any plans to nuke this consensus api? I remember discussing this a while back, don't remember the outcome of that conversation.

task-impls/src/consensus.rs
```diff
@@ -263,51 +266,80 @@ where
...
+                                    vote_token,
+                                );
+                        } else {
+                            let _message: GeneralConsensusMessage<TYPES, I> =
+                                self.quorum_exchange.create_yes_message(
+                                    proposal.justify_qc.commit(),
+                                    proposal.justify_qc.leaf_commitment,
+                                    cert.view_number,
+                                    vote_token,
+                                );
```
DieracDelta, line 303: What are these messages used for? Do we need to hook this into something?

task-impls/src/consensus.rs
```diff
@@ -263,51 +266,80 @@ where
...
+                    return;
+                }
+                let view_leader_key = self.quorum_exchange.get_leader(view);
                 if view_leader_key != sender {
                     return;
                 }
 
-                let vote_token = self.quorum_exchange.make_vote_token(self.cur_view);
+                let vote_token = self.quorum_exchange.make_vote_token(view);
+                // TODO: do some of this logic without the vote token check, only do that when voting.
```
DieracDelta, line 332: Should we add an issue for this?

task-impls/src/consensus.rs
```diff
@@ -475,33 +528,52 @@ where
...
                                 quorum_exchange: self.quorum_exchange.clone(),
                                 accumulator,
                                 cur_view: vote.current_view,
                                 event_stream: self.event_stream.clone(),
                             };
                             let name = "Quorum Vote Collection";
-                            let filter = FilterEvent::default();
-                            let _builder =
+                            let filter = FilterEvent(Arc::new(|event| match event {
+                                SequencingHotShotEvent::QuorumVoteRecv(_, _) => true,
```
DieracDelta, line 540:
```suggestion
                                matches!(event, SequencingHotShotEvent::QuorumVoteRecv(_, _))
```

testing/tests/consensus_task.rs
```diff
@@ -0,0 +1,317 @@
...
+    let node_id = 1;
+    let network_generator = Arc::new((launcher.generator.network_generator)(node_id));
+    let quorum_network = (launcher.generator.quorum_network)(network_generator.clone());
+    let committee_network = (launcher.generator.committee_network)(network_generator);
+    let storage = (launcher.generator.storage)(node_id);
+    let config = launcher.generator.config.clone();
+    let initializer =
+        HotShotInitializer::<TYPES, I::Leaf>::from_genesis(I::block_genesis()).unwrap();
+
+    let node_id = 1;
```
DieracDelta, line 227: duplicates `let node_id`s

testing/tests/consensus_task.rs
```diff
@@ -0,0 +1,317 @@
...
+        private_key,
+        node_id,
+        config,
+        storage,
+        exchanges,
+        initializer,
+        NoMetrics::boxed(),
+    )
+    .await
+    .expect("Could not init hotshot");
```
DieracDelta, line 259: Is this for the unit tests? Or is this the way consensus is initialized? In the latter case, we should discuss this. I'm already initializing the systemcontexts in the testing framework.

task-impls/src/events.rs
```diff
@@ -10,19 +11,22 @@ use hotshot_types::vote::{DAVote, QuorumVote};
...
     QuorumVoteSend(QuorumVote<TYPES, I::Leaf>),
     DAProposalSend(Proposal<DAProposal<TYPES>>),
     DAVoteSend(DAVote<TYPES, I::Leaf>),
+    DACertificateSend(DACertificate<TYPES>),
     QCFormed(QuorumCertificate<TYPES, I::Leaf>),
+    /// A view has timeouted out locally or we received/formed a timeout cert
+    /// contains the view we are moving to
     ViewChange(ViewNumber),
     ViewSyncTimeout(ViewNumber, u64, ViewSyncPhase),
     ViewSyncMessageSend(ViewSyncMessageType<TYPES>),
```
elliedavidson, line 32: I'll have these working in my next PR! Already broke them out into ViewSyncVoteSend/Recv and ViewSyncCertificateSend/Recv

task-impls/src/events.rs
```diff
@@ -10,19 +11,22 @@ use hotshot_types::vote::{DAVote, QuorumVote};
...
     DAVoteRecv(DAVote<TYPES, I::Leaf>, TYPES::SignatureKey),
-    ViewSyncMessage(ViewSyncMessageType<TYPES>),
+    DACertificateRecv(DACertificate<TYPES>),
     QuorumProposalSend(Proposal<QuorumProposalType<TYPES, I>>),
     QuorumVoteSend(QuorumVote<TYPES, I::Leaf>),
     DAProposalSend(Proposal<DAProposal<TYPES>>),
     DAVoteSend(DAVote<TYPES, I::Leaf>),
+    DACertificateSend(DACertificate<TYPES>),
     QCFormed(QuorumCertificate<TYPES, I::Leaf>),
+    /// A view has timeouted out locally or we received/formed a timeout cert
```
elliedavidson, line 28: This event should also be emitted anytime we go to the next view, even in the optimistic case where we vote and immediately go to the next view.  View Sync needs to know the current view.

src/lib.rs
bfish713: It's a hack to allow for the unit test stuff

bfish713: So yes we could add getters instead

src/tasks/mod.rs
bfish713: Hmm I'm not sure, I don't remember the discussion from before.  Here it's just needed to get the exchanges from consensus object.  I'd like to discuss this part since it looks like every tasks needs something like this passed in.

src/lib.rs
```diff
@@ -354,6 +351,11 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> SystemContext<TYPES::Consens
             .get_state()
     }
 
+    /// Returns a copy of the consensus struct
+    pub fn get_consensus(&self) -> Arc<RwLock<Consensus<TYPES, I::Leaf>>> {
```
elliedavidson, line 355: Useful!

task-impls/src/consensus.rs
bfish713: yep, we should emit this as an event, I think I missed this when I moved the logic around, good catch.

elliedavidson (approved):

task-impls/src/consensus.rs
bfish713: Nice! I had done this somewhere else but didn't think to apply it here.

testing/tests/consensus_task.rs
bfish713: Yeah this is just for unit tests, again kind of a hack to get the exchanges to create the consensus task.

PR #1339, opened 2023-07-05
Testing tasks

DieracDelta:
Very rough PR that comes with one compiling test. More to come.

Caveats/missing things:
- All node safety features have not been implemented. Currently, only safety on a per-node basis is done.
- Custom overrides
- Moving of the entire test suite on top of the new test harness. Both test harnesses currently exist.


src/lib.rs
```diff
@@ -236,8 +233,6 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> SystemContext<TYPES::Consens
             transactions: Arc::default(),
             saved_leaves,
             saved_blocks,
-            // TODO this is incorrect
-            // https://github.com/EspressoSystems/HotShot/issues/560
```
elliedavidson, line 240: Probably should keep this comment in

DieracDelta: Oops, I thought this was resolved since the issue is closed. Should I reopen it?

PR #1345, opened 2023-07-06
Ed/view sync task misc

elliedavidson:
This PR: 
* Addresses comments from my last PR
* Adds ViewSyncCertSend/Recv and ViewSyncVoteSend/Recv events
  * Adds handling for the above in the network task
* Adds some extra utility functions on view sync types

Still todo
* Initialize view sync task in mod.rs

### DIFF
--- a/task-impls/src/consensus.rs
+++ b/task-impls/src/consensus.rs
@@ -574,9 +574,7 @@ where
 
             SequencingHotShotEvent::ViewChange(new_view) => {
                 // update the view in state to the one in the message
-                if self.cur_view < TYPES::Time::new(*new_view) {
-                    self.cur_view = TYPES::Time::new(*new_view);
-                }
+                self.update_view(new_view);
             }
             SequencingHotShotEvent::Timeout(view) => {
                 self.update_view(view);
--- a/task-impls/src/view_sync.rs
+++ b/task-impls/src/view_sync.rs
@@ -88,7 +88,6 @@ pub struct ViewSyncTaskState<
         Commitment = ViewSyncData<TYPES>,
     >,
 {
-
     pub registry: GlobalRegistry,
     pub event_stream: ChannelStream<SequencingHotShotEvent<TYPES, I>>,
     pub filtered_event_stream: UnboundedStream<SequencingHotShotEvent<TYPES, I>>,
@@ -331,9 +330,9 @@ where
                                 .register_state(replica_state)
                                 .register_event_handler(replica_handle_event);
 
-                            let _view_sync_replica_task =
-                            async_spawn(
-                                ViewSyncReplicaTaskStateTypes::build(builder).launch());
+                            let _view_sync_replica_task = async_spawn(async move {
+                                ViewSyncReplicaTaskStateTypes::build(builder).launch().await
+                            });
                         }
                     }
                     ViewSyncMessageType::Vote(vote) => {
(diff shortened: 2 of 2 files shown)


task-impls/src/view_sync.rs
```diff
@@ -386,12 +393,18 @@ where
...
+                                    .register_event_stream(relay_state.event_stream.clone(), filter)
+                                    .await
+                                    .register_registry(&mut self.registry.clone())
                                     .await
                                     .register_state(relay_state)
                                     .register_event_handler(relay_handle_event);
+
+                            // TODO ED For now we will not await these futures, in the future we can await them only in the case of shutdown
+                            let _view_sync_relay_task =
+                                ViewSyncRelayTaskStateTypes::build(builder).launch();
```
bfish713, line 407: I think this is where we need to async_spawn and await or the future won't start running

elliedavidson: yes, you're right!  I'll add that in! Slipped my mind

bfish713 (approved):
just one small comment on vote collection, otherwise looks good!

task-impls/src/view_sync.rs
```diff
@@ -317,15 +320,20 @@ where
...
                                 .await
+                                .register_registry(&mut self.registry.clone())
+                                .await
                                 .register_state(replica_state)
                                 .register_event_handler(replica_handle_event);
-                            // TODO ED Launch task
+
+                            let _view_sync_replica_task =
+                            async_spawn(
+                                ViewSyncReplicaTaskStateTypes::build(builder).launch());
```
DieracDelta, line 336:
```suggestion
                            async_spawn(
                                ViewSyncReplicaTaskStateTypes::build(builder).launch()).await;
```
`launch` returns a `BoxFuture` that needs to be awaited in order to be executed

task-impls/src/consensus.rs
```diff
@@ -563,18 +563,20 @@ where
...
                 }
             }
-            SequencingHotShotEvent::ViewChange(_) => nll_todo(),
-            SequencingHotShotEvent::ViewSyncMessageRecv(_) => {
+
+            SequencingHotShotEvent::ViewChange(new_view) => {
                 // update the view in state to the one in the message
-                // TODO ED This info should come in the form of a ViewChange message, update
-                nll_todo()
+                if self.cur_view < TYPES::Time::new(*new_view) {
```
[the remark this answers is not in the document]
bfish713: I think it's ok for now, we probably should cancel some tasks, but they will get cancelled either way when other messages come in.  I need to think it through a little

task-impls/src/view_sync.rs
```diff
@@ -386,12 +393,18 @@ where
...
+                                    .register_event_stream(relay_state.event_stream.clone(), filter)
+                                    .await
+                                    .register_registry(&mut self.registry.clone())
                                     .await
                                     .register_state(relay_state)
                                     .register_event_handler(relay_handle_event);
+
+                            // TODO ED For now we will not await these futures, in the future we can await them only in the case of shutdown
+                            let _view_sync_relay_task =
+                            async_spawn(ViewSyncRelayTaskStateTypes::build(builder).launch());
```
DieracDelta, line 407: Is the idea to conditionally await this?

DieracDelta (approved):

PR #1352, opened 2023-07-07
Fix Signature types + Add Send QC logic

bfish713:
Got rid of the places we used an encoded signature in the network task and replaced it with SignatureKey type.  In doing this I realized that we weren't sending the QC proposal (the `QCSend` event was really just emitted when we collected the votes to form a QC). 

This also adds in that Proposal logic.  I also noticed that we never send a DA proposal, and I think we need to add this logic to the DA task.  As it stands now we will just propose empty blocks, but we should be getting the block from the DAProposalSend event in the Consensus task.

I merged in Elle's branch: https://github.com/EspressoSystems/HotShot/pull/1348, so we should merge that first.  The tests don't build but the errors seem to be related to ViewSync and are now revealed because all the libs are building.


task-impls/src/events.rs
```diff
@@ -20,17 +19,19 @@ pub enum SequencingHotShotEvent<TYPES: NodeType, I: NodeImplementation<TYPES>> {
...
     DAVoteRecv(DAVote<TYPES>),
     DACRecv(DACertificate<TYPES>),
-    ViewSyncMessageRecv(ViewSyncMessageType<TYPES>),
-    QuorumProposalSend(Proposal<QuorumProposalType<TYPES, I>>),
+    QuorumProposalSend(Proposal<QuorumProposalType<TYPES, I>>, TYPES::SignatureKey),
     QuorumVoteSend(QuorumVote<TYPES, I::Leaf>),
     DAProposalSend(Proposal<DAProposal<TYPES>>, TYPES::SignatureKey),
     DAVoteSend(DAVote<TYPES>),
-    QCSend(QuorumCertificate<TYPES, I::Leaf>),
+    QCFormed(QuorumCertificate<TYPES, I::Leaf>),
```
shenkeyao, line 26: Nit: I feel `QCSend` is a more consistent name since we have `DACSend` and `ViewSyncCertificateSend`?

shenkeyao (approved):
LGTM!

DieracDelta (approved):

task-impls/src/events.rs
bfish713: I changed it to send since this is not he actual send on the network, it's just the event for the main consensus task to know that enough votes were collected to make a QC and we can now propose.

PR #1357, opened 2023-07-08
Debugging

elliedavidson:
Just a debugging branch; we likely won't commit all these changes.  

This PR: 
* Added a TransactionRecv event that the network task properly handles
* Added a TransactionSend event, but it is not currently used (api.broadcast is used instead)
* Updated network task to handle direct messages (seems like it was only handling broadcast messages before)
* Extended the CompletionTask time to 10 seconds
* Added logic in the DA task to handle the TransactionRecv event and create a DA proposals
* Added a ViewChange(1) event to "start" HotShot
* Updated logic to handle genesis block and DAC in consensus task
* Made separate network filters for each exchange to avoid sending duplicate messages to the network
  * ^ It might be good to do the same for receiving messages as well, since each network task receives all the messages and therefore sends out too many events, but all event handlers should be idempotent anyway (need to check if they actually are)
* Add logic to consensus task to store leaves from proposals
* Debugged consensus task
* Updates self.block in consensus task with the DA block using a new event

Current state: 
* A race condition where events come in before a task is launched, and therefore the task misses those events.  (Currently seen in the consensus task, but will show up in other tasks once we get to them)
* Lots of logging statements
* I think we still need to add logic to emit decide events, update locked_view, etc. I copied it over from the old consensus run_view but it is commented out

This is a messy branch, but I vote to merge it so we can all work on debugging.  We can remove comments and logging at a later date.

### DIFF
--- a/src/tasks/mod.rs
+++ b/src/tasks/mod.rs
@@ -551,6 +551,7 @@ pub async fn add_da_task<
     task_runner: TaskRunner,
     event_stream: ChannelStream<SequencingHotShotEvent<TYPES, I>>,
     committee_exchange: CommitteeEx<TYPES, I>,
+    handle: SystemContextHandle<TYPES, I>,
 ) -> TaskRunner
 where
     I::Exchanges: SequencingExchangesType<TYPES, Message<TYPES, I>>,
@@ -565,6 +566,8 @@ where
     let registry = task_runner.registry.clone();
     let da_state = DATaskState {
         registry: registry.clone(),
+        consensus: handle.hotshot.get_consensus(),
+        high_qc: QuorumCertificate::<TYPES, I::Leaf>::genesis(),
         cur_view: TYPES::Time::new(0),
         committee_exchange: committee_exchange.into(),
         vote_collector: None,
--- a/src/tasks/mod.rs
+++ b/src/tasks/mod.rs
@@ -323,7 +323,7 @@ pub fn quorum_filter<
     match event {
         SequencingHotShotEvent::QuorumProposalSend(_, _)
         | SequencingHotShotEvent::QuorumVoteSend(_)
-        | SequencingHotShotEvent::DACRecv(_)
+        | SequencingHotShotEvent::SendDABlockData(_)
         | SequencingHotShotEvent::Shutdown
         | SequencingHotShotEvent::ViewChange(_)=> true,
 
--- a/src/tasks/mod.rs
+++ b/src/tasks/mod.rs
@@ -310,6 +310,69 @@ pub enum GlobalEvent {
 }
 impl PassType for GlobalEvent {}
 
+pub fn quorum_filter<
+    TYPES: NodeType<ConsensusType = SequencingConsensus>,
+    I: NodeImplementation<
+        TYPES,
+        Leaf = SequencingLeaf<TYPES>,
+        ConsensusMessage = SequencingMessage<TYPES, I>,
+    >,
+>(
+    event: &SequencingHotShotEvent<TYPES, I>,
+) -> bool {
+    match event {
+        SequencingHotShotEvent::QuorumProposalSend(_, _)
+        | SequencingHotShotEvent::QuorumVoteSend(_)
+
+        | SequencingHotShotEvent::Shutdown
+        | SequencingHotShotEvent::ViewChange(_)=> true,
+
+
+        _ => false,
+    }
+}
+
+pub fn da_filter<
+    TYPES: NodeType<ConsensusType = SequencingConsensus>,
+    I: NodeImplementation<
+        TYPES,
+        Leaf = SequencingLeaf<TYPES>,
+        ConsensusMessage = SequencingMessage<TYPES, I>,
+    >,
+>(
+    event: &SequencingHotShotEvent<TYPES, I>,
+) -> bool {
+    match event {
+        | SequencingHotShotEvent::DAProposalSend(_, _)
+        | SequencingHotShotEvent::DAVoteSend(_)
+        | SequencingHotShotEvent::Shutdown
+        | SequencingHotShotEvent::ViewChange(_)
+        | SequencingHotShotEvent::TransactionSend(_) => true,
+
+        _ => false,
+    }
+}
+
+pub fn view_sync_filter<
+    TYPES: NodeType<ConsensusType = SequencingConsensus>,
+    I: NodeImplementation<
+        TYPES,
+        Leaf = SequencingLeaf<TYPES>,
+        ConsensusMessage = SequencingMessage<TYPES, I>,
+    >,
+>(
+    event: &SequencingHotShotEvent<TYPES, I>,
+) -> bool {
+    match event {
+        | SequencingHotShotEvent::ViewSyncVoteSend(_)
+        | SequencingHotShotEvent::ViewSyncCertificateSend(_, _)
+        | SequencingHotShotEvent::Shutdown
+        | SequencingHotShotEvent::ViewChange(_) => true,
+
+        _ => false,
+    }
+}
+
 /// add the networking task
 /// # Panics
 /// Is unable to panic. This section here is just to satisfy clippy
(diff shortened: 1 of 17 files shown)


src/tasks/mod.rs
```diff
@@ -310,6 +310,70 @@ pub enum GlobalEvent {
...
+        | SequencingHotShotEvent::SendDABlockData(_)
+        | SequencingHotShotEvent::Shutdown
+        | SequencingHotShotEvent::ViewChange(_)=> true,
+
+
+        _ => false,
+    }
+}
+
+pub fn da_filter<
```
shenkeyao, line 335: Nit: maybe `committee_filer` is a better name, since it's for the committee networking task, not the DA task?

src/tasks/mod.rs
```diff
@@ -351,7 +417,7 @@ where
                     .expect("Failed to receive broadcast messages"),
             );
             async_sleep(Duration::new(0, 500)).await;
-            network.shut_down().await;
+            // network.shut_down().await;
```
shenkeyao, line 420: I'm a bit unsure about this. Don't we still need to shut down the two `network`s somewhere else?

task-impls/src/da.rs
```diff
@@ -286,8 +339,84 @@ where
...
+                let signature = self.committee_exchange.sign_da_proposal(&block.commit());
+                let data: DAProposal<TYPES> = DAProposal {
+                    deltas: block.clone(),
+                    view_number: self.cur_view,
+                };
+                let message = SequencingMessage::<TYPES, I>(Right(
+                    CommitteeConsensusMessage::DAProposal(Proposal { data, signature }),
+                ));
+                // Brodcast DA proposal
+                // TODO ED We should send an event to do this, but just getting it to work for now
```
shenkeyao, line 409: Isn't the event just the `SendDABlockData` below? Or do we need an additional event?

task-impls/src/network.rs
```diff
@@ -107,9 +117,17 @@ impl<
                     }
                 },
             },
-            MessageKind::Data(_) => {
-                warn!("Got unexpected message type in network task!");
-                return;
+            MessageKind::Data(message) => {
+                match message {
+                    // ED Why do we need the view number in the transaction?
```
shenkeyao, line 122: Found this comment in this enum:
```
/// TODO rethink this when we start to send these messages
/// we only need the view number for broadcast
```
which is suggesting quite the opposite--keeping `view_number` and removing `transaction`. I guess we should rethink this. :) This is fine for now though.

task-impls/src/network.rs
```diff
@@ -190,7 +264,9 @@ impl<
             | SequencingHotShotEvent::ViewSyncVoteSend(_)
             | SequencingHotShotEvent::ViewSyncCertificateSend(_, _)
             | SequencingHotShotEvent::Shutdown
-            | SequencingHotShotEvent::ViewChange(_) => true,
+            | SequencingHotShotEvent::ViewChange(_)
+            | SequencingHotShotEvent::TransactionSend(_) => true,
+
```
shenkeyao, line 269: Do we still need this function, now that we've added different filters for the three types of networking tasks?

task-impls/src/consensus.rs
```diff
@@ -248,29 +284,100 @@ where
...
-    async fn vote_if_able(&self) {
+    async fn vote_if_able(&self) -> bool {
+        error!("In vote if able");
+
         if let Some(proposal) = &self.current_proposal {
-            if let Some(cert) = self.certs.get(&proposal.get_view_number()) {
+            // ED Need to account for the genesis DA cert
+            // // error!("in vote if able for proposal view {:?}", proposal.view_number);
+
+            if proposal.justify_qc.is_genesis() && proposal.view_number == ViewNumber::new(1) {
```
bfish713, line 315: I had kept this logic in the handling of the proposal so we didn't duplicate the vote token and proposal check when we get a DA cert since his is called in both places.

task-impls/src/consensus.rs
```diff
@@ -248,29 +284,100 @@ where
...
+                        message = self.quorum_exchange.create_yes_message(
+                            proposal.justify_qc.commit(),
+                            leaf.commit(),
+                            view,
+                            vote_token,
+                        );
+
+                        if let GeneralConsensusMessage::Vote(vote) = message {
+                            error!("Sending vote to next leader {:?}", vote.current_view());
+                            self.event_stream
```
bfish713, line 369: I think we can't vote here because we haven't checked the DA cert

task-impls/src/consensus.rs
```diff
@@ -486,31 +753,41 @@ where
                                     .await;
                             }
                         });
-                        if let GeneralConsensusMessage::Vote(vote) = message {
```
bfish713, line 489: This was in here for sending no votes if the proposal was incorrect/invalid.  The message is created above only if the vote is a no

task-impls/src/consensus.rs
```diff
@@ -551,49 +830,83 @@ where
...
-                                VoteCollectionTypes::build(builder).launch().await
+                                VoteCollectionTypes::build(builder).launch().await;
                             });
-                            self.vote_collector = Some((vote.current_view, id));
+                            error!("Starting vote handle for view {:?}", vote.current_view);
                         }
                     }
                     QuorumVote::Timeout(_) | QuorumVote::No(_) => {
-                        warn!("The next leader has received an unexpected vote!");
+                        panic!("The next leader has received an unexpected vote!");
```
bfish713, line 843: I guess the idea is we are just removing no voting because it's just not implemented.  As a side note I wasn't sure which task view sync or consensus would handle forming the no quorum, but we can solve that in a later sprint

task-impls/src/consensus.rs
```diff
@@ -461,14 +628,114 @@ where
                                 );
                             }
                         }
-                        // self.update_view(view);
+
+                        self.high_qc = leaf.justify_qc.clone();
+
+                        //             let mut new_anchor_view = consensus.last_decided_view;
```
bfish713, line 634: Yep this was the logic I knew was missing I'll work off this to get it to compile/mostly work before the pair programming hopefully

src/tasks/mod.rs
elliedavidson: Yes, we should shut down them down somewhere else, though not sure exactly where

task-impls/src/da.rs
elliedavidson: The SendDABlockData event is only for the DA leader in the DA task to send the block data to itself in the consensus task.  The DAProposalSend event is to send the proposal to the DA committee over the network.  There might be a better way to do this in the future, but sending an event seemed like the easiest way to get things working.

task-impls/src/consensus.rs
elliedavidson: I ended up running into an error where the `vote_if_able` function needs the leaf commitment to put in the vote.  In the future there is probably a cleaner way to handle this, but for now I duplicated some of the proposal checking code so I would have access to `leaf.commit()` inside `vote_if_able`  😄

task-impls/src/consensus.rs
elliedavidson: Yes, you are right!  I'll add a comment for now and we can come back to it later.

task-impls/src/consensus.rs
elliedavidson: Yes, you are right!  I'll add it back in

task-impls/src/consensus.rs
elliedavidson: Yes!

task-impls/src/da.rs
```diff
@@ -212,10 +263,11 @@ where
                             vote_token,
                         );
 
-                        self.cur_view = view;
+                        // ED Don't think this is necessary?
+                        // self.cur_view = view;
```
shenkeyao, line 267: Yeah I think we could remove this.

PR #1366, opened 2023-07-10
Update message handling and event filters for network tasks

shenkeyao:
- Adds `NetworkTaskKind` enum for the three kinds of network tasks.
- Updates `handle_message` to avoid sending duplicate events from different task kinds.
- Moves filter functions from `lib.rs` to `NetworkTaskState` and adds an upper-level `filter` function to choose the corresponding filter by the task kind.
- Also includes auto-formatting.

### DIFF
--- a/task-impls/src/network.rs
+++ b/task-impls/src/network.rs
@@ -132,11 +132,11 @@ impl<
                         },
                         Either::Right(committee_message) => match committee_message {
                             CommitteeConsensusMessage::DAProposal(proposal) => {
-                                SequencingHotShotEvent::DAProposalRecv(proposal.clone(), sender)
+                                SequencingHotShotEvent::DAProposalRecv(proposal, sender)
                             }
                             CommitteeConsensusMessage::DAVote(vote) => {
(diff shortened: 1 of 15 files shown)


elliedavidson (approved):
Looks great!

task-impls/src/network.rs
```diff
@@ -77,57 +84,102 @@ impl<
...
+                                error!("Got unexpected message type in network task!");
+                                return;
+                            }
+                        },
+                        Either::Right(committee_message) => match committee_message {
+                            CommitteeConsensusMessage::DAProposal(proposal) => {
+                                SequencingHotShotEvent::DAProposalRecv(proposal.clone(), sender)
+                            }
+                            CommitteeConsensusMessage::DAVote(vote) => {
+                                SequencingHotShotEvent::DAVoteRecv(vote.clone())
```
DieracDelta, line 138: does this need to be clone?

shenkeyao: No it doesn't! I've also removed another `clone` above it.

PR #1369, opened 2023-07-11
Ed/more debugging

elliedavidson:


bfish713 (approved):

task-impls/src/da.rs
```diff
@@ -403,9 +415,10 @@ where
                 let signature = self.committee_exchange.sign_da_proposal(&block.commit());
                 let data: DAProposal<TYPES> = DAProposal {
                     deltas: block.clone(),
-                    view_number: self.cur_view,
+                    // Upon entering a new view we want to send a DA Proposal for the next view -> Is it always the case that this is cur_view + 1? 
+                    view_number: self.cur_view + 1,
```
shenkeyao, line 419: Yeah, I think we need to send it to the next leader. Not sure if it's `self.cur_view + 1` though.

PR #1378, opened 2023-07-12
Add signature serialization function

shenkeyao:
- A follow-up PR for https://github.com/EspressoSystems/HotShot/pull/1377.
- Adds a helper function to serialize signatures that can be used in multiple places.


dailinsubjam (approved):
Looks good to me.

PR #1381, opened 2023-07-13
Orchestrator with Testing PR

RohanAhluwalia:
Orchestrator is running with simple unit tests. Currently averaging a couple of values.

To run the tests, `cd testing/tests/orchestrator` and run `cargo test --features="full-ci, channel-async-std" -p hotshot-testing --test orchestrator`


elliedavidson (changes requested):
When you have a chance could you rebase this to paper_benchmarking so that the only changed files are the ones you updated (and maybe the Cargo.lock)

PR #1397, opened 2023-07-18
Jr/flushing out testing harness

DieracDelta:


elliedavidson (approved):
These are very wonderful changes!  Thank you so much!

PR #1406, opened 2023-07-20
Kaley/update examples

elliedavidson:


examples/infra/modDA.rs
```diff
@@ -0,0 +1,775 @@
...
+        //                     |block| total_transactions += block.txn_count(),
+        //                     |_| total_commitments += 1,
+        //                 );
+        //             }
+        //         }
+        //         Err(e) => {
+        //             timed_out_views += 1;
+        //             error!("View: {:?}, failed with : {:?}", round, e);
+        //         }
+        //     }
```
DieracDelta, line 438: this can be deleted

examples/infra/modDA.rs
```diff
@@ -0,0 +1,775 @@
...
+                            if num_successful_commits >= rounds {
+                                break;
+                            }
+                            // when we make progress, submit new events
+                        }
+                        EventType::ReplicaViewTimeout { view_number } => {
+                            error!("Timed out as a replicas in view {:?}", view_number);
+                        }
+                        EventType::NextLeaderViewTimeout { view_number } => {
+                            error!("Timed out as the next leader in view {:?}", view_number);
```
DieracDelta, line 392: We should be able to drop this part too if we integrate the timeout task (or a configured safety task, which tracks successful views).

examples/infra/modDA.rs
```diff
@@ -0,0 +1,775 @@
...
+        Exchanges = SequencingExchanges<
+            TYPES,
+            Message<TYPES, NODE>,
+            QuorumExchange<
+                TYPES,
+                SequencingLeaf<TYPES>,
+                QuorumProposal<TYPES, SequencingLeaf<TYPES>>,
+                MEMBERSHIP,
+                QUORUMNETWORK,
+                Message<TYPES, NODE>,
```
DieracDelta, line 711: these are some crazy trait bounds! I wish there was an easier way to do this.

examples/web-server-da/README.md
```diff
@@ -0,0 +1,29 @@
+Commands to run da examples: 
+1a)Start web servers by either running 3 servers:
+cargo run --example web-server --profile=release-lto --features="full-ci" <PORT_FOR_CDN> 
+cargo run --example web-server --profile=release-lto --features="full-ci" <PORT_FOR_DA> 
+cargo run --example web-server --profile=release-lto --features="full-ci" <PORT_FOR_VIEW_SYNC>
```
DieracDelta, line 5: nit: might be worth using backticks

examples/web-server-da/multi-web-server.rs
```diff
@@ -0,0 +1,56 @@
...
+}
+#[async_std::main]
+async fn main() {
+    let args = MultiWebServerArgs::parse();
+    let (server_shutdown_sender_cdn, server_shutdown_cdn) = oneshot();
+    let (server_shutdown_sender_da, server_shutdown_da) = oneshot();
+    let (server_shutdown_sender_view_sync, server_shutdown_view_sync) = oneshot();
+    let _sender = Arc::new(server_shutdown_sender_cdn);
+    let _sender = Arc::new(server_shutdown_sender_da);
+    let _sender = Arc::new(server_shutdown_sender_view_sync);
```
DieracDelta, line 23: is there a reason for these?

src/types/handle.rs
```diff
@@ -202,7 +202,8 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES> + 'static> SystemContextHandl
     }
 
     /// begin consensus by sending a genesis event
-    pub async fn start_consensus(&self) {
+    #[deprecated]
+    pub async fn start_consensus_deprecated(&self) {
```
DieracDelta, line 206: Might be worth documenting what should be used instead

task-impls/src/consensus.rs
```diff
@@ -851,7 +851,7 @@ where
                             }
 
                             // warn!("Inserting leaf into storage {:?}", leaf.commit());
-                            if *view % 10 == 0 && self.quorum_exchange.is_leader(view){
+                            if *view % 10 == 0 && self.quorum_exchange.is_leader(view) {
```
DieracDelta, line 854: Out of curiosity, why `% 10`?

DieracDelta (changes requested):

examples/infra/modDA.rs
```diff
@@ -0,0 +1,775 @@
...
+            >>::Membership::default_election_config(config.config.total_nodes.get() as u64)
+        });
+
+        let _committee_election_config = <CommitteeEx<TYPES, NODE> as ConsensusExchange<
+            TYPES,
+            Message<TYPES, NODE>,
+        >>::Membership::default_election_config(
+            config.config.total_nodes.get() as u64
+        );
```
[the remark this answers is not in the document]
bfish713: Looks like we do use it right below to create the exchanges, should probably jus remove the leading `_`

task-impls/src/consensus.rs
bfish713: I actually had put this logging to check that we were committing unique transactions.  I just wanted to log relatively less so it wasn't crowding out everything else in the logs.  We probably don't need this logging.  I think thought the most updated run_view_refactor code didn't have this tho.

examples/web-server-da/README.md
```diff
@@ -0,0 +1,29 @@
+Commands to run da examples:
```
shenkeyao, line 1: I forgot to build before running the examples, and got a `no example target` error--might be something that's trivial but worth mentioning here.

examples/infra/modDA.rs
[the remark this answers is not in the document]
DieracDelta: oops! Bamboolzed by the _ bahaha

src/types/handle.rs
elliedavidson: This change is on me!  We're supposed to use start_consensus on SystemContext instead.  I can add a comment.

task-impls/src/consensus.rs
kmchicoine: I changed it to `% 50` for now- if you'd rather I delete it, just say the word!

examples/infra/modDA.rs
kmchicoine: I did try to not have to include the VIEWSYNCNETWORK as one of the bounds, but it was going to take a lot longer and we wanted to get examples running. I could make a ticket to look into reducing the bounds for the next time we have spare cycles?

DieracDelta: I think we've devoted significant effort to this, and am not sure it's possible. +1 for making an issue, but I'm not sure how we could simplify bounds, sadly.

PR #1407, opened 2023-07-20
Refactor to expose `OrchestratorClient`

nomaxg:
- Exposes OrchestratorClient for use in external crates
- Load Orchestrator as TOML value instead of from a file path so that external crates can run the orchestrator


kmchicoine (approved):
LGTM! I like the separation of the client and lib files, nice

PR #1409, opened 2023-07-20
Orchestrator with multiple tests 

RohanAhluwalia:
New merge for orchestrator into the paper benchmarking branch. Should be able to merge into the paper benchmarking branch as of 7/20. 

To run tests for orchestrator run, `cargo test --features="full-ci, channel-async-std" -p hotshot-testing --test orchestrator -- --nocapture`

### DIFF
--- a/orchestrator/src/lib.rs
+++ b/orchestrator/src/lib.rs
@@ -4,6 +4,7 @@ use async_lock::RwLock;
 use hotshot_types::traits::election::ElectionConfig;
 use hotshot_types::traits::signature_key::SignatureKey;
 use std::io;
+use std::vec::Vec;
 use std::io::ErrorKind;
 use std::net::IpAddr;
 use std::net::SocketAddr;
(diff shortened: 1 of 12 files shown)


orchestrator/src/lib.rs
```diff
@@ -194,12 +248,8 @@ where
     KEY: serde::Serialize,
     ELECTION: serde::Serialize,
 {
-    let api_toml = toml::from_str::<toml::Value>(include_str!(concat!(
```
nomaxg, line 197: let's leave the instantiation of the API as it is, if we create the API from a file, external crates wont be able to run the orchestrator

RohanAhluwalia: Yes! Makes sense. Updated it

elliedavidson (commented):
Looks good!  The tests pass!  I'd like to wait to merge until the sequencer team is stable on using this branch for their code.

rob-maron:
Closing for now but keeping the branch open !

PR #1414, opened 2023-07-21
Doppio benchmarking

elliedavidson:
Some updates from benchmarking: 
* Remove panic from failed message send in the network task
* Update testing harness and examples to use `send_transaction` instead of `submit_transaction`.  The former makes use of the event for sending transactions and allows the transactions to be submitted to the committee network
  * Unfortunately the above (at least to get it working quickly for benching) required making the testing harness sequencing consensus specific.  I think this is fine, but let's discuss in our next meeting
*  Updates some logging since we only log ERROR level in the benchmarks
* Logs the number of transactions committed upon a decide event and upon getting a DA proposal.  

We may not merge all of these changes in, but I wanted feedback to make sure this branch doesn't diverge too much from `run_view_refactor`

### DIFF
--- a/consensus/src/lib.rs
+++ b/consensus/src/lib.rs
@@ -17,7 +17,7 @@ mod next_leader;
 mod replica;
 mod sequencing_leader;
 mod sequencing_replica;
-mod traits;
+pub mod traits;
 pub mod utils;
 
 use async_compatibility_layer::async_primitives::subscribable_rwlock::SubscribableRwLock;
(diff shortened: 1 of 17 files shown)


bfish713 (approved):
I'm fine with merging this.  I had already updated some places where transactions were shared via broadcast instead of via the network task but must have missed this one.  And we will need to clean up the logging anyway once we are confident everything works 100%

PR #1442, opened 2023-07-27
Fix txn proposing logic for DA task

shenkeyao:
Fixes the logic of proposing transactions in the DA task, to avoid skipping transactions when there's a timeout.

### DIFF
--- a/task-impls/src/da.rs
+++ b/task-impls/src/da.rs
@@ -568,10 +568,10 @@ where
 
         let receiver = consensus.transactions.subscribe().await;
 
-        while task_start_time.elapsed() < self.api.propose_max_round_time() {
-            let txns = consensus.transactions.cloned().await;
-            error!("Size of transactions: {}", txns.len());
-            let unclaimed_txns: Vec<_> = txns
+        loop {
+            let all_txns = consensus.transactions.cloned().await;
+            error!("Size of transactions: {}", all_txns.len());
+            let unclaimed_txns: Vec<_> = all_txns
                 .iter()
                 .filter(|(txn_hash, _txn)| !previous_used_txns.contains(txn_hash))
                 .collect();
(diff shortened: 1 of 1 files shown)


task-impls/src/da.rs
```diff
@@ -597,13 +597,20 @@ where
                     Ok(Ok(_)) => continue,
                 }
             }
-            let mut txns = vec![];
-            for (_hash, txn) in unclaimed_txns {
-                txns.push(txn.clone());
-            }
-            return Some(txns);
+            break;
```
jbearer, line 600: IMO it's clearer to put this `break` in the `Err(_)` case in the `match` above, to make it clear that this only happens in one scenario (timeout). That's mostly personal preference though.

task-impls/src/da.rs
```diff
@@ -597,13 +597,20 @@ where
...
+        let txns: Vec<TYPES::Transaction> = all_txns
+            .iter()
+            .filter_map(|(txn_hash, txn)| {
+                if !previous_used_txns.contains(txn_hash) {
+                    Some(txn.clone())
+                } else {
+                    None
+                }
+            })
+            .collect();
```
jbearer, line 612: Does this compute the same list as `unclaimed_txns` above? If so I think you can do `break unclaimed_txns` inside the loop and `let unclaimed_txns = loop { ... }`, to extract the value from inside the loop instead of computing it again

jbearer (approved):
Looks good, left a couple mostly cosmetic suggestions

task-impls/src/da.rs
shenkeyao: Agree! I put it there because there was some code after the `match`, but that's not the case anymore. Will move it.

shenkeyao: Oh actually I think the `break` should be put here, otherwise we aren't handling the case where the `if` conditions (line 580) aren't satisfied.

task-impls/src/da.rs
shenkeyao: Ahh, I really like this idea and gave it a try, but got an `all_txns does not live long enough` error that I wasn't able to fix.

jbearer: Yeah I think you'd have to do something like `break all_txns.into_iter().cloned()` to take ownership of the txns when exiting the loop. Not a big deal though

task-impls/src/da.rs
jbearer: Oh yeah you're right

task-impls/src/da.rs
shenkeyao: Oh I merged the PR before seeing your reply. I'll do this on my lint fix branch then. :)

PR #1447, opened 2023-07-28
fix: async executor bug

DieracDelta:


task/src/task.rs
```diff
@@ -411,21 +411,39 @@ impl<'pin, HSTT: HotShotTaskTypes> ProjectedHST<'pin, HSTT> {
...
-                                    *self.state = Some(state);
                                     if let Some(completed) = result {
+                                        *self.in_progress_fut = None;
+                                        *self.state = Some(state);
                                         *self.r_val = Some(completed);
                                         let result = self.launch_shutdown_fut(cx);
                                         *self.event_stream = Some(inner_event_stream);
                                         return Left(result);
+                                    } else {
+                                        let mut fut = async move {
```
jbearer, line 430: This deserves a comment

task/src/task.rs
```diff
@@ -411,21 +411,39 @@ impl<'pin, HSTT: HotShotTaskTypes> ProjectedHST<'pin, HSTT> {
...
+                                            async_yield_now().await;
+                                            (None, state)
+                                        }
+                                        .boxed();
+                                        if let Poll::Ready((_, state)) = fut.as_mut().poll(cx) {
+                                            *self.state = Some(state);
```
jbearer, line 436: Comment: explain why it's ok to continue (executor decided no other tasks need to run)

task/src/task.rs
```diff
@@ -480,13 +496,27 @@ impl<'pin, HSTT: HotShotTaskTypes> ProjectedHST<'pin, HSTT> {
                             let mut fut = handle_msg(msg, state);
                             match fut.as_mut().poll(cx) {
                                 Poll::Ready((result, state)) => {
-                                    *self.in_progress_fut = None;
-                                    *self.state = Some(state);
                                     if let Some(completed) = result {
                                         *self.r_val = Some(completed);
```
jbearer, line 500: Do we need to reset `self.in_progress_fut` and `self.state` here, as we do above?

jbearer (commented):
While we're at it, can you change the generator closures in `broadcast_stream` and `direct_stream` in `add_network_message_task()` to something more like

```rust
        let closure = async move {
            loop {
                let msgs = Messages(
                    network
                        .recv_msgs(TransmitType::Broadcast)
                        .await
                        .expect("Failed to receive broadcast messages"),
                );
                if msgs.0.is_empty() {
                    async_sleep(Duration::new(0, 500)).await;
                } else {
                    break msgs;
                }
            }
        };
```

So we don't spam the stream with empty messages in the first place?

task/src/task.rs
DieracDelta: Yes! This is a bug. Pushing fix. We never actually use the state, which is why nothing broke.

PR #1449, opened 2023-07-28
Discard Seen Transactions

bfish713:
Current logic only considers the parent leaf's transactions when building blocks and removes transactions from our transaction store when they are decided.  New transactions are always added to the our transaction store.  The assumption was that the replica would keep it's transaction list constantly up to date with the network.  However we only poll transactions when we know we need them.  This means we see transactions committed which never our transaction store.  Late when we poll transactions we may download a bunch of transactions we previously saw and then propose them.  

The solution in this PR is to filter transactions before adding to our store if we know the transaction was in a previous block.  This should prevent us from knowingly proposing duplicate transactions.  

There may be a memory leak with this solution.  If the buffer on the webserver is removing old transactions before new leaders download them, Then old transactions will never be removed.   We're only storing the commitments so it shouldn't be a major issue.  We will want to eventually come up with a proper solution to this.  Probably the next thing to do would be make this hash set have a bounded number of entries and remove FIFO

### DIFF
--- a/consensus/src/lib.rs
+++ b/consensus/src/lib.rs
@@ -28,6 +28,7 @@ pub use next_leader::NextValidatingLeader;
 pub use replica::Replica;
 pub use sequencing_leader::{ConsensusLeader, ConsensusNextLeader, DALeader};
 pub use sequencing_replica::SequencingReplica;
+use std::collections::HashSet;
 pub use traits::{ConsensusSharedApi, SequencingConsensusApi, ValidatingConsensusApi};
 pub use utils::{View, ViewInner};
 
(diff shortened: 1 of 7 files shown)


types/src/traits/election.rs
```diff
@@ -337,7 +337,7 @@ pub trait ConsensusExchange<TYPES: NodeType, M: NetworkMsg>: Send + Sync {
                         if matches!(signature.1 .1, VoteData::Yes(thing) if thing == leaf_commitment)
                         {
                             yes_votes += u64::from(signature.1 .2.vote_count());
-                        } else if matches!(signature.1 .1, VoteData::Yes(thing) if thing == leaf_commitment)
+                        } else if matches!(signature.1 .1, VoteData::No(thing) if thing == leaf_commitment)
```
shenkeyao, line 340: Good catch!

shenkeyao (approved):

PR #1480, opened 2023-08-02
feat: move in QC and stake table traits and impl from hotshot-primitives 

alxiong:
Major changes include:

- move `trait QuorumCertificate` and `trait StakeTableScheme` in `hotshot_types::traits`
- move implementations to separate crates `hotshot-qc` and `hotshot-stake-table`
  - move in QC circuit impl under `hotshot_qc::snarked` module


chancharles92 (commented):
LGTM on my side. Do we have any TODO issues in the old repo? Shall we move the important TODO issues if there are any?

crates/hotshot-qc/Cargo.toml
```diff
@@ -0,0 +1,38 @@
...
+typenum = { version = "1.16.0" }
+
+[dev-dependencies]
+hotshot-stake-table = { path = "../hotshot-stake-table" }
+
+[features]
+default = ["parallel"]
+std = ["ark-std/std"]
+parallel = ["jf-primitives/parallel", "jf-utils/parallel"]
+# TODO: (alex) what other features should I add
```
DieracDelta, line 37: `full-ci` is an alias for async-std channels and executor. I'm afraid having `full-ci` here will confuse cargo when it builds `tokio-ci`, since `full-ci` and `tokio-ci` are mutually exclusive. It might be better to just leave these features on by default and have all other crates that use this enable this crate's default features.

crates/hotshot-stake-table/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+
+[dev-dependencies]
+rand_chacha = { version = "0.3.1", default-features = false }
+
+[features]
+default = ["parallel"]
+std = ["ark-std/std", "ark-serialize/std", "ark-ff/std"]
+parallel = ["jf-primitives/parallel", "jf-utils/parallel", "ark-ff/parallel"]
+# TODO: (alex) what other features should I add to follow the same pattern as other crates?
+full-ci = ["std", "parallel"]
```
DieracDelta, line 36: same thought as above

crates/hotshot-qc/Cargo.toml
alxiong: sounds good, I will just remove `full-ci` here. 

I guess the main question I have, which `just` command to run so that I can run test only on my specific crate while able to compile others?
for example right now I can't just run `just _basic_test` since this is not running my test, I can't just `cargo test -p my-package` because compiler will complain that I gave the wrong feature flag for other crates. 

any idea @DieracDelta or should we add another Just command for what I ask?

DieracDelta: For now, `just test_async_std_test my_test` will build the workspace and run a test. I'm not sure there's a cargo invocation that will test only that package that works right now. I'm working on changing our build flags per your suggestion to use `RUSTFLAGS`. That should fix this issue. In the mean time maybe we should merge?

DieracDelta (approved):

crates/hotshot-qc/Cargo.toml
alxiong: sounds good, Thanks Justin!

PR #1482, opened 2023-08-03
Clean up signature schemes, only use BN254

dailinsubjam:
This PR:

- Implement trait Serialization, Deserialization, PartialOrd, Ord, SignatureKey, and Key Generation for BN254 Key Pair.
- Delete and substitute `TestableSignatureKey` as the comment said "TODO we don't need this if we have `generated_from_seed_indexed`".
- Discard use of Ed25519, update all of them to BN254 (besides part related to libp2p). And Ed25519 can be a parameter substituting all places that use BN254 in the future.
- Clean up some warnings related to keys and signatures.

### DIFF
--- a/src/traits/election/vrf.rs
+++ b/src/traits/election/vrf.rs
@@ -251,7 +251,6 @@ where
     // Sishan NOTE TODO: this one should be substitude by the one in bn254_priv.rs or ed25519_priv.rs or balabala, use SIGSCHEME;
     // Maybe the trait SignatureScheme should be deleted...At least for BLSoverBN254CurveSignatureScheme, the function inside is useless.
     fn generated_from_seed_indexed(seed: [u8; 32], index: u64) -> (Self, Self::PrivateKey) {
-        println!("Inside generated_from_seed_indexed");//Sishan NOTE TODO: delete this line
 
         let mut hasher = blake3::Hasher::new();
         hasher.update(&seed);
--- a/src/traits/election/vrf.rs
+++ b/src/traits/election/vrf.rs
@@ -248,15 +248,19 @@ where
         })
     }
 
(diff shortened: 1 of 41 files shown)


src/traits/election/vrf.rs
```diff
@@ -262,13 +248,18 @@ where
         })
     }
 
-    fn generated_from_seed_indexed(_seed: [u8; 32], index: u64) -> (Self, Self::PrivateKey) {
+    // Sishan NOTE TODO: this one should be substitude by the one in bn254_priv.rs or ed25519_priv.rs or balabala, use SIGSCHEME;
+    // Maybe the trait SignatureScheme should be deleted...At least for BLSoverBN254CurveSignatureScheme, the function inside is useless.
+    fn generated_from_seed_indexed(seed: [u8; 32], index: u64) -> (Self, Self::PrivateKey) {
```
elliedavidson, line 253: +1, Let's make an issue to do this in the future!

types/src/traits/signature_key/bn254/bn254_pub.rs
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    fn sign(key_pair: QCKeyPair, data: &[u8]) -> EncodedSignature {
+        let generic_msg = GenericArray::from_slice(data);
+        let agg_signature_test = BitVectorQC::<BLSOverBN254CurveSignatureScheme>::sign(
+            &(),
+            &generic_msg,
+            key_pair.sign_key_ref(),
+            &mut rand::thread_rng(),
+        ).unwrap();
```
elliedavidson, line 76: Small nit: let's add an error check here instead of directly unwrapping.  😁

elliedavidson (changes requested):

src/traits/election/vrf.rs
dailinsubjam: Just made one https://github.com/EspressoSystems/HotShot/issues/1512

elliedavidson (approved):

PR #1520, opened 2023-08-07
Doppio benchmarking merge

elliedavidson:
This PR merges in changes made while benchmarking, including: 
* Update examples to submit transactions on each view change event
* Update examples so only the last 10 nodes of the network submit transactions
* Update DA task so that only DA members poll for DA proposals
* Fix bug where PollForTransaction tasks were not getting properly canceled  
* Update orchestrator to properly set up small DA committees

This PR does not: 
* Fix any stylistic errors other than clippy errors.


src/traits/networking/web_server_network.rs
```diff
@@ -412,32 +506,48 @@ impl<
...
-                    .await
-                {
-                    error!(
-                        "Background receive proposal polling encountered an error: {:?}",
-                        e
-                    );
-                }
-            }
-        });
+        // let tx_handle = async_spawn({
```
bfish713, line 523: Nit: Remove

bfish713 (approved):

bfish713 (approved):

PR #1522, opened 2023-08-07
Ed/debug web server

elliedavidson:
This PR: 
* Reduces the number of successful views needed for the web server, since it will naturally be slower than the memory network
* Removes old tests from the web server crate


shenkeyao (approved):
LGTM. Maybe sync with main again in case `test_with_failures` causes the CI to fail.

shenkeyao:
> LGTM. Maybe sync with main again in case `test_with_failures` causes the CI to fail.

Oh there are also format errors.

PR #1528, opened 2023-08-08
Ed/web server debug 2

elliedavidson:


examples/infra/mod.rs
```diff
@@ -348,9 +348,9 @@ pub trait Run<
         loop {
             if should_submit_txns {
                 for _ in 0..transactions_per_round {
-                    let txn = txns.pop_front().unwrap();
+                    let _txn = txns.pop_front().unwrap();
                     tracing::info!("Submitting txn on round {}", round);
-                    context.submit_transaction(txn).await.unwrap();
+                    // context.submit_transaction(txn).await.unwrap();
```
shenkeyao, line 353: Looks like we removed `submit_transaction` here and `DataMessage::SubmitTransaction` in lib.rs but didn't replace them with something else. I thought we would need either `submit_transaction` or `send_transaction` to add transactions, otherwise wouldn't we get empty blocks?

shenkeyao: Or is the `send_transaction` call in txn_task.rs enough to handle all txn submissions?

elliedavidson: This is in the non-DA examples, which I figure we should probably delete anyway?

bfish713 (approved):

examples/infra/mod.rs
shenkeyao: Oh, yes that makes sense!

PR #1530, opened 2023-08-09
Discard the part related to SignatureScheme and SIGSCHEME and JfPubKey

dailinsubjam:
This PR:

- Totally remove JfPubKey, and substitute it with BN254Pub.
- Help ensure all the places across the code except libp2p use and only use the signature scheme {BN254Priv, BN254Pub}.
- Get rid of duplicate keys.
- Rename `known_nodes_qc` to `knwon_nodes_with_stake`.


src/traits/election/vrf.rs
```diff
@@ -68,208 +53,6 @@ impl<VRF, VRFHASHER, VRFPARAMS> Clone for VRFStakeTable<VRF, VRFHASHER, VRFPARAM
...
-
-        (
-            Self {
-                pk: pk.clone(),
-                _pd: PhantomData,
-            },
-            (sk, pk),
-        )
-    }
-}
```
elliedavidson, line 272: Yay for finally removing this!

elliedavidson (approved):

Issue #1466, opened 2023-08-02
Tests for Nodes Leaving the Nework


bfish713:
This is related to #1464 but specifically we need a test which can shows we make expected progress when one or more nodes leaves the network in an arbitrary round.  This test doesn't need to include the node coming back to the network

DieracDelta:
https://github.com/EspressoSystems/HotShot/blob/2830bf17b95b790791f99d3160a187d3119c1401/testing/tests/basic.rs#L24 this exists but doesn't pass.

PR #1540, opened 2023-08-10
Test nodes leaving the network

shenkeyao:
Adds tests for 1, f/2, and f nodes leaving the network.

Closes https://github.com/EspressoSystems/HotShot/issues/1466.


testing/tests/basic.rs
```diff
@@ -5,74 +5,150 @@
...
+}
+
+/// Test f nodes leaving the network.
+#[cfg(test)]
+#[cfg_attr(
+    feature = "tokio-executor",
+    tokio::test(flavor = "multi_thread", worker_threads = 2)
+)]
+#[cfg_attr(feature = "async-std-executor", async_std::test)]
+async fn test_with_failures_f() {
```
elliedavidson, line 108: Nit: Maybe we could add a quick comment that these failures purposefully don't include DA nodes.  I feel like we may forget this in the future when we switch to a rotating DA committee, and then we'll wonder why these tests fail. :)

elliedavidson (approved):
A few minor comments, otherwise looks good!

testing/tests/basic.rs
shenkeyao: Oh, the failures actually include the DA nodes. The DA committee is the entire network for this test, so those 6 nodes to be killed are also DA nodes.

shenkeyao: Now failures only include DA nodes, so I added a comment.

DieracDelta (approved):
LGTM

elliedavidson (approved):
LGTM!

PR #1541, opened 2023-08-11
Nuke validating consensus

DieracDelta:
We're still failing the libp2p tests, but I think this is a good checkpoint. Validating consensus has been nuked, and the `TestExchange` trait has been introduced to generate the entire set of comm channels.


elliedavidson (approved):
🔥

Issue #1524, opened 2023-08-08
Update view sync task to properly handle view change events


elliedavidson:
There currently exists a race condition where consensus proposes just as a view is timing out.  View sync triggers and then never cancels itself accordingly.

PR #1550, opened 2023-08-14
Ed/view sync polling fix

elliedavidson:
This PR: 
* Fixes the view sync polling bug where tasks were not correctly canceled and garbage collected if a `ViewChange` event occurred during, but not because of, view sync. 
* Minor clean up to view sync task code 

Closes: https://github.com/EspressoSystems/HotShot/issues/1524

### DIFF
--- a/justfile
+++ b/justfile
@@ -26,8 +26,12 @@ _test_basic:
   echo Testing with async std executor
   RUST_LOG="" cargo test  --features=full-ci --lib --bins --tests --benches --workspace --no-fail-fast test_basic -- --test-threads=1 --nocapture
 
-_test_basic_tokio:
+test_web_server:
   echo Testing with async std executor
+  RUST_LOG="" cargo test  --features=full-ci --lib --bins --tests --benches --workspace --no-fail-fast web_server_network -- --test-threads=1 --nocapture
+
+_test_basic_tokio:
+  echo Testing with tokio executor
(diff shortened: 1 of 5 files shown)


bfish713 (approved):

justfile
```diff
@@ -26,8 +26,12 @@ _test_basic:
   echo Testing with async std executor
   RUST_LOG="" cargo test  --features=full-ci --lib --bins --tests --benches --workspace --no-fail-fast test_basic -- --test-threads=1 --nocapture
 
-_test_basic_tokio:
+test_web_server:
   echo Testing with async std executor
+  RUST_LOG="" cargo test  --features=full-ci --lib --bins --tests --benches --workspace --no-fail-fast web_server_network -- --test-threads=1 --nocapture
+
+_test_basic_tokio:
```
DieracDelta, line 33: maybe we should remove the `_` here too

testing/src/overall_safety_task.rs
```diff
@@ -437,6 +437,7 @@ impl OverallSafetyPropertiesDescription {
                                         }
                                     }
                                 }
+                                // TODO Emit this event in the consensus task once canceling the timeout task is implemented
```
DieracDelta, line 440: Should we add a github issue for this?

DieracDelta (approved):
LGTM, nice!

Issue #1492, opened 2023-08-04
Unit Tests for Consensus


bfish713:
Should test the basic case of proposal received, vote, and collect vote workflows

PR #1565, opened 2023-08-15
Unit Tests for Consensus

bfish713:
Adds very basic tests for the consensus task.  Tests that a node can send a proposal and vote after receiving a propsal.

This is a good base to build more complex test off of.  We still need code for creating some other events, like DAC but adding tests should go much faster starting from this point

closes: #1492


task-impls/src/consensus.rs
```diff
@@ -1410,5 +1411,6 @@ pub fn consensus_event_filter<TYPES: NodeType, I: NodeImplementation<TYPES>>(
             | SequencingHotShotEvent::ViewChange(_)
             | SequencingHotShotEvent::SendDABlockData(_)
             | SequencingHotShotEvent::Timeout(_)
+            | SequencingHotShotEvent::Shutdown,
```
shenkeyao, line 1414: Oh good catch!

task-impls/src/harness.rs
```diff
@@ -92,10 +90,16 @@ pub fn handle_event<TYPES: NodeType, I: NodeImplementation<TYPES>>(
     std::option::Option<HotShotTaskCompleted>,
     TestHarnessState<TYPES, I>,
 ) {
-    if !state.expected_output.contains(&event) {
+    error!("got event: {:?}", event);
```
shenkeyao, line 93: NIt: same here, maybe use `info!` instead.

testing/tests/consensus_task.rs
```diff
@@ -1,346 +1,344 @@
...
+use hotshot::traits::election::vrf::JfPubKey;
+use hotshot_task_impls::harness::run_harness;
+use hotshot_types::traits::signature_key::EncodedSignature;
+use hotshot_types::traits::{
+    election::ConsensusExchange, node_implementation::NodeType, state::ConsensusTime,
+};
+
+use std::collections::HashMap;
+
+async fn build_consensus_api(
```
shenkeyao, line 51: I'll probably move this function to a separate file and refactor it a bit, so it can be used for other tasks as well. Just a heads-up so you won't think it's disappeared. 😃

shenkeyao (commented):
Looks good! Only nit comments about tracing.

justfile
```diff
@@ -22,6 +22,10 @@ test_async_std_all:
   echo Testing with async std executor
   cargo test  --features=full-ci --lib --bins --tests --benches --workspace --no-fail-fast -- --test-threads=1
 
+_test_consensus:
```
shenkeyao, line 25: Forgot to comment on this earlier--I think we could remove the underscore like we did to `test_basic`.

testing/tests/consensus_task.rs
```diff
@@ -1,346 +1,344 @@
...
+
+    let propsal_recv = build_proposal_recv(&handle, &private_key, &public_key).await;
+
+    input.push(SequencingHotShotEvent::ViewChange(ViewNumber::new(1)));
+    input.push(propsal_recv.clone());
+
+    input.push(SequencingHotShotEvent::Shutdown);
+
+    output.insert(propsal_recv.clone(), 1);
+    if let SequencingHotShotEvent::QuorumProposalRecv(proposal, _) = propsal_recv {
```
shenkeyao, line 327: This `if` seems to be redundant since `proposal_recv` is always `SequencingHotShotEvent::QuorumProposalRecv`. Would it be better if we just make the `build_proposal_recv` return `(message, key)` rather than the `SequencingHotShotEvent` enum, then use it to build the enum to be pushed to `input` in this function?

However, except for the additional `if` check, I do think the way you implement looks nicer, so I don't have a strong opinion here. 😃

DieracDelta (approved):

shenkeyao (approved):
LGTM!

PR #1577, opened 2023-08-18
Update audit.toml to remove RUSTSEC-2022-0093 ignore

elliedavidson:
The issue tracking this audit finding in libp2p has been closed.  https://github.com/libp2p/rust-libp2p/issues/4327

We can merge this once we update libp2p

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -42,12 +42,4 @@ ignore = [
     # Waiting on a new release of https://github.com/mxinden/if-watch
     # https://rustsec.org/advisories/RUSTSEC-2022-0008
     "RUSTSEC-2022-0008",
-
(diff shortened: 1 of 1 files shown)


DieracDelta:
Updating libp2p-identity doesn't fix the problem. I think we need to wait for libp2p to make a new release.

elliedavidson:
> 

The markdown issue is just a warning.  Audit will pass with it.  

Yes, it looks like we'll need to resolve some Jellyfish conflicts first before we are able to update libp2p-identity.  Here's an issue for this: https://github.com/EspressoSystems/HotShot/issues/1578

elliedavidson:
This will probably be partially resolved once we remove the dependency on `hotshot-primitives` crate

PR #1589, opened 2023-08-21
Ed/merge sig agg

elliedavidson:
This PR is a wrapper around this PR: https://github.com/EspressoSystems/HotShot/pull/1568


shenkeyao (approved):
LGTM!

src/traits/storage/memory_storage.rs
```diff
@@ -150,8 +149,8 @@ mod test {
     impl NodeType for DummyTypes {
         type Time = ViewNumber;
         type BlockType = DummyBlock;
-        type SignatureKey = Ed25519Pub;
-        type VoteTokenType = StaticVoteToken<Ed25519Pub>;
+        type SignatureKey = BN254Pub;
```
philippecamacho, line 152: I think we should use another naming here, something like BLSVerKey (see https://github.com/EspressoSystems/jellyfish/blob/903d32a4a4c65a1e41fc9f2da602e323779659e0/primitives/src/signatures/bls_over_bn254.rs#L239C1-L239C1) as we are refering to the public key of some signature scheme. Also BN254 is not specific enough as there are two possible groups.

types/src/traits/signature_key/bn254/bn254_priv.rs
```diff
@@ -0,0 +1,71 @@
+use custom_debug::Debug;
+use jf_primitives::signatures::bls_over_bn254::{KeyPair as QCKeyPair, SignKey as QCSignKey};
+use rand::SeedableRng;
+use rand_chacha::ChaCha20Rng;
+use serde::{Deserialize, Serialize};
+use std::cmp::Ordering;
+
+/// Private key type for a bn254 keypair
+#[derive(PartialEq, Eq, Clone, Serialize, Deserialize, Debug)]
+pub struct BN254Priv {
```
philippecamacho, line 10: As mentioned above, the naming for this wrapping should mention the signature scheme instead of the curve.

types/src/traits/signature_key/bn254/bn254_pub.rs
```diff
@@ -0,0 +1,168 @@
...
+                let generic_msg: &GenericArray<u8, U32> = GenericArray::from_slice(data);
+                BLSOverBN254CurveSignatureScheme::verify(&(), &ver_key, generic_msg, &s).is_ok()
+            }
+            Err(_) => false,
+        }
+    }
+
+    fn sign(sk: &Self::PrivateKey, data: &[u8]) -> EncodedSignature {
+        let generic_msg = GenericArray::from_slice(data);
+        let agg_signature_test_wrap = BitVectorQC::<BLSOverBN254CurveSignatureScheme>::sign(
```
philippecamacho, line 74: Why is the word "test" mentioned several time in this function?

src/traits/storage/memory_storage.rs
elliedavidson: I made an issue for this here: https://github.com/EspressoSystems/HotShot/issues/1592

types/src/traits/signature_key/bn254/bn254_pub.rs
elliedavidson: Fixed this in the latest commit!

src/traits/storage/memory_storage.rs
philippecamacho: Thanks!

types/src/lib.rs
```diff
@@ -46,6 +46,8 @@ pub struct HotShotConfig<K, ELECTIONCONFIG> {
     pub max_transactions: NonZeroUsize,
     /// List of known node's public keys, including own, sorted by nonce ()
     pub known_nodes: Vec<K>,
+    /// List of known node's public keys and stake value for certificate aggregation, serving as public parameter
+    pub known_nodes_with_stake: Vec<ENTRY>,
```
alxiong, line 50: it felt underconstrained w/o specifying the equivalence between the key in `ENTRY` and `K`. 

I wonder if you want to use:
```rust
use hotshot-types::traits::stake_table::StakeTableScheme;

pub struct HotShotConfig<K,ST: StakeTableScheme, ELECTIONCONFIG> 
where ST::Key: K, {
  pub known_nodes: Vec<K>,
  pub stake_table: ST,
}
```

philippecamacho:
I get this failure when I [run the tests](https://github.com/EspressoSystems/HotShot#testing) on my machine.

```
    at src/traits/networking/libp2p_network.rs:549

The application panicked (crashed).
  2023-08-21T22:45:30.303380Z ERROR libp2p_networking::network::behaviours::gossip: error publishing gossip message Err(Duplicate)
    at libp2p-networking/src/network/behaviours/gossip.rs:218

Message:  TEST FAILED! Results: [("Test Overall Safety Task", MismatchedLeaf)]
Location: /home/leloup/repositories/espressosys/HotShot/testing/src/test_runner.rs:168

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
FAILED

failures:

failures:
    webserver_libp2p_network

test result: FAILED. 0 passed; 1 failed; 1 ignored; 0 measured; 0 filtered out; finished in 11.11s

```

alxiong (commented):
1. I thought @dailinsubjam is going to move `SignatureKey` into its own crate under `/crates/` and out of `/types`? does that affect this PR as well?
2. I got lost in the many abstractions of crypto-team-defined `QuorumCertificate`, jellyfish's `AggregatableSignature`, and here HotShot's `SignatureKey`, `AssembledSignature`, `SignedCertificate`, and HotShot's own `QuorumCertificate`, we should clarify them and remove unnecessary layers of abstraction. (maybe in a separate PR) 
in fact, I'm lost because I don't know which trait is for what purposes and thus cannot comment on what API I expect it to have.
3. would be great if @elliedavidson could provide more details in the PR description, at least highlight the main changes, and what changes deserve special attention (if any)

elliedavidson:
> 

This is fine; the fix for that test is being upstreamed in another PR.

types/src/lib.rs
elliedavidson: 1. Yes, that will happen in a separate PR!
2. I agree.  In general HotShot traits have gotten messy.  Here is an issue for this, but we unfortunately have other priorities to work on right now: https://github.com/EspressoSystems/HotShot/issues/1595
3. Normally a PR would have more details!  Sishan's original PR did have more details.  I only made this PR since Sishan can't make it while she is off.

elliedavidson: Re this particular comment - I agree - good catch.  I think we should probably just remove the known_nodes since it effectively the same thing as known_nodes_with_stake.  Issue here: https://github.com/EspressoSystems/HotShot/issues/1596

philippecamacho (approved):
LGTM

Issue #1567, opened 2023-08-15
Update message broadcasting to avoid hanging


shenkeyao:
`broadcast_message` may hang when sending messages to nodes that are shut down. We should update it to terminate if the recipient is killed.

Issue #1526, opened 2023-08-08
Update `submit_transaction` function to work like `send_transaction` function


elliedavidson:
The `submit_transaction` function needs to be updated to submit transactions to the committee network instead of the quorum network.  Currently the `send_transaction` function does this, but it has poor UX since it requires a `SystemContext` to use.

PR #1600, opened 2023-08-22
Ed/update timeout vote and cert

elliedavidson:
I want to get this merged before I start reworking the view change logic based on this discussion: https://espresso.zulipchat.com/#narrow/stream/311646-Consensus/topic/.E2.9C.94.20Timeout.20Logic

This PR: 

- [x] Updates view sync pre-commit certificates to properly work with signature aggregation by changing their threshold to `failure_threshold`, which is correct since pre-commit certificates only need f + 1 votes instead of the usual 2f + 1
- [x] Fixes https://github.com/EspressoSystems/HotShot/issues/1567 by making transaction submission create an async task to send the transaction.  This is how it was previously implemented before the event-based architecture refactor.  Also fixes https://github.com/EspressoSystems/HotShot/issues/1526 by un-deprecating `submit_transaction` and deprecating `send_transaction`
  * In the future we should support a client waiting for the result of the transaction submission
- [x] Creates basic test for timeouts
  * This test is currently ignored since it fails.  That is expected because timeout is not fully implemented
- [x] Add basic `TimeoutCertificate`

### DIFF
--- a/consensus/src/sequencing_leader.rs
+++ b/consensus/src/sequencing_leader.rs
@@ -518,7 +518,7 @@ where
                                 }
                             }
                             QuorumVote::Timeout(vote) => {
-                                qcs.insert(vote.justify_qc);
+                                qcs.insert(vote.high_qc);
                             }
                             QuorumVote::No(_) => {
                                 warn!("The next leader has received an unexpected vote!");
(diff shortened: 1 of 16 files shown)


shenkeyao (commented):
Not to block this PR since it's already merged, just a question about the use of `async spwan`. 😃

PR #1606, opened 2023-08-23
temporarily remove arm build

elliedavidson:
Temporarily disables the ARM build and re-adds the flake upgrades that caused the ARM build to fail.


DieracDelta:
I think this can be closed since the arm issue is resolved

PR #1620, opened 2023-08-24
Add unit test for network event handling and refactor testing harness

shenkeyao:
- Adds a unit test for the event handling of the network task.
- Moves helper functions for unit tests to `helpers.rs`.
- Updates `run_harness` to take an optional event stream as an input to support consistent use of stream when needed. 
- Updates `justfile`.

Closes https://github.com/EspressoSystems/HotShot/issues/1495.


elliedavidson (approved):
Looks great!  Thank you so much for adding so many code comments.  And also for abstracting the task_helpers file.  These will be so helpful as we add more tests.

bfish713 (approved):

PR #1627, opened 2023-08-26
chore(tedious): remove hotshot-primitives dependency

### LINKED ISSUE #1570: API Change after hotshot-primitives moved into hotshot

ggutoski:
close #1570 

`hotshot-primitives` dependency is gone!

# Addendum to discussion in #1570 

Recall that `hotshot-qc` is a port from `hotshot-primitives` into the local `hotshot` workspace. Due to historical accident, code in `bit_vector.rs` from `hotshot-qc` is newer than the code in `bit_vector.rs` from the now-removed `hotshot-primitives` dependency. I could not easily update `SignatureKey` code to use `hotshot-qc` so instead I copied old code from `hotshot-primitives` into a new module `bit_vector_old.rs` for easy compatibility.

`SignatureKey` is old and needs to be removed. But that's a job for another PR.


crates/hotshot-signature-key/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+tracing = { workspace = true }
+typenum = { workspace = true }
+
+[features]
+# default = ["parallel"]
+# std = ["ark-std/std"]
+# parallel = ["jf-primitives/parallel", "jf-utils/parallel"]
+full-ci = ["demo"]  # needed only because `just test_*` needs `demo` code
+tokio-ci = ["demo"] # needed only because `just test_*` needs `demo` code
+demo = []
```
alxiong, line 31: someone know better how we should declare flags in this repo vet this please? 

plus could we write a section in CONTRIBUTING.md to include a simple explanation of these feature flag and how to use them?

alxiong (approved):
The quick hacky fix looks good to me, we will clean up the technical debt later.

alxiong (approved):

Issue #1624, opened 2023-08-25
Update libp2p to use QUIC


elliedavidson:
https://github.com/libp2p/rust-libp2p/blob/master/transports/quic/CHANGELOG.md

DieracDelta:
I think this is a duplicate of https://github.com/EspressoSystems/HotShot/issues/725. Should I close #725?

elliedavidson:
> I think this is a duplicate of #725. Should I close #725?

Yes, that sounds good.

PR #1650, opened 2023-08-30
Enable quic networking (#1624)

rob-maron:
references issue #1624


DieracDelta:
Nice work, LGTM! Wow, seems like it's essentially a drop in replacement for TCP.

DieracDelta (approved):

PR #1665, opened 2023-09-05
Remove all audit ignores

elliedavidson:
Removes all ignore statements from `audit.toml`.  There are 5 warnings that stem from `tide/surf-disco` still present.  

Link to successful `cargo audit` run: https://github.com/EspressoSystems/HotShot/actions/runs/6085713033/job/16510443653


rob-maron (approved):
Looks good to me !

Issue #1463, opened 2023-08-02
Implement Catchup Mode for Replica


bfish713:
When a replica starts up it should be in a "catchup mode" where it won't vote on proposals until it knows that it has enough information.

We probably don't need a special mode for this but rather make sure we don't have any issue starting back up and seeing a QC from arbitrarily far in the future

PR #1682, opened 2023-09-07
Catchup Tests and Fixes

### LINKED ISSUE #1464: Tests for Catchup
Create a unit tests and integration tests to test catchup

The integration test should do the following:

- Start a committee of nodes.
- Disable one of the nodes for more than 3 rounds
- Re connect the node and expect it to start voting an expected number of rounds later
- Make sure it is able to send valid proposals as well.

bfish713:
Adds tests for catchup by adding the ability to start nodes after the tests is already running.  The tests verify that a late starting node can join the network and can successfully lead views

The view sync tests fails for the same reason as the timeout tests currently fail for.

Also fixes a bug where the same node id was being double counted in the certificate, which eventually causes a panic when trying to assemble the signature.

Closes #1464 and #1463


elliedavidson (approved):
Looks good to me!  I agree that the view sync issue should be fixed with my timeout work.

elliedavidson (approved):

PR #1711, opened 2023-09-11
feat(block_contents): Change Block::add_transaction_raw function signature to accept &mut self.

### LINKED ISSUE #1588: Block::add_transaction_raw function signature
`add_transaction_raw` returns `Self` but does not consume `self`.

https://github.com/EspressoSystems/HotShot/blob/e157601188c03ff6136a017b6ac03fb145c60eb2/types/src/traits/block_contents.rs#L53-L54

Thus, any impl of this method must clone self. We don't want to copy the whole block each time we add a single tx.

Seems to me that this method is a natural analogue to `Vec::push`: it should take `mut &self` and return something like `Result<(), Self::Error>`

PatStiles:
Aims to close #1588. Appreciate any nits and if this is stale let me know.

### DIFF
--- a/crates/hotshot/src/demos/sdemo.rs
+++ b/crates/hotshot/src/demos/sdemo.rs
@@ -206,15 +206,14 @@ impl Block for SDemoBlock {
     }
 
     fn add_transaction_raw(
-        &self,
+        &mut self,
         tx: &Self::Transaction,
-    ) -> std::result::Result<Self, Self::Error> {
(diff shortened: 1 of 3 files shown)


rob-maron:
Hey PatStiles,

Thanks for the PR!
Apologies for not seeing this sooner. I believe we removed that function in commit 26e24518bf85fc48398cce01934107b8de4c6839, so it is indeed stale. 

I'm closing for now but you are encouraged to open new ones if you would like! 

Thanks again!

Issue #1672, opened 2023-09-06
Remove current block building API


shenkeyao:
Remove the current `Block` trait which will be replaced by `BlockHeader` and `BlockPayload`.

Issue #1673, opened 2023-09-06
Add block payload building API


shenkeyao:
- Create an empty payload
- Append txn
- Compute VID payload commitment
- External API - distribute raw data block to DA (the same as submitting an entire block)

PR #1714, opened 2023-09-12
Replace block building with block payload

shenkeyao:
Instead of removing the current block-building API as planned in https://github.com/EspressoSystems/HotShot/issues/1672, this PR renames `Block` to `BlockPayload` with an additional unimplemented `commit` function.

This allows us to be ready to implement the block payload and header without breaking the build, and we can add the VID commitment computation once the VID integration is in.

Closes https://github.com/EspressoSystems/HotShot/issues/1672. Also is part 1 of https://github.com/EspressoSystems/HotShot/issues/1673 (part 2 is in progress).


crates/hotshot/src/demos/sdemo.rs
```diff
@@ -229,6 +234,13 @@ impl Block for SDemoBlock {
                 .collect(),
         }
     }
+
```
elliedavidson, line 237: Such cleanliness!  Thank you!

elliedavidson (dismissed):

bfish713 (approved):

elliedavidson (approved):

ggutoski (approved):

PR #1747, opened 2023-09-14
Add endpoint to get a recent proposal to catchup w/ webserver

bfish713:
The way we do polling by view means a node that comes back will online will poll for old views.  This PR adds an endpoint which allows the node to poll for the most recent view the webserver has received.  That way the node will update to a very recent view and subsequently poll for the correct view.

A proper solution is to transform the webserver into a stream like interface, or at least not depend completely on fetching by a specific view.  For now this is a stop gap that allows for catchup to work in the pull model.


elliedavidson (approved):
LGTM!

PR #1749, opened 2023-09-14
Split Transaction Collection and block Creation to Separate Task

bfish713:
All transaction handling should be moved to one place and out of the main consensus/da tasks


crates/hotshot/src/tasks/mod.rs
```diff
@@ -407,6 +408,78 @@ where
...
+) -> TaskRunner
+where
+    CommitteeEx<TYPES, I>: ConsensusExchange<
+        TYPES,
+        Message<TYPES, I>,
+        Certificate = DACertificate<TYPES>,
+        Commitment = TYPES::BlockType,
+    >,
+{
+    // build the da task
```
shenkeyao, line 435: Nit: `da` -> `transaction`.

crates/testing/tests/network_task.rs
```diff
@@ -102,6 +108,8 @@ async fn test_network_task() {
     // view number check in `publish_proposal_if_able` in consensus.rs, and we will see an error in
     // logging, but that is fine for testing as long as the network task is correctly handling
     // events.
+    output.insert(SequencingHotShotEvent::Timeout(ViewNumber::new(1)), 1);
+    output.insert(SequencingHotShotEvent::Timeout(ViewNumber::new(2)), 1);
```
shenkeyao, line 112: Nit: move these two lines up to line 106 (put before the comment).

crates/task-impls/src/transactions.rs
```diff
@@ -0,0 +1,346 @@
...
+use hotshot_utils::bincode::bincode_opts;
+use snafu::Snafu;
+use std::{collections::HashSet, sync::Arc, time::Instant};
+use tracing::{debug, error, instrument, warn};
+
+#[derive(Snafu, Debug)]
+/// Error type for consensus tasks
+pub struct ConsensusTaskError {}
+
+/// Tracks state of a DA task
```
shenkeyao, line 36: Nit: `DA` -> `transaction`.

crates/task-impls/src/transactions.rs
```diff
@@ -0,0 +1,346 @@
...
+            .collect();
+        Some(txns)
+    }
+
+    /// Event filter for the transaction task
+    pub fn filter(event: &SequencingHotShotEvent<TYPES, I>) -> bool {
+        matches!(
+            event,
+            SequencingHotShotEvent::TransactionsRecv(_)
+                | SequencingHotShotEvent::DAVoteRecv(_)
```
shenkeyao, line 313: I think `DAVoteRecv` should be replaced by `LeafDecided`.

shenkeyao (changes requested):
I think the event filter needs to be updated. Other comments are nit issues.

crates/hotshot/src/tasks/mod.rs
shenkeyao: Or `transactions`? Looks like we use the singular form in struct names, but the plural form in comments.

crates/task-impls/src/transactions.rs
bfish713: Yes good catch, took me a second to realize why the test were passing.  We only remove transactions on decide so we might get duplicates with this bug left in but still will be able to form blocks with transactions.  Fixed now

shenkeyao (dismissed):

rob-maron (approved):

PR #1750, opened 2023-09-14
hotfix publish_proposal_if_able function

elliedavidson:


bfish713:
Is only the network task failing?  I had this issue when I created: https://github.com/EspressoSystems/HotShot/pull/1749.  I think we should disable this test, it seems to be relying on this timeout bug for the consensus task to actually not be able to propose in view 2.  I also don't think it's a very intuitive test to test the network layer since it runs a the full system of consensus tasks and feeds one of those tasks some additional input

PR #1759, opened 2023-09-14
Ed/vote accumulator

elliedavidson:
This PR: 
* Extra tasks not done in this PR are listed here: https://github.com/EspressoSystems/HotShot/issues/1760
* Updates the `Accumulator` trait to take a `VoteType` as an argument, which is cleaner than the previous 7 arguments it used to take
* Makes `Accumulator` types for each certificate type, which allows us to separate certificate forming logic instead of having it all in one, special-cased function
* Updates the accumulators in the consensus, view sync, and da tasks
* Adds more functions to the `VoteType` trait
* Consolidates accumulation logic into only `accumulate_vote_2` and `append`; gets rid of `accumulate_internal`
* Removes `PROPOSAL` and `VOTE` from the `CommunicationChannel` trait 

These changes allow the `TimeoutVote` and `TimeoutCertificates` (coming in the next PR) to more cleanly fit into the existing code.  

Key places to review are `election.rs`, `certificate.rs` and `vote.rs`

This PR does not: 
* Fix all tech debt related to the accumulator and exchanges.  In the future we can further refine these traits.  
* Perfectly generalize `VoteType`; in the future we should add more functions to this trait so that we never call directly into the vote field data outside of a getter function
* Aim to be perfectly clean, since the TimeoutCert PR will follow it.  
* Implement a default accumulator.  That will come in the Timeout PR (Default accumulator could work for DA, Timeout, and VID since they all have the same accumulation rules) 
* Does not remove all dead code


elliedavidson:
@ggutoski just tagging you to let you know that the accumulator logic changed.  I updated the VID sections accordingly.  You should be able to use the default accumulator for VID purposes, but let me know if there are any issues.

crates/types/src/vote.rs
```diff
@@ -255,7 +319,448 @@ pub trait Accumulator<T, U>: Sized {
     fn append(self, val: T) -> Either<Self, U>;
 }
 
+/// Accumulator trait used to accumulate votes into an `AssembledSignature`
+pub trait Accumulator2<
```
dailinsubjam, line 323: Do we want to rename this `Accumulator2` and the old `Accumulator` to something like `Accumulator_basic_vote` and `Accumulator_other_votes`? So that their functionalities are more readable. But this is not a hurry thing I think.

crates/types/src/vote.rs
```diff
@@ -255,7 +319,448 @@ pub trait Accumulator<T, U>: Sized {
...
+            .or_insert_with(|| (0, BTreeMap::new()));
+
+        if finalize_vote_map.contains_key(&encoded_key) {
+            return Either::Left(self);
+        }
+
+        // update the active_keys and sig_lists
+        // TODO ED Possible bug where a node sends precommit vote and then commit vote after
+        // precommit cert is formed, their commit vote won't be counted because of this check
+        // Probably need separate signers vecs.
```
dailinsubjam, line 649: Agree! Looks like they need separate accumulators.

crates/types/src/vote.rs
```diff
@@ -255,7 +319,448 @@ pub trait Accumulator<T, U>: Sized {
...
+        da_vote_map.insert(
+            encoded_key,
+            (vote.get_signature(), vote.get_data(), vote.get_vote_token()),
+        );
+
+        if *da_stake_casted >= u64::from(self.success_threshold) {
+            // Assemble QC
+            let real_qc_pp = <TYPES::SignatureKey as SignatureKey>::get_public_parameter(
+                // TODO ED Something about stake table entries.  Might be easier to just pass in membership?
+                stake_table_entries.clone(),
```
dailinsubjam, line 413: This is relevant to what parameters are needed when [assembling](https://github.com/EspressoSystems/HotShot/blob/main/crates/hotshot-qc/src/bit_vector_old.rs#L71) signatures. If we pass in membership here, then we'll need to extract these stake table entries somewhere deeper, but it's true that our input format of the append val is a little bit messy :( Not sure whether there's a better way to relay the information into append().

crates/types/src/vote.rs
```diff
@@ -255,7 +319,448 @@ pub trait Accumulator<T, U>: Sized {
...
+        stake_table_entries: Vec<<TYPES::SignatureKey as SignatureKey>::StakeTableEntry>,
+    ) -> Either<Self, AssembledSignature<TYPES>> {
+        let VoteData::DA(vote_commitment) = vote.get_data() else {
+            return Either::Left(self);
+        };
+
+        let encoded_key = vote.get_key().to_bytes();
+
+        // Deserialize the signature so that it can be assembeld into a QC
+        // TODO ED Update this once we've gotten rid of EncodedSignature
```
dailinsubjam, line 377: BTW, why do we have `EncodedSignature` originally? Will relaying the serialized format help speed up?

crates/types/src/vote.rs
dailinsubjam, line 1: LGTM! That's a lot of work! There might be some conflicts after I merge in the parameterized curve, but I'll take care of them. I'm also thinking about re-structuring the codes in `append()` so that there aren't too many repeated codes, sorry for not doing it earlier :-(

crates/types/src/vote.rs
dailinsubjam: But this val format is already much better than before! I guess it's fine.

crates/types/src/vote.rs
elliedavidson: I agree!  I put that task in in this issue: https://github.com/EspressoSystems/HotShot/issues/1760. I will work on this issue next! I just wanted to get this merged in a timely fashion since it will cause conflicts with other peoples work.

crates/types/src/vote.rs
elliedavidson: Awesome! And yeah, the append stuff has gotten messy, so it's a handful to deal with.  We can always have someone else take on that task too!  It might be a good onboarding task for someone new, so don't feel you have to do it.

dailinsubjam: Also created an issue for this: https://github.com/EspressoSystems/HotShot/issues/1761

dailinsubjam (approved):

PR #1764, opened 2023-09-15
Sishan/pin jellyfish to previous workable version

dailinsubjam:
Pin to Jellyfish's latest workable and compat version


rob-maron (approved):
Looks good to me!

ggutoski (approved):

Issue #1513, opened 2023-08-07
Parameterizing the curve signature scheme


dailinsubjam:
- [x] Set StakeTableEntry as a type in BN254 SignatureScheme.
- [x] #1514
- [x] Let `validate()` and `sign()` of `SignatureKey` in `hotshot` use the `validate()` and `sign()` from `SignatureScheme` in `jellyfish`. This will close #1515
- [x] https://github.com/EspressoSystems/HotShot/issues/1592
- [x] Cleanup.

dailinsubjam:
Partially done in https://github.com/EspressoSystems/HotShot/pull/1551

PR #1768, opened 2023-09-16
Sishan/merge parameterizing

dailinsubjam:
This PR is a wrap of PR https://github.com/EspressoSystems/HotShot/pull/1767.
And also finish the issue https://github.com/EspressoSystems/HotShot/issues/1513.


dailinsubjam:
> Looks good!
> 
> I think we should make an issue to actually implement the other BLS curve key type and test that this parameterization works. But it isn't a high priority to do this sprint.

Just made one https://github.com/EspressoSystems/HotShot/issues/1771.

PR #1770, opened 2023-09-18
Refactor demo and block payload implementations

shenkeyao:
- Removes `vdemo` (for validating consensus).
- Moves `sdemo` to the upper crate, and renames it to `demo` since it's now the only demo.
- Extract block payload-related implementations from `demo` to `block_impl`.

Closes https://github.com/EspressoSystems/HotShot/issues/1673.


elliedavidson (approved):
Looks good!

crates/hotshot/src/block_impl.rs
```diff
@@ -0,0 +1,174 @@
+//! This module provides an implementation of the `HotShot` suite of traits.
```
ggutoski, line 1: Is this file ported from elsewhere or is it 100% new code?

elliedavidson: It's ported

shenkeyao: Yeah it's from `demo.rs` (originally `demos/sdemo.rs`).

crates/hotshot/examples/web-server-da/multi-validator.rs
```diff
@@ -3,7 +3,7 @@ use async_compatibility_layer::{
     logging::{setup_backtrace, setup_logging},
 };
 use clap::Parser;
-use hotshot::demos::sdemo::SDemoTypes;
+use hotshot::demo::SDemoTypes;
```
jbearer, line 6: Since we renamed `sdemo` to `demo` maybe we should also change `SDemoTypes` to `DemoTypes`?

crates/hotshot/src/block_impl.rs
```diff
@@ -0,0 +1,174 @@
...
+use serde::{Deserialize, Serialize};
+use snafu::Snafu;
+
+/// The transaction in a [`VIDBlockPayload`].
+#[derive(PartialEq, Eq, Hash, Serialize, Deserialize, Clone, Debug)]
+pub struct VIDTransaction {
+    /// identifier for the transaction
+    pub id: u64,
+    /// padding to add to txn (to make it larger and thereby more realistic)
+    pub padding: Vec<u8>,
```
jbearer, line 19: If this is going to be a more generic implementation that is used outside the demo, in real applications, we shouldn't impose any structure beyond just bytes, ie `struct VIDTransaction(Vec<u8>)`

crates/hotshot/src/block_impl.rs
```diff
@@ -0,0 +1,174 @@
...
+
+/// genesis block
+#[derive(PartialEq, Eq, Hash, Serialize, Deserialize, Clone, Debug)]
+pub struct GenesisBlockPayload {}
+
+/// Any block after genesis
+#[derive(PartialEq, Eq, Hash, Serialize, Deserialize, Clone, Debug)]
+pub struct NormalBlockPayload {
+    /// [`BlockPayload`] state commitment
+    pub previous_state: (),
```
jbearer, line 80: Similar comment: maybe I'm misunderstanding the point of this PR, but if this is to be used in other applications we should have any structure in the block payload, just `transactions`

shenkeyao: Sounds good! I'll remove `previous_state`.

jbearer: In a similar vein I don't think we should distinguish between the genesis block and regular blocks here, we should do it at the application level. Unless there's a specific reason within HotShot that this distinction needs to be hard-coded?

crates/hotshot/src/block_impl.rs
shenkeyao: I'll double-check how `id` is used since I think we might still need it for logging/testing purposes. If so, maybe I could add a feature flag.

crates/hotshot/src/block_impl.rs
shenkeyao: There used to be a reason to distinguish them because only a normal block would have a previous state, but that's not the case anymore. I'll combine them, thanks for catching this!

crates/hotshot/src/block_impl.rs
jbearer: Ideally we would build the demo application (which uses `id`?) on top of the VID primitive in the same way we will build the sequencer application on top of it: have a demo-specific transaction type with internal structure, but serialize/deserialize when sending it to/receiving it from HotShot APIs

shenkeyao: It turned out that `id` wasn't used in many places so I just removed it. Agree that we can have a demo-specific transaction type if needed.

crates/hotshot/src/block_impl.rs
```diff
@@ -12,30 +11,17 @@ use snafu::Snafu;
...
+pub struct VIDTransaction(pub Vec<u8>);
 
 impl Committable for VIDTransaction {
     fn commit(&self) -> Commitment<Self> {
-        commit::RawCommitmentBuilder::new("SDemo Txn Comm")
-            .u64_field("id", self.id)
-            .finalize()
+        // TODO: Use use VID block commitment.
+        // <https://github.com/EspressoSystems/HotShot/issues/1730>
+        commit::RawCommitmentBuilder::new("Txn Comm").finalize()
```
jbearer, line 20: How *does* the transaction commitment relate to the block commitment? Does it at all? My understanding is VID just treats the whole block as a blob, without regard for transaction boundaries. Then I think the only use of the transaction hash would be as a unique index, say for filtering duplicate transactions and stuff, peripheral to the core of consensus? In that case I think this can just be a regular Keccak commitment (ie using the commit library as we normally do)

crates/hotshot/src/block_impl.rs
```diff
@@ -44,11 +30,14 @@ impl Transaction for VIDTransaction {}
...
-            padding: vec![],
-        }
+    pub fn new() -> Self {
+        Self(Vec::new())
+    }
+}
+
+impl Default for VIDTransaction {
+    fn default() -> Self {
+        Self::new()
```
jbearer, line 40: Nit: this could be derived

jbearer (approved):
LGTM

crates/hotshot/src/block_impl.rs
shenkeyao: Agree! Ideally, this function could just return the keccak hash, but this would need updates to the `Committable` trait and some other structs that implement `Committable`, so as a short-term fix, I calculated the `Keccak256` hash and then converted it into a `Commitment`.

crates/hotshot/src/block_impl.rs
```diff
@@ -7,17 +7,20 @@ use std::{
...
 pub struct VIDTransaction(pub Vec<u8>);
 
 impl Committable for VIDTransaction {
     fn commit(&self) -> Commitment<Self> {
-        // TODO: Use use VID block commitment.
-        // <https://github.com/EspressoSystems/HotShot/issues/1730>
-        commit::RawCommitmentBuilder::new("Txn Comm").finalize()
+        let builder = commit::RawCommitmentBuilder::new("Txn Comm");
+        let mut hasher = Keccak256::new();
+        hasher.update(self.0.clone());
```
jbearer, line 21: Should be possible to borrow here?

jbearer (dismissed):

crates/hotshot/src/block_impl.rs
shenkeyao: Made an issue for commitment refactoring: https://github.com/EspressoSystems/HotShot/issues/1774.

jbearer (approved):

Issue #673, opened 2022-10-28
Persist libp2p kad state


DieracDelta:
This is for way in the future but it would be better to just save and recover a node's kademlia/DHT state to/from disk to reduce the load on bootstrap nodes.

PR #1773, opened 2023-09-18
Persist Kademlia cache to disk

rob-maron:
aims to close issue #673

### DIFF
--- a/crates/libp2p-networking/src/network/behaviours/dht/cache.rs
+++ b/crates/libp2p-networking/src/network/behaviours/dht/cache.rs
@@ -36,6 +36,12 @@ pub enum CacheError {
         /// source of error
         source: Box<bincode::ErrorKind>,
     },
+
+    /// General cache error
+    GeneralCache {
+        /// source of error
+        source: Box<bincode::ErrorKind>,
+    },
 }
 
 pub struct Config {
@@ -115,8 +121,16 @@ impl Cache {
         let mut cache_to_write = HashMap::new();
         let expiries = self.expiries.read().await;
         for (expiry, key) in &*expiries {
-            let entry = self.cache.get(key).unwrap();
-            cache_to_write.insert(expiry, (key, entry.value().clone()));
+            if let Some(entry) = self.cache.get(key) {
+                cache_to_write.insert(expiry, (key, entry.value().clone()));
+            } else {
+                tracing::warn!("key not found in cache: {:?}", key);
+                Err(CacheError::GeneralCache {
+                    source: Box::new(bincode::ErrorKind::Custom(
+                        "key not found in cache".to_string(),
+                    )),
+                })?;
+            };
         }
 
         let encoded = bincode_opts()
@@ -247,7 +261,7 @@ mod test {
         });
 
         // add 10 key-value pairs to the cache
-        for i in 0 as u8..10 as u8 {
+        for i in 0u8..10u8 {
             let (key, value) = (vec![i; 1], vec![i + 1; 1]);
             cache.insert(key, value).await;
         }
@@ -263,13 +277,13 @@ mod test {
         });
 
         // check that the cache has the 10 key-value pairs
-        for i in 0 as u8..10 as u8 {
+        for i in 0u8..10u8 {
             let (key, value) = (vec![i; 1], vec![i + 1; 1]);
             assert_eq!(cache.get(&key).await.unwrap().value(), &value);
         }
 
         // delete the cache file
-        let _ = std::fs::remove_file("test.cache").unwrap();
+        let _ = std::fs::remove_file("test.cache");
     }
 
     #[cfg_attr(
@@ -303,6 +317,6 @@ mod test {
         assert!(std::path::Path::new("test.cache").exists());
 
         // delete the cache file
-        let _ = std::fs::remove_file("test.cache").unwrap();
+        _ = std::fs::remove_file("test.cache");
     }
 }
--- a/crates/libp2p-networking/src/network/behaviours/dht/cache.rs
+++ b/crates/libp2p-networking/src/network/behaviours/dht/cache.rs
@@ -56,7 +56,7 @@ pub struct Config {
 
 impl Default for Cache {
     fn default() -> Self {
-        Self::new(Config::default())
+        async_block_on(Self::new(Config::default()))
     }
 }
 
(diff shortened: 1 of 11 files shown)


rob-maron:
It's failing some tests right now, so drafting this

crates/libp2p-networking/src/network/behaviours/dht/cache.rs
```diff
@@ -0,0 +1,322 @@
...
+        /// source of error
+        source: Box<bincode::ErrorKind>,
+    },
+}
+
+#[derive(Clone, derive_builder::Builder, custom_debug::Debug, Default)]
+pub struct Config {
+    #[builder(default = "Some(\"dht.cache\".to_string())")]
+    pub filename: Option<String>,
+    #[builder(default = "Duration::from_secs(KAD_DEFAULT_REPUB_INTERVAL_SEC * 16)")]
```
DieracDelta, line 51: Nice use of derive_builder!

crates/libp2p-networking/src/network/behaviours/dht/cache.rs
```diff
@@ -0,0 +1,322 @@
...
+    pub fn new(config: Config) -> Self {
+        let cache = Self {
+            cache: Arc::new(DashMap::new()),
+            expiries: Arc::new(RwLock::new(BTreeMap::new())),
+            config,
+            disk_parity_delta: Arc::new(AtomicU32::new(0)),
+        };
+
+        // try loading from file
+        if let Err(err) = async_block_on(cache.load()) {
```
DieracDelta, line 86: Why `async_block_on` instead of making the function async?

crates/libp2p-networking/src/network/behaviours/dht/cache.rs
```diff
@@ -0,0 +1,322 @@
...
+            if now > expires {
+                self.cache.remove(&key);
+                removed += 1;
+            } else {
+                expiries.insert(expires, key);
+                break;
+            }
+        }
+
+        // todo: test speed on this as opposed to inline
```
DieracDelta, line 160: > as opposed to inline
What would be inlined?

crates/libp2p-networking/src/network/behaviours/dht/cache.rs
```diff
@@ -0,0 +1,322 @@
...
+        // save if reached max disk parity delta
+        self.disk_parity_delta.fetch_add(1, Ordering::Relaxed);
+        self.save_if_necessary().await;
+    }
+
+    async fn save_if_necessary(&self) {
+        let cur_disk_parity_delta = self.disk_parity_delta.load(Ordering::Relaxed);
+        if cur_disk_parity_delta >= self.config.max_disk_parity_delta {
+            if let Err(err) = self.save().await {
+                tracing::warn!("failed to save cache to file: {}", err);
```
DieracDelta, line 192: Might be worth making this an error. If it's not able to save the cache to a file, there's likely some sort of configuration failure going on (or the disk is out of space). Either way, loud failure seems good.

crates/libp2p-networking/src/network/behaviours/dht/cache.rs
```diff
@@ -0,0 +1,322 @@
...
+            };
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
```
DieracDelta, line 199: 🎉 you added tests!!

crates/libp2p-networking/src/network/node/handle.rs
```diff
@@ -51,10 +48,6 @@ pub struct NetworkNodeHandle<S, K: Debug + Eq + Hash + Serialize + Clone> {
     peer_id: PeerId,
     /// human readable id
     id: usize,
-    /// the cache for peers we've looked up
-    peer_cache: Arc<DashMap<K, PeerId>>,
-    /// the expiries for the peer cache, in order
-    peer_cache_expiries: Arc<RwLock<BTreeMap<SystemTime, K>>>,
```
DieracDelta, line 57: I'm glad to see this gone!

flake.nix
```diff
@@ -173,7 +173,7 @@
             zlib.out
             fenix.packages.${system}.rust-analyzer
             just
-            pkgconfig
```
DieracDelta, line 176: Oof did pkgconfig not exist? What's the difference between the two :eyes:

DieracDelta (approved):
LGTM! Great work.

flake.nix
rob-maron: The name had just been replaced with pkg-config! But otherwise it's the same. We were getting build errors from it

crates/libp2p-networking/src/network/behaviours/dht/cache.rs
rob-maron: Good point 👍

rob-maron: Fixed this, but had to pin the `network_handle` because it was a large future. Let me know if that's something we're not okay with!

crates/libp2p-networking/src/network/behaviours/dht/cache.rs
rob-maron: I had meant doing a `fetch_add` inside the loop as opposed to tracking how many changes and then doing one atomic add at the end. Definitely wouldn't make a big difference, probably don't have to "todo" this lol, will remove it

flake.nix
DieracDelta: Good catch!

crates/libp2p-networking/src/network/behaviours/dht/cache.rs
DieracDelta: If that's what rustc says then it should be fine. But, I'm kind of surprised that it needs to be pinned. Could it just be a Box? The idea is to stick it on the heap so we don't clog up the stack, right?

rob-maron: It's clippy, with the warning located here: 
https://rust-lang.github.io/rust-clippy/master/index.html#/large_futures

According to the warning, it can sometimes cause stack overflows. I wasn't able to make it happy with a Box. We can ignore this clippy warning, but the consequence listed in the clippy error is possible stack overflow 😮 

I think we can probably do one of the following:
1. Put the `async_block_on` back for the `cache.load()` function, or perhaps further down
2. Keep it pinned
3. Refactor cache to use only synchronous operations

What do you think?

DieracDelta: Upon further inspection, (2) makes sense to me. We're pinning the future, then immediately resolving it. So, there's no funky reasoning about whether or not it remains pinned. Seems fine to leave as is!

DieracDelta (approved):

PR #1777, opened 2023-09-19
chore: Remove Committable from VoteData

### LINKED ISSUE #1772: T2.1.1: Unnecessary double-commit in VoteData
Currently we are needlessly computing commitments of commitments in the impl of `Committable` for `VoteData`.

All variants of `VoteData` are of type `Commitment<COMMITTABLE>`. In other words: they are already commitments.

https://github.com/EspressoSystems/HotShot/blob/598b22e33a402913c2e2d89af3c4359015e1b61c/crates/types/src/traits/election.rs#L77-L79

However, the impl of `Committable` computes another commitment of the commitment:

https://github.com/EspressoSystems/HotShot/blob/598b22e33a402913c2e2d89af3c4359015e1b61c/crates/types/src/traits/election.rs#L95-L102

Seems that `VoteData` should not impl `Committable`. Or, at the very least, its impl of `Committable` should just return the commitment that's already stored in `VoteData`.

ggutoski:
fix #1772

### DIFF
--- a/crates/types/src/traits/election.rs
+++ b/crates/types/src/traits/election.rs
@@ -91,6 +91,20 @@ pub enum VoteData<COMMITTABLE: Committable + Serialize + Clone> {
     ViewSyncFinalize(Commitment<COMMITTABLE>),
 }
 
+impl<COMMITTABLE> VoteData<COMMITTABLE>
+where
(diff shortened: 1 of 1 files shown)


shenkeyao (approved):
LGTM!

PR #1779, opened 2023-09-19
chore: remove Committable from VoteData type param

### LINKED ISSUE #1778: T2.1.2: VoteData type param should not be Committable
This issue is a baby step toward #1757. The idea is to do a sequence of small changes, each of which passes tests and is easy to review.

Currently `VoteData` takes a type param `COMMITTABLE: Committable`, but all enum variants use only `Commitment<COMMITTABLE>`.

https://github.com/EspressoSystems/HotShot/blob/8ba608416b8f20c6d6a86d676588844ea1d69c23/crates/types/src/traits/election.rs#L77-L79

Instead, the type param should be used directly in the enum variants. This allows `VoteData` to be used for things that are not `Committable` as required by #1757 .

Moreover, I think it's good practice to avoid `Committable` type params when we really need only their commitments. Doing so makes it hard for future edits to accidentally rely nontrivially on the underlying `COMMITTABLE` type.

ggutoski:
fix #1778 

Also, fix libp2p examples build break.


shenkeyao (approved):

ggutoski:
[CI again](https://github.com/EspressoSystems/HotShot/actions/runs/6240391458/job/16940398765?pr=1779#step:7:1141) this is a new one:

> LLVM ERROR: IO failure on output stream: No space left on device
> error: could not compile hotshot-task (lib test)

Tests pass locally. Merging despite CI.

PR #1783, opened 2023-09-20
Reduce build times and disk size for CI

rob-maron:


elliedavidson (approved):
LGTM!  We add `nll` back in the future, but I'm fine deleting it for now.

Issue #1503, opened 2023-08-04
Remove Transaction Handling logic


bfish713:
any place we update `consensus.transactions` should be removed and only happen in the builder task

PR #1786, opened 2023-09-20
Remove `transactions` from `Consensus` struct

bfish713:
We only use transactions in the new transactions task so it doesn't need to be in the consensus type.  This avoids write locking consensus in the transactions task.  We still need a read lock to update the metrics but this should still be an improvement.  Also transactions no longer belonged in the main consensus struct because transactions should be the responsibility of the builder.

Closes: https://github.com/EspressoSystems/HotShot/issues/1503


shenkeyao (approved):
LGTM!

PR #1789, opened 2023-09-20
chore: remove Committable from 3 more types

### LINKED ISSUE #1788: T2.1.3: Remove Committable from type params for VoteMap, VoteAccumulator, VoteType
This issue is like #1778, except for `VoteMap`, `VoteAccumulator`, and `VoteType`.

Currently those types each take a type param `COMMITTABLE: Committable`, but they all use only `Commitment<COMMITTABLE>`.

https://github.com/EspressoSystems/HotShot/blob/2afb7e544d597088905a70e987733a5f99867126/crates/types/src/vote.rs#L761-L767

https://github.com/EspressoSystems/HotShot/blob/2afb7e544d597088905a70e987733a5f99867126/crates/types/src/vote.rs#L771

https://github.com/EspressoSystems/HotShot/blob/2afb7e544d597088905a70e987733a5f99867126/crates/types/src/vote.rs#L31-L32

Instead, the type param should be used directly in all these types. This allows these types to be used for things that are not `Committable` as required by #1757 .

Moreover, I think it's good practice to avoid `Committable` type params when we really need only their commitments. Doing so makes it hard for future edits to accidentally rely nontrivially on the underlying `COMMITTABLE` type.

ggutoski:
close #1788

### DIFF
--- a/crates/hotshot/src/tasks/mod.rs
+++ b/crates/hotshot/src/tasks/mod.rs
@@ -5,7 +5,7 @@ use crate::{
     QuorumCertificate, SequencingQuorumEx,
 };
 use async_compatibility_layer::art::async_sleep;
-use commit::Committable;
+use commit::{Commitment, Committable};
 use futures::FutureExt;
 use hotshot_task::{
     boxed_sync,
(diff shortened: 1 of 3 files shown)


shenkeyao (approved):
LGTM

ggutoski:
Tests pass locally. I assume the CI failure is a symptom of our broader CI issues, so I'll merge anyway.

PR #1794, opened 2023-09-21
Ed/timeout

elliedavidson:
This PR: 
* Starts hotshot with a `QCFormed` event instead of `ViewChange` event
* Updates view logic such that replicas enter new views when they see a QC, TC, or ViewSyncCert for the previous view.  They no longer increment their view after successfully voting on a proposal (see this thread [here](https://espresso.zulipchat.com/#narrow/stream/311646-Consensus/topic/.E2.9C.94.20Timeout.20Logic) for more detail)
* Updates view sync logic to handle the new view change logic - view sync is now triggered after 3 timeouts: 1st timeout indicates we didn't see a QC for the next view and should send a timeout certificate, second timeout is waiting for said timeout certificate, third time is when a second leader in a row is unresponsive, meaning we should view sync
* Removes lots of dead code
* Replaces "2" types with their native types
* Updates `VoteType` trait with more getter functions
* Renames `from_signatures_and_commitment` to `create_certificate`
* Adds a `TimeoutExchange` to facilitate `TimeoutVotes` but does not expose it outside of `NodeImplementation`.  We can do the same for the other exchanges in this issue: https://github.com/EspressoSystems/HotShot/issues/1799
* Adds timeout tests for both the web server and libp2p networks 

This PR does not: 
* Accumulate accumulators: https://github.com/EspressoSystems/HotShot/issues/1797. We should move toward a single default accumulator implementation to get rid of redundant code
* Unit tests for timeout: https://github.com/EspressoSystems/HotShot/issues/1798
* Fix previous known inefficiencies in the consensus code (even though it touches most of that code) since that would be too large a PR

Use `just async_std test_timeout` to run the timeout test


crates/types/src/vote.rs
```diff
@@ -721,249 +820,11 @@ impl<TYPES: NodeType, COMMITMENT: CommitmentBounds, VOTE: VoteType<TYPES, COMMIT
...
         u64,
         BTreeMap<EncodedPublicKey, (EncodedSignature, VoteData<COMMITMENT>, TOKEN)>,
     ),
 >;
-
-/// Describe the process of collecting signatures on block or leaf commitment, to form a DAC or QC,
-/// respectively.
-///
-/// TODO GG used only in election.rs; move this to there and make it private?
-pub struct VoteAccumulator<TOKEN, COMMITMENT: CommitmentBounds, TYPES: NodeType> {
```
bfish713, line 761: Love getting rid of this!

crates/types/src/vote.rs
```diff
@@ -721,249 +820,11 @@ impl<TYPES: NodeType, COMMITMENT: CommitmentBounds, VOTE: VoteType<TYPES, COMMIT
...
-                ),
-            ),
-        ),
-        AssembledSignature<TYPES>,
-    > for VoteAccumulator<TOKEN, COMMITMENT, TYPES>
-where
-    TOKEN: Clone + VoteToken,
-{
-    #![allow(clippy::too_many_lines)]
-    fn append(
```
bfish713, line 807: I see append implemented for TimeCertificate, did we already have it for the other certificate types, I can't find it in this PR and am confused since it obviously works

bfish713 (changes requested):

crates/task-impls/src/consensus.rs
```diff
@@ -572,241 +672,157 @@ where
...
-                                error!(?proposal.signature, "Could not verify proposal.");
-                                return;
-                            }
+                // Justify qc's leaf commitment is not the same as the parent's leaf commitment, but it should be (in this case)
+                let Some(parent) = parent else {
+                    error!(
+                        "Proposal's parent missing from storage with commitment: {:?}",
+                        justify_qc.leaf_commitment()
+                    );
+                    return;
```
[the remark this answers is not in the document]
elliedavidson: Latest push should fix this. Instead of return right away here, it first updates the state map and then returns. Let me know if it still incorrect.

crates/task-impls/src/da.rs
```diff
@@ -510,10 +508,11 @@ where
...
                 // cause an overflow error.
-                if view < self.cur_view - 1 {
-                    warn!("Throwing away VID disperse data that is more than one view older");
-                    return None;
-                }
+                // TODO ED Revisit this
+                // if view < self.cur_view - 1 {
+                //     warn!("Throwing away VID disperse data that is more than one view older");
+                //     return None;
+                // }
```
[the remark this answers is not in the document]
elliedavidson: Yes, put it back in with logic to account for the genesis view.

crates/testing/tests/timeout.rs
```diff
@@ -21,28 +23,41 @@ async fn test_timeout() {
...
     };
-    let mut metadata = TestMetadata::default();
+
+    // TODO ED Reduce down to 5 nodes once memory network issues is resolved
+    // https://github.com/EspressoSystems/HotShot/issues/1790
+    let mut metadata = TestMetadata {
+        total_nodes: 10,
+        start_nodes: 10,
+        ..Default::default()
+    };
```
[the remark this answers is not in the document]
elliedavidson: I still get the same error.  I actually ran the timeout test on the `develop` branch.  Looks like the same thing happens.  The test works with the web server network but fails with the memory network.  Looking at the logs it looks like some messages just don't make it to the nodes like they are supposed to.

crates/testing/tests/timeout.rs
```diff
@@ -21,28 +23,41 @@ async fn test_timeout() {
...
         idx: 0,
         updown: UpDown::Down,
     }];
 
     metadata.timing_data = timing_data;
 
-    metadata.spinning_properties = SpinningTaskDescription {
-        node_changes: vec![(Duration::new(0, 5000), dead_nodes)],
+    metadata.overall_safety_properties = OverallSafetyPropertiesDescription {
+        num_successful_views: 25,
```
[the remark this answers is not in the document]
elliedavidson: Tried this.  I think we should keep it 25 for now.  Since views are timing out very regularly, it doesn't go through views as fast.  Also the test uses the web server, which is a bit slower anyway.  If I change the test to run for 60 seconds or so, then it can get through 50 views.

crates/types/src/vote.rs
elliedavidson: Yes, they were in a previous PR!

crates/testing/tests/timeout.rs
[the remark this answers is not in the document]
bfish713: Hmm ok I'll look into this one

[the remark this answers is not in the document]
elliedavidson: I did update the timeout_test to have both libp2p and the webserver, though. Just to ensure something weird wasn't also happening with one of them.

crates/hotshot-signature-key/src/bn254/bn254_priv.rs
```diff
@@ -54,6 +54,7 @@ impl BLSPrivKey {
     }
 }
 
+// #[allow(clippy::incorrect_partial_ord_impl_on_ord_type)]
```
dailinsubjam, line 57: Do we want to get rid of this dead code?

crates/libp2p-networking/src/network/node/handle.rs
```diff
@@ -136,6 +136,9 @@ impl<S: Default + Debug> NetworkNodeHandle<S> {
     ///
     /// Will panic if a handler is already spawned
     #[allow(clippy::unused_async)]
+    // // Tokio and async_std disagree how this function should be linted
+    // #[allow(clippy::ignored_unit_patterns)]
```
dailinsubjam, line 140: Also dead code here?

dailinsubjam (approved):
Except for some dead codes, all LGTM!

crates/hotshot-signature-key/src/bn254/bn254_priv.rs
elliedavidson: I kept this in because my local clippy needed it.  I expect once CI's clippy is upgraded we'll need to uncomment these.  :)

bfish713 (approved):

crates/task-impls/src/consensus.rs
```diff
@@ -581,245 +681,192 @@ where
...
-                                timestamp: time::OffsetDateTime::now_utc().unix_timestamp_nanos(),
-                                proposer_id: sender.to_bytes(),
-                            };
-                            if !view_leader_key
-                                .validate(&proposal.signature, leaf.commit().as_ref())
-                            {
-                                error!(?proposal.signature, "Could not verify proposal.");
-                                return;
-                            }
+                    return;
```
bfish713, line 763: In the fix branch we were updating the view here to avoid timing out in the view that might have actually been successful.  I'm not 100% sure either way so I'll leave up to you.  If we don't update the view like in this PR, then the view will time out and the network might move on to future views without us and we'll be stuck trying to catch back up.

If we do update the view then we might time out on a future view when everyone else times out because this is a bogus QC that nobody voted on.

bfish713: If after this is deployed and we see one node constantly behind, this is probably the culprit.  On the other hand if we see nodes entering view sync at different times in the current deployment this change would fix it.

Issue #1787, opened 2023-09-20
Refactor block payload and VID disperse to get the VID commitment


shenkeyao:
- Add VID commitment to `VIDBlockPayload`
- Update `Committable` impl to use the pre-computed VID commitment
- Get the VID disperse before constructing the block
  - Calculate it from the txn list, instead of the message bytes
  - Probably just delete `VidDisperse::commitment` if it's not needed
- Remove add_transaction_raw from BlockPayload trait
- Move VID block construction to transaction.rs task

Issue #1763, opened 2023-09-15
Investigate the use of `DummyBlock`


shenkeyao:
Check if we still need `DummyBlock`. If we can remove it, `SDemoBlock` (may be renamed to `VIDBlockPayload`) will be the only struct that implements `BlockPayload`. If we don't need other implementations (e.g., for `TimeBoost`), then we can convert `BlockPayload` from a trait to a struct.

PR #1814, opened 2023-09-25
Refactor block payload and VID disperse

shenkeyao:
- Updates `VIDBlockPayload`.
  - Adds VID commitment.
  - Updates `Committable` impl for `VIDBlockPayload`.
- Removes `add_transaction_raw` from `BlockPayload` trait.
- Updates VID disperse.
  - Moves the construction from the DA task to the transaction task.
  - Gets VID first, then the block and proposal.
  - Uses transaction bytes instead of message bytes for VID construction.

Closes https://github.com/EspressoSystems/HotShot/issues/1787.
Closes https://github.com/EspressoSystems/HotShot/issues/1588.


crates/hotshot/src/demo.rs
```diff
@@ -107,9 +101,9 @@ impl TestableState for SDemoState {
     fn create_random_transaction(
         _state: Option<&Self>,
         _rng: &mut dyn rand::RngCore,
-        padding: u64,
+        _padding: u64,
     ) -> <Self::BlockType as BlockPayload>::Transaction {
-        VIDTransaction(vec![0; padding as usize])
```
jbearer, line 112: Why did we get rid of padding? Isn't this used for benchmarking?

crates/hotshot/src/lib.rs
```diff
@@ -634,7 +634,7 @@ pub trait HotShotType<TYPES: NodeType, I: NodeImplementation<TYPES>> {
 
 #[async_trait]
 impl<
-        TYPES: NodeType,
+        TYPES: NodeType<Transaction = VIDTransaction, BlockType = VIDBlockPayload>,
```
jbearer, line 637: This was my concern about making these abstract when we generally only want one implementation. If we are fixing `Transaction` and `BlockType` in library code as general-purpose as this, they probably shouldn't be associated types to begin with. WDYT?

crates/task-impls/src/transactions.rs
```diff
@@ -233,18 +235,44 @@ where
...
                     }
+                    let vid_disperse = vid.disperse(&txns_flatten).unwrap();
+                    let block = VIDBlockPayload::new(txns, vid_disperse.commit);
+
+                    self.event_stream
+                        .publish(SequencingHotShotEvent::BlockReady(block.clone(), view + 1))
+                        .await;
+
+                    self.event_stream
+                        .publish(SequencingHotShotEvent::VidDisperseSend(
```
ggutoski, line 258: Should we publish `VidDisperseSend` here, or later in the DA task? Note that `DAProposalSend` is still published in the DA task, and `VidDisperseSend` is intended to be a mirror of `DAProposalSend` for VID. ie. the leader should send the DA proposal at the same time he sends VID disperse data.

If we move it to DA task then the we'll need to add the VID shares to the `BlockReady` event. That raises another question: should we limit the volume of data passed through events? eg. do we do something silly in event processing that takes O(n) time like serialize?

We could address these questions in a future issue, in which case a simple TODO would suffice here.

crates/task-impls/src/transactions.rs
```diff
@@ -254,29 +282,52 @@ where
...
+        A: SequencingConsensusApi<TYPES, SequencingLeaf<TYPES>, I> + 'static,
+    > TransactionTaskState<TYPES, I, A>
+where
+    CommitteeEx<TYPES, I>: ConsensusExchange<
+        TYPES,
+        Message<TYPES, I>,
+        Certificate = DACertificate<TYPES>,
+        Commitment = TYPES::BlockType,
+    >,
+{
```
ggutoski, line 303: Why a new impl block here? Did you relax trait bounds or something? Please comment.

crates/testing/tests/atomic_storage.rs
```diff
@@ -13,76 +13,6 @@ use rand::thread_rng;
 
 type AtomicStorage = hotshot::traits::implementations::AtomicStorage<DEntryState>;
 
-#[cfg_attr(
```
ggutoski, line 16: Why delete this test?

crates/hotshot/src/lib.rs
ggutoski: I agree that we should not over-generalize. If we only ever expect 1 tx type and 1 block type then just hard-code them. Perhaps an argument could be made that traits are useful so that it's easy to mock things in tests. But I don't like adding complexity to production code merely to facilitate tests.

Regarding associated types: This is a question about idiomatic rust. In this instance we are using assoc types as if they were type params. Presumably that's not idiomatic, but there are reasons why people do it. eg. If there are many type params then the code becomes unreadable---I assume that's why we have a `NodeType` trait to collect all the type params into a single trait so that every single type doesn't need 47 type params. I have not yet formed a strong opinion here.

jbearer: Sorry, I wasn't complaining about these being associated types specifically. I actually don't think they should be generic at all (whether as type parameters or associated types) if we have core library code that doesn't work unless one specific instantiation (`VIDTransaction`, `VIDBlockPayload`) is used. Like, HotShot is generic over `Transaction` and `BlockType` but it seems impossible to use HotShot unless your are using the VID implementations...we should either fix that or just hard code these two specific types.

crates/hotshot/src/demo.rs
shenkeyao: Good catch! I wanted to avoid the `padding = 0` case, since the VID calculation wouldn't work for an empty vector. And I thought the padding wasn't used so I just removed it, which now I realized wasn't correct. Will update this.

crates/hotshot/src/lib.rs
shenkeyao: There are two cases where we may want to keep the two traits.
1. We have `DummyBlock` that also implements `BlockPayload`.

2. We may need another `BlockPayload` impl in the future. Pasting the note from an earlier [discussion on Zulip](https://espresso.zulipchat.com/#narrow/stream/311646-Consensus/topic/.E2.9C.94.20use.20of.20VID.20block.20commitment/near/391205141):

> I just had a quick discussion with the consensus team and Ellie mentioned we might need two BlockPayload implementations for TimeBoost (undecided yet, but it's possible), so I'll do the renaming but keep BlockPayload as a trait for more flexibility.
We could convert it into a struct if we figure we only need one implementation later.

For 1, we have a GitHub [issue](https://github.com/EspressoSystems/HotShot/issues/1763) to investigate whether we can remove the dummy struct, and I could do it soon. For 2, can we wait until @elliedavidson is back from vacation, then decide whether we can simplify this?

crates/hotshot/src/demo.rs
ggutoski: > ... the VID calculation wouldn't work for an empty vector.

Sounds like a bug in jellyfish. We can use a single zero-byte instead until it's fixed.

crates/types/src/block_impl.rs
```diff
@@ -55,14 +61,44 @@ pub enum BlockPayloadError {
...
+    /// Create a genesis block payload with transaction bytes `vec![0]`.
+    /// # Panics
+    /// If the `VidScheme` construction fails.
+    #[must_use]
+    pub fn genesis() -> Self {
+        // TODO <https://github.com/EspressoSystems/HotShot/issues/1686>
+        let srs = test_srs(NUM_STORAGE_NODES);
+        let vid = VidScheme::new(NUM_CHUNKS, NUM_STORAGE_NODES, &srs).unwrap();
+        let txn = vec![0];
+        let vid_disperse = vid.disperse(&txn).unwrap();
```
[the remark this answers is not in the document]
shenkeyao: Yeah in `add_consensus_task` in mod.rs, a genesis block is needed to build the consensus task. I'll add a TODO!

I agree that we don't really need the VID dispersal for the genesis block, but to construct a block, we need the `commitment` field which requires the VID computation.

crates/types/src/block_impl.rs
```diff
@@ -91,20 +127,10 @@ impl BlockPayload for VIDBlockPayload {
 
     type Transaction = VIDTransaction;
 
-    fn new() -> Self {
-        <Self as TestableBlock>::genesis()
-    }
-
-    fn add_transaction_raw(
```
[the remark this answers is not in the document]
shenkeyao: I added #1588 to the description of this PR! Merging this PR should close it.

crates/types/src/block_impl.rs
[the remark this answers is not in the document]
shenkeyao: > serialized genesis block body will not be a single zero-byte

Sorry can I double-check on this--did you mean that we would replace the zero-byte with something else?

crates/types/src/traits/block_contents.rs
```diff
@@ -63,6 +49,8 @@ pub trait BlockPayload:
 pub trait Transaction:
     Clone + Serialize + DeserializeOwned + Debug + PartialEq + Eq + Sync + Send + Committable + Hash
 {
+    /// Get the transaction bytes.
+    fn bytes(&self) -> Vec<u8>;
```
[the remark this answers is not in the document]
shenkeyao: I think `bytes` here isn't the same as serialization, since it's getting the inner vector of `Transaction` (i.e., `self.0` for `VIDTransaction`), rather than serializing the whole `Transaction`.

Maybe `inner` or `inner_bytes` is a better name to avoid the confusion.

crates/task-impls/src/transactions.rs
shenkeyao: Yes! One of them has additional bounds on `TYPES: NodeType`. I'll add a comment.

crates/testing/tests/atomic_storage.rs
shenkeyao: It's unrelated to the purpose of this PR, but when trying to modify the types in this test, I found it outdated and confirmed with the team that we could just delete it.

crates/hotshot/src/demo.rs
shenkeyao: I'm going to add a `panic!` if `padding == 0` (I think panic is fine here since it's test code.), then use the given padding rather than fixing the size.

crates/types/src/block_impl.rs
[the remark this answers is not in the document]
ggutoski: It's fine for now. Anything we put here will be changed later

crates/task-impls/src/transactions.rs
ggutoski: Any opinion here @bfish713 ? I'm not yet familiar with the philosophy behind the event model and the purpose of `BlockReady` .

crates/types/src/traits/block_contents.rs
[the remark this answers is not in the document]
ggutoski: > I think `bytes` here isn't the same as serialization, since it's getting the inner vector of `Transaction` (i.e., `self.0` for `VIDTransaction`), rather than serializing the whole `Transaction`.

What's the difference? A transaction is just a vec<u8> so wouldn't its serialization be identical to self.0?

Do we have reason to require two different ways to get the bytes of a tx?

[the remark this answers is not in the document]
shenkeyao: Agreed. Actually, this function is only used to chain txn bytes for VID dispersal computation, which IIUC will be unnecessary once jellyfish supports passing in a vec of transactions. But anyway, I think it makes sense to choose either `self.0` or `serde`, not both.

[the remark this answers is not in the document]
ggutoski: Jellyfish will take an iterator over bites, but you still need bites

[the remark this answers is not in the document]
ggutoski: An alternative is to impl AsRef<u8>

shenkeyao:
PR comments have been addressed. I'll sync with main and resolve merge conflicts later.

crates/hotshot/src/demo.rs
ggutoski: This new `panic!` causes `test_success` to fail. (I don't know how you got CI to skip tests in commit https://github.com/EspressoSystems/HotShot/pull/1814/commits/2d51285e4f05033f0e156823310d8dba95622a26 ! I thought you need to prefix the commit message with `[no ci]` to accomplish that.)

At first glance it seems the purpose of `padding` is to append some extra noise to the end of a (randomly generated) tx with the understanding that the original tx already has nonzero size and could be further enlarged via `padding`.

But the impl of `create_random_transaction` in this PR starts with a *zero-size* tx, so the *only* data in the tx is due to `padding`. Thus, if `padding` is zero then the entire tx is zero.

Seems we could solve the problem by picking a nonzero minimum size `s` for this "random" tx and then return a tx with size `s + padding`. Make sense?

ggutoski: > > ... the VID calculation wouldn't work for an empty vector.
> 
> Sounds like a bug in jellyfish. We can use a single zero-byte instead until it's fixed.

upstream fix: https://github.com/EspressoSystems/jellyfish/pull/377

shenkeyao: With this update, `test_success` passes without error logs! I'll update the jellyfish dependency in the cargo file after your PR is merged.

shenkeyao: Hmm but `test_catchup_web` fails. I'll investigate.

shenkeyao: The second run passes, so looks like it's the nondeterministic issue we have on main as well.

ggutoski (approved):

PR #1825, opened 2023-09-27
Fix transactions

### LINKED ISSUE #1766: Investigate examples not processing TXes
Both the `web-server` and `libp2p` examples report 0 transactions being processed upon successful views.

### LINKED ISSUE #1821: Fix transaction submission in examples
Fix examples to make them submit transactions every few seconds for benchmarking.

rob-maron:
Examples were not processing transactions properly.
Temporary fix is to factor in the payload for the commitment

closes #1821 and closes #1766


crates/hotshot/examples/infra/modDA.rs
```diff
@@ -313,6 +313,13 @@ pub trait RunDA<
                                 }
                             }
 
+                            // send transactions
+                            for _ in 0..(total_nodes_u64 / transactions_per_round as u64) {
```
shenkeyao, line 317: I'm wondering why it's `total_nodes_u64 / transactions_per_round`. E.g., if there are 10 nodes and it's 2 txn/round, do we submit transactions 5 times? I think it should be `total_transactions / transactions_per_round`, but maybe I'm misunderstanding something.

rob-maron: Yes sorry, totally missed that . Fixed !

crates/hotshot/examples/infra/modDA.rs
```diff
@@ -284,6 +283,9 @@ pub trait RunDA<
         let mut num_successful_commits = 0;
 
         let total_nodes_u64 = total_nodes.get() as u64;
+        let mut total_transactions_sent = 0;
+        let transactions_to_send_per_round =
+            (transactions_per_round as f64 / total_nodes_u64 as f64).ceil() as u64;
```
shenkeyao, line 288: Sorry I'm still confused about this. Why do we divide `transactions_per_round` by `total_nodes_u64` rather than use `transactions_per_round` directly? I don't quite understand why the number of transactions is related to the number of nodes.

rob-maron: Ah, that's because each node running it enters that loop, and the txs_per_view I believe is used independently of the number of nodes. So we have to factor that in.

So if we have 10 rounds and 10 transactions per view, it should be around 100 total txs committed. And each node would commit approx 10 (1 per round)

shenkeyao: Oh I see! So a round isn't a view, but multiple views with different nodes as the leader.

Let me try to use some other numbers to distinguish which "10" is which: If `transactions_per_round` is 120, and there are 10 nodes, then `transactions_to_send_per_round` is 12, meaning each node would submit 12--is this example correct? If so I think this implementation makes sense!

rob-maron: Correct ! But I believe round is still synonymous with view in this case. `transaction_per_round` is more the total transactions per round, so we need to split between the nodes.

I realized that the implementation I had is quite primitive, so I just split out a function to do it. It also has the first node send the remaining transactions that did not split out evenly. Let me know what you think!

shenkeyao (approved):
Looks good!

shenkeyao (approved):

PR #1835, opened 2023-09-29
fix: revert to using fixed val for transactions

rob-maron:
The CI for async_std is operating with odd behavior with the randomness

### DIFF
--- a/crates/hotshot/src/demo.rs
+++ b/crates/hotshot/src/demo.rs
@@ -105,9 +105,7 @@ impl TestableState for SDemoState {
     ) -> <Self::BlockType as BlockPayload>::Transaction {
         /// clippy appeasement for `RANDOM_TX_BASE_SIZE`
         const RANDOM_TX_BASE_SIZE: usize = 8;
-        let mut bytes = vec![0; RANDOM_TX_BASE_SIZE + (padding as usize)];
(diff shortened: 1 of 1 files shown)


crates/hotshot/src/demo.rs
```diff
@@ -100,14 +100,12 @@ impl State for SDemoState {
...
-        rng: &mut dyn rand::RngCore,
+        _rng: &mut dyn rand::RngCore,
         padding: u64,
     ) -> <Self::BlockType as BlockPayload>::Transaction {
         /// clippy appeasement for `RANDOM_TX_BASE_SIZE`
         const RANDOM_TX_BASE_SIZE: usize = 8;
-        let mut bytes = vec![0; RANDOM_TX_BASE_SIZE + (padding as usize)];
-        rng.fill_bytes(&mut bytes);
-        VIDTransaction(bytes)
+        VIDTransaction(vec![0; RANDOM_TX_BASE_SIZE + (padding as usize)])
```
shenkeyao, line 108: I thought without the randomness, there was a test failure. Was it fixed?

rob-maron: Yeah, I think the implementing the nonstatic commit() for VID fixed this :)

shenkeyao (approved):

PR #1847, opened 2023-10-02
quick fix: in_flight test nondeterminism

### LINKED ISSUE #1857: memory_network_test_in_flight_message_count hangs nondeterministically on main branch
Make it easier to reproduce: add the following to your `justfile`:
```
test_memory_network:
  echo Testing the DA task with async std executor
  ASYNC_STD_THREAD_COUNT=1 cargo test --lib --bins --tests --benches --workspace --no-fail-fast memory_network -- --test-threads=1 --nocapture
```
Then run
```
just async_std test_memory_network
```
Need to run ~5 times to observe the test hang, otherwise it passes with no error.

Assigning to @rob-maron because he was the last person to touch this code. 😄

rob-maron:
Test was occasionally hanging when using the Tokio runtime 

> aims to close #1857

### DIFF
--- a/crates/testing/tests/memory_network.rs
+++ b/crates/testing/tests/memory_network.rs
@@ -358,9 +358,12 @@ async fn memory_network_test_in_flight_message_count() {
         network1.recv_msgs(TransmitType::Broadcast).await.unwrap();
     }
 
+    while network2.in_flight_message_count().unwrap() > messages.len() {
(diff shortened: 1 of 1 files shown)


ggutoski (approved):
No idea how this test works. But the change is tiny and if it fixes the hanging issue then LGTM. Thanks for fixing!

PR #1872, opened 2023-10-09
Ed/timeout hotfix merge

elliedavidson:
Merges the changes in the 'ed/timeout-hotfix' branch into `main'
Also updates Nix flake for the new MacOS update


crates/task-impls/src/consensus.rs
```diff
@@ -426,8 +428,13 @@ where
                         // Validate the DAC.
                         if self
                             .committee_exchange
-                            .is_valid_cert(cert, proposal.block_commitment)
+                            .is_valid_cert(cert)
                         {
+                            // Validate the block commitment for non-genesis DAC.
+                            if !cert.is_genesis() && cert.leaf_commitment() != proposal.block_commitment {
```
shenkeyao, line 434: `is_genesis` always returns `false` so is redundant. But the fact that it always returns false might be an issue. I'll make a GH issue for further discussion/fix, so no need to change anything, just a heads-up. :)

shenkeyao (approved):

Issue #1900, opened 2023-10-10
[hotshot-state-prover] threshold checking


mrain:
Checks whether the signer's accumulated stake amount exceed a threshold.

PR #1901, opened 2023-10-10
threshold checking for hotshot state prover

mrain:
closes: #1900 

Should be simple enough


mrain:
Circuit logic is done.

We may still need to alter the stake table design. As every user now has 2 keys.

crates/hotshot-state-prover/src/circuit/types.rs
```diff
@@ -0,0 +1,40 @@
...
+pub struct HotShotState<F: PrimeField> {
+    pub view_number: usize,
+    pub block_height: usize,
+    pub block_comm: F,
+    pub fee_ledger_comm: F,
+    pub stake_table_comm: F,
+}
+
+///
+pub trait IntoFields<F: PrimeField> {
```
alxiong, line 37: is this used somewhere?

crates/hotshot-state-prover/src/circuit/types.rs
```diff
@@ -0,0 +1,40 @@
...
+
+/// HotShot state
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct HotShotState<F: PrimeField> {
+    pub view_number: usize,
+    pub block_height: usize,
+    pub block_comm: F,
+    pub fee_ledger_comm: F,
+    pub stake_table_comm: F,
+}
```
alxiong, line 34: this shouldn't be defined here. but in `types` ? 

cc @jbearer

crates/hotshot-state-prover/src/circuit/types.rs
```diff
@@ -0,0 +1,40 @@
+//! Types for verification circuit
```
alxiong, line 1: nit: personally, I prefer to just keep these into `circuit.rs` file. don't see much benefits of keeping it separate as we don't have lots of types anyway.

crates/hotshot-state-prover/src/circuit.rs
```diff
@@ -0,0 +1,123 @@
...
+    /// - stake table entries (`Vec<(BLSVerKey, SchnorrVerKey, Amount)>`)
+    /// - schnorr signatures of the updated states (`Vec<SchnorrSignature>`)
+    /// - updated hotshot state (`(view_number, block_height, block_comm, fee_ledger_comm, stake_table_comm)`)
+    /// - signer bit vector
+    /// - quorum threshold
+    /// checks that
+    /// - the signer's accumulated weight exceeds the quorum threshold
+    /// - the commitment of the stake table
+    /// - all schnorr signatures are valid
+    pub fn build<ST, P>(
```
alxiong, line 47: this is `fn build_for_preprocessing` instead? (alternatively, maybe via an enum input argument to indicate preprocessing mode or proving mode)

alxiong (commented):
would be great if we can add unit tests incrementally as we build along.

> We may still need to alter the stake table design. As every user now has 2 keys.

our current stake table impl is generic over `K: Key` type, which could be just a tuple of `(BLSVerKey, SchnorrVerKey)`.

also when we are computing the stake table commitment, we can commit over the compressed BLS points concatenated with the uncompressed Schnorr (since we need the full for verification), this will slightly improve the circuit size as we need to hash fewer elements.
(ofc, need consistency inside and outside circuit, luckily, we can hide this optimization under a uniformed abstracted `StakeTable::commitment()` API)

crates/hotshot-state-prover/src/circuit/types.rs
```diff
@@ -0,0 +1,40 @@
...
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct HotShotState<F: PrimeField> {
+    pub view_number: usize,
+    pub block_height: usize,
+    pub block_comm: F,
+    pub fee_ledger_comm: F,
+    pub stake_table_comm: F,
+}
+
+///
```
chancharles92, line 36: add comment for the trait?

crates/hotshot-state-prover/src/circuit/types.rs
```diff
@@ -0,0 +1,40 @@
...
+#[derive(Clone, Debug)]
+pub struct StakeTableEntryVar {
+    pub bls_ver_key: (Variable, Variable),
+    pub schnorr_ver_key: VerKeyVar,
+    pub stake_amount: Variable,
+}
+
+/// HotShot state Variable
+#[derive(Clone, Debug)]
+pub struct HotShotStateVar {
```
chancharles92, line 18: Maybe add a From() method converting between `HotShotStateVar` and `HotshotState`.

crates/hotshot-state-prover/src/circuit.rs
```diff
@@ -0,0 +1,123 @@
...
+                let stake_amount = circuit.create_public_variable(u256_to_field::<F>(&amount))?;
+                Ok(StakeTableEntryVar {
+                    bls_ver_key: (0, 0), // TODO(Chengyu)
+                    schnorr_ver_key,
+                    stake_amount,
+                })
+            })
+            .collect::<Result<Vec<_>, CircuitError>>()?;
+        let dummy_ver_key_var =
+            VerKeyVar(circuit.create_public_point_variable(TEPoint::default())?);
```
chancharles92, line 83: Do we need to make it a public point variable? Can we make it a constant point variable?

crates/hotshot-state-prover/src/circuit.rs
```diff
@@ -0,0 +1,123 @@
...
+        // - The commitment of the stake table as [https://www.notion.so/espressosys/Light-Client-Contract-a416ebbfa9f342d79fccbf90de9706ef?pvs=4#6c0e26d753cd42e9bb0f22db1c519f45]
+        // - Batch Schnorr signature verification
+
+        // creating variables for stake table entries
+        let mut stake_table_var = stake_table
+            .try_iter(SnapshotVersion::LastEpochStart)?
+            .map(|((_bls_ver_key, schnorr_ver_key), amount)| {
+                // TODO(Chengyu): create variable for bls_key_var
+                let schnorr_ver_key = VerKeyVar(
+                    circuit.create_public_point_variable(schnorr_ver_key.to_affine().into())?,
```
chancharles92, line 72: Why is it a public point variable?

crates/hotshot-state-prover/src/circuit.rs
```diff
@@ -0,0 +1,123 @@
...
+        _sigs: &[Signature<P>],
+        _hotshot_state: &HotShotState<F>,
+        signer_bit_vec: &[bool],
+        threshold: &U256,
+    ) -> Result<PlonkCircuit<F>, PlonkError>
+    where
+        ST: StakeTableScheme<Key = (BLSVerKey, SchnorrVerKey<P>), Amount = U256>,
+        P: TECurveConfig<BaseField = F>,
+    {
+        let mut circuit = PlonkCircuit::new_turbo_plonk();
```
chancharles92, line 58: What are the public input of the circuit? I thought it's only
1. The stake table commitment
2. The hotshot state.
3. The threshold.

All others could be secret witness.

crates/hotshot-state-prover/src/circuit.rs
mrain: I'm thinking whether we should only export a single API for proof generation only, w/o exposing these details about circuit.
These interface issues could be finalized when doing proof generation: #1985

crates/hotshot-state-prover/src/circuit/types.rs
mrain: Moved to `types`.
There's already some `State` trait definition. I'm just abruptly adding my stuff at the end.

crates/hotshot-state-prover/src/circuit/types.rs
mrain: Everything is merged in one file now.

crates/hotshot-state-prover/src/circuit/types.rs
mrain: Such `From()` also needs to take a circuit as input. Doesn't seems possible.

crates/hotshot-state-prover/src/circuit.rs
mrain: Done in the newest commit.

crates/hotshot-state-prover/src/circuit.rs
```diff
@@ -0,0 +1,219 @@
...
+        bls_over_bn254::VerKey as BLSVerKey,
+        schnorr::{Signature, VerKey as SchnorrVerKey},
+    },
+};
+use jf_relation::{
+    errors::CircuitError, gadgets::ecc::TEPoint, BoolVar, Circuit, PlonkCircuit, Variable,
+};
+
+/// convert a U256 to a field element.
+pub(crate) fn u256_to_field<F: PrimeField>(v: &U256) -> F {
```
chancharles92, line 27: What if `v` is larger than the field size? Should we return an error in that case?

crates/hotshot-state-prover/src/circuit.rs
```diff
@@ -0,0 +1,219 @@
...
+            .collect::<Vec<_>>()
+    }
+
+    fn stake_table_for_testing(
+        keys: &[(BLSVerKey, SchnorrVerKey)],
+    ) -> StakeTable<BLSVerKey, SchnorrVerKey, F> {
+        let mut st = StakeTable::<BLSVerKey, SchnorrVerKey, F>::new();
+        // Registering keys
+        keys.iter()
+            .for_each(|key| st.register(key.0, U256::from(100), key.1.clone()).unwrap());
```
chancharles92, line 182: Maybe have some different amounts for diff keys.

crates/hotshot-state-prover/src/circuit.rs
```diff
@@ -0,0 +1,219 @@
...
+        st
+    }
+
+    #[test]
+    fn test_circuit_building() {
+        let keys = key_pairs_for_testing();
+        let st = stake_table_for_testing(&keys);
+
+        let bit_vec_6 = [
+            true, true, true, true, true, true, false, false, false, false,
```
chancharles92, line 195: Maybe add a `false` in the consecutive `true` interval and add a `true` in the consecutive `false` interval.

crates/hotshot-state-prover/src/circuit.rs
mrain: This is only for stake amount, and we assume that it fits into the native field. Maybe I should call it "lossy conversion" for clarity.

crates/hotshot-state-prover/src/circuit.rs
mrain: Done in de3a697

crates/hotshot-state-prover/src/circuit.rs
mrain: Done in de3a697

chancharles92 (commented):
LGTM

crates/types/src/traits/state.rs
```diff
@@ -218,3 +219,18 @@ pub mod dummy {
         }
     }
 }
+
+/// A serialized consensus state for proof generation
+#[derive(Clone, Debug, serde::Serialize, serde::Deserialize, Default)]
+pub struct HotShotStateForProver<F: PrimeField> {
```
alxiong, line 225: this should simply be `HotShotState`? is there a conflicting struct? 

also `ForProver` is unclear in general context.

alxiong: maybe `HotShotStateDigest` or `HotshotLightClientState` ?

mrain: changed to `LightClientState` in ccb06e9

crates/hotshot-state-prover/src/circuit.rs
```diff
@@ -0,0 +1,228 @@
...
+
+        let threshold = u256_to_field::<F>(threshold);
+        let threshold_var = circuit.create_public_variable(threshold)?;
+
+        // TODO(Chengyu): put in the hotshot state
+        let public_inputs = vec![threshold];
+
+        // Checking whether the accumulated weight exceeds the quorum threshold
+        // We assume that NUM_ENTRIES is always a multiple of 2
+        let signed_amount_var = (0..STAKE_TABLE_CAPACITY / 2)
```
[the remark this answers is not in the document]
mrain: Add logic for non-multiple of 2 in ccb06e9

crates/hotshot-state-prover/src/circuit.rs
```diff
@@ -134,6 +133,13 @@ where
                 )
             })
             .collect::<Result<Vec<_>, CircuitError>>()?;
+        // Adding the last if STAKE_TABLE_CAPACITY is not a multiple of 2
+        if STAKE_TABLE_CAPACITY % 2 == 1 {
+            signed_amount_var.push(circuit.mul(
+                stake_table_var[STAKE_TABLE_CAPACITY - 1].stake_amount,
+                signer_bit_vec_var[STAKE_TABLE_CAPACITY - 1].0,
+            )?);
+        }
```
alxiong, line 142: wait, wouldn't this cause double counting the last pk's vote twice (if he voted)?

mrain: No, for example if `STAKE_TABLE_CAPACITY` is 3, then previous computation only takes first 2 indices, and this one takes the last.

alxiong: ah, make sense!

alxiong (approved):

chancharles92 (approved):

alxiong (approved):

PR #1908, opened 2023-10-12
Build overhaul

Ancient123:


elliedavidson (approved):
LGTM!

.github/workflows/lints.yml
```diff
@@ -4,33 +4,37 @@ on:
...
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y protobuf-compiler
+
+      - name: Install Just
+        run: |
+          wget https://github.com/casey/just/releases/download/1.14.0/just-1.14.0-x86_64-unknown-linux-musl.tar.gz
+          tar -vxf just-1.14.0-x86_64-unknown-linux-musl.tar.gz just
+          sudo cp just /usr/bin/just
```
DieracDelta, line 31: Nit: could we do `apt-get install -y just` here? If so maybe we could just bundle it in with the previous step

Ancient123: ( https://docs.makedeb.org/prebuilt-mpr/getting-started/ the apt repo info )

DieracDelta: Ah, makes sense!

PR #1917, opened 2023-10-13
Make Test Stricter

bfish713:
- Increase the expected view success for the failure tests.  At 10 they were not really testing what happened after a bad node
- Increase duration of failure tests to allow for timeouts and view sync to happen
- Decrease default failures and expect the correct number of failures based on specific tests.
- Tighten up the default timings.  Makes propose time much shorter which allows for timeout to be decreased significantly.  Rounds should complete faster in general now, and the for tests wit expected timeouts they should run 5 times as fast.
- Add more delay at startup.  I noticed the first leader was proposing before we spun up all nodes triggering weird behavior which isn't realistic (or what the tests want to test).  This fixes that.

### DIFF
--- a/crates/testing/src/overall_safety_task.rs
+++ b/crates/testing/src/overall_safety_task.rs
@@ -367,7 +367,7 @@ impl Default for OverallSafetyPropertiesDescription {
             check_leaf: false,
             check_state: true,
             check_block: true,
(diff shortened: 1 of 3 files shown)


shenkeyao (approved):
LGTM

elliedavidson (approved):

Issue #1571, opened 2023-08-16
[stake-table] Add vector + HashMap implementation


alxiong:
A simple implementation of keeping all (pk, amount) pair in a single vector, and using a `HashMap<PubicKey, u32>` to keep track of index for efficient lookup.

See: https://www.notion.so/espressosys/Stake-Table-Design-d18e9d40182b446ca11387eb2c9d1bc2?pvs=4#f827662f005e4310a5136b2cbb40bf3d

This issue mainly require instantiating the existing `trait StakeTableTrait` with another trivial impl.

PR #1932, opened 2023-10-19
feat: Stake table refactor & vector based implementation

mrain:
closes: #1902, #1571


crates/hotshot-stake-table/src/vec_based.rs
```diff
@@ -0,0 +1,455 @@
+//! A vector based stake table implementation. The commitment is the rescue hash of the list of (key, amount) pairs;
```
chancharles92, line 1: For this hash-based implementation, are we ever going to use any update/lookup APIs?

crates/hotshot-stake-table/src/mt_based/config.rs
```diff
@@ -13,9 +13,19 @@ pub(crate) type FieldType = ark_bn254::Fq;
...
-    v.to_little_endian(&mut bytes);
-    F::from_le_bytes_mod_order(&bytes)
+impl ToFields<FieldType> for FieldType {
+    const SIZE: usize = 1;
+    fn to_fields(&self) -> Vec<FieldType> {
+        vec![FieldType::default()]
+    }
+}
+
+impl ToFields<FieldType> for bls_over_bn254::VerKey {
```
chancharles92, line 23: Are we also going to implement it for Schnorr verification keys?

crates/hotshot-stake-table/src/vec_based.rs
mrain: We have update and lookup, w/o proof.

crates/hotshot-stake-table/src/mt_based/config.rs
mrain: Check line 334 of vec_based.rs

crates/hotshot-stake-table/src/vec_based.rs
```diff
@@ -0,0 +1,451 @@
...
+    use ark_std::{rand::SeedableRng, vec::Vec};
+    use ethereum_types::U256;
+    use hotshot_types::traits::stake_table::{SnapshotVersion, StakeTableError, StakeTableScheme};
+    use jf_primitives::signatures::bls_over_bn254::{
+        BLSOverBN254CurveSignatureScheme, VerKey as BLSVerKey,
+    };
+    use jf_primitives::signatures::schnorr::VerKey as SchnorrVerKey;
+    use jf_primitives::signatures::{SchnorrSignatureScheme, SignatureScheme};
+
+    // KeyType is a pair of BLS verfication key and Schnorr verification key
```
chancharles92, line 329: Is it better to move it out from the test module?

chancharles92 (commented):
Finished reviewing files except for vec_based.rs

mrain:
Just realized that we may need to redesign the interfaces for the following reason:

- The application, especially light client state prover requires the stake table to be dual key (a bls key for qc verification and a schnorr key for state proof generation) indexed. 
- QC verification requires a commitment that binds the mapping between bls key and the stake amount. And the state prover  needs a commitment that bind the mapping between the schnorr key and the stake amount.

I'm proposing the following new design:

- The stake table now contains 2 (hash) mappings, one between bls keys and indices and the other between schnorr keys and indices.
- bls keys, schnorr keys and stake amount are stored in 3 separate lists.
- The actually commitments are computed in the following: we first (rescue) hash three lists, and then hash the needed list hashes for the final commitment. E.g. commitment for qc verification is computed as `rescue_hash([rescue_hash(bls keys), rescue_hash(stake_amount_list), 0])`.

WDYT? @alxiong @chancharles92

chancharles92:
> Just realized that we may need to redesign the interfaces for the following reason:
> 
> * The application, especially light client state prover requires the stake table to be dual key (a bls key for qc verification and a schnorr key for state proof generation) indexed.
> * QC verification requires a commitment that binds the mapping between bls key and the stake amount. And the state prover  needs a commitment that bind the mapping between the schnorr key and the stake amount.
> 
> I'm proposing the following new design (not really interfaces, but implementation):
> 
> * The stake table now contains 2 (hash) mappings, one between bls keys and indices and the other between schnorr keys and indices.
> * bls keys, schnorr keys and stake amount are stored in 3 separate lists.
> * The actually commitments are computed in the following: we first commit three lists, and then hash the needed list hashes for the final commitment. E.g. commitment for qc verification is computed as `rescue_hash([commit(bls keys), commit(stake_amount_list), 0])`.
> 
> Notes about upgradability:
> 
> * In the future if we want to persistentize the stake table to support querying for really old snapshots, we could just update the committed list implementation.
> * Lookup proof could be supported by implementing those 3 lists as Merkle trees. However, the proof is now twice as large as before, b/c you need two Merkle paths now.
> 
> WDYT? @alxiong @chancharles92

Looks fine for me, we can even just define the commitment as the *concatenation* of 3 hashes. And the light client contract will only pick the hash for the stake_amount_list and the hash for the Schnorr keys.

crates/hotshot-stake-table/src/vec_based/config.rs
```diff
@@ -0,0 +1,22 @@
...
+
+/// Hashable representation of a key
+/// NOTE: commitment is only used in light client contract.
+/// For this application, we needs only hash the Schnorr verfication key.
+impl ToFields<FieldType> for KeyType {
+    const SIZE: usize = 2;
+
+    fn to_fields(&self) -> Vec<FieldType> {
+        let p = self.1.to_affine();
+        vec![p.x, p.y]
```
alxiong, line 20: similar comment, can we hash compressed points instead and using a single field?

crates/hotshot-stake-table/src/vec_based.rs
```diff
@@ -0,0 +1,436 @@
...
+                Ok(())
+            }
+        }
+    }
+
+    fn deregister(&mut self, existing_key: &Self::Key) -> Result<(), StakeTableError> {
+        match self.mapping.get(existing_key) {
+            Some(pos) => {
+                self.head_total_stake -= self.head[*pos].1;
+                self.head[*pos].1 = U256::zero();
```
alxiong, line 74: should we _remove_ or set it to zero here? 
the former clean up the memory usage and will affect updated stake table commitment.

crates/hotshot-stake-table/src/vec_based.rs
```diff
@@ -0,0 +1,436 @@
...
+    fn lookup(
+        &self,
+        version: SnapshotVersion,
+        key: &Self::Key,
+    ) -> Result<(Self::Amount, Self::LookupProof), StakeTableError> {
+        match self.mapping.get(key) {
+            Some(&pos) => match version {
+                SnapshotVersion::Head => {
+                    if pos >= self.head.len() {
+                        Err(StakeTableError::KeyNotFound)
```
alxiong, line 122: we can remove lots of code duplication, by defining a `Self::get_table(&self, version) -> Result<Vec<(K, U256)>>`, see a similar [`get_root` helper function](https://github.com/EspressoSystems/HotShot/blob/develop/crates/hotshot-stake-table/src/mt_based.rs#L178-L188) here.

the `pos>=..` check can be write there once.

crates/hotshot-stake-table/src/vec_based.rs
```diff
@@ -0,0 +1,436 @@
...
+            return RescueCRHF::sponge_with_zero_padding(&[], 1)[0];
+        }
+        let mut to_be_hashed = vec![];
+        self.head.iter().for_each(|(key, amount)| {
+            to_be_hashed.extend(key.to_fields());
+            to_be_hashed.push(u256_to_field(amount));
+        });
+        let mut comm = to_be_hashed[0];
+        for i in (1..self.head.len()).step_by(2) {
+            comm = RescueCRHF::sponge_with_zero_padding(
```
alxiong, line 291: we shouldn't use this lower-level API, we should use `VariableLengthRescueCRHF::evaluate` instead.
(same for the usage a couple of lines above)

alxiong (changes requested):

alxiong:
> we can even just define the commitment as the concatenation of 3 hashes. And the light client contract will only pick the hash for the stake_amount_list and the hash for the Schnorr keys.

what about: `Commit(table) := H( Com(Schnorr_keys, Amounts),  Com(Bls_keys))` where the circuit only prove the first half of the preimage.

crates/hotshot-stake-table/src/vec_based.rs
```diff
@@ -0,0 +1,436 @@
...
+                        Ok((self.last_epoch_start[pos].1, ()))
+                    }
+                }
+                SnapshotVersion::BlockNum(_) => Err(StakeTableError::SnapshotUnsupported),
+            },
+            None => Err(StakeTableError::KeyNotFound),
+        }
+    }
+
+    fn simple_lookup(
```
chancharles92, line 147: What's the difference from lookup instantiation?

mrain: Originally with Merkle tree design, this is the one w/o a membership proof.

chancharles92: In this case where two API implementations are the same, maybe we can reduce code duplications by directly calling `lookup()` here.

alxiong: nope, I meant to use `simple_lookup` for looking up `amount` only, and `lookup` will return together with a membership proof. 

```rust
    /// Returns the stakes withhelded by a public key, None if the key is not registered.
    /// If you need a lookup proof, use [`Self::lookup()`] instead (which is usually more expensive).
    ///
    /// # Errors
    /// Return err if the `version` is not supported or `key` doesn't exist.
    fn simple_lookup(
        &self,
        version: SnapshotVersion,
        key: &Self::Key,
    ) -> Result<Self::Amount, StakeTableError>;
```

meanwhile, I agree with Binyi that we can reduce duplication whenever possible.

crates/hotshot-stake-table/src/mt_based/config.rs
```diff
@@ -13,9 +13,19 @@ pub(crate) type FieldType = ark_bn254::Fq;
...
+    const SIZE: usize = 2;
+    fn to_fields(&self) -> Vec<FieldType> {
```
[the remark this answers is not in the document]
mrain: By compression do you mean we hash them into a field element? I felt like concat and hash is overall more efficient.

crates/hotshot-stake-table/src/vec_based.rs
mrain: This is a debatable question. I vaguely remember that @jbearer brought this idea up before. We may settle for it right now, or we may have to implement some garbage collection mechanism.

crates/hotshot-stake-table/src/mt_based.rs
```diff
@@ -86,7 +88,18 @@ impl<K: Key> StakeTableScheme for StakeTable<K> {
...
+        match self.mapping.get(key) {
+            Some(index) => {
+                let branches = to_merkle_path(*index, self.height);
+                root.simple_lookup(self.height, &branches)
+            }
+            None => Err(StakeTableError::KeyNotFound),
+        }
+    }
+
+    fn lookup_with_proof(
```
chancharles92, line 102: We can add a private func `internal_lookup(self, version, key, is_simple_lookup: bool)` to remove code redundancy.

crates/hotshot-stake-table/src/mt_based/config.rs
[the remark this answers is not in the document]
chancharles92: @alxiong means to represent an elliptic curve point in its compressed form, which is a single coordinate plus a bit. I think it can be represented as a single U256, but probably hard to represented as a single base field element.

chancharles92 (commented):
LGTM, some minor comments.

crates/hotshot-stake-table/src/mt_based.rs
mrain: It already calls `lookup_pos` internally, and the rest is just fetching items.

mrain: Idk why I can't reply to the comment about compression. But currently the key is over G2, thus the compressed form is already 2 Fq elements + 1 bit. Which requires 3 Fr elements to hold. And I believe the canonical serialization is already doing the job.

chancharles92: cc @alxiong

crates/hotshot-stake-table/src/vec_based.rs
```diff
@@ -0,0 +1,495 @@
...
+        key: &Self::Key,
+    ) -> Result<Self::Amount, StakeTableError> {
+        let table = self.get_version(version)?;
+        let pos = self.lookup_pos(key)?;
+        if pos >= table.bls_keys.len() {
+            Err(StakeTableError::KeyNotFound)
+        } else {
+            Ok(table.stake_amount[pos])
+        }
+    }
```
alxiong, line 157: I have a slight concern w/ this.
the `self.bls_mapping` is not per-snapshot, which means if our deregistering will prune entries with zero amount, then the `pos` of a certain bls key might be shifted. 

This logic is only correct if we assume ever-growing vectors for `Snapshot::bls_keys/schnorr_keys/amounts` (even the non-zero staker is always bounded at 1000, the length of the vector will be growing)

this will lead to some awkward memory management issue imo.

crates/hotshot-stake-table/src/vec_based.rs
```diff
@@ -0,0 +1,495 @@
...
+
+    fn commitment(&self, version: SnapshotVersion) -> Result<Self::Commitment, StakeTableError> {
+        match version {
+            // IMPORTANT: we don't support committing the head version b/c it's not finalized.
+            SnapshotVersion::Head => Err(StakeTableError::SnapshotUnsupported),
+            SnapshotVersion::EpochStart => Ok(self.epoch_start_comm),
+            SnapshotVersion::LastEpochStart => Ok(self.last_epoch_start_comm),
+            SnapshotVersion::BlockNum(_) => Err(StakeTableError::SnapshotUnsupported),
```
alxiong, line 124: the first and last cases can be simplified as: `_ => Err()` (all other cases)

crates/hotshot-stake-table/src/vec_based.rs
```diff
@@ -0,0 +1,495 @@
...
+        let owned = (0..table.bls_keys.len())
+            .map(|i| {
+                (
+                    table.bls_keys[i].clone(),
+                    table.stake_amount[i],
+                    table.schnorr_keys[i].clone(),
+                )
+            })
+            .collect::<Vec<_>>();
+        Ok(owned.into_iter())
```
alxiong, line 232: this looks slightly strange to me, why taking an iterator, collect it, and then into an iterator again?
I think we can directly return `(0..table.bls_keys.len()).map()` (or maybe adding some `.to_owned()` or `.into_iter()` in the middle to make the compiler happy?

crates/hotshot-stake-table/src/vec_based.rs
```diff
@@ -0,0 +1,495 @@
...
+            .collect::<Vec<_>>()
+            .concat();
+        to_be_hashed.resize(
+            STAKE_TABLE_CAPACITY * <K1 as ToFields<F>>::SIZE,
+            F::default(),
+        );
+        let bls_comm = VariableLengthRescueCRHF::<F, 1>::evaluate(to_be_hashed).unwrap()[0];
+
+        // Compute rescue hash for Schnorr keys
+        let mut to_be_hashed = self
```
alxiong, line 310: I prefer a more descriptive name like `bls_comm_preimage`, same comments for almost all other variable named `to_be_hashed` in this file

crates/hotshot-stake-table/src/vec_based.rs
```diff
@@ -0,0 +1,495 @@
...
+            .head
+            .bls_keys
+            .iter()
+            .map(|key| key.to_fields())
+            .collect::<Vec<_>>()
+            .concat();
+        to_be_hashed.resize(
+            STAKE_TABLE_CAPACITY * <K1 as ToFields<F>>::SIZE,
+            F::default(),
+        );
```
alxiong, line 306: Some concerns:
- since during `deregister`, we never remove that entry, that key will be included in this commitment computation
- similarly, the stake table will never be truly "full" (at 1000 stakers) due to withdrawn stakers taking up the space/indexes.

alxiong (commented):
I still have some concerns over how we handle deregistered entries and its implication for stake table commitment computation. see below

crates/types/src/traits/stake_table.rs
```diff
@@ -5,6 +5,9 @@ use digest::crypto_common::rand_core::CryptoRngCore;
 use displaydoc::Display;
 use jf_primitives::errors::PrimitivesError;
 
+/// Capacity of a stake table
+pub const STAKE_TABLE_CAPACITY: usize = 1000;
+
```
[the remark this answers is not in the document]
mrain: I don't have a good idea about where to put it. Shall there be something like HotShot configuration crate for all these parameters?

crates/hotshot-stake-table/src/vec_based.rs
mrain: That's the problem, I tried but couldn't make the compiler happy :(

crates/hotshot-stake-table/src/vec_based.rs
mrain: Even if you remove the entry, in the commitment computation some default value will take that place.
For commitment computation, I'm padding the stake table to its capacity considering we need to do it in a circuit

crates/hotshot-stake-table/src/vec_based.rs
mrain: changed in acfdc82

crates/hotshot-stake-table/src/vec_based.rs
mrain: changed in acfdc82

crates/hotshot-stake-table/src/vec_based.rs
mrain: I assume this problem is in couple of whether we remove the entry when deregistering the key. 
I would suggest we keep everything as is for now (we simply set the stake amount to zero without actual deregistration). And figure out a better way in the future.

crates/types/src/traits/stake_table.rs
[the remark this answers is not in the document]
chancharles92: Yes, do we have a `config.rs` or something?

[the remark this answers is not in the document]
alxiong: > Shall there be something like HotShot configuration crate for all these parameters?

imo, 2 options, one is our code is generic over this constant (either via generic param or via associated type),
the other is we put this in the `config.rs` file of our _implementation_ crate, and let HotShot assemble these into a single file (via re-export). 

We should do the latter, as it's simpler.

crates/hotshot-stake-table/src/vec_based.rs
alxiong: Let me paste another related thread here, for easier future back reference:

> [alxiong](https://github.com/alxiong)(https://github.com/EspressoSystems/HotShot/pull/1932#discussion_r1385291474)
> Some concerns:
>
> since during deregister, we never remove that entry, that key will be included in this commitment computation
similarly, the stake table will never be truly "full" (at 1000 stakers) due to withdrawn stakers taking up the space/indexes.
> 
> @[mrain](https://github.com/mrain)(https://github.com/EspressoSystems/HotShot/pull/1932#discussion_r1385369341)
>Even if you remove the entry, in the commitment computation some default value will take that place.
For commitment computation, I'm padding the stake table to its capacity considering we need to do it in a circuit

crates/hotshot-stake-table/src/vec_based.rs
alxiong: closing this thread as we will centralize our discussion in https://github.com/EspressoSystems/HotShot/pull/1932#discussion_r1385931169

crates/hotshot-stake-table/src/vec_based.rs
alxiong: I understand the "padding for circuit purpose" part, Also, I don't mind the "zero-balance pk included in the stake table commitment computation" aspect that much for now (altho it incur some useless hashing)

but I'm pointing out the fact that imagine we cap the table size to be 1000, there are 900 registered then deregistered, now, we only have 100 slots left for actual active stakers.

This is a problem.

(I don't mind merging this PR now and dealing with this in a separate issue, but need acknowledgement from @jbearer @chancharles92 and @philippecamacho)

alxiong: a naive approach is: 
- when deregister, remove the entry from the stake table vector (not just mark balance to 0)
- keep an extra `pruned_indices: Vec<usize>` list of pruned positions, so the next register will fill up these first
- to enable `lookup` on `EpochStart` and `LastEpochStart` snapshots of the stake table, keep another 2 super small HashMap, of the pruned bls_key and their original pos. 


another approach with slightly lower overhead on state management is:
- keep 3 (small) lists: `pruned_indices_cur_epoch`,  `pruned_indices_last_epoch`, and `pruned_indices_sec_last_epoch` (maybe think of better names)
- during register, instead of immediately filling up all available pruned positions, we only fill up those in second last epoch (i.e. 2 epochs ago), if there are more registration, we simply append to the current head.
  - this obviate the need for the 2 small hashmaps to enable efficient lookup on the 3 snapshots we want to support.

crates/hotshot-stake-table/src/vec_based.rs
alxiong: an idea: change the `Self::IntoIter = Zip<Zip<Self::Key, Self::Amount>, Self::Aux>`, then you can just return `table.bls_keys.iter().zip(table.stake_amount.iter()).zip(table.schnorr_keys.iter())`?

mrain: Nah, you are missing lifetime specifiers.

crates/hotshot-stake-table/src/vec_based.rs
mrain: I had this garbage collection idea before, but never filled in the details. Let's do it in another PR (in the future)

crates/types/src/traits/stake_table.rs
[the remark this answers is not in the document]
mrain: > > Shall there be something like HotShot configuration crate for all these parameters?
> 
> imo, 2 options, one is our code is generic over this constant (either via generic param or via associated type), the other is we put this in the `config.rs` file of our _implementation_ crate, and let HotShot assemble these into a single file (via re-export).
> 
> We should do the latter, as it's simpler.

Second this. Let's make a wrap-up PR after all features are done.

crates/hotshot-stake-table/src/vec_based.rs
philippecamacho: > I understand the "padding for circuit purpose" part, Also, I don't mind the "zero-balance pk included in the stake table commitment computation" aspect that much for now (altho it incur some useless hashing)
> 
> but I'm pointing out the fact that imagine we cap the table size to be 1000, there are 900 registered then deregistered, now, we only have 100 slots left for actual active stakers.
> 
> This is a problem.
> 
> (I don't mind merging this PR now and dealing with this in a separate issue, but need acknowledgement from @jbearer @chancharles92 and @philippecamacho)

I am ok as well to address this issue later.

chancharles92 (approved):

PR #1954, opened 2023-10-27
Add VID to webserver

rob-maron:
Also fixes the webserver error per view

### DIFF
--- a/crates/hotshot/src/traits/networking/web_server_network.rs
+++ b/crates/hotshot/src/traits/networking/web_server_network.rs
@@ -122,6 +122,15 @@ struct Inner<M: NetworkMsg, KEY: SignatureKey, TYPES: NodeType> {
     /// Task map for quorum votes.
     vote_task_map:
         Arc<RwLock<HashMap<u64, UnboundedSender<ConsensusIntentEvent<TYPES::SignatureKey>>>>>,
+    /// Task map for vid votes
+    vid_vote_task_map:
+        Arc<RwLock<HashMap<u64, UnboundedSender<ConsensusIntentEvent<TYPES::SignatureKey>>>>>,
+    /// Task map for VID certs
+    vid_cert_task_map:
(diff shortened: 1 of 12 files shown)


crates/hotshot/src/traits/networking/web_server_network.rs
```diff
@@ -927,6 +1058,44 @@ impl<
...
+
+            //     if let Some((_, sender)) = task_map.remove_entry(&(view_number)) {
+            //         // Send task cancel message to old task
+
+            //         // If task already exited we expect an error
+            //         let _res = sender
+            //             .send(ConsensusIntentEvent::CancelPollForVIDCertificate(view_number))
+            //             .await;
+            //     }
+            // }
```
elliedavidson, line 1098: Nit: Can we delete this code if we aren't going to use it?

crates/task-impls/src/da.rs
```diff
@@ -292,11 +292,7 @@ where
                 }
             }
             HotShotEvent::DAVoteRecv(vote) => {
-                // warn!(
-                //     "DA vote recv, Main Task {:?}, key: {:?}",
-                //     vote.current_view,
-                //     self.committee_exchange.public_key()
-                // );
+                debug!("DA vote recv, Main Task {:?}", vote.current_view,);
```
elliedavidson, line 295: Very nice!

crates/task-impls/src/vid.rs
```diff
@@ -361,6 +369,9 @@ where
                     }
                 }
             }
+            HotShotEvent::VidCertRecv(_) => {
```
elliedavidson, line 372: I actually think we'd want to put this event into the consensus task, since replicas need to check if they have it before voting on a quorum proposal.  But this can be done in a separate issue.

elliedavidson (dismissed):

crates/hotshot/src/traits/networking/web_server_network.rs
rob-maron: Sure can ! I forgot to remove this

elliedavidson (approved):

PR #1961, opened 2023-10-30
Create New Cert, Vote, and Vote Accumulator types

bfish713:
Adds new traits for Vote and Certificate and uses them to implement as generic vote accumulator.


elliedavidson (approved):
This looks great!

PR #1967, opened 2023-11-01
Create new Certificate and Vote types and use them for Quorum

### LINKED ISSUE #1965: Create New QC Vote and Certificate
- [ ] Use the New Vote and Certificate traits to create the structs
- [ ] Integrate the structures into the consensus code

bfish713:
I created mainly 2 new structures: `SimpleCertificate` and `SimpleVote`.  These implement the most basic vote and certificate pattern.  

`SimpleVote` is some voteable (something we can vote on) data, a signature and a view.  It's really important to note that the voteable data is very similar to the old `VoteData` type except it is no longer an enum.  Basically the `data` member of `SimpleVote` is the thing that the vote actually votes on (e.g. it is a leaf commitment for the quorum vote) and an implementation of commit.  All the voteable implementations of commit add in a tag which commits to which type of vote it is.  This prevents forming a QC with different vote types over the same data.  (one example would be if we had no votes, both a yes and no vote would be over the same leaf commitment, but the votes would commit to different values).

`SimpleCertificate` is the basic implementation of a certificate.  Like `SimpleVote` we should be able to use it for all our certificate types (`threshould()` is the only thing that differs).  A certificate is over some voteable data, and commitment to that data.

For votes I created type alias for all the vote types so other code doesn't need to specify the correct `Voteable` type everywhere in the code.

For certificates I only created the Quorum type because not all the types will use the implementation of `threshold` in `SimpleCertificate` I want to follow up to make this function more generic to prevent copypasta of the rest of the `Certificate1` trait impl.  

I also only implemented the `genesis` function for the Quorum certificate.  

Closes #1965


crates/types/src/simple_certificate.rs
```diff
@@ -0,0 +1,90 @@
...
+    traits::{
+        election::Membership, node_implementation::NodeType, signature_key::SignatureKey,
+        state::ConsensusTime,
+    },
+    vote2::Certificate2,
+};
+
+/// A certificate which can be created by aggregating many simple votes on the commitment.
+#[derive(Eq, Hash, PartialEq, Debug, Clone)]
+pub struct SimpleCertificate<TYPES: NodeType, VOTEABLE: Voteable, MEMBERSHIP: Membership<TYPES>> {
```
elliedavidson, line 21: I'm branching off this branch for my work.  Is this supposed to represent all the SimpleCertificates, or just the QuorumCertificate type?

crates/types/src/simple_vote.rs
```diff
@@ -0,0 +1,214 @@
...
+        node_implementation::NodeType,
+        signature_key::{EncodedPublicKey, EncodedSignature, SignatureKey},
+    },
+    vote2::{HasViewNumber, Vote2},
+};
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Hash, Eq)]
+pub struct YesData<LEAF: Committable> {
+    pub leaf_commit: Commitment<LEAF>,
+}
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Hash, Eq)]
```
elliedavidson, line 22: I feel we should remove all references to "No" votes, since we are not going to implement them in the near future.

crates/types/src/simple_certificate.rs
bfish713: It should be eventually all simple certificates.  The exception right now is that the threshold part isn't properly implemented for certificates that use `failure_threshold()` of the `MEMBERSHIP`

crates/types/src/simple_vote.rs
bfish713: Yeah we can remove, I'm going to push very soon, hopefully something that builds, and I will remove the NoVote stuff

crates/types/src/simple_certificate.rs
elliedavidson: If that is the case, then I don't think it makes sense to have leaf_commitment be a field of the certificate?

bfish713: It's just poorly named, it is always the thing which is voted on

elliedavidson: Could we make the SimpleCertificate generic over the threshold?  That way we can use the same SimpleCertificate for all certs?

bfish713: Ideally we can.  I didn't think of a good way to do this which works with the types, and the `VoteAccumulator`

elliedavidson (commented):
This really is SO much cleaner than what we had before!

crates/types/src/simple_certificate.rs
```diff
@@ -0,0 +1,129 @@
...
+}
+
+impl<
+        TYPES: NodeType,
+        LEAF: Committable + Committable + Clone + Serialize + Debug + PartialEq + Hash + Eq + 'static,
+    > SimpleCertificate<TYPES, YesData<LEAF>>
+{
+    #[must_use]
+    /// Creat the Genisis certificate
+    pub fn genesis() -> Self {
```
elliedavidson, line 112: This is only used for the QuorumCertificate.  It shouldn't ever be used on other certificates.  Because of that I feel like we shouldn't have it here.  Ideally we'd want to protect from accidentally calling it on the other certificate types.

crates/types/src/simple_vote.rs
```diff
@@ -0,0 +1,225 @@
...
+    pub relay: EncodedPublicKey,
+    /// The view number we are trying to sync on
+    pub round: TYPES::Time,
+}
+
+/// Marker trait for data or commitments that can be voted on.
+/// Only structs in this file can implement voteable.  This is enforced with the `Sealed` trait
+/// Sealing this trait prevents creating new vote types outside this file.
+pub trait Voteable:
+    sealed::Sealed + Committable + Clone + Serialize + Debug + PartialEq + Hash + Eq
```
elliedavidson, line 70: I really like this use of sealed!  Such a good idea!

crates/types/src/simple_vote.rs
```diff
@@ -0,0 +1,225 @@
...
+
+/// A simple yes vote over some votable type.  
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Hash, Eq)]
+pub struct SimpleVote<TYPES: NodeType, DATA: Voteable, MEMBERSHIP: Membership<TYPES>> {
+    /// The signature share associated with this vote
+    pub signature: (EncodedPublicKey, EncodedSignature),
+    /// The leaf commitment being voted on.
+    pub data: DATA,
+    /// The view this vote was cast for
+    pub current_view: TYPES::Time,
```
elliedavidson, line 95: Nit: We should change current_view to just view_number, since "current_view" isn't the best descriptor.

crates/types/src/vote2.rs
```diff
@@ -85,20 +96,19 @@ pub struct VoteAccumulator2<
...
         let key = vote.get_signing_key();
 
         let vote_commitment = vote.get_data_commitment();
-        if !key.validate(
-            &vote.get_signature(),
-            &bincode_opts().serialize(&vote_commitment).unwrap(),
-        ) {
+        if !key.validate(&vote.get_signature(), vote_commitment.as_ref()) {
             error!("Vote data is {:?}", vote.get_data_commitment());
             error!("Invalid vote! Data");
```
elliedavidson, line 113: Nit: We should combine this into one error message

elliedavidson (changes requested):

crates/task-impls/src/vid.rs
```diff
@@ -309,7 +309,7 @@ where
                 // `self.cur_view` should be at least 1 since there is a view change before getting
                 // the `DAProposalRecv` event. Otherewise, the view number subtraction below will
                 // cause an overflow error.
-                if view < self.cur_view - 1 {
+                if *self.cur_view == 0 || view < self.cur_view - 1 {
```
shenkeyao, line 312: If we are going to use two conditions here, I think the first one should be `if *self.cur_view != 0`, and `||` should be `&&`.

Anyway, this is equivalent to my change in the [block payload PR](https://github.com/EspressoSystems/HotShot/pull/1968/files#diff-73ecf69c8525d26ad29ee60d0cfc62de3a8316ca47ab2946f0c64e42ee976590), as discusses on Zulip.

crates/testing/tests/consensus_task.rs
```diff
@@ -34,11 +35,11 @@ async fn build_vote(
         inner: handle.hotshot.inner.clone(),
     };
     let quorum_exchange = api.inner.exchanges.quorum_exchange().clone();
-    let vote_token = quorum_exchange.make_vote_token(view).unwrap().unwrap();
+    let _vote_token = quorum_exchange.make_vote_token(view).unwrap().unwrap();
```
shenkeyao, line 38: Nit: can we remove this line?

crates/types/src/simple_vote.rs
```diff
@@ -0,0 +1,225 @@
...
+// impl votable for all the data types in this file sealed marker should ensure nothing is accidently
+// implemented for structs that aren't "voteable"
+impl<V: sealed::Sealed + Committable + Clone + Serialize + Debug + PartialEq + Hash + Eq> Voteable
+    for V
+{
+}
+
+// Type aliases for simple use of all the main votes.  We should never see `SimpleVote` outside this file
+/// Yes vote Alias
+pub type YesVote<TYPES, LEAF, M> = SimpleVote<TYPES, YesData<LEAF>, M>;
```
shenkeyao, line 213: Similar to @elliedavidson's comment about `YesData`, can we just call this `QuorumVote`?

crates/types/src/simple_vote.rs
```diff
@@ -0,0 +1,225 @@
...
+    /// Only structs in this file can impl `Sealed`
+    pub trait Sealed {}
+
+    // TODO: Does the implement for things outside this file that are commitable?
+    impl<C: Committable> Sealed for C {}
+}
+
+/// A simple yes vote over some votable type.  
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Hash, Eq)]
+pub struct SimpleVote<TYPES: NodeType, DATA: Voteable, MEMBERSHIP: Membership<TYPES>> {
```
shenkeyao, line 89: Not a change request, just for my own knowledge -- why is this called _simple_ vote, not just _vote_?

crates/types/src/simple_vote.rs
```diff
@@ -0,0 +1,225 @@
...
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Hash, Eq)]
+/// Data used for a yes vote.
+pub struct YesData<LEAF: Committable> {
+    /// Commitment to the leaf
+    pub leaf_commit: Commitment<LEAF>,
+}
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Hash, Eq)]
+/// Data used for a DA vote.
+pub struct DAData<BLOCK: Committable> {
```
shenkeyao, line 25: Is the `BLOCK` a block payload, or can it be something else? If it's a block payload, since I'm refactoring the payload to a struct, we could then remove the `BLOCK` generic. Before the refactor, I think we could rename `block_commit` and `BLOCK` to `payload_commit` and `PAYLOAD`.

shenkeyao (commented):
Thanks for making these changes!

crates/types/src/simple_certificate.rs
bfish713: I believe the way this is implemented is that it's only for `SimpleCertificate`s with VOTABLE type `YesData` which means any other certificate types calling genesis it will be unimplemented.  I'll use the `QurorumCertificate2` type alias here instead to make it more clear but it's functionally the same.

crates/task-impls/src/vid.rs
bfish713: I'll take the changes from your PR

crates/types/src/simple_vote.rs
bfish713: For 2 somewhat weak reasons:
1 - it makes the relationship between simple vote and simple certificate clear
2 - to allow for more complex vote types in the future that maybe have some additional fields

crates/types/src/simple_vote.rs
bfish713: I honestly hadn't really thought it through.  It's really the thing that the DA committee receives and votes on.  I think this should be the block payload, so I'll rename to that for now unless you have a different thought.

crates/types/src/simple_certificate.rs
elliedavidson: Oh I see, yes you are right! I didn't look closely enough at the type signature.

crates/types/src/simple_vote.rs
shenkeyao: Sounds good!

shenkeyao (dismissed):
LGTM!

shenkeyao (dismissed):

shenkeyao (approved):

shenkeyao (approved):
LGTM

elliedavidson (approved):

Issue #1971, opened 2023-11-02
Scripts for generating key pair


dailinsubjam:

dailinsubjam:
Steps to run the script:

1. Build the Hotshot repo following its [readme](https://github.com/EspressoSystems/HotShot#readme).
2. Run `just async_std gen_key_pair`.

The input config file is in `config/ValidatorConfigFile.toml`.
The output validator key pair is in: `config/ValidatorConfigOutput`.
And an example for validator key pair easier visualization is in `config/ValidatorConfigExample`.

PR #1990, opened 2023-11-03
Key Management and Config File for Validator Onboarding

dailinsubjam:
This PR:
- Added a struct for ValidatorConfigFile and ValidatorConfig.
- Added a flow from reading config seed of file to generating a node's own key pairs.
- Make sure key generation only appear in test initialization or config initialization.
- Added a script to generate key pair. The steps are in https://github.com/EspressoSystems/HotShot/issues/1971.

Designs:
- The config file for validators contains two parts: seed & node_id. They together will server as a final seed for key generation. They can be used either for input of scripts for key generation or validator's config for test. The reason that it's not directly private key & public key is that the form of the keys are very complex (see`config/ValidatorConfigExample` for reference), it would be better to generate it rather than read it from config file, and also our `HotshotConfig` does contain the secret key originally for signning, so it's fine.
- An parameter added for gen_launcher like `gen_launcher::<TestTypes, MemoryImpl>(0)` is to identify `node_id` for key pair (pk & sk) generation, since `build_system_handle()` in `crates/testing/src/task_helpers.rs` needs launcher with different node id.

Main review parts: `orchestrator/src/config.rs`, `crates/types/src/lib.rs`.


bfish713 (dismissed):
Nice, LGTM!

bfish713 (approved):

crates/hotshot-signature-key/src/bn254/bn254_priv.rs
```diff
@@ -54,12 +54,9 @@ impl BLSPrivKey {
     }
 }
 
-#[allow(clippy::incorrect_partial_ord_impl_on_ord_type)]
```
rob-maron, line 57: I'm good with this as long as it still works as intended ! I wonder why we implemented it this way to begin with

crates/hotshot/examples/infra/mod.rs
```diff
@@ -94,38 +94,14 @@ pub struct OrchestratorArgs {
...
-    TYPES::SignatureKey,
-    <TYPES::SignatureKey as SignatureKey>::StakeTableEntry,
-    TYPES::ElectionConfigType,
-> {
+) -> NetworkConfig<TYPES::SignatureKey, TYPES::ElectionConfigType> {
     let config_file_as_string: String = fs::read_to_string(config_file.as_str())
         .unwrap_or_else(|_| panic!("Could not read config file located at {config_file}"));
-    let config_toml: NetworkConfigFile =
-        toml::from_str::<NetworkConfigFile>(&config_file_as_string)
+    let config_toml: NetworkConfigFile<TYPES::SignatureKey> =
```
rob-maron, line 100: This is clean! Love it

rob-maron (approved):
Looks good to me !

crates/hotshot-signature-key/src/bn254/bn254_priv.rs
dailinsubjam: I'm also wondering about that. In the beginning the implementation on this was very long ago, and later implementation just follow that. There are more discussions on whether we want to keep this [on zullip](https://espresso.zulipchat.com/#narrow/stream/311645-Engineering/topic/Clippy.20incorrect.20partial_cmp).

PR #2022, opened 2023-11-08
[hotshot-state-prover] circuit for verification stake table commitment

mrain:
closes: #1983 

blocked by #1901 

I also changed the way to pad the schnorr keys list when computing the stake table commitment, b/c in circuit it's easier to create a neutural point variable (0,1) rather than (0,0).


crates/hotshot-state-prover/src/circuit.rs
```diff
@@ -199,6 +246,14 @@ mod tests {
         let keys = key_pairs_for_testing();
         let st = stake_table_for_testing(&keys);
 
+        let lightclient_state = LightClientState {
```
chancharles92, line 249: Instead of using 0 and F::default(), maybe use some non-default values.

crates/hotshot-state-prover/src/circuit.rs
```diff
@@ -207,9 +262,9 @@ mod tests {
         let (circuit, public_inputs) = StateUpdateBuilder::<F>::build(
             &st,
             &[],
-            &LightClientState::default(),
+            &lightclient_state,
             &bit_vec,
-            &U256::from(25u32),
+            &U256::from(26u32),
```
chancharles92, line 267: Maybe 25 is fine as threshold is not supposed to be exactly equal to the signed weight. (Or maybe use both 26 and a threshold less than 26).

chancharles92 (commented):
LGTM, only a few minor comments.

crates/hotshot-state-prover/src/circuit.rs
mrain: They are irrelevant right now. I can do it in the next PR.

crates/hotshot-state-prover/src/circuit.rs
mrain: Added one in 9927ac8

chancharles92 (approved):

alxiong (approved):

crates/hotshot-state-prover/src/circuit.rs
```diff
@@ -12,16 +12,14 @@ use hotshot_types::traits::{
 };
 use jf_plonk::errors::PlonkError;
 use jf_primitives::{
-    circuit::signature::schnorr::VerKeyVar,
+    circuit::{rescue::RescueNativeGadget, signature::schnorr::VerKeyVar},
     rescue::RescueParameter,
     signatures::{
         bls_over_bn254::VerKey as BLSVerKey,
         schnorr::{Signature, VerKey as SchnorrVerKey},
```
alxiong, line 19: question: why aren't we using BLS12-381 here?

PR #2031, opened 2023-11-09
Cancel polling for retrieved items

rob-maron:


shenkeyao (dismissed):
LGTM!

crates/task-impls/src/consensus.rs
```diff
@@ -642,6 +642,21 @@ where
...
 
+                // stop polling for the received proposal
+                self.quorum_exchange
+                    .network()
+                    .inject_consensus_info(ConsensusIntentEvent::CancelPollForProposal(
+                        *proposal.data.view_number,
+                    ))
+                    .await;
+
+                self.quorum_exchange
```
bfish713, line 653: I don't think we can cancel the DAC poll here.  Usually the DAC comes before the proposal, but it isn't always the case.  We should Cancel the DAC polling on the DAC received event

bfish713 (changes requested):

shenkeyao (approved):

bfish713 (approved):

Issue #1985, opened 2023-11-03
[hotshot-state-prover] proof generation


mrain:

mrain:
We might also need to change the interface, that provides only a proof generation API.
cc @alxiong

PR #2033, opened 2023-11-09
[hotshot-state-prover] Schnorr signature verification

mrain:
closes: #1984 

Please review it after #2022 
Test is taking very long :(


crates/hotshot-state-prover/src/circuit.rs
```diff
@@ -172,11 +180,9 @@ where
             &schnorr_ver_key_preimage_vars,
             1,
         )?[0];
-        circuit.enforce_equal(
-            schnorr_ver_key_comm,
-            lightclient_state_var.stake_table_comm_var.1,
-        )?;
+        circuit.enforce_equal(schnorr_ver_key_comm, lightclient_state_var[5])?;
```
chancharles92, line 183: Instead of using 5 (and 6 below), better to set it as a parameter, otherwise it's easy to lead to bugs if we change the `lightclient_state_var` later.

crates/hotshot-state-prover/src/circuit.rs
```diff
@@ -243,35 +283,71 @@ mod tests {
...
+            VariableLengthRescueCRHF::<F, 1>::evaluate(vec![F::from(3u32), F::from(5u32)]).unwrap()
+                [0];
 
         let lightclient_state = LightClientState {
-            view_number: 0,
-            block_height: 0,
-            block_comm: F::default(),
-            fee_ledger_comm: F::default(),
+            view_number: 24,
+            block_height: 738,
```
chancharles92, line 300: Should block height usually be smaller than the view number?

crates/hotshot-state-prover/src/circuit.rs
```diff
@@ -304,9 +391,17 @@ mod tests {
...
+            .map(|(key, _)| {
+                SchnorrSignatureScheme::<Config>::sign(&(), key, bad_state_msg, &mut prng)
+            })
+            .collect::<Result<Vec<_>, PrimitivesError>>()
+            .unwrap();
         let (bad_circuit, public_inputs) = StateUpdateBuilder::<F>::build(
             &st,
-            &[],
+            &sig_for_wrong_state,
             &bad_lightclient_state,
```
chancharles92, line 405: Why do we need to use `bad_lightclient_state` if we already set `bad_state_msg` to be signed?

crates/hotshot-state-prover/src/circuit.rs
```diff
@@ -304,9 +391,17 @@ mod tests {
             fee_ledger_comm: F::default(),
             stake_table_comm: (F::default(), F::default(), F::default()),
         };
+        let bad_state_msg = [F::default(); 7];
```
chancharles92, line 394: Maybe we can set everything correct except for the bls key commitment.

crates/hotshot-state-prover/src/circuit.rs
```diff
@@ -315,5 +410,21 @@ mod tests {
         assert!(bad_circuit
             .check_circuit_satisfiability(&public_inputs)
             .is_err());
+
+        // bad path: incorrect signatures
+        let wrong_sigs = (0..10)
```
chancharles92, line 415: 10 -> num_validators

chancharles92 (commented):
Some minor comments.

crates/hotshot-state-prover/src/circuit.rs
```diff
@@ -37,8 +38,8 @@ pub struct StakeTableEntryVar {
     pub stake_amount: Variable,
 }
 
-/// HotShot state Variable
-/// The stake table commitment is a triple (bls_keys_comm, stake_amount_comm, schnorr_keys_comm).
+/// Light client state Variable
+/// The stake table commitment is a triple (bls_keys_comm, schnorr_keys_comm, stake_amount_comm).
 #[derive(Clone, Debug)]
 pub struct LightClientStateVar {
```
alxiong, line 44: Can we define internal `struct StakeTableCommitmentVar` to avoid the `.0, .1, .2` access? 
we could have `stake_table_comm_var.bls_key_comm()` instead.

crates/hotshot-state-prover/src/circuit.rs
```diff
@@ -67,8 +68,8 @@ where
     /// - all schnorr signatures are valid
     pub fn build<ST, P>(
         stake_table: &ST,
-        _sigs: &[Signature<P>],
-        _hotshot_state: &LightClientState<F>,
+        sigs: &[Signature<P>],
+        lightclient_state: &LightClientState<F>,
         signer_bit_vec: &[bool],
```
alxiong, line 73: sry for the late question: why do we define `StateUpdateBuilder` struct, instead of either directly offering a `fn build()` ?

crates/hotshot-state-prover/src/circuit.rs
```diff
@@ -116,8 +123,28 @@ where
...
+        let block_height_f = F::from(lightclient_state.block_height as u64);
+        let lightclient_state_var = [
+            circuit.create_public_variable(view_number_f)?,
+            circuit.create_public_variable(block_height_f)?,
+            circuit.create_public_variable(lightclient_state.block_comm)?,
+            circuit.create_public_variable(lightclient_state.fee_ledger_comm)?,
+            circuit.create_public_variable(lightclient_state.stake_table_comm.0)?,
+            circuit.create_public_variable(lightclient_state.stake_table_comm.1)?,
+            circuit.create_public_variable(lightclient_state.stake_table_comm.2)?,
+        ];
```
alxiong, line 136: this variable should be of type `LightClientStateVar` instead of an array, right?

also that struct should have a `LightClientStateVar::new(state) -> Self` to help.

crates/hotshot-state-prover/src/circuit.rs
alxiong: it's not even gonna be parameter, there should be `struct StakeTableCommVar` and accessing to certain field should be clear from the getter name

crates/hotshot-stake-table/src/vec_based.rs
```diff
@@ -293,33 +293,31 @@ where
...
             .bls_keys
             .iter()
-            .map(|key| key.to_fields())
-            .collect::<Vec<_>>()
-            .concat();
+            .flat_map(|key| key.to_fields())
+            .collect::<Vec<_>>();
         bls_comm_preimage.resize(
             STAKE_TABLE_CAPACITY * <K1 as ToFields<F>>::SIZE,
             F::default(),
```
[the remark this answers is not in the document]
mrain: 1. That also works.
2. B/c in the stake table commitment computation, i.e. the non-circuit implementation, Schnorr key list is padded with neutural TE point (0, 1), not (0, 0).

crates/hotshot-state-prover/src/circuit.rs
mrain: This is for later when you need to offer all these vars as a input to Schnorr signature verification. Maybe I should have a struct that has `get_*()` interfaces and implement `AsRef<[Variable]>`.

mrain: check 599a75d

crates/hotshot-state-prover/src/circuit.rs
mrain: check https://github.com/EspressoSystems/HotShot/commit/599a75df2a83eeddfadb783d23cfccb30ad2d494

crates/hotshot-state-prover/src/circuit.rs
mrain: done in 7669200

crates/hotshot-state-prover/src/circuit.rs
mrain: done in 7669200

crates/hotshot-state-prover/src/circuit.rs
mrain: done in the newest commitment

crates/hotshot-state-prover/src/circuit.rs
mrain: check https://github.com/EspressoSystems/HotShot/commit/599a75df2a83eeddfadb783d23cfccb30ad2d494

crates/hotshot-state-prover/src/circuit.rs
mrain: I'm also thinking whether to expose the circuit builder interface. We could discuss it in proof generation #1985

crates/hotshot-state-prover/src/circuit.rs
mrain: Wait you cannot do this.
The stake table commitment check ignores the bls key part. If everyone sign this, the validation shall pass.

chancharles92: But the signature verification circuit is checked over the message that contains the bls key commitment in the public input, the validation should fail in this case.

mrain: a better test is provided in 6ff6fcf

crates/hotshot-state-prover/src/circuit.rs
alxiong: so you wanna add an item or comment in that issue, or change to a plain function here in this PR?

e.g. our [proof_of_exp example code](https://github.com/EspressoSystems/jellyfish/blob/main/plonk/examples/proof_of_exp.rs#L98) has a single function to build the circuit, I feel we could follow this.

mrain: Added a comment here: https://github.com/EspressoSystems/HotShot/issues/1985#issuecomment-1806101480

chancharles92 (approved):

alxiong (approved):

PR #2039, opened 2023-11-10
Remove Vote Token for all types except View sync

bfish713:
Simplifies the logic to simply check if we have stake/can vote rather than a match on vote token.

Also Adds a way to get the `StakeTableEntry` from the key to the membership trait.  This will be needed when we don't have a static fixed stake comittee.


crates/types/src/vote2.rs
```diff
@@ -112,7 +112,9 @@ impl<
         }
 
         // TODO: Lookup the actual stake
```
rob-maron, line 114: Does this resolve the above comment?

rob-maron (approved):
Looks good to me !

crates/types/src/vote2.rs
bfish713: Yeah I should remove

rob-maron (approved):

crates/hotshot/src/traits/election/static_committee.rs
```diff
@@ -124,6 +123,22 @@ where
         let signature = PUBKEY::sign(private_key, &message);
         Ok(Some(StaticVoteToken { signature, pub_key }))
     }
+    fn has_stake(&self, pub_key: &PUBKEY) -> bool {
+        let entry = pub_key.get_stake_table_entry(1u64);
```
dailinsubjam, line 127: Is this line serves as assigning stake value (1u64) to a public key to make them searchable in committee? I feel rather than doing this, we'd better propagate in a complete node info (pub key + stake value) through the function parameter or design the committee only saving the pub key (which looks more reasonable), so that stake value is not something we assign. How do you think? We can leave it as a future issue, not an urgent thing for now.

crates/hotshot/src/traits/election/static_committee.rs
```diff
@@ -124,6 +123,22 @@ where
...
+        self.committee_nodes_with_stake.contains(&entry)
+    }
+
+    fn get_stake(
+        &self,
+        pub_key: &<TYPES as NodeType>::SignatureKey,
+    ) -> Option<<TYPES::SignatureKey as SignatureKey>::StakeTableEntry> {
+        let entry = pub_key.get_stake_table_entry(1u64);
```
dailinsubjam, line 135: Same here.

crates/hotshot-qc/src/bit_vector_old.rs
dailinsubjam, line 1: Is it a sure thing we're still using `bit_vector_old.rs`? Because I saw last commit on `bit_vector.rs` is much newer than `bit_vector_old.rs`. I can work on removing the `_old` and transfer to the new one sometime if needed.

dailinsubjam (approved):

crates/hotshot/src/traits/election/static_committee.rs
bfish713: We could, but I don't want to hard code 1u64 throughout the code and we have no way to look up stake from the stake table.  We could change the static committee to to only save the public key.  I do want to rework the `get_stake_table_entry` function a bit but I think we should leave this as future work.

crates/hotshot-qc/src/bit_vector_old.rs
bfish713: I'm not sure on the status of this.  I was just making the changes need to get rid of the vote token checks everywhere in the code.

crates/hotshot/src/traits/election/static_committee.rs
dailinsubjam: Yeah I was trying to remove `get_stake_table_entry` and all these kinds of `1u64` in validator onboarding task, therefore added [this](https://github.com/EspressoSystems/HotShot/pull/1990/files#diff-cbe058cb32b3323389374453c6b2db7d52432aacf1a7e59d3ee2ee99364bccc4R294) so that stake value 1 will be a default setting. However not sure at which middle point this value together with the pub key are not propagated together :( But yeah this could be a future issue.

bfish713: I agree that this should be a future issue to clean up and make sure we are fetching the stake properly from the table in the production code and not using `get_stake_table_entry` (except in tests)

dailinsubjam (approved):

Issue #2025, opened 2023-11-09
Fix transaction flattening


shenkeyao:
We currently concatenate bytes of transactions into one vector and use it to calculate the VID disperse data, which isn't safe. We should update the (un)flattening to include the transaction length as well.

(Related [Zulip discussion](https://espresso.zulipchat.com/#narrow/stream/330967-Sequencer/topic/VID.20namespace.20integration/near/400814568).)

Issue #2024, opened 2023-11-09
Associate block header and payload with metadata type


shenkeyao:
[Pasting from [Zulip discussion](https://espresso.zulipchat.com/#narrow/stream/330967-Sequencer/topic/VID.20namespace.20integration/near/400815880)]
```rust
trait BlockPayload {
  type Transaction;
  type Metadata; // Data created during block building which feeds into the block header, e.g. namespace table

  fn build(transactions: impl IntoIterator<Item = Transaction>) -> (Self, Self::Metadata);
  fn encode(&self) -> Vec<u8>;
  fn decode(bytes: &[u8]) -> Self;
}

trait BlockHeader<Metadata> {
  fn new(payload_commitment: VidCommitment, metadata: Self::Metadata, parent_leaf: Leaf) -> Self;
}
```
Then the workflow would be like this:
- Transactions task uses build to create a block payload and namespace table
- Transactions task uses encode to flatten block payload into bytes, computes VID commitment
- Leader sends flat bytes and metadata to DA committee members
- Committee members compute VID commitment over flat bytes, sign hash of VID commitment and metadata
- Leader uses BlockHeader::new to construct block header (and thus proposal) from VID commitment and metadata
- Leader sends proposal including header and DAC to replicas
- Replicas extract metadata and VID commitment from header and check against the signatures in DAC

Issue #2026, opened 2023-11-09
Fix the payload sending and receiving for the DA proposal


shenkeyao:
We should not assume the payload commitment from BlockPayload is correct, so in the DA proposal, we should use the transaction bytes to recompute the payload commitment, rather than take the commitment directly.

(Related [Zulip discussion](https://espresso.zulipchat.com/#narrow/stream/330967-Sequencer/topic/VID.20namespace.20integration/near/400814568).)

PR #2045, opened 2023-11-10
Associate block with metadata and fix transaction encoding

shenkeyao:
* Updates the `BlockPayload` trait.
  * Add a `Metadata` type and a constructor to build the payload and metadata.
  * Add transaction encoding and decoding functions.
* Updates the `BlockHeader` constructor to take `metadata` as an input.
* Implements the new `BlockPayload` functions for the `VIDBlockPayload` struct.
* Fixes transaction flattening to include the length of each transaction.

Closes https://github.com/EspressoSystems/HotShot/issues/2024. Closes https://github.com/EspressoSystems/HotShot/issues/2025.


crates/types/src/traits/block_contents.rs
```diff
@@ -50,7 +46,17 @@ pub trait BlockPayload:
...
+    type Metadata;
+
+    /// Build a payload and associated metadata with the transactions.
+    fn build(transactions: impl IntoIterator<Item = Self::Transaction>) -> (Self, Self::Metadata);
+
+    /// Encode the payload
+    fn encode(&self) -> Vec<u8>;
+
+    /// Decode the payload
+    fn decode(encoded_transactions: &[u8]) -> Self;
```
jbearer, line 59: I think `decode` has to take the `Metadata` as well. E.g. in the sequencer, we can't find the transaction boundaries without the namespace table.

crates/types/src/block_impl.rs
```diff
@@ -25,10 +25,35 @@ pub const NUM_STORAGE_NODES: usize = 8;
...
 
+impl VIDTransaction {
+    #[must_use]
+    /// Encode a list of transactions into bytes.
+    pub fn encode(transactions: Vec<Self>) -> Vec<u8> {
+        let mut encoded = Vec::new();
+
+        for txn in transactions {
+            // Encode the length of the inner transaction and the transaction bytes.
+            if let Ok(len) = txn.0.len().try_into() {
```
jbearer, line 47: Since we're converting the length to a `u8`, this means we can't have any transactions bigger than 256 bytes. Would this be a problem for benchmarking? If so I think you can take the length as a `u64` or `u32` and use `to_le_bytes()` to get a binary representation of the length.

Also, probably not good to silently skip transactions here when the length is too large. Better to panic or at least log something

crates/task-impls/src/consensus.rs
```diff
@@ -1319,7 +1319,7 @@ where
                 view_number: view,
                 justify_qc: consensus.high_qc.clone(),
                 parent_commitment: parent_leaf.commit(),
-                block_header: TYPES::BlockHeader::new(*payload_commitment, &parent_header),
+                block_header: TYPES::BlockHeader::new(*payload_commitment, (), &parent_header),
```
jbearer, line 1322: I'm confused how we can pass `()` here for the metadata, when we don't know that `TYPES::BlockHeader::Metadata = ()`. We should really be getting the metadata from whoever built the block, ie wherever `self.payload_commitment` came from, we should get `self.payload_metadata` from the same place.

crates/task-impls/src/consensus.rs
```diff
@@ -1331,7 +1331,7 @@ where
                 .sign_validating_or_commitment_proposal::<I>(&leaf.commit());
             // TODO: DA cert is sent as part of the proposal here, we should split this out so we don't have to wait for it.
             let proposal = QuorumProposal {
-                block_header: TYPES::BlockHeader::new(*payload_commitment, &parent_header),
+                block_header: TYPES::BlockHeader::new(*payload_commitment, (), &parent_header),
```
jbearer, line 1334: Same here, but actually probably better to do `leaf.block_header.clone()` rather than rebuild the header from scratch

crates/task-impls/src/transactions.rs
```diff
@@ -219,30 +219,15 @@ where
                 // TODO (Keyao) Determine whether to allow empty blocks.
                 // <https://github.com/EspressoSystems/HotShot/issues/1822>
                 let txns = self.wait_for_transactions(parent_leaf).await?;
+                let encoded_txns = VIDTransaction::encode(txns.clone());
```
jbearer, line 222: We need to use the application-specific encoding function here, ie `let (payload, payload_metadata) = TYPES::BlockPayload::build(txns); let encoded_txns = block.encode();`. This also means we can restrict to using `VIDTransaction` anymore, we have to deal with the application-specific transaction type (in the sequencer, the transaction type will have an extra namespace field, which we will use to populate the namespace table in `BockPayload::build`.

crates/types/src/traits/block_contents.rs
```diff
@@ -50,7 +46,17 @@ pub trait BlockPayload:
...
 
-    // type Header: BlockHeader;
+    /// Data created during block building which feeds into the block header
+    type Metadata;
+
+    /// Build a payload and associated metadata with the transactions.
+    fn build(transactions: impl IntoIterator<Item = Self::Transaction>) -> (Self, Self::Metadata);
+
+    /// Encode the payload
+    fn encode(&self) -> Vec<u8>;
```
ggutoski, line 56: Maybe lend out a reference instead of allocating new memory with each call.
```suggestion
    fn encode(&self) -> &[u8];
```

crates/task-impls/src/transactions.rs
```diff
@@ -219,30 +219,15 @@ where
...
-
-                // TODO Wasteful flattening of tx bytes to accommodate VID API
-                // https://github.com/EspressoSystems/jellyfish/issues/375
-                let mut txns_flatten = Vec::new();
-                for txn in &txns {
-                    txns_flatten.extend(txn.0.clone());
-                }
-
-                let vid_disperse = vid.disperse(&txns_flatten).unwrap();
                 let block = VIDBlockPayload {
```
jbearer, line 231: Then here we would use the application-specific `block` we built above.

Then in the `BlockReady` event below, we would add `block_metadata` in addition to `block`, so that the metadata gets sent to the DA task, who can in turn send it to the consensus task, which will use it to construct the header.

crates/types/src/traits/block_contents.rs
ggutoski, line 59: Is it correct for me to view `build` and `decode` as two different constructors? Maybe rename them to `from_txs` and `from_bytes`. Why does `build` return `(Self, Metadata)` whereas `decode` returns only `Self`?

jbearer (changes requested):

crates/types/src/traits/block_contents.rs
```diff
@@ -50,7 +46,17 @@ pub trait BlockPayload:
...
+
+    /// Decode the payload
+    fn decode(encoded_transactions: &[u8]) -> Self;
 
     /// returns hashes of all the transactions in this block
     /// TODO make this ordered with a vec
```
ggutoski, line 62: Github won't let me comment on line 63 so I'll do it here. Does `transaction_commitments` method need to exist? If so then I think we want to change the return type so that it's no longer a `HashSet`.

crates/types/src/traits/block_contents.rs
```diff
@@ -64,8 +70,12 @@ pub trait BlockHeader:
     /// Block payload associated with the commitment.
     type Payload: BlockPayload;
 
-    /// Build a header with the payload commitment and parent header.
-    fn new(payload_commitment: Commitment<Self::Payload>, parent_header: &Self) -> Self;
+    /// Build a header with the payload commitment, metadata, and parent header.
```
ggutoski, line 73: It's not clear to me that `BlockHeader` needs to be a trait. Will the sequencer impl this trait? Does the sequencer even know anything about block headers?

Maybe it should be a struct like
```rust
struct BlockHeader<METADATA> {
  payload_commitment: VID::Commitment, // ...or something
  metadata: METADATA,
  parent: &Self,
}
```

crates/types/src/block_impl.rs
```diff
@@ -45,21 +70,6 @@ impl Committable for VIDTransaction {
...
-    GenesisFailed,
-    /// Genesis reencountered after initialization
-    GenesisAfterStart,
-    /// invalid block
-    InvalidBlock,
-}
-
 /// A [`BlockPayload`] that contains a list of `VIDTransaction`.
 #[derive(PartialEq, Eq, Hash, Serialize, Deserialize, Clone, Debug)]
 pub struct VIDBlockPayload {
```
ggutoski, line 75: I'm a little confused about this `VIDBlockPayload` struct. My understanding is that any impl of `BlockPayload` inside hotshot is purely for example or testing purposes. So why does this file `block_impl.rs` even exist in production code? Where is it used?

crates/task-impls/src/transactions.rs
ggutoski, line 222: My understanding is that here we should instead call `build(txns)` for some generic `BlockPayload`. I don't think the struct `VIDTransaction` should exist in production code.

crates/task-impls/src/transactions.rs
```diff
@@ -219,30 +219,15 @@ where
...
                 // <https://github.com/EspressoSystems/HotShot/issues/1822>
                 let txns = self.wait_for_transactions(parent_leaf).await?;
+                let encoded_txns = VIDTransaction::encode(txns.clone());
+                // TODO <https://github.com/EspressoSystems/HotShot/issues/1686>
+                let srs = test_srs(NUM_STORAGE_NODES);
+                // TODO We are using constant numbers for now, but they will change as the quorum size
+                // changes.
+                // TODO <https://github.com/EspressoSystems/HotShot/issues/1693>
+                let vid = VidScheme::new(NUM_CHUNKS, NUM_STORAGE_NODES, &srs).unwrap();
+                let vid_disperse = vid.disperse(encoded_txns.clone()).unwrap();
```
ggutoski, line 229: Here I think we should call `disperse(b.encode())` where `b` is something that impls `BlockPayload`.

ggutoski (commented):
Some quick comments. Perhaps I'll post more later when I have time. We'll definitely need feedback from @jbearer here.

crates/types/src/traits/block_contents.rs
ggutoski: @jbearer I'm starting to think it's bad form for `build` to return `(Self, Metadata)`. I think we should view it as a constructor, so it should return only `Self`. We can add a getter to the trait to allow people to get `Metadata` if they want it. wdyt?

crates/types/src/traits/block_contents.rs
ggutoski: btw @jbearer what will be in `Metadata` other than the namespace table? If we put the namespace table inside the payload like we've been discussing then perhaps `Metadata` will just be `()` in the sequencer. Seems kinda crufty to keep this `Metadata` around if we don't actually need it.

shenkeyao:
All the task implementations (for consensus, transactions, etc.) are still using `VIDBlockPayload` which used to be the only payload struct. I should have replaced all them in the production code with the app-specified payload type.

Will update this. Thanks for the comments! @jbearer @ggutoski

crates/types/src/traits/block_contents.rs
jbearer: I'm torn here. The latter forces the implementation to store the flat bytes as a field, so it doesn't even have the option of generating it on the fly. However, for the sequencer, we probably are going to do this anyways, so we have convenient access to the flat bytes for generating namespace proofs. In which case this interface would save a lot of copying.

There is a way around this, which allows the implementation to generate the flat bytes on the fly if desired, but doesn't require a copy if it already has them lying around. But it's not especially pretty:
```rust
fn encode<F, R>(&self, f: F) -> R
where
  F: FnOnce(&[u8]) ->R;

// Use like
let dispersal = block.encode(|bytes| vid_scheme.disperse(bytes));
```

crates/types/src/traits/block_contents.rs
jbearer: That requires you to store the `Metadata` in the block payload type, which we don't necessarily want. It's redundant because the whole point of the metadata is to later put it in the header type. I don't think it's necessarily bad form to return `(Self, OtherStuff)`. The nice thing about Rust constructors is they're just functions.

I do like the rename `from_txs` and `from_bytes`. The reason for the different types is that these functions have different purposes: `build` or `from_txs` is a one-time method for bringing a block into the world, used by the builder. But once a block is in the world, we have these two different canonical representations of it: `BlockPayload` and `(Vec<u8>, Metadata)`. We can freely convert between them with `encode`/`decode`.

crates/types/src/traits/block_contents.rs
jbearer: The header is extremely important to the sequencer. It's where we put all the data that gets broadcast to replicas and validated. E.g. fee payments, L1 info, timestamps, etc....

If we put the namespace table in the payload, `Metadata` will be something like `u32`, not `()`. Because we need the size of the namespace table in the block header as a starting point for interpreting the payload.

jbearer: I guess we could define the size as the first 4 bytes of payload, and that would still give a unique interpretation. But we can save a lookup/evaluation proof for 4 bytes in the header, which still might be useful

crates/types/src/traits/block_contents.rs
jbearer: Thinking about this more, maybe it's fine to require `Metadata` to be a field of the block type. This is after all just an in-memory representation. Whenever we store the payload on disk or send it over the network, we always have the option of converting to flat bytes, and then using `decode` on the other end to reconstruct it from the flat bytes and the metadata which we already have (e.g. from a header)

crates/types/src/traits/block_contents.rs
shenkeyao: Yeah `transaction_commitments` is needed for the consensus and transactions tasks to update their `included_txns`, but it doesn't have to be a `HashSet`. I'll change it.

crates/types/src/traits/block_contents.rs
shenkeyao: Good question! I view them as constructors that do the transaction encoding/decoding in opposite ways. `build` _encodes_ the transaction, then outputs `Self` and `Metadata`, whereas `decode` _decodes_ the transactions _with_ the `Metadata` (which I missed in the input) and outputs `Self`.

I don't have a strong opinion re: whether to make `build` return `Metadata` or add a getter instead, but for `decode`, because of the opposite encoding/decoding logic, I actually think `decode` should have `Metadata` as input, rather than output (as @jbearer also suggested in [this comment](https://github.com/EspressoSystems/HotShot/pull/2045#discussion_r1389979517)).

shenkeyao: Ahh, I didn't see Jeb's replies until now.

shenkeyao: So in summary things to update are:
* Rename functions
* Add a `Metadata` getter
* `from_txns`: return `Self` only
* `from_bytes`: add `Metadata` to input

ggutoski: > ... to reconstruct it from the flat bytes and the metadata which we already have (e.g. from a header)

Ah, so metadata is on-chain and known to all. What's the difference between metadata and header? Can we unify these two concepts?

jbearer: Metadata is a part of the header. Basically the header construction is allowed to take input from two different sources: `metadata`, which relates to the payload, and `parent`, which allows us to chain state updates together. These are separate until the header is built, after which you only care about the header. They are separate initially because they come from different places, metadata from the block builder and parent header from the consensus proposer.

crates/types/src/traits/block_contents.rs
jbearer: Ok so I think, what we basically have with the transactions table idea, is a canonical way to encode a generic list of variably sized elements, with (somewhat) efficient inclusion proofs.

Why don't we do the exact same thing with the namespace table in the beginning of the payload, where the namespace table is a list of (offset, size, rollup1, rollup2, ....) elements. This lets us do transaction inclusion proofs by combining a list inclusion proof for the namespace table entry and a list inclusion proof for the transaction data. In this case we don't care about finding all namespaces that contain a particular rollup, we just need the prover to point us at the one particular namespace that includes our transaction.

For namespace proofs, ie rollup full nodes, we will have the verifier download the entire namespace table, which we had them doing before anyways. This lets them find all namespaces relevant to their rollup, and check that each bundle is included in every applicable rollup. The only extra cost is the computation of one additional KZG commitment -- assuming the namespace table fits within a single polynomial, which seems a reasonable enough assumption.

But a malicious builder could really screw over a rollup by making the namespace table huge and forcing the rollup to do a bunch of extra KZG commitments even if they don't have any transactions in the block. I think this is a decent reason to put the namespace table size in the block header as `Metadata`: by _defining_ the namespace table as the first `size` bytes of payload, with `size` coming from the header, we can enforce a maximum size or fee per size in consensus.

crates/types/src/traits/block_contents.rs
ggutoski: Ok I can't resist the call of the nerd snipe. You're going to need to explain that one to me. I don't see how your example relieves the implementation of the burden to store the flat bytes as a field. How is the implementation supposed to use `f`? Why does `f` take a `&[u8]`? Seems like the implementation still needs flat bytes to feed to `f`, no?

ggutoski: An alternative is to return `impl Iterator<Iterm = u8>` and pray that compiler optimizations will eliminate any extra scans if the underlying type is something convenient like `&[u8]` or `Vec<u8>`.

crates/types/src/traits/block_contents.rs
ggutoski: Hmm. Why is hotshot using the `BlockPayload` trait to update their included txs? Seems like this task is not in the scope of `BlockPayload`.

crates/types/src/traits/block_contents.rs
jbearer: > I don't see how your example relieves the implementation of the burden to store the flat bytes as a field.

Ah, because the lifetime of the `&[u8]` in the argument of `F` is only required to live as long as the call to `f`. So the implementation can compute the flat byte array on the stack and pass that reference into `f`, instead of taking a reference from `&self`, like so:

```rust
fn encode<F, R>(&self, f: F) -> R {
  let flat_bytes = self.implementation_defined_flattening();
  f(&flat_bytes)
}
```

`impl Iterator<Item = u8>` would work nicely if `disperse` and `commit_only` took an iterator. But if they require a `&[u8]`, the caller is going to have to `collect()` that iterator into a `Vec<u8>` before it can be passed to VID, and I don't see how the compiler can eliminate that unless it's doing some really crazy inlining.

ggutoski: > `impl Iterator<Item = u8>` would work nicely if `disperse` and `commit_only` took an iterator.

They once did. As I recall you recently convinced me not to bother taking an iterator in `disperse` and `commit_only`, so I changed it to `&[u8]` 😛 . But I could reinstate it without much difficulty. The only tiny difficulty is I need a convenient way to get the iterator length (so that payload reconstruction can truncate appropriately). I could use `ExactSizeIterator`, or I could hack the `bytes_to_field` function to return the number of bytes consumed.

jbearer: Yeah, I remember 🙃 I think I was thinking let's use the simplest type until we have a use case for something more general. And lo, a use case!

I think `ExactSizeIterator` might pose a problem, because the `encode` function may not know the exact size of the encoding without doing a full scan over the data, in which case why not just return a `Vec`? I don't really understand the suggestion about `bytes_to_field`.

ggutoski: Ok this is a tangent unrelated to this PR (sorry Keyao).

The exact promise of `ExactSizeIterator` is not clear to me, and perhaps you're correct that it cannot promise not to do a linear scan of the data.

Moreover, `ExactSizeIterator` is not preserved by many common iterator compositions, which renders it a pretty useless trait. https://github.com/rust-lang/rust/pull/66531

To expound on `bytes_to_field`: This is an iterator adaptor that turns an iterator over `u8` to an iterator over `F`. It is possible to hack this function to return the number of bytes consumed, but any such hack would be an ergonomics nightmare because it can no longer be a pure iterator adaptor, so it's no longer composable with iterator design patterns.

Basically: I'm not aware of a clean solution here so I capitulated and enforced `&[u8]`.  The purist in me wants to make it an iterator but Rust needs a clean solution for this problem.

ggutoski: @shenkeyao offline conversation with Jeb results in the following recommendation for `encode`:
```rust
trait BlockPayload {
  type Encode<'a>: 'a + Iterator<Item = u8>;

  fn encode(&self) -> Self::Encode<'_>;
}
```
Explanation:
- It's compatible with both methods of generating the flattened bytes: pre-stored or on-the-fly.
- Need the assoc type `Encode` because Rust has not yet stabilized impl trait return types in trait definitions

crates/types/src/traits/block_contents.rs
shenkeyao: `transaction_commitments` is mainly used for metrics, e.g., `outstanding_transactions` and `outstanding_transactions_memory_size`. I could move this function to another trait since it's not in the scope of `BlockPayload`, but we still need to implement it as long as the metrics need it.

Is there any concern about having it in `BlockPayload` besides the extra implementation? Otherwise I'll probably just keep it there. I've changed it to a vector.

crates/task-impls/src/da.rs
```diff
@@ -394,18 +394,18 @@ where
...
                     .await;
 
-                let payload_commitment = block.commit();
+                let payload_commitment = payload.commit();
                 let signature = self
                     .committee_exchange
                     .sign_da_proposal(&payload_commitment);
                 let data: DAProposal<TYPES> = DAProposal {
-                    block_payload: block.clone(),
+                    block_payload: payload.clone(),
```
jbearer, line 408: Should we send ` payload.encode()` here? As we discussed before the DA member task will need to validate the raw payload anyways, so we can save them a step of calling `encode()` to get the raw payload, and also avoid serializing extra metadata in the structured form of `payload` that the receiver of this message cannot trust anyways.

jbearer: Speaking of which, elsewhere in this file when we handle `DAProposalRecv`, we need to call `decode()` to compute the commitment from the raw payload, rather than trusting the commitment in the proposal as we do now

crates/task-impls/src/transactions.rs
```diff
@@ -219,24 +216,32 @@ where
                 // TODO (Keyao) Determine whether to allow empty blocks.
                 // <https://github.com/EspressoSystems/HotShot/issues/1822>
                 let txns = self.wait_for_transactions(parent_leaf).await?;
-                let encoded_txns = VIDTransaction::encode(txns.clone());
+                // TODO (Keyao) We encode the transactions and compute the VID disperse data twice,
```
jbearer, line 219: I'm not sure this is actually redundant. `from_transactions` and `encode` are doing semantically different things: `from_transactions` is converting a list of transactions into an application-specific block type, which may be represented a variety of ways. `encode` is requesting a specific representation (flat bytes) of that payload.

In practice, in the sequencer, we plan to have our implementation of `BlockPayload` store the flat bytes directly, since we need it anyways for doing VID proofs. Then `from_transactions` will do the work of flattening, and `encode` will just return the flat bytes, so no work is wasted.

crates/types/src/block_impl.rs
```diff
@@ -44,10 +44,9 @@ impl VIDTransaction {
 
         for txn in transactions {
             // Encode the length of the inner transaction and the transaction bytes.
-            if let Ok(len) = txn.0.len().try_into() {
-                encoded.extend::<Vec<u8>>(vec![len]);
-                encoded.extend(txn.0);
-            }
+            let txn_size = txn.0.len().to_le_bytes();
```
jbearer, line 47: `len()` here returns a `usize`, which doesn't have the same number of bytes on every platform. Thus if we encode this way on a 64-bit machine and send this payload to a 32-bit machine, it won't deserialize properly. Better to do `(txn.0.len() as u32).to_le_bytes()` so that the length is always 4 bytes no matter what

crates/types/src/block_impl.rs
```diff
@@ -153,29 +155,38 @@ impl BlockPayload for VIDBlockPayload {
...
+    fn from_bytes(encoded_transactions: Self::Encode<'_>, _metadata: Self::Metadata) -> Self {
+        let encoded_vec: Vec<u8> = encoded_transactions.collect();
         let mut transactions = Vec::new();
         let mut current_index = 0;
-        while current_index < encoded_transactions.len() {
-            let txn_len = encoded_transactions[current_index] as usize;
-            let next_index = current_index + txn_len;
+        while current_index < encoded_vec.len() {
+            // Decode the length of the transaction and the transaction bytes.
+            let txn_start_index = current_index + size_of::<usize>();
```
jbearer, line 164: Same here, we should replace `usize` with something platform-independent, e.g. `u32`

crates/types/src/traits/block_contents.rs
```diff
@@ -47,20 +46,27 @@ pub trait BlockPayload:
     type Transaction: Transaction;
 
     /// Data created during block building which feeds into the block header
-    type Metadata;
+    type Metadata: Clone + Debug + Eq + Hash + Send + Sync;
+
+    /// Encoded payload.
+    type Encode<'a>: 'a + Iterator<Item = u8> + AsRef<[u8]> + Send;
```
jbearer, line 52: Do we need `AsRef` here? That basically forces the implementation to be a contiguous array of bytes which kind of defeats the purpose of using `Iterator`

crates/types/src/traits/block_contents.rs
```diff
@@ -47,20 +46,27 @@ pub trait BlockPayload:
...
     /// Build a payload and associated metadata with the transactions.
-    fn build(transactions: impl IntoIterator<Item = Self::Transaction>) -> (Self, Self::Metadata);
+    fn from_transactions(
+        transactions: impl IntoIterator<Item = Self::Transaction>,
+    ) -> (Self, Self::Metadata);
 
-    /// Encode the payload
-    fn encode(&self) -> Vec<u8>;
+    /// Build a payload with the encoded transaction bytes and metadata.
+    fn from_bytes(encoded_transactions: Self::Encode<'_>, metadata: Self::Metadata) -> Self;
```
jbearer, line 60: `Self::Encode` may be a problem here. This means the method can only be called with a specific type of iterator, a type which can only be constructed by calling `encode`. What if we want to serialize the results of `encode`, send it over the network, receive it as a `Vec<u8>`, and then decode that? I think this function should be generic:
```suggestion
    fn from_bytes<I>(encoded_transactions: I, metadata: Self::Metadata) -> Self
    where I: Iterator<Item = u8>;
```

jbearer (commented):
I think the overall shape of things is right now. All of my comments are just details

crates/task-impls/src/da.rs
shenkeyao: Yeah, I'm going to do this in the next PR. Issue: https://github.com/EspressoSystems/HotShot/issues/2026.

crates/types/src/block_impl.rs
shenkeyao: I thought using `size_of::<usize>()` would make the deserialization correct, but yeah, I think using `u32` is better so that we don't have to handle the platform difference.

crates/types/src/traits/block_contents.rs
shenkeyao: There was a build error before adding `AsRef`, but I'll try removing it and see if the error occurs again.

crates/task-impls/src/transactions.rs
shenkeyao: Makes sense! How about the VID disperse calculation? I think both `from_transactions` (in the sequencer implementation) and `vid.disperse` will do it.

ggutoski: I think we need to clarify. The sequencer (ie. the impl of `BlockPayload`) will never call VID `disperse` or `commit_only`. It's the job of hotshot to compute the VID commitment when needed. (The leader gets the commitment from `disperse`, the DA committee needs to call `commit_only` to verify the payload bytes.)

crates/types/src/block_impl.rs
ggutoski: Are we assuming that `usize` > `u32`? If not then we shouldn't use `as u32` because that will silently wrap. (Why is it that the easy conversion using `as` is also the most dangerous?? Thanks Rust.)

ggutoski: Edited previous comment. As a rule I don't like anything `usize as u32`. In this case it's probably ok because the length will never be large enough to overflow a u32. But I'd rather not rely on that.

crates/task-impls/src/transactions.rs
shenkeyao: Got it, thanks for the clarification. Then there's no redundancy here.

crates/types/src/block_impl.rs
shenkeyao: Thinking more about this, if we are going to use `usize as`, I think we should use `u64` instead, to make sure the length is enough.

crates/types/src/traits/block_contents.rs
shenkeyao: Ahh, it's here in the transactions task:
```rust
let vid_disperse = vid.disperse(encoded_txns).unwrap();
```
`disperse` requires `encoded_txns` to be `AsRef<[u8]>`.

crates/types/src/block_impl.rs
ggutoski: uh, are we willing to send an additional 4 bytes of zeros over the wire just to save us the trouble of properly converting usize to u32? If panic is unacceptable then perhaps the method could return a `Result`.

crates/types/src/block_impl.rs
ggutoski: @jbearer why does `block_impl.rs` need to exist? Any production code implementing `BlockPayload` should not be in hotshot, correct?

crates/types/src/block_impl.rs
ggutoski: What's the purpose of this `encode` method? I'm not aware of any reason why hotshot should flatten a vec of tx bytes like this. Indeed, I it's not clear to me that anything in this file needs to exist. See my [other comment](https://github.com/EspressoSystems/HotShot/pull/2045#discussion_r1389992991)

crates/types/src/block_impl.rs
shenkeyao: All the implementations in `block_impl.rs` are for examples/tests only. Ideally, it should be moved to a place explicitly for non-production code, but files that depend on it are not well-organized, so moving it may cause dependency issues. I'll create an issue to organize our non-production code.

crates/types/src/block_impl.rs
shenkeyao: Ah, this is a good point. Since Everything in this file is non-production code (but does need to exist), I think panic should be acceptable and there's no need to use u64.

crates/types/src/traits/block_contents.rs
ggutoski: Yes we should remove `AsRef` here. In the future I'll change jellyfish to take an iterator over `u8` instead of `AsRef<[u8]>` so that you can pass `encoded_txns` directly into `disperse`. In the meantime please just collect the iterator yourself into a `Vec` and then pass to `disperse`.

crates/types/src/block_impl.rs
ggutoski: > (but does need to exist for HotShot tests)

In that case, is it ok to move everything here into test code?

jbearer: We'll still have this problem in the sequencer, though. I think it is probably ok for `encode` to return a `Result`. It is only called in block building, so if it returns `Err` it basically means you have not built a valid block. In this case the leader could try a different builder or just not proposer. However, it might be slightly better to have `from_transactions` return a `Result`, so you do all the necessary error checking there and then the rest of the functions are pure.

I hate to say it, but we could also bring back the incremental `add_transaction` API, so that you can reject individual transactions (e.g. if their size is more than 2^32 bytes). But that's more of an optimization, we definitely don't need it now.

shenkeyao: >In that case, is it ok to move everything here into test code?

Yes! I've created an [issue](https://github.com/EspressoSystems/HotShot/issues/2059). Will work on that next.

> However, it might be slightly better to have from_transactions return a Result
Yeah, I think this is better. Will update `encode` and `from_transactions` (not bringing back `add_transaction` for now).

crates/task-impls/src/transactions.rs
```diff
@@ -216,54 +216,61 @@ where
...
-                            },
-                            // TODO (Keyao) This is also signed in DA task.
-                            signature: self
-                                .quorum_exchange
-                                .sign_payload_commitment(payload.commit()),
-                        },
-                        self.quorum_exchange.public_key().clone(),
-                    ))
-                    .await;
+                match <TYPES::BlockPayload as BlockPayload>::from_transactions(txns) {
```
jbearer, line 219: Nit: I feel like we can clean up this nesting slightly by returning early on errors. Like
```rust
let x = match fallible_function() {
  Ok(x) => x,
  Err(e) => {
    error!("something");
    return None;
  }
};
let y = match another_fallible_function(x) {
  ...
};
```

crates/task-impls/src/transactions.rs
```diff
@@ -216,54 +216,61 @@ where
...
+                    Ok((payload, metadata)) => {
+                        match <TYPES::BlockPayload as BlockPayload>::encode(&payload) {
```
jbearer, line 221: Nit: can we not just do `payload.encode()`?

crates/types/src/traits/block_contents.rs
```diff
@@ -49,21 +59,33 @@ pub trait BlockPayload:
...
 
     /// Build the genesis payload and metadata.
     fn genesis() -> (Self, Self::Metadata);
 
     /// Encode the payload
-    fn encode(&self) -> Self::Encode<'_>;
+    ///
+    /// # Errors
+    /// If the transaction length conversion fails.
+    fn encode(&self) -> Result<Self::Encode<'_>, BlockError>;
```
jbearer, line 88: I'm kind of waffling on this, but I'm not sure this needs to return `Result`. I feel like there should be some consistency property where if `from_transactions` succeeds, `encode` will never fail, so all the error checking is done up front in `from_transactions`. I'm not sure, though, and I guess it's not so bad if we do return a `Result` here, since this is only called in the builder and doesn't introduce a lot of annoying error handling into HotShot proper

crates/types/src/traits/block_contents.rs
```diff
@@ -49,21 +59,33 @@ pub trait BlockPayload:
...
+        Self: 'a;
 
     /// Build a payload and associated metadata with the transactions.
+    ///
+    /// # Errors
+    /// If the transaction length conversion fails.
     fn from_transactions(
         transactions: impl IntoIterator<Item = Self::Transaction>,
-    ) -> (Self, Self::Metadata);
+    ) -> Result<(Self, Self::Metadata), BlockError>;
```
jbearer, line 72: Downstream crates (ie the sequencer) might want more control over what kinds of errors can happen here then just the error types that are predefined in `BlockError`. There are many ways to do this because Rust error handling sucks. I don't have a strong opinion over which is the least bad:
* Make the error an associated type `type Error: Debug + std::error::Error`, then return `Self::Error`
* Make the error something generalized like `anyhow::Error`
* Add a `Custom { message: String }` variant to `BlockError`

crates/types/src/traits/block_contents.rs
shenkeyao: > if from_transactions succeeds, encode will never fail

Yeah, I think property makes sense. On the other hand, if `from_transactions` fails, that means `encode` fails, which should return an error, right? So I'm not sure how we could make `encode` not return `Result`, unless we do the length check in `from_transactions` and assume `u32::try_from` won't fail in `encode`, which I think is a bit confusing as well.

crates/types/src/traits/block_contents.rs
shenkeyao: Ah, actually the first solution was the plan. I've added an associated type `Error: Error + Debug + Send + Sync` to the `Payload` trait, but forgot its existence 😅  and used `BlockError` instead. Will update this!

crates/types/src/traits/block_contents.rs
jbearer: If `from_transactions` fails you have no block payload to call `encode` on. You just failed to build a block. But this way there is only one way to fail to build a block instead of two ways.

But you are correct, the way to implement this for `VIDBlockPayload` would be to do all the length checks in `from_transactions` and then panic if `u32::try_from` fails in `encode`, which is ugly and error prone. In the sequencer, our block type will actually contain the flat bytes, which we can just return directly in `encode`, so we don't have this problem.

I'm fine keeping the `Result`, though

jbearer (dismissed):
I think it looks good, except see [this thread](https://github.com/EspressoSystems/espresso-sequencer/pull/756#discussion_r1394543815) about possibly removing the `commit()` method of `BlockPayload` and having HotShot keep track of the commit on its own. WDYT about that @shenkeyao @ggutoski

shenkeyao:
> I think it looks good, except see [this thread](https://github.com/EspressoSystems/espresso-sequencer/pull/756#discussion_r1394543815) about possibly removing the `commit()` method of `BlockPayload` and having HotShot keep track of the commit on its own. WDYT about that @shenkeyao @ggutoski

@jbearer As mentioned in that [thread](https://github.com/EspressoSystems/espresso-sequencer/pull/756#discussion_r1394896694), I think it makes sense to remove the `Commitable` bound. I've made an issue and will do that after merging this PR.

jbearer (approved):

crates/types/src/traits/block_contents.rs
ggutoski: @shenkeyao [more recent conversation with Jeb](https://github.com/EspressoSystems/hotshot-query-service/issues/266) suggests an even better solution: [`Cow`](https://doc.rust-lang.org/std/borrow/enum.Cow.html). Flagging this tip for future work.

PR #2054, opened 2023-11-14
Ed/view sync cert vote

elliedavidson:
This PR: 
* Separates the `ViewSyncVote` and `ViewSyncCertificate` types into their 3 respective types.  
* Updates the view sync task accordingly

This PR does not: 
* Implement the `failure_threshold` for `PreCommit` Certificates. That will be done in [this](https://github.com/EspressoSystems/HotShot/issues/2056) issue since it will require touching additional code. I wanted to keep this PR smaller and unblock other work to remove `Exchanges` traits.


crates/types/src/traits/election.rs
```diff
@@ -687,143 +682,99 @@ impl<
...
-                error!("Vote token was invalid");
-                false
-            }
-            Ok(Checked::Valid(_)) => true,
-            Ok(Checked::Inval(_) | Checked::Unchecked(_)) => false,
-        };
-
-        is_valid_signature && is_valid_vote_token
-    }
+    // /// A vote on a [`Proposal`](Self::Proposal).
```
bfish713, line 685: Let's remove the commented code

crates/testing/tests/view_sync_task.rs
```diff
@@ -20,44 +17,37 @@ use std::collections::HashMap;
...
-        .unwrap_or_else(|| panic!("Not chosen for the committee"));
-    let vote_data_internal: ViewSyncData<TestTypes> = ViewSyncData {
-        relay: relay_pub_key.clone(),
-        round: ViewNumber::new(5),
-    };
-    let vote_data_internal_commitment = vote_data_internal.commit();
-    let signature = view_sync_exchange.sign_precommit_message(vote_data_internal_commitment);
-    let vote = ViewSyncVote::PreCommit(ViewSyncVoteInternal {
-        relay_pub_key,
+    let _relay_pub_key = api.public_key().to_bytes();
```
bfish713, line 34: Nit: remove unused

crates/task-impls/src/view_sync.rs
```diff
@@ -945,110 +1103,230 @@ impl<
             Leaf = Leaf<TYPES>,
             ConsensusMessage = SequencingMessage<TYPES, I>,
         >,
-    > ViewSyncRelayTaskState<TYPES, I>
-where
-    ViewSyncEx<TYPES, I>: ViewSyncExchangeType<
+    >
+    ViewSyncRelayTaskState<
```
bfish713, line 1107: Just so I don't miss anything in the review: we basically have 3 copies of this task for each vote + cert pair?

crates/task-impls/src/view_sync.rs
```diff
@@ -344,14 +316,94 @@ where
...
+                let event_stream_id = builder.get_stream_id().unwrap();
+
+                self.relay_task_map
+                    .insert(vote.get_view_number(), ViewSyncTaskInfo { event_stream_id });
+                let _view_sync_relay_task = async_spawn(async move {
+                    ViewSyncRelayTaskStateTypes::build(builder).launch().await
+                });
+            }
+
+            HotShotEvent::ViewSyncCommitVoteRecv(vote) => {
```
bfish713, line 405: Same question as well here, are all the vote handler blocks the same but with different generic types for the relay task?

bfish713 (commented):
Overall looks good, a few nits.  We talked about the code duplication just want to make sure I'm not missing some subtle differences in the vote handling if the code is not exactly replicated.  We can go back and refactor the common stuff after

crates/task-impls/src/view_sync.rs
elliedavidson: Yes, that is correct.  All this task does is collect votes into a certificate, so we can probably replace it with the generic vote collection task we talked about.

crates/task-impls/src/view_sync.rs
elliedavidson: Yes, they should all be the same.  I tried to make the `handle_event` function generic over the VOTE and CERT types to avoid this duplication, but the compiler complained when matching on events.  I think the solution is to break up the `handle_event` function so that each implementation of it only handles events related to 1 type of vote and certificate.  Definitely part of the clean up I need to do for view sync.

bfish713 (approved):

bfish713 (commented):
LGTM! but a few lint errors

bfish713 (approved):

PR #2055, opened 2023-11-14
[hotshot-state-prover] proof generation

mrain:
closes: #1985 

I tried but cannot make functions generics because of the StakeTable trait bounds. It's too complicated.

The tests are not implemented yet for srs problem. I'm handling it now.

### DIFF
--- a/crates/hotshot-state-prover/src/circuit.rs
+++ b/crates/hotshot-state-prover/src/circuit.rs
@@ -2,25 +2,20 @@
 
 use ark_ec::twisted_edwards::TECurveConfig;
 use ark_ff::PrimeField;
+use ark_std::borrow::Borrow;
 use ethereum_types::U256;
 use hotshot_stake_table::config::STAKE_TABLE_CAPACITY;
-use hotshot_types::traits::{
-    stake_table::{SnapshotVersion, StakeTableScheme},
-    state::LightClientState,
-};
+use hotshot_types::traits::state::LightClientState;
 use jf_plonk::errors::PlonkError;
 use jf_primitives::{
     circuit::{
         rescue::RescueNativeGadget,
-        signature::schnorr::{SignatureGadget, SignatureVar, VerKeyVar},
+        signature::schnorr::{SignatureGadget, VerKeyVar},
     },
     rescue::RescueParameter,
-    signatures::{
-        bls_over_bn254::VerKey as BLSVerKey,
-        schnorr::{Signature, VerKey as SchnorrVerKey},
-    },
+    signatures::schnorr::{Signature, VerKey as SchnorrVerKey},
 };
-use jf_relation::{errors::CircuitError, BoolVar, Circuit, PlonkCircuit, Variable};
+use jf_relation::{errors::CircuitError, Circuit, PlonkCircuit, Variable};
 
 /// Lossy conversion of a U256 into a field element.
 pub(crate) fn u256_to_field<F: PrimeField>(v: &U256) -> F {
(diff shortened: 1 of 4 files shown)


crates/hotshot-state-prover/src/circuit.rs
```diff
@@ -116,236 +114,191 @@ impl AsRef<[Variable]> for LightClientStateVar {
     }
 }
 
-#[derive(Clone, Debug)]
-pub struct StateUpdateBuilder<F: RescueParameter>(PhantomData<F>);
-
-impl<F> StateUpdateBuilder<F>
+/// A function that takes as input:
+/// - a list of stake table entries (`Vec<(BLSVerKey, Amount, SchnorrVerKey)>`)
```
chancharles92, line 118: Minor: mention that `Vec<BLSVerKey>` is not used here.

crates/hotshot-state-prover/src/circuit.rs
```diff
@@ -116,236 +114,191 @@ impl AsRef<[Variable]> for LightClientStateVar {
...
+    let mut signer_bit_vec_var = signer_bit_vec
+        .iter()
+        .map(|&b| circuit.create_boolean_variable(b))
+        .collect::<Result<Vec<_>, CircuitError>>()?;
+    signer_bit_vec_var.resize(STAKE_TABLE_CAPACITY, BoolVar(circuit.zero()));
+
+    let threshold = u256_to_field::<F>(threshold);
+    let threshold_var = circuit.create_public_variable(threshold)?;
+
+    let lightclient_state_var = LightClientStateVar::new(&mut circuit, lightclient_state)?;
```
chancharles92, line 189: Maybe rename it to `lightclient_state_pub_input_var` to remind the user that they are public input variables.

crates/hotshot-state-prover/src/lib.rs
```diff
@@ -1,3 +1,91 @@
...
+    R: CryptoRng + RngCore,
+{
+    let (circuit, public_inputs) = build_state_verifier_circuit(
+        stake_table,
+        sigs,
+        lightclient_state,
+        signer_bit_vec,
+        threshold,
+    )?;
+    let proof = PlonkKzgSnark::<Bn254>::prove::<_, _, StandardTranscript>(rng, &circuit, pk, None)?;
```
chancharles92, line 75: Is `StandardTranscript` using Keccak (that's cheap in contracts)?

crates/hotshot-state-prover/src/lib.rs
```diff
@@ -1,3 +1,91 @@
 //! SNARK-assisted light client state update verification in HotShot
 
+/// State verifier circuit builder
```
chancharles92, line 3: Remember to add minimal tests.

crates/hotshot-state-prover/src/circuit.rs
```diff
@@ -116,236 +114,191 @@ impl AsRef<[Variable]> for LightClientStateVar {
...
-
-        circuit.finalize_for_arithmetization()?;
-        Ok((circuit, public_inputs))
+        })
+        .collect::<Result<Vec<_>, CircuitError>>()?;
+    let dummy_ver_key_var = VerKeyVar(circuit.neutral_point_variable());
+    stake_table_var.resize(
+        STAKE_TABLE_CAPACITY,
+        StakeTableEntryVar {
+            schnorr_ver_key: dummy_ver_key_var,
```
chancharles92, line 161: Is `dummy_ver_key_var` a constant variable? If so, the circuit will be different if the number of *valid* staking schnorr keys is different.

crates/hotshot-state-prover/src/circuit.rs
```diff
@@ -116,236 +114,191 @@ impl AsRef<[Variable]> for LightClientStateVar {
...
+            stake_amount: 0,
```
chancharles92, line 162: Should it be a variable rather than zero?

crates/hotshot-state-prover/src/circuit.rs
```diff
@@ -116,236 +114,191 @@ impl AsRef<[Variable]> for LightClientStateVar {
...
+    );
+
+    // creating variables for signatures
+    let mut sig_vars = sigs
+        .iter()
+        .map(|sig| circuit.create_signature_variable(sig))
+        .collect::<Result<Vec<_>, CircuitError>>()?;
+    sig_vars.resize(
+        STAKE_TABLE_CAPACITY,
+        SignatureVar {
```
chancharles92, line 173: Same here, the circuit will be different if the number of valid signatures differs.

chancharles92 (changes requested):

crates/hotshot-state-prover/src/circuit.rs
mrain: You are right. Maybe it's better to pad the input outside the circuit building function. Let me fix it.

crates/hotshot-state-prover/src/circuit.rs
mrain: I changed the input type. It's not taking `BLSVerKey` anymore.

crates/hotshot-state-prover/src/circuit.rs
mrain: Fixed now.

crates/hotshot-state-prover/src/lib.rs
mrain: Test added now

crates/hotshot-state-prover/src/lib.rs
mrain: I'm not sure about the details. But we can fix this during integration (should be easy)

crates/hotshot-state-prover/src/circuit.rs
mrain: Done in 3c40ff1

mrain:
This PR is ready for view. cc @alxiong @chancharles92 

1. padding is done in the circuit building function to avoid more memory copies.
2. It's hard to make proof generation function generic. There are too many trait bounds. Although, this is really tempting to me.
3. Only minimal test for proof generation, many bad cases are checked inside the circuit building tests.
4. Other cleanup tasks will be handled during integration (e.g. struct alignment with consensus implementation)

crates/hotshot-state-prover/src/circuit.rs
```diff
@@ -116,236 +109,244 @@ impl AsRef<[Variable]> for LightClientStateVar {
...
+    P: TECurveConfig<BaseField = F>,
+    STIter: IntoIterator,
+    STIter::Item: Borrow<(SchnorrVerKey<P>, U256)>,
+    STIter::IntoIter: ExactSizeIterator,
+    BitIter: IntoIterator,
+    BitIter::Item: Borrow<bool>,
+    BitIter::IntoIter: ExactSizeIterator,
+    SigIter: IntoIterator,
+    SigIter::Item: Borrow<Signature<P>>,
+    SigIter::IntoIter: ExactSizeIterator,
```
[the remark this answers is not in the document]
alxiong: Can I also propose to change `STIter` to `ST::IntoIter where ST: StakeTableScheme`? 
I want to avoid the extra iterating over stake table entries just to build an input for this function from the outside:

```rust
// I'd prefer avoid building `stake_table_entries`
    let stake_table_entries = stake_table
        .try_iter(SnapshotVersion::LastEpochStart)
        .unwrap()
        .map(|(_, stake_amount, schnorr_key)| (schnorr_key, stake_amount))
        .collect::<Vec<_>>();
    let (circuit, public_inputs) = build_state_verifier_circuit(
        stake_table_entries, // here I want to directly pass in `st.try_iter()?`

``` 

Is this possible?

crates/hotshot-state-prover/src/lib.rs
alxiong: we should change it to `SolidityTranscript` now.

There's more than just use of hash, but also whether `label` fields are hashed into the transcript, (StandardTranscript does, thus creates mismatch with what contract expect)

crates/hotshot-state-prover/src/circuit.rs
```diff
@@ -116,236 +109,244 @@ impl AsRef<[Variable]> for LightClientStateVar {
...
+    circuit.enforce_true(sig_ver_result.0)?;
+
+    circuit.finalize_for_arithmetization()?;
+    Ok((circuit, public_inputs))
 }
 
 #[cfg(test)]
 mod tests {
-    use super::{LightClientState, StateUpdateBuilder};
+    use super::{build_state_verifier_circuit, LightClientState};
```
alxiong, line 336: Should add `From<LightClientState> for Vec<F>` or maybe `LightClient::to_fields()` so that we don't need to manually do `let state_msg = [ /* manually constructing each items */ ]` later in the test

crates/hotshot-state-prover/src/lib.rs
```diff
@@ -1,3 +1,322 @@
...
+            powers_of_g,
+            h,
+            beta_h,
+        };
+        end_timer!(setup_time);
+        Ok(pp)
+    }
+
+    /// Internal function for helping generate the proving/verifying key
+    fn get_num_of_gates() -> Result<usize, PlonkError> {
```
alxiong, line 199: instead of this, remove it, and directly use `build_for_preprocessing()?.0.num_gates()` instead

alxiong (commented):
👍 good work!

Apart from concrete comments, I'd like to see better/clearer organization of bad path in circuit tests.
e.g. `let bit_masked_sig = |sigs| { .. }` could be an internal lambda function, and can summarize some helper function to make the test code reads more smoothly.

^^ just my opinion, improve at @mrain's discretion.

crates/hotshot-state-prover/src/lib.rs
```diff
@@ -1,3 +1,322 @@
 //! SNARK-assisted light client state update verification in HotShot
 
+/// State verifier circuit builder
 pub mod circuit;
+mod utils;
```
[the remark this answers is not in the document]
mrain: Tried before, but how do you deal with the imports?

crates/hotshot-state-prover/src/lib.rs
```diff
@@ -1,3 +1,322 @@
...
+        signatures,
+        lightclient_state,
+        threshold,
+    )?;
+    let proof = PlonkKzgSnark::<Bn254>::prove::<_, _, StandardTranscript>(rng, &circuit, pk, None)?;
+    Ok((proof, public_inputs))
+}
+
+/// Internal function for helping generate the proving/verifying key
+fn build_dummy_circuit_for_preprocessing(
```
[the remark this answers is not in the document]
mrain: This function is implemented for concrete types. But things in `circuit.rs` is very generic, so this function doesn't quite fit there.
Maybe in the future if we have everything generic, we can take such move.

crates/hotshot-state-prover/src/circuit.rs
```diff
@@ -116,236 +109,244 @@ impl AsRef<[Variable]> for LightClientStateVar {
...
+/// The list of stake table entries, bit indicators and signatures will be padded to the `STAKE_TABLE_CAPACITY`.
+/// It checks that
+/// - the signer's accumulated weight exceeds the quorum threshold
+/// - the stake table corresponds to the one committed in the light client state
+/// - all signed schnorr signatures are valid
+/// and returns
+/// - A circuit for proof generation
+/// - A list of public inputs for verification
+/// - A PlonkError if any error happens when building the circuit
+pub(crate) fn build_state_verifier_circuit<F, P, STIter, BitIter, SigIter>(
```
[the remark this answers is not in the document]
mrain: Done in febaee5

crates/hotshot-state-prover/src/circuit.rs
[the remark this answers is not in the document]
mrain: Borrow is for more generic input types, e.g. iterator of references.
I add an extra bound `ST::IntoIterator: ExactIterator`, so I can directly pass in the `map` without collecting the entries. 
check febaee5

crates/hotshot-state-prover/src/circuit.rs
```diff
@@ -116,236 +109,244 @@ impl AsRef<[Variable]> for LightClientStateVar {
...
+    stake_table_entries: STIter,
+    signer_bit_vec: BitIter,
+    signatures: SigIter,
+    lightclient_state: &LightClientState<F>,
+    threshold: &U256,
+) -> Result<(PlonkCircuit<F>, Vec<F>), PlonkError>
```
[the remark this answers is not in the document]
mrain: In febaee5, add a wrapper `PublicInputs`. We may only need to implement `AsRef<[F]>`.

crates/hotshot-state-prover/src/circuit.rs
```diff
@@ -116,236 +109,244 @@ impl AsRef<[Variable]> for LightClientStateVar {
...
+        .collect::<Result<Vec<_>, CircuitError>>()?;
+    stake_table_var.extend(
+        (0..stake_table_entries_pad_len)
+            .map(|_| {
+                let schnorr_ver_key =
+                    circuit.create_signature_vk_variable(&SchnorrVerKey::<P>::default())?;
+                let stake_amount = circuit.create_variable(F::default())?;
                 Ok(StakeTableEntryVar {
                     schnorr_ver_key,
                     stake_amount,
```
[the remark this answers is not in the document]
mrain: The problem here is, internal function is returning a `Result` type. How do you capture the error with `resize_with`?

crates/hotshot-state-prover/src/lib.rs
mrain: Done in 88784b9

crates/hotshot-state-prover/src/circuit.rs
mrain: Implements a `to_array()` for `LightClientState` in 88784b9

crates/hotshot-state-prover/src/circuit.rs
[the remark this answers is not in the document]
alxiong: I was hoping to have more semantically meaningful fields, rather than simply wrapping a vector.

crates/hotshot-state-prover/src/lib.rs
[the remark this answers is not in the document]
alxiong: hmm... I see. But my gut feeling is we shouldn't locate stuff based on "how generic" they are. We group them by functionality and module, ideally.

crates/hotshot-state-prover/src/circuit.rs
alxiong: may I ask why not `to_fields`? that seems clearer than just generic "array"

crates/hotshot-state-prover/src/circuit.rs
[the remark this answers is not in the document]
alxiong: Right, but we do we need the flexibility of accepting iterator of references for now? 

Let's think about the actual usage, the prover will have these lay flat out in its memory, everything owned, 
to even have a vector of reference the code will even need to build a separate vector, which I don't think the prover code would do.

the `.iter()` by default only takes references of `Item` and won't consume the ownership to start with. So I prefer removing these heavy trait bounds including the `Borrow` here.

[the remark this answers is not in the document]
mrain: First, there's no harm in make it more generic right now. 
Second, in the test where you are building multiple good or bad circuits, if you take the ownership, you need then construct multiple stake table entry lists. Doesn't look pretty that way

crates/hotshot-state-prover/src/circuit.rs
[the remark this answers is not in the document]
mrain: I thought about it. But do we really need to access the element inside? The only usage case is to feed them into proof generation/verification.

crates/hotshot-state-prover/src/circuit.rs
mrain: That's just my preference of array over vector. Since it's fixed length.

alxiong (changes requested):
we need to change `block_comm` name as it is wrong, (both in this PR and in `types/traits/state.rs` 

suggestions:
- `block_comms_root`
- `block_comm_history_root`

alxiong (approved):
LGTM! 🚀

elliedavidson (approved):

PR #2071, opened 2023-11-15
Remove Message Traits

bfish713:
Removing this from Node Implementation will simplify the Exchanges work I'm doing a little bit.  This was left over from when we had two different consensus types and therefore different message types.  We used the `SequencingMessage` struct everywhere in the code right now


crates/hotshot/src/lib.rs
```diff
@@ -254,28 +254,19 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> SystemContext<TYPES, I> {
     pub async fn timeout_view(
```
elliedavidson, line 254: Maybe outside this PR's scope, but do we ever use this function anymore?

elliedavidson (approved):

crates/hotshot/src/lib.rs
bfish713: Good catch!, nope it's not used, I'll clean it up in a different PR

PR #2077, opened 2023-11-17
VID: Reorganize and create new events

### LINKED ISSUE #1817: Determine and update where to publish `VidDisperseSend`
We should determine whether `VidDisperseSend` is supposed to be published in the transaction task as is now, or in the DA task, together with `DAProposalSend`. We may need to update the structure of the `BlockReady` event.

rob-maron:
Aims to close #1817


crates/task-impls/src/vid.rs
```diff
@@ -355,6 +364,53 @@ where
...
+
+            HotShotEvent::TransactionsSequenced(payload, _metadata, view) => {
+                // encode transactions for disperse calculation
+                let encoded_txns = match payload.encode() {
+                    Ok(encoded) => encoded,
+                    Err(e) => {
+                        error!("Failed to encode the block payload: {:?}.", e);
+                        return None;
+                    }
+                };
```
ggutoski, line 376: Is there no way to replace 6 lines of code with the `?` operator?

crates/task-impls/src/vid.rs
```diff
@@ -355,6 +364,53 @@ where
...
+                // TODO We are using constant numbers for now, but they will change as the quorum size
+                // changes.
+                // TODO <https://github.com/EspressoSystems/HotShot/issues/1693>
+                let vid = VidScheme::new(NUM_CHUNKS, NUM_STORAGE_NODES, &srs).unwrap();
+                let vid_disperse = vid
+                    .disperse(encoded_txns.into_iter().collect::<Vec<u8>>())
+                    .unwrap();
+
+                // TODO never clone a block
+                // https://github.com/EspressoSystems/HotShot/issues/1858
```
ggutoski, line 389: I think we can remove this comment?

crates/task-impls/src/vid.rs
```diff
@@ -355,6 +364,53 @@ where
...
+                self.event_stream
+                    .publish(HotShotEvent::BlockReady(
+                        Proposal {
+                            data: VidDisperse {
+                                view_number: view,
+                                payload_commitment: payload.commit(),
```
ggutoski, line 395: Not sure whether this comment is with scope of this PR but...

I think this should be `vid_disperse.commit` instead of `payload.commit()`. That means we need to change the `VidDisperse` struct so that the type of `payload_commitment` is `<VidScheme as VidSchemeTrait>::Commit` just like the other fields `shares`, `common`.

The goal is to remove `Committable` from `BlockPayload`, correct? In that case we should never be calling `commit()`.

crates/task-impls/src/vid.rs
rob-maron: Unfortunately, the `handle_event` function this is in returns an option. I could change the return type to a `Result<Option<>>` to pass the error and print it up the stack if wanted

crates/task-impls/src/vid.rs
rob-maron: Yeah, looks like we don't clone the block anymore 😃

crates/task-impls/src/events.rs
```diff
@@ -94,17 +94,19 @@ pub enum HotShotEvent<TYPES: NodeType, I: NodeImplementation<TYPES>> {
     TransactionsRecv(Vec<TYPES::Transaction>),
     /// Send transactions to the network
     TransactionSend(TYPES::Transaction, TYPES::SignatureKey),
+    /// Transactions have been sequenced
+    TransactionsSequenced(
+        TYPES::BlockPayload,
+        <TYPES::BlockPayload as BlockPayload>::Metadata,
+        TYPES::Time,
+    ),
```
elliedavidson, line 102: Doesn't the `BlockPayload` type currently require the vid commitment?  If so, then I think we need to change this `BlockPayload` type here to something else, like a type that just contains the transactions.

elliedavidson (changes requested):

elliedavidson:
I'd like both @shenkeyao and @ggutoski to approve this before merging :)

crates/task-impls/src/events.rs
rob-maron: Ah yes, looks like it _does_ calculate the vid commitment at that step. Let me fix it

ggutoski: `BlockPayload` should not need the VID commitment.  Quoting from [Jeb's zulip post](https://github.com/EspressoSystems/HotShot/pull/2077#discussion_r1397848207):

I think a minimal interface required to completely solve this problem would be something like

```rust
trait BlockPayload {
  type Transaction;
  type Metadata; // Data created during block building which feeds into the block header, e.g. namespace table

  fn build(transactions: impl IntoIterator<Item = Transaction>) -> (Self, Self::Metadata);
  fn encode(&self) -> Vec<u8>;
  fn decode(bytes: &[u8]) -> Self;
}

trait BlockHeader<Metadata> {
  fn new(payload_commitment: VidCommitment, metadata: Self::Metadata, parent_leaf: Leaf) -> Self;
}
```

Then the workflow would be like this:
* Transactions task uses `build` to create a block payload and namespace table
* Transactions task uses `encode` to flatten block payload into bytes, computes VID commitment
* Leader sends flat bytes *and* metadata to DA committee members
* Committee members compute VID commitment over flat bytes, sign hash of VID commitment and metadata
* Leader uses `BlockHeader::new` to construct block header (and thus proposal) from VID commitment and metadata
* Leader sends proposal including header and DAC to replicas
* Replicas extract metadata and VID commitment from header and check against the signatures in DAC

ggutoski: Relevant discussion: https://github.com/EspressoSystems/HotShot/pull/2045#discussion_r1389986724 (Sorry I'm in a rush and unsure of the final design atm.)

ggutoski: There was debate in the above linked thread on whether to put `Metadata` in the payload or header. But we can leave it as-is for now. The important part is that the VID commitment goes with the parent leaf in the header, *not in the payload*.

shenkeyao: I think this should also include the encoded transactions, which with the metadata the leader can use to get the block payload and the commitment.

crates/task-impls/src/vid.rs
shenkeyao: Yeah after removing `Committable` from `BlockPayload`, calling `commit()` will no longer work.

crates/task-impls/src/vid.rs
```diff
@@ -355,6 +364,53 @@ where
...
+                                shares: vid_disperse.shares,
+                                common: vid_disperse.common,
+                            },
+                            // TODO (Keyao) This is also signed in DA task.
```
shenkeyao, line 399: Nit: I think we could remove this comment since the signing isn't the same as what's in the DA task anymore.

crates/task-impls/src/vid.rs
shenkeyao: But if it's too complicated to change this, you could leave it to me. I'm going to update other `commit()` calls anyway.

rob-maron: Okay, I will leave it to your changes (removing commit)

crates/task-impls/src/events.rs
rob-maron: Okay, so I think the plan is now as follows:

Since @shenkeyao is removing commit(), the changes for _this_ PR will just be to include the encode transactions in `TransactionsSequenced`.

shenkeyao (dismissed):
LGTM!

crates/task-impls/src/events.rs
```diff
@@ -94,17 +94,20 @@ pub enum HotShotEvent<TYPES: NodeType, I: NodeImplementation<TYPES>> {
     TransactionsRecv(Vec<TYPES::Transaction>),
     /// Send transactions to the network
     TransactionSend(TYPES::Transaction, TYPES::SignatureKey),
+    /// Transactions have been sequenced
+    TransactionsSequenced(
+        TYPES::BlockPayload,
+        <TYPES::BlockPayload as BlockPayload>::Metadata,
+        Vec<u8>,
```
ggutoski, line 101:
```suggestion
```

I missed this earlier comment https://github.com/EspressoSystems/HotShot/pull/2077#discussion_r1397951288 .

I don't think we should include the encoded txs in this event. If we do this then we're sending 2 copies of the block payload. Anyone receiving this event can get the encoded bytes if they want it simply by calling `encode`.

ggutoski (changes requested):

crates/task-impls/src/events.rs
rob-maron: I agree; I think before we change the payload and header items this makes the most sense. @shenkeyao what do you think?

rob-maron: Okay, sounds good. I'll put this one on hold until that PR is merged

ggutoski: > Oh looks like @ggutoski and I were suggesting different things. Gus was suggesting removing the encoded txns, by I the payload.
> 
> I think the reason we change the `BlockReady` event to `TransactionsSequenced` is that we want to handle the txns rather than the block (payload), so here we should use the encoded txns?

I'm not sure whether to keep the bytes or the payload, but I know we don't want both. In discussion elsewhere it seems like we prefer to keep the bytes everywhere but I'm not yet confident enough to express a strong preference.

Issue #1799, opened 2023-09-22
Remove `Exchanges`


elliedavidson:
Goal is to remove the `Exchanges` trait and remove the `Exchange` trait

PR #2081, opened 2023-11-17
Completely Remove `Exchanges` and `Exchange` traits

bfish713:
Depends on: https://github.com/EspressoSystems/HotShot/pull/2074.  Review that one first since this is a branch on top of that one.

This PR completely removes the `Exchanges` trait and the `Exchange` trait  

To accomplish this 2 types were added to `NodeImplemenation`. They are the two network types for Comittee and Quorum, both of which are `CommunicationChannels`.  If in the future we want to have more network types we should just add them tot the `NodeImplementationTrait`.

In essence all my PRs replace the use of `Exchange` with:
- The new `Vote`, `Certificate`, and `VoteAccumulator` types
- The `Membership` associated type on `NodeTpes`
- The 2 new associated types `QuorumNetwork` and `CommitteeNetwork` on `NodeImplementation`
- Replacing usages of `ConsensusExchange` functions by directly using the network, membership, or other types

Also this PR does some minor reworking in the launcher code.  Most notably the networks and memberships are passed in directly to SystemContext rather than exchanges.

closes: #1799


crates/testing/src/node_types.rs
```diff
@@ -64,572 +58,82 @@ pub struct CombinedImpl;
...
             ChannelMaps::new(start_view),
             Some(ChannelMaps::new(start_view)),
         )
     }
 }
 
-// man these generics are big oof
-// they're a LOT
-// when are we getting HKT for rust
-// smh my head
```
shenkeyao, line 347: Whoever wrote this comment must be happy to see this gone. 😛

shenkeyao (approved):
Just some nit comments.

shenkeyao (approved):
LGTM

crates/task-impls/src/consensus.rs
```diff
@@ -332,11 +297,7 @@ where
...
 
     async fn vote_if_able(&self) -> bool {
-        if !self
-            .quorum_exchange
-            .membership()
-            .has_stake(self.quorum_exchange.public_key())
-        {
+        if !self.quorum_membership.has_stake(&self.public_key) {
             debug!(
                 "We were not chosen for consensus committee on {:?}",
```
elliedavidson, line 302: Nit: Perhaps change this error message to reflect that we don't have stake.

crates/task-impls/src/consensus.rs
```diff
@@ -453,15 +414,15 @@ where
...
-                    let vote = QuorumVote::<TYPES, QuorumMembership<TYPES, I>>::create_signed_vote(
+                    let vote = QuorumVote::<TYPES>::create_signed_vote(
                         QuorumData {
                             leaf_commit: leaf.commit(),
                         },
                         view,
-                        self.quorum_exchange.public_key(),
-                        self.quorum_exchange.private_key(),
+                        &self.public_key,
+                        &self.private_key,
```
elliedavidson, line 423: I feel like we'll need to have public and private key as fields on every task.  Would it make sense to make them part of the Membership type so we don't have to pass them in as fields all the time?

crates/task-impls/src/consensus.rs
```diff
@@ -513,31 +474,27 @@ where
 
             // Poll the future leader for lookahead
             let lookahead_view = new_view + LOOK_AHEAD;
-            if !self.quorum_exchange.is_leader(lookahead_view) {
-                self.quorum_exchange
-                    .network()
+            if self.quorum_membership.get_leader(lookahead_view) != self.public_key {
```
elliedavidson, line 477: I feel like it might be nice to make an "is_leader" function on the membership type, since we will call this same logic often.  That can be for a later PR, though.

crates/task-impls/src/consensus.rs
```diff
@@ -1016,28 +985,29 @@ where
...
-                    let quorum_accumulator = VoteAccumulator2 {
+                    let quorum_accumulator = VoteAccumulator {
                         vote_outcomes: HashMap::new(),
                         sig_lists: Vec::new(),
-                        signers: bitvec![0; self.quorum_exchange.total_nodes()],
+                        signers: bitvec![0; self.quorum_membership.total_nodes()],
                         phantom: PhantomData,
                     };
 
                     // self.timeout_accumulator = accumulator;
```
elliedavidson, line 1005: Nit: remove this comment

crates/task-impls/src/consensus.rs
```diff
@@ -1164,23 +1133,19 @@ where
...
                     return;
                 }
-                if !self
-                    .timeout_exchange
-                    .membership()
-                    .has_stake(self.timeout_exchange.public_key())
-                {
+                if !self.timeout_membership.has_stake(&self.public_key) {
                     debug!(
                         "We were not chosen for consensus committee on {:?}",
```
elliedavidson, line 1138: Nit: Change this debug message to reflect that we don't have stake

crates/testing/src/node_types.rs
```diff
@@ -64,568 +58,82 @@ pub struct CombinedImpl;
...
-    ) {
+    ) -> (ChannelMaps<TestTypes>, Option<ChannelMaps<TestTypes>>) {
         (
             ChannelMaps::new(start_view),
             Some(ChannelMaps::new(start_view)),
         )
     }
 }
 
-impl TestableExchange<TestTypes, Message<TestTypes, Libp2pImpl>> for SequencingLibp2pExchange {
```
elliedavidson, line 140: Wonderful!

elliedavidson (approved):
A few comments, but they don't need to block merging.

bfish713:
> A few comments, but they don't need to block merging.

I will follow up with a PR to fix all these since this PR is already massive enough

PR #2083, opened 2023-11-17
Fix CI Failure: Vote sending timed out in View Sync related event

dailinsubjam:
Set `view_sync_timeout` from 10(s) to 30(s). And increase the task completion time from 200(s) to 250(s).


dailinsubjam:
Will close it as it Somehow fixed by https://github.com/EspressoSystems/HotShot/commit/b6f1b08f62392f567ad21f769d2a0a5ce2940a28.

PR #2093, opened 2023-11-20
Fix Develop Lints for Clippy Update

rob-maron:


DieracDelta (dismissed):

elliedavidson:
I added `#[allow(clippy::should_panic_without_expect)]` instead of specifying the panic message because the panic message is non-deterministic and very long.

DieracDelta (approved):

Issue #2072, opened 2023-11-15
Remove commitment computation for `BlockPayload`


shenkeyao:
Currently `BlockPayload` implements `fn commit` which takes a VID commitment computed from the payload construction. We should get or store the VID commitment in HotShot instead, to avoid the computation waste in the app-specific implementation.

Related discussions: https://github.com/EspressoSystems/espresso-sequencer/pull/756#discussion_r1394811330.

ggutoski:
A long time ago I tried to remove `Committable` from `BlockPayload` but it was infeasible. Perhaps now that the refactor is near done this might be feasible.

PR #2102, opened 2023-11-21
Remove commitment computation from payload and fix DA proposal

shenkeyao:
- Updates commitment computation from payload.
  - Removes `Committable` bound from `BlockPayload`.
  - Adds a function to compute the VID commitment from encoded transactions.
  - Replaces all commitment retrieval calls with the commitment computation function.
- Fixes DA proposal.
  - Replaces the payload commitment in the `BlockReady` event and `DAProposal` struct with encoded transactions and metadata.
  - Computes the VID commitment in the DA task, instead of taking the received info directly.
- Updates test parameters to make byzantine tests pass.

Closes https://github.com/EspressoSystems/HotShot/issues/2072. Closes https://github.com/EspressoSystems/HotShot/issues/2026.


crates/task-impls/src/da.rs
```diff
@@ -371,18 +370,17 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
                 return None;
             }
-            HotShotEvent::BlockReady(payload, metadata, view) => {
+            HotShotEvent::BlockReady(encoded_transactions, metadata, view) => {
                 self.da_network
                     .inject_consensus_info(ConsensusIntentEvent::CancelPollForTransactions(*view))
                     .await;
 
-                let payload_commitment = payload.commit();
+                let payload_commitment = vid_commitment(encoded_transactions.clone());
```
jbearer, line 378: Maybe we will fix this in the other VID-related tasks(?) but I don't think we should be computing the VID commitment in the DA leader task at all. It's unnecessary, because the leader is already going to compute the commitment in the VID task. So maybe it should be the VID task, not the DA task, that sends `SendPayloadCommitmentAndMetadata`

crates/task-impls/src/transactions.rs
```diff
@@ -233,15 +233,13 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
-                    .disperse(encoded_txns.into_iter().collect::<Vec<u8>>())
-                    .unwrap();
+                let vid_disperse = vid.disperse(encoded_transactions.clone()).unwrap();
 
                 // TODO never clone a block
                 // https://github.com/EspressoSystems/HotShot/issues/1858
                 self.event_stream
                     .publish(HotShotEvent::BlockReady(
-                        payload.clone(),
+                        encoded_transactions.into_iter().collect::<Vec<u8>>(),
```
jbearer, line 242: It looks like above on line 224 we already collected `encoded_transactions` into a `Vec<u8>`. Unless I'm missing something in between this `into_iter().collect()` looks like a noop.

crates/types/src/consensus.rs
```diff
@@ -372,28 +373,32 @@ impl<PAYLOAD: BlockPayload> BlockPayloadStore<PAYLOAD> {
...
     /// multiple owning references to the payload commitment. [`remove`](Self::remove) must be
     /// called once for each reference before the payload commitment will be deallocated.
-    pub fn insert(&mut self, payload: PAYLOAD) {
+    ///
+    /// # Errors
+    /// If the transaction length conversion fails.
+    pub fn insert(&mut self, payload: PAYLOAD) -> Result<(), PAYLOAD::Error> {
         self.0
-            .entry(payload.commit())
+            .entry(vid_commitment(payload.clone().encode()?.collect()))
```
jbearer, line 381: I don't think we should compute the VID commitment here. This is called from the DA task where we already have it. So just add it as a parameter to `insert`

jbearer: In fact, maybe we shouldn't store `PAYLOAD` here at all. Maybe it should just be `VidCommitment -> Vec<u8>`. Because `PAYLOAD` is derived from the `Vec<u8>` _plus_ the metadata. But the DA task isn't really promising anything about the metadata (I don't think it signs the metadata when it votes, does it?) The DA task is really promising "I will make available a certain sequence of bytes with this VID commitment". So maybe this whole mapping should deal in bytes, and we reconstruct the `PAYLOAD` from the bytes and the metadata (which comes from the block header) when we emit the decide event, or possibly when the query service consumes the decide event.

WDYT @shenkeyao ?

jbearer: I think maybe if we did that, it would clean up error handling a bit too, since we don't have to constantly go back and forth between bytes and `PAYLOAD` -- everything is bytes, right up until the very last step.

crates/types/src/traits/block_contents.rs
```diff
@@ -91,6 +79,20 @@ pub trait BlockPayload:
...
+/// If the VID computation fails.
+#[must_use]
+pub fn vid_commitment(encoded_transactions: Vec<u8>) -> <VidScheme as VidSchemeTrait>::Commit {
+    // TODO <https://github.com/EspressoSystems/HotShot/issues/1686>
+    let srs = test_srs(NUM_STORAGE_NODES);
+    // TODO We are using constant numbers for now, but they will change as the quorum size
+    // changes.
+    // TODO <https://github.com/EspressoSystems/HotShot/issues/1693>
+    let vid = VidScheme::new(NUM_CHUNKS, NUM_STORAGE_NODES, srs).unwrap();
+    vid.disperse(encoded_transactions).unwrap().commit
```
jbearer, line 93: Use `commit_only`, it's faster

jbearer (changes requested):
Looking good, few suggestions and questions but nothing major

crates/task-impls/src/transactions.rs
shenkeyao: I think you are right! I thought lint would catch things like this. 🤔

crates/task-impls/src/da.rs
shenkeyao: We have an in-progress PR by @rob-maron to split the `BlockReady` event but `SendPayloadCommitmentAndMetadata` isn't in the plan. I think this is a good idea to move it to the VID task! I'll do it in this PR then.

crates/types/src/consensus.rs
shenkeyao: > I don't think we should compute the VID commitment here. This is called from the DA task where we already have it.

Just wanted to note that besides the DA task, we also call this function with the payload instead of the VID commitment in `crates/hotshot/src/lib.rs`. But I agree we change the input from the payload to VID commitment, and in places where we have the payload, we can just compute its commitment.

> Maybe it should just be VidCommitment -> Vec<u8>.

Agreed!

ggutoski: I have trouble following everything. Is this payload store maintained only by DA committee nodes, or everyone? In any case, Jeb's suggestion to store only the bytes instead of the payload struct makes sense to me.

Perhaps it's confusing because you cannot get the bytes without first creating a payload. Yet it sounds like all we really want is the bytes. So the flow becomes:
1. build payload from txs
2. encode to bytes, forget about the payload, use bytes everywhere until the end.
3. re-construct the payload at the end.

If that's the case then perhaps we should rethink the payload API?

Finally: as a general rule to everyone: please do not clone payloads, do not call `encode()?.collect()` more than once, do not compute the commitment more than once. If you find the need to violate this constraint then there's something wrong with our design and we need to fix it.

crates/task-impls/src/da.rs
```diff
@@ -196,11 +195,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
                     return None;
                 }
 
-                debug!(
-                    "Got a DA block with {} transactions!",
-                    proposal.data.block_payload.transaction_commitments().len()
-                );
-                let payload_commitment = proposal.data.block_payload.commit();
+                let payload_commitment = vid_commitment(proposal.data.encoded_transactions.clone());
```
ggutoski, line 198: Re-computing VID commit here. Is this a DA committee member preparing to vote? If so then it's okay, otherwise we should not be computing it here.

Perhaps this computation should move down to line 219 so it's the last thing you do after all error checking and before signing the DA vote.

Why is the payload_commitment needed in line 207 for the call to `validate`? DA committee members do not check the VID commitment against anything---they compute it for themselves and sign it.

crates/task-impls/src/da.rs
```diff
@@ -244,15 +239,19 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
-                    },
+                    view_inner: ViewInner::DA { payload_commitment },
                 });
 
                 // Record the block payload we have promised to make available.
-                consensus
+                if let Err(e) = consensus
                     .saved_block_payloads
-                    .insert(proposal.data.block_payload);
+                    .insert(BlockPayload::from_bytes(
```
ggutoski, line 248: This is what Jeb mentioned in another comment. We are converting bytes to payload, then we convert payload to bytes at the other side. Maybe instead just stay with bytes and save all the conversion.

crates/task-impls/src/transactions.rs
jbearer: Yeah me too...weird

crates/types/src/consensus.rs
jbearer: This store is maintained by everyone but I think after genesis it's only populated when you're on the DA committee. It's only temporary, so that nodes who are on DA can include the entire payload in the `Decide` events, whereafter they become the responsibility of the consumer (e.g. query service) to persist. So entries are created when a DA member signs a block and deleted after that block is finalized, or after it gets forked out of the chain.

Regarding the workflow, it is true that on the builder side, we construct the block, encode it, and then immediately forget the structured representation. On the other end, we do want the structured payload: HotShot will construct it right before emitting a decide event, use the `Blockpayload` interface to update its mempool statistics, and then send it off to the query service, where we definitely want it structured.

Maybe we can come up with a better API for the builder, like `encode_txns(Vec<Transaction>) -> Vec<u8>`, skipping the intermediate step, or `into_bytes(self) -> Vec<u8>` (keeping the intermediate step, but moving out of `self` to ensure there is no unnecessary cloning). But I don't think this is a high priority for right now, since there is actually not very much cost to `encode()` once you have a `BlockPayload` (and once we fix the API to return `Cow` or have Jellyfish take an iterator)

crates/task-impls/src/da.rs
shenkeyao: > Is this a DA committee member preparing to vote?

Yes!

> Why is the `payload_commitment` needed in line 207 for the call to validate?

I think the VID commitment is needed as the signed data to be used to verify the signature, to make sure the DA proposal is from the expected leader. So it's not really the commitment that's being checked, but the signature of the proposal.

ggutoski: > I think the VID commitment is needed as the signed data to be used to verify the signature, to make sure the DA proposal is from the expected leader. So it's not really the commitment that's being checked, but the signature of the proposal.

Ok I think I finally get it. We recently discussed that the DA leader no longer sends the commitment to the DA committee. However, the DA leader's *signature* is on the VID commitment, even though he doesn't send it. Instead each committee member re-computes the VID commitment and then checks the DA leader's signature as a signature of that re-computed commitment.

jbearer: Hmm. I feel like the signature check should come first (ie signature is on data, not commitment). Otherwise anybody can force a DA member to do the expensive VID commitment before they realize the message was not sent by the leader. This is probably something we can play around with later though, no need to change it now

ggutoski: If you're signing the raw data then basically you have to hash the entire payload. So you're effectively computing two different commitments of the same payload one fast one slow.

jbearer: Yep. But one is _really_ fast...

ggutoski: True. Alright I'm convinced. I'm also convinced there's no need to change it now.

ggutoski: New issue #2106 . Let's make an effort to post an issue whenever we discuss something like this. I feel like these insights get forgotten in the bowels of github PR convos.

crates/hotshot/src/lib.rs
```diff
@@ -218,12 +221,19 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> SystemContext<TYPES, I> {
         );
 
         let mut saved_leaves = HashMap::new();
-        let mut saved_block_payloads = BlockPayloadStore::default();
+        let mut saved_payload_commitments = PayloadCommitmentStore::default();
```
jbearer, line 224: This naming seems off to me. We're not saving payload commitments. We're saving payloads, _indexed by_ their commitments. I think just `PayloadStore`/`saved_payloads` would be better

jbearer (dismissed):
Looks good. Just one suggestion on naming

crates/hotshot/src/lib.rs
shenkeyao: Oh, we are actually saving the encoded transaction bytes instead of the payload, since it's a commitment -> bytes mapping. I'll rename the variable and struct.

shenkeyao: On second thought, `PayloadStore` and `saved_payloads`are indeed better. We are mapping the payload commitment to its encoded format, so it's basically a store for payloads.

jbearer (approved):

Issue #2067, opened 2023-11-15
Vote logic for VID cert


ggutoski:

dailinsubjam:
Some notes from pair programming meeting:

- currently we’re doing VID computations but not using them in consensus
- NOT in scope for this issue: deliver different shares to different storage nodes
- Each replica node does not vote on a proposal until it has seen it’s VID share
    - QC = VIDCert
- Handle VidShareRecv in the VID task
- Handle VIDCertRecv in the consensus task
- Get rid of VIDVote events and VIDCert events
- Tasks for this issue:
    - Handle VIDShareRecv event in the VID task
        - Store that you have gotten a VID share for a specific block
        - clarify: for this issue VIDShareRecv contains ALL shares. The task to deliver one share to each node is a separate issue
    - Update vote logic so replicas don’t vote until they’ve seen a VIDShare

dailinsubjam:
- When a node receives a `QuorumProposal` it does not vote until all of the following conditions are met
    - It has seen a DAC, `vote_if_able()` will decide.
    - It has seen its VID share (not VIDCert) --> This part this issue will need to implement. Will add to `vote_if_able()`.

Issue #1732, opened 2023-09-13
T4: Different VID shares to each node instead of all shares to all nodes


ggutoski:
- [x] #2104 
- [x] #1696

Issue #1583, opened 2023-08-18
[parent] Integrate VID


ggutoski:
- [x] Resolve tech debt to unlock progress:
    - [x] Add trait bounds in jellyfish: https://github.com/EspressoSystems/jellyfish/pull/362
    - [x] Eliminate hotshot-primitives dep from hotshot. https://github.com/EspressoSystems/HotShot/issues/1570
    - [x] Eliminate hotshot-compat branch of jellyfish https://github.com/EspressoSystems/HotShot/issues/1628
- [x] #1662 
- [x] #1730 
- [x] #1799 
- [x] #2067 
- [x] #1817 
- [ ] #1732 
- [ ] #2068 
- [ ] #2069

PR #2107, opened 2023-11-22
[VID] Vid Vote Logic in Consensus Task

dailinsubjam:
Closes issue: https://github.com/EspressoSystems/HotShot/issues/2067

### This PR: 
- Update vote logic by updating `vote_if_able()` so that replicas don’t vote until they’ve seen a VIDShare(event `VIDDisperseRecv` in implementation).
- Get rid of VIDVote, VIDCert, VIDVoteSend/Recv, VIDCertSend/Recv, VIDCertificate, VIDVoteCollection.
- Update `test_vid_task`.

### This PR does not: 
- Deliver different shares to different storage nodes, for this issue VIDDisperseRecv contains ALL shares.

### Key places to review:

- `task-impls/src/consensus.rs`: added `vid_shares` to save received vid shares for each view, added one more condition in `vote_if_albe()`, dealt with `HotShotEvent::VidDisperseRecv`.

 ### How to test this PR: 
`just async_std test_vid_task`
`just async_std test_success`

### Comments requested:
- Currently both consensus task and vid task will deal with `HotShotEvent::VidDisperseRecv`, and there is a high overlapping of codes in validating the received shares. We have to keep the one in consensus task to determine when can vote, but do we still want to keep the one in vid task?


crates/task-impls/src/consensus.rs
```diff
@@ -365,6 +365,16 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
                 }
             }
 
+            // Only vote if you has seen the VID share for this view
+            if let Some(_vid_share) = self.vid_shares.get(&proposal.view_number) {
+            } else {
+                error!(
```
elliedavidson, line 371: Can we reduce this to a `debug` level message?

crates/task-impls/src/consensus.rs
```diff
@@ -1100,20 +1110,63 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
+                // Allow VID disperse date that is one view older, in case we have updated the
+                // view.
+                // Adding `+ 1` on the LHS rather than `- 1` on the RHS, to avoid the overflow
+                // error due to subtracting the genesis view number.
+                if view + 1 < self.cur_view {
+                    warn!("Throwing away VID disperse data that is more than one view older");
+                    return;
+                }
+
+                debug!("VID disperse data is fresh.");
```
elliedavidson, line 1137: @ggutoski what does this message mean? :) Maybe we can update it to something more understandable?

crates/task-impls/src/consensus.rs
```diff
@@ -1100,20 +1110,63 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
+                );
 
-                // RM TODO: VOTING
-            }
+                // stop polling for the received disperse
+                self.quorum_network
+                    .inject_consensus_info(ConsensusIntentEvent::CancelPollForVIDDisperse(
+                        *disperse.data.view_number,
+                    ))
+                    .await;
```
elliedavidson, line 1126: I think we should only stop polling for VIDDisperse after we've verified that this is a valid share.  So I think we should move this block of code further down in the function.

crates/task-impls/src/consensus.rs
```diff
@@ -1100,20 +1110,63 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
+                if view + 1 < self.cur_view {
+                    warn!("Throwing away VID disperse data that is more than one view older");
+                    return;
+                }
+
+                debug!("VID disperse data is fresh.");
+                let payload_commitment = disperse.data.payload_commitment;
 
+                // Check whether the sender is the right leader for this view
+                let view_leader_key = self.committee_membership.get_leader(view);
```
elliedavidson, line 1141: Technically either `committee_membership` or `quorum_membership` work here, but to be consistent, I think we should always use quorum membership unless we explicitly need to use committee membership.  In a future, low-priority issue we should standardize this somehow - if the leaders are always the same between memberships, we should encode that.  But that is for a later issue.

crates/task-impls/src/consensus.rs
```diff
@@ -1100,20 +1110,63 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
+                if view_leader_key != sender {
+                    error!(
```
elliedavidson, line 1143: I think this can be a warn level

crates/task-impls/src/consensus.rs
```diff
@@ -1100,20 +1110,63 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
+                        "VID dispersal/share is not from expected leader key for view {} \n",
+                        *view
+                    );
+                    return;
+                }
+
+                if !view_leader_key.validate(&disperse.signature, payload_commitment.as_ref()) {
+                    error!("Could not verify VID dispersal/share sig.");
```
elliedavidson, line 1151: This should also be a warn level message.

crates/task-impls/src/network.rs
```diff
@@ -98,18 +98,11 @@ impl<TYPES: NodeType> NetworkMessageTaskState<TYPES> {
                                 HotShotEvent::DAVoteRecv(vote.clone())
                             }
                             CommitteeConsensusMessage::DACertificate(cert) => {
-                                // panic!("Recevid DA C! ");
```
elliedavidson, line 101: Thanks for cleaning this up, even though it was out of scope for your PR!

crates/task-impls/src/vid.rs
```diff
@@ -179,88 +80,10 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
-                        VIDVoteCollectionTypes::build(builder).launch().await
-                    });
-                    self.vote_collector = Some((view, id, stream_id));
-                } else if let Some((_, _, stream_id)) = self.vote_collector {
-                    self.event_stream
-                        .direct_message(stream_id, HotShotEvent::VidVoteRecv(vote))
-                        .await;
-                };
-            }
             HotShotEvent::VidDisperseRecv(disperse, sender) => {
```
elliedavidson, line 83: Now that we have this event handler in the consensus task, we can remove it entirely from the vid task.

crates/testing/src/overall_safety_task.rs
```diff
@@ -366,7 +366,7 @@ impl Default for OverallSafetyPropertiesDescription {
             check_leaf: false,
             check_state: true,
             check_block: true,
-            num_failed_views: 0,
+            num_failed_views: 5,
```
elliedavidson, line 369: Why do we increase this to 5?  I feel like it _should_ be 0, and if tests fail then we have some error in our logic.

crates/web_server/api.toml
```diff
@@ -52,13 +52,6 @@ DOC = """
 Return the DA certificate for a given view number
 """
 
-# GET the VID certificate for a view, where the view is passed as an argument
-[route.getvidcertificate]
-PATH = ["vid_certificate/:view_number"]
-":view_number" = "Integer"
-DOC = """
-Return the VID certificate for a given view number
-"""
```
elliedavidson, line 61: Thanks for remembering to remove these!

crates/task-impls/src/consensus.rs
ggutoski: Yep. We have definite future plans to make committee membership different from quorum. We should never use committee membership for anything other than the DA committee.

crates/task-impls/src/consensus.rs
```diff
@@ -1100,20 +1110,63 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
+                    return;
+                }
+
+                // Add to the storage that we have received the VID disperse for a specific view
+                self.vid_shares.insert(view, disperse.clone());
```
ggutoski, line 1156: should not need to clone here. I'm surprised clippy did not complain.

Should we remove `Clone` from things like block payload and `Disperse`? That would make it impossible to accidentally copy a full block payload.

crates/task-impls/src/consensus.rs
dailinsubjam: Changed to `info` level.

crates/task-impls/src/consensus.rs
dailinsubjam: Changed to `warn` level.

crates/testing/src/overall_safety_task.rs
dailinsubjam: Good catch! Moved back to 0. The reason is that before polling strict testing parameters from `main` to `develop`, this number is set to 5. And when running this PR I accidentally changed some numbers on this branch to see whether switching to the old one could make the test results better. Agree that it should always be 0.

crates/task-impls/src/consensus.rs
dailinsubjam: Changed to `info!("VID disperse data is s not more than one view older.");`.

crates/task-impls/src/consensus.rs
dailinsubjam: Surprised too. Removed this clone(). I also checked other `Clone`s for things like block payload and `Disperse` and find out they are all necessary :(

shenkeyao (commented):
Looks good! Just a question about the endpoint.

crates/task-impls/src/consensus.rs
```diff
@@ -117,8 +117,8 @@ pub struct ConsensusTaskState<
     /// All the DA certs we've received for current and future views.
     pub da_certs: HashMap<TYPES::Time, DACertificate<TYPES>>,
 
-    /// All the VID certs we've received for current and future views.
-    pub vid_certs: HashMap<TYPES::Time, VIDCertificate<TYPES>>,
+    /// All the VID shares we've received for current and future views.
+    pub vid_shares: HashMap<TYPES::Time, Proposal<TYPES, VidDisperse<TYPES>>>,
```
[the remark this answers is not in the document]
dailinsubjam: Created one here: https://github.com/EspressoSystems/HotShot/issues/2146. Do we want to add it to this parent issue https://github.com/EspressoSystems/HotShot/issues/1583?

crates/task-impls/src/consensus.rs
```diff
@@ -295,7 +295,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
     }
 
     #[instrument(skip_all, fields(id = self.id, view = *self.cur_view), name = "Consensus vote if able", level = "error")]
-
+    // Check if we are able to vote, like whether the proposal is valid, whether we have DAC and VID share, and if so, vote
```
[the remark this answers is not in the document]
dailinsubjam: Agree it's too long. Let me see whether there's any easy cleanup.
- Line 356 is an action for genesis view, it will not check VID share nor DA cert.
- Those two same lines, one is for creating vote for genesis view, which will `return` after the vote, therefore we need the other one for the normal cases.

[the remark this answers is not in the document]
dailinsubjam: Not that easy to turn those common parts into a function :( Maybe for a future issue. I added more comments for the clarification of this function.

crates/web_server/api.toml
```diff
@@ -52,13 +52,6 @@ DOC = """
 Return the DA certificate for a given view number
 """
 
-# GET the VID certificate for a view, where the view is passed as an argument
-[route.getvidcertificate]
```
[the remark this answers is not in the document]
dailinsubjam: Yes! And we do have one at line 15.

crates/task-impls/src/consensus.rs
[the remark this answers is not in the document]
ggutoski: Thank you for indulging me. Seems like a silly request now. (We also have a sub-parent issue here https://github.com/EspressoSystems/HotShot/issues/1732 ) I guess all I really wanted was to plant a flag for the future that this line is one point in the code that needs to be updated when we switch to one-share-per-node.

dailinsubjam:
> Could we update the consensus_task unit tests to also test for receiving VID shares? I don't think they test that now, and after this change we'll want them to test that.

@elliedavidson  Good point! I know `test_success` also test this, but `consensus_task` doesn't. I took a look and found that test input & output are also relevant to the event `TransactionsSequenced` and `BlockReady`, I have to take some time to see how to sequence them and how they change the view number. How about we create an issue for this? We can also discuss this in the next meeting.

elliedavidson:
> > Could we update the consensus_task unit tests to also test for receiving VID shares? I don't think they test that now, and after this change we'll want them to test that.
> 
> @elliedavidson Good point! I know `test_success` also test this, but `consensus_task` doesn't. I took a look and found that test input & output are also relevant to the event `TransactionsSequenced` and `BlockReady`, I have to take some time to see how to sequence them and how they change the view number. How about we create an issue for this? We can also discuss this in the next meeting.

Yes, sounds good!  This can be done in a separate issue.  Let's get this merged first.

elliedavidson (dismissed):

elliedavidson:
Let's try updating the view timeout for the half_f failures timeout like we did for the merge yesterday.

dailinsubjam:
> Could we update the consensus_task unit tests to also test for receiving VID shares? I don't think they test that now, and after this change we'll want them to test that.

Created the issue here: https://github.com/EspressoSystems/HotShot/issues/2148

elliedavidson (approved):

Issue #2040, opened 2023-11-10
Simplify Vote Logic


bfish713:
We duplicate a lot of the vote collection, checking and certificate event creation logic.  We should consolidate this code into one place and use it everywhere.

One Idea is to create a generic vote accumulator task and a function to spawn it.

Also we should create functions to encapsulate the vote logic

PR #2122, opened 2023-11-28
Refactor Vote accumulator tasks to one General implementation

bfish713:
Closes issue: <ISSUE>

### This PR: 

- Creates a generic vote accumulator task
- Creates a function to which generically spawns the task (`spawn_vote_accumulator`)
- Uses traits (`AggregatableVote` and `HandleVoteEvent`) to map input and output events to different vote types
    - `AggregatableVote` - defines how we get the right leader for that vote and maps the certificate formed to the appropriate `HotShotEvent`
    - `HandleVoteEvent` - defines how we map from `HotShotEvent` to the specific Vote type, and defines the event filter
- Replaces the duplicated vote accumulator task and task creation logic in all of the task_impls

### This PR does not: 
Replace any other potential duplicated logic in the View sync task for spawning Replicas

### Key places to review: 
The new code in crates/task_impls/src/vote.rs is the main place to look.  If anyone has ideas how to make that file have less boilerplate that would be great.  Fortunately the boilerplate in that file abstracts away pretty much completely in the other task impls

Closes: [#2040](https://github.com/EspressoSystems/HotShot/issues/2040)

### DIFF
--- a/crates/task-impls/src/consensus.rs
+++ b/crates/task-impls/src/consensus.rs
@@ -780,7 +780,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> + 
                         id: self.id,
                         registry: self.registry.clone(),
                     };
-                    let name = "Quorum Vote Collection";
+                    let name = "Timeout Vote Collection";
                     self.vote_collector =
                         spawn_vote_accumulator(&info, vote.clone(), event, name.to_string()).await;
                 } else if let Some((_, _, stream_id)) = self.vote_collector {
--- a/crates/task-impls/src/consensus.rs
+++ b/crates/task-impls/src/consensus.rs
@@ -723,6 +723,20 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> + 
             }
             HotShotEvent::QuorumVoteRecv(ref vote) => {
                 debug!("Received quroum vote: {:?}", vote.get_view_number());
+                if self
+                    .quorum_membership
+                    .get_leader(vote.get_view_number() + 1)
+                    != self.public_key
+                {
+                    error!(
+                        "We are not the leader for view {} are we the leader for view + 1? {}",
+                        *vote.get_view_number() + 1,
+                        self.quorum_membership
+                            .get_leader(vote.get_view_number() + 2)
+                            == self.public_key
+                    );
+                    return;
+                }
 
                 let collection_view =
                     if let Some((collection_view, collection_task, _)) = &self.vote_collector {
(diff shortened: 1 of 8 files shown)


crates/task-impls/src/consensus.rs
```diff
@@ -894,86 +736,31 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
-                        .direct_message(stream_id, HotShotEvent::QuorumVoteRecv(vote))
+                        .direct_message(stream_id, HotShotEvent::QuorumVoteRecv(vote.clone()))
                         .await;
                 }
             }
-            HotShotEvent::TimeoutVoteRecv(vote) => {
-                if self
-                    .timeout_membership
-                    .get_leader(vote.get_view_number() + 1)
-                    != self.public_key
```
shenkeyao, line 962: Is the reason that we no longer check the leadership that it's done in `spawn_vote_accumulator` in the vote task? I think this is indeed cleaner but may cause unnecessary operations since `spawn_vote_accumulator` is called at the end of this event handling after other checks. Would it be better if we keep the leadership check in individual tasks?

crates/task-impls/src/da.rs
```diff
@@ -247,23 +164,18 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
-            HotShotEvent::DAVoteRecv(vote) => {
-                debug!("DA vote recv, Main Task {:?}", vote.get_view_number(),);
+            HotShotEvent::DAVoteRecv(ref vote) => {
+                debug!("DA vote recv, Main Task {:?}", vote.get_view_number());
                 // Check if we are the leader and the vote is from the sender.
                 let view = vote.get_view_number();
                 if self.da_membership.get_leader(view) != self.public_key {
                     error!("We are not the committee leader for view {} are we leader for next view? {}", *view, self.da_membership.get_leader(view + 1) == self.public_key);
                     return None;
                 }
```
shenkeyao, line 174: If we are going to remove the leadership check from each task (not sure if we should though, see my other comment), should we also move it here?

shenkeyao (commented):
Some questions about where to check the leadership. Everything else looks good.

crates/task-impls/src/vote.rs
```diff
@@ -0,0 +1,514 @@
...
+        }
+        let Some(accumulator) = self.accumulator else {
+            return (None, self);
+        };
+        match accumulator.accumulate(vote, &self.membership) {
+            Either::Left(acc) => {
+                self.accumulator = Some(acc);
+                (None, self)
+            }
+            Either::Right(qc) => {
```
elliedavidson, line 102: nit: I feel like we should use `cert` here instead of `qc` since we aren't always actually forming a quorum certificate.

crates/task-impls/src/vote.rs
```diff
@@ -0,0 +1,514 @@
...
+/// Info needed to create a vote accumulator task
+#[allow(missing_docs)]
+pub struct AccumulatorInfo<TYPES: NodeType> {
+    pub public_key: TYPES::SignatureKey,
+    pub membership: Arc<TYPES::Membership>,
+    pub view: TYPES::Time,
+    pub event_stream: ChannelStream<HotShotEvent<TYPES>>,
+    pub id: u64,
+    pub registry: GlobalRegistry,
+}
```
elliedavidson, line 168: nit: Can we add docs here and remove the clippy ignore?

crates/task-impls/src/consensus.rs
bfish713: I agree it's better to keep it in the individual tasks and remove it from the spawn function

elliedavidson (dismissed):

shenkeyao (approved):

shenkeyao (approved):

Issue #2119, opened 2023-11-28
[light-client-integration] adding/generating schnorr keys


mrain:
Include a schnorr key in the hotshot config.
Also generate those keys during initialization.

PR #2123, opened 2023-11-28
[light-client-integration] add state signature keys to the config

mrain:
Closes issue: #2119 

### This PR: 

Defines the types for Schnorr signatures of the light client state.

### This PR does not: 
Contains any logic related to signature generations, etc.
Since validators need not verify these Schnorr signatures, didn't add the verification keys to the config file. 

### Key places to review: 
* Types and structs related to schnorr signatures are moved to `crates/types/src/light_client_state.rs`
* Add a `state_sig_key_pair` field inside the `ValidatorConfig` struct, see line 51-52 `crates/types/src/lib.rs`
* Are these modification enough so that the validators will be able to sign a state change?

### Comments requested
* I didn't wrap these key types, nor define any related traits. Because they are pretty isolated and are only used in light client related logics.
* For stake table, I renamed some bls keys to qc keys, and Schnorr keys to state keys. Which ones do you think is better? Does it confuse people?

### DIFF
--- a/crates/hotshot/examples/infra/mod.rs
+++ b/crates/hotshot/examples/infra/mod.rs
@@ -884,16 +884,12 @@ pub async fn main_entry_point<
 
     run_config.node_index = node_index.into();
 
-    let (public_key, private_key) =
-        <<TYPES as NodeType>::SignatureKey as SignatureKey>::generated_from_seed_indexed(
+    run_config.config.my_own_validator_config =
+        ValidatorConfig::<<TYPES as NodeType>::SignatureKey>::generated_from_seed_indexed(
             run_config.seed,
             node_index.into(),
+            1,
         );
-    run_config.config.my_own_validator_config = ValidatorConfig {
-        public_key,
-        private_key,
-        stake_value: 1,
-    };
     //run_config.libp2p_config.as_mut().unwrap().public_ip = args.public_ip.unwrap();
 
     error!("Initializing networking");
--- a/crates/orchestrator/src/config.rs
+++ b/crates/orchestrator/src/config.rs
@@ -292,13 +292,7 @@ fn default_transaction_size() -> usize {
 impl<K: SignatureKey> From<ValidatorConfigFile> for ValidatorConfig<K> {
     fn from(val: ValidatorConfigFile) -> Self {
         // here stake_value is set to 1, since we don't input stake_value from ValidatorConfigFile for now
-        let validator_config =
-            ValidatorConfig::generated_from_seed_indexed(val.seed, val.node_id, 1);
-        ValidatorConfig {
-            public_key: validator_config.public_key,
-            private_key: validator_config.private_key,
-            stake_value: validator_config.stake_value,
-        }
+        ValidatorConfig::generated_from_seed_indexed(val.seed, val.node_id, 1)
     }
 }
 impl<KEY: SignatureKey, E: ElectionConfig> From<ValidatorConfigFile> for HotShotConfig<KEY, E> {
--- a/crates/types/src/lib.rs
+++ b/crates/types/src/lib.rs
@@ -10,6 +10,7 @@
 #![allow(clippy::module_name_repetitions)]
 
 use displaydoc::Display;
+use light_client_state::{generat_key_pair_from_seed_indexed, StateSigKeyPairs};
 use std::{num::NonZeroUsize, time::Duration};
 use traits::{election::ElectionConfig, signature_key::SignatureKey};
 pub mod block_impl;
@@ -17,6 +18,7 @@ pub mod consensus;
 pub mod data;
 pub mod error;
 pub mod event;
+pub mod light_client_state;
 pub mod message;
 pub mod simple_certificate;
 pub mod simple_vote;
@@ -46,6 +48,8 @@ pub struct ValidatorConfig<KEY: SignatureKey> {
     pub private_key: KEY::PrivateKey,
     /// The validator's stake
     pub stake_value: u64,
+    /// the validator's key pairs for state signing/verification
+    pub state_key_pairs: StateSigKeyPairs,
 }
 
 impl<KEY: SignatureKey> ValidatorConfig<KEY> {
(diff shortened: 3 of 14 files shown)


crates/orchestrator/src/config.rs
```diff
@@ -292,13 +292,7 @@ fn default_transaction_size() -> usize {
 impl<K: SignatureKey> From<ValidatorConfigFile> for ValidatorConfig<K> {
     fn from(val: ValidatorConfigFile) -> Self {
         // here stake_value is set to 1, since we don't input stake_value from ValidatorConfigFile for now
-        let validator_config =
-            ValidatorConfig::generated_from_seed_indexed(val.seed, val.node_id, 1);
-        ValidatorConfig {
-            public_key: validator_config.public_key,
-            private_key: validator_config.private_key,
-            stake_value: validator_config.stake_value,
-        }
```
dailinsubjam, line 301: The removal is nice!

crates/types/src/lib.rs
```diff
@@ -46,17 +48,21 @@ pub struct ValidatorConfig<KEY: SignatureKey> {
...
+    /// the validator's key pairs for state signing/verification
+    pub state_key_pairs: StateSigKeyPairs,
 }
 
 impl<KEY: SignatureKey> ValidatorConfig<KEY> {
     /// generate validator config from input seed, index and stake value
     #[must_use]
     pub fn generated_from_seed_indexed(seed: [u8; 32], index: u64, stake_value: u64) -> Self {
         let (public_key, private_key) = KEY::generated_from_seed_indexed(seed, index);
+        let state_key_pairs = generat_key_pair_from_seed_indexed(seed, index);
```
dailinsubjam, line 60: Are we using the same signature key scheme for `state_key_pairs`? I'm a little bit confused on these three generation functions. I know  `generated_from_seed_indexed` is for key pair generation inside `hotshot`, and `KEY::generated_from_seed_indexed` is calling key pair generation with given seed inside `jellyfish`, what's the third one `generat_key_pair_from_seed_indexed` used for? Do we want to give it a different naming?

dailinsubjam (approved):
LGTM! One concern I have regarding the renaming of `bls keys` and `Schnorr keys` is, if we want to test a different signature scheme for qc keys or state keys in the future, and still want to keep the old ones, we may want to keep markers like `bls` or `schnorr` in the name, but if these testing situation will not likely to happen I'm fine with the renaming :)

crates/types/src/lib.rs
mrain: Here I'm just copying things from `hotshot-signature-keys`. In Jellyfish, it's always taking some `rng` as an input. This function just wraps around generating `rng` from some seed.

dailinsubjam: Oh sorry I forgot the previous two are all for generation in `hotshot`, but one for ValidatorConfig and the other for general usage by calling the one inside jellyfish. I didn't find this function: `generat_key_pair_from_seed_indexed()`, just wondering is it the same as calling `KEY::generated_from_seed_indexed()`?

mrain: Just noticed it's a typo here.
https://github.com/EspressoSystems/HotShot/pull/2123/files#diff-65484181673c3c8c0a2c4b52a749c666a9028434c97f571ccd5b34d758eb2788R62-R74

dailinsubjam: That makes sense, thanks!

crates/types/src/light_client_state.rs
```diff
@@ -0,0 +1,74 @@
...
+
+/// Generate key pairs from seed
+#[must_use]
+pub fn generate_state_key_pair_from_seed(seed: [u8; 32]) -> StateSigKeyPairs {
+    StateSigKeyPairs::generate(&mut ChaCha20Rng::from_seed(seed))
+}
+
+/// Generate key pairs from an index and a seed
+#[must_use]
+pub fn generate_state_key_pair_from_seed_indexed(seed: [u8; 32], index: u64) -> StateSigKeyPairs {
```
alxiong, line 68: An alternative is using `impl Deref` for newtypes (which is somewhat [controversial](https://stackoverflow.com/questions/45086595/is-it-considered-a-bad-practice-to-implement-deref-for-newtypes) as a practice), considering that we are fine with people directly manipulate and access all inner field's methods, we could use it.

i.e. 
```rust
pub struct LightClientStateVerKey(schnorr::VerKey<Config>);
impl LightClientStateVerKey {
  fn generate_from_seed() -> Self;
  fn generate_from_seed_indexed() -> Self;
}

impl Deref for LightClientStateVerKey {
  type Target = schnorr::VerKey<Config>;
  fn deref(&self) -> &Self::Target {
     &self.0
  }
}
```

I think in this way, you can use `state_ver_key::schnorr_original_methods()`.

crates/hotshot/examples/infra/mod.rs
```diff
@@ -884,16 +884,12 @@ pub async fn main_entry_point<
 
     run_config.node_index = node_index.into();
 
-    let (public_key, private_key) =
-        <<TYPES as NodeType>::SignatureKey as SignatureKey>::generated_from_seed_indexed(
+    run_config.config.my_own_validator_config =
+        ValidatorConfig::<<TYPES as NodeType>::SignatureKey>::generated_from_seed_indexed(
             run_config.seed,
             node_index.into(),
+            1,
```
alxiong, line 891: nit: `stake_value: 1` would be clearer

crates/types/src/light_client_state.rs
mrain: Good idea. Did it with `StateKeyPair`in 6cfee79

crates/hotshot/examples/infra/mod.rs
mrain: This is a function call, cannot do this.

alxiong (approved):

crates/types/src/lib.rs
```diff
@@ -46,17 +47,21 @@ pub struct ValidatorConfig<KEY: SignatureKey> {
...
+    /// the validator's key pairs for state signing/verification
+    pub state_key_pair: light_client::StateKeyPair,
 }
 
 impl<KEY: SignatureKey> ValidatorConfig<KEY> {
     /// generate validator config from input seed, index and stake value
     #[must_use]
     pub fn generated_from_seed_indexed(seed: [u8; 32], index: u64, stake_value: u64) -> Self {
         let (public_key, private_key) = KEY::generated_from_seed_indexed(seed, index);
+        let state_key_pairs = light_client::StateKeyPair::generate_from_seed_indexed(seed, index);
```
dailinsubjam, line 59: Love this!

dailinsubjam (approved):

Issue #2094, opened 2023-11-20
[Stability] Rework the Spinning Task to be a function of view


rob-maron:
* What is the reason for this task? 

Right now, the spinning (spinning up/down) task works as a function of time. This can create some nondeterminism where depending on what hardware/codebase we run on, all views may finish before any of these "spins" finish.

It makes more sense for the spinning task to work as a function of the view, so that it is independent of hardware and software execution context.

### Acceptance criteria
* What work will need to be done to complete this task?  If this is unknown, describe what information is needed to determine the work needed

- [x] Figure out how to inject the view number into the spinning task
- [x] Use the injected view number to spin up/down nodes and networks
- [x] Update all tests to use this constraint instead

All tests are applicable to be changed.

PR #2130, opened 2023-11-29
[Stability] Rework spinning task to be function of view

rob-maron:
Closes issue: #2094 

### This PR: 

- Reworks the spinning task to be a function of the latest view
- Changes the tests to work under this new description

### This PR does not: 

- Change any other functionality
- Address any test failures

### Key places to review: 

- `spinning_task.rs`: where the functionality is changed
- the tests, to make sure they are (re)configured properly

### DIFF
--- a/crates/testing/src/spinning_task.rs
+++ b/crates/testing/src/spinning_task.rs
@@ -1,40 +1,35 @@
-use std::{
-    collections::HashMap,
-    sync::{atomic::AtomicUsize, Arc},
-    time::Duration,
-};
+use either::Either;
+use hotshot_task::Merge;
+use hotshot_task_impls::events::HotShotEvent;
+use std::{collections::HashMap, sync::Arc};
 
-use crate::{test_launcher::TaskGenerator, test_runner::Node, GlobalTestEvent};
-use async_compatibility_layer::art::async_sleep;
+use async_compatibility_layer::channel::UnboundedStream;
 use futures::FutureExt;
 use hotshot::{traits::TestableNodeImplementation, HotShotType, SystemContext};
 use hotshot_task::{
-    boxed_sync,
     event_stream::ChannelStream,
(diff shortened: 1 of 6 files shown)


crates/testing/src/spinning_task.rs
```diff
@@ -86,80 +79,120 @@ impl SpinningTaskDescription {
...
+                        .get_event_stream_known_impl(FilterEvent::default())
+                        .await
+                        .0;
+                    let s2 = handle
+                        .handle
+                        .get_internal_event_stream_known_impl(FilterEvent::default())
+                        .await
+                        .0;
+                    streams.push(Merge::new(s1, s2));
+                }
```
bfish713, line 170: I'm a bit confused by this, we are merging the internal and external streams, but above it looks like we only use the external one?

crates/testing/src/spinning_task.rs
```diff
@@ -86,80 +79,120 @@ impl SpinningTaskDescription {
...
         })
     }
 }
+
+/// types for safety task
+pub type SpinningTaskTypes<TYPES, I> = HSTWithEventAndMessage<
+    SpinningTaskErr,
+    GlobalTestEvent,
+    ChannelStream<GlobalTestEvent>,
+    (usize, Either<Event<TYPES>, HotShotEvent<TYPES>>),
```
bfish713, line 195: Same as above, can't this just be `(usize,  HotShotEvent<TYPES>)`

bfish713 (changes requested):

bfish713:
Change looks good, but I had some question about why we handle messages from 2 streams instead of 1

crates/testing/src/spinning_task.rs
rob-maron: Fixed !

bfish713 (approved):

elliedavidson (approved):

Issue #2059, opened 2023-11-14
Organize non-production code


shenkeyao:
Everything in `block.impl.rs` is for examples and tests only, so it should not be put under `types/src`. However, simply moving it to the testing crate may cause dependency errors since several files outside `crates/testing` are also using structs in `block_impl.rs`.  

We should organize code for demos, examples, and tests, to avoid confusion with production code.

PR #2134, opened 2023-11-29
#2059: Organize non-prod code

QuentinI:
Closes issue: #2059

### This PR: 
Addresses #2059 by moving everything from `types/block_impl` and some code from `types/traits/state` and `hotshot/demo` to a separate file in `testing`.
Also collapses three separate `NodeType` implementations which were using basically the same types into one.

### This PR does not: 
Code to move was selected mainly by moving `block_impl` into testing first and seeing what broke, so it is probably not a comprehensive list of all demo/testing implementations.

### Key places to review: 
`crates/testing/src/demo.rs`


crates/testing/src/lib.rs
```diff
@@ -27,6 +27,14 @@ pub mod node_types;
 /// task to spin nodes up and down
 pub mod spinning_task;
 
+/// Implementations for testing/examples
+#[cfg_attr(
```
QuentinI, line 31: Kinda hack-ish but can be useful IMO.
Not sure what's our general policy for this sort of thing though

crates/testing/src/demo.rs
QuentinI, line 1: Kind of uninspired, but I don't yet feel like I understand the codebase well enough to come up with a better name.

crates/testing/src/demo.rs
```diff
@@ -0,0 +1,510 @@
...
+            state::TestableBlock,
+            BlockPayload,
+        },
+    };
+    use serde::{Deserialize, Serialize};
+    use sha3::{Digest, Keccak256};
+
+    /// The transaction in a [`VIDBlockPayload`].
+    #[derive(Default, PartialEq, Eq, Hash, Serialize, Deserialize, Clone, Debug)]
+    pub struct VIDTransaction(pub Vec<u8>);
```
shenkeyao, line 47: Nit: Probably rename `VID*` structs to `Demo*`. "VID" used to indicate that they were associated with VID commitment, but since they've moved to `demo`, I think it's better to rename them.

crates/testing/src/lib.rs
shenkeyao: Is this `cfg_attr` needed because we are using demo implementations in non-test code? I think this is okay as a short-term fix but we'll need to investigate the relevant non-test code.

Do we have a list of places where demo implementations are used? If so we could probably investigate and fix them in a separate PR.

shenkeyao (changes requested):
Looks good! Some minor comments about naming and configuration.

crates/testing/src/demo.rs
shenkeyao: What name it is? This comment doesn't seem to appear in the changed file (maybe because it's "outdated"?), so I'm not sure which name it's referring to.

QuentinI: @shenkeyao I meant the `demo` itself.

crates/testing/src/lib.rs
QuentinI: It is not needed at all, it just generates warning if something from `demo` module is detected in context that is not one of test build, example, debug build

shenkeyao: Oh, I see. This is nice!

crates/testing/src/demo.rs
shenkeyao: Got it. Since we have `node_types.rs` in this testing crate, I think ideally we could split this `demo.rs` to `block_types.rs`, `state_types.rs` etc. If we want to keep this file, maybe `test_types.rs` is a better name.

crates/testing/src/demo.rs
QuentinI: Done

crates/testing/src/block_types.rs
```diff
@@ -83,21 +81,32 @@ impl VIDBlockPayload {
...
+        let encoded = TestTransaction::encode(vec![TestTransaction(txns.clone())]).unwrap();
+        TestBlockPayload {
+            transactions: vec![TestTransaction(txns)],
             payload_commitment: vid_commitment(&encoded),
         }
     }
 }
 
-impl Display for VIDBlockPayload {
+impl Committable for TestBlockPayload {
```
shenkeyao, line 92: Is this needed? I think we no longer need `Committable` for block payload. Maybe this was added due to some merge issue.

crates/testing/src/state_types.rs
shenkeyao, line 1: Thanks for splitting the files!

shenkeyao (changes requested):
Just one question about `Committable`. There are also merge conflicts to be resolved.

crates/testing/src/lib.rs
ggutoski: I like the way you think. I don't fully understand what's achieved. We're already in the `testing` crate. So nothing from this crate should be used in non-test/example/debug contests. Can't we replace all these `cfg_attr` with a single one at the `testing` crate root?

ggutoski (commented):
Thank you! I don't have much insight here. Some dumb questions:

- The vast majority of edits are merely renaming, correct?
- Seems that you were able to completely remove all these impls from production code to test, correct? Fantastic! Is there anything that you couldn't move?

crates/testing/src/lib.rs
QuentinI: @ggutoski right, that makes sense, I'll see if gating entirety of the testing crate will cause any issues.

crates/testing/src/block_types.rs
QuentinI: Oh, git sneaked that in while rebasing I think. Thank you for the catch!

QuentinI:
> Thank you! I don't have much insight here. Some dumb questions:
> 
>     * The vast majority of edits are merely renaming, correct?

Yes, it's mostly renaming and moving stuff around. One other thing is that I deleted a couple of identical dummy `NodeTypes` impls in favor of a single one in `testing` crate

> 
>     * Seems that you were able to completely remove all these impls from production code to test, correct? Fantastic! Is there anything that you couldn't move?

There wasn't anything that I encountered that I wanted to move but couldn't, but there may still be some impls that could/should be moved to testing that I just didn't stumble upon.

ggutoski:
I'm happy but I'll let someone from the consensus team give final approval. Some notes:
- I think force-push confuses github so I can't see what's changed since my previous review. I don't know that we have a firm policy on force-push. I prefer to avoid it in favor of merge main if it's after others have already reviewed, but each individual case is a judgement call.
- In the future PRs should be posted directly to the hotshot repo as opposed to from a fork. Make some noise if you lack permission for this.

QuentinI:
@ggutoski oh okay. Is there an up-to-date document describing our git(hub) workflow?

shenkeyao (approved):

shenkeyao (approved):

PR #2143, opened 2023-11-30
Updates orchestrator and web to use url instead of IP address or hostname

### LINKED ISSUE #2142: [TECH-DEBT] - Parameterize hostname for web server and orchestrator so HTTPS is supported

Currently, both the orchestrator and web server hardcode the `http` protocol.  We want to update both services so that we can support `https` as well.  

- [ ] Update sequencer repo with changes
- [x] Update web server and orchestrator to use `Url` type instead of `String`

_No response_

- [x] All web server tests pass
- [ ] @Ancient123 can successfully deploy both the web server and orchestrator using HTTPS. 
- [ ] PR in sequencer repo is accepted

elliedavidson:
Closes #2142

### This PR: 
* Updates the orchestrator and web server to take `url` arguments instead of `host` arguments.  The `url` arguments include the network protocol, such as HTTP or HTTPS.  Previously everything was hard-coded to always use `HTTP`.  
* Updates the network configuration file accordingly so that it also takes in url instead of hostname or IP address.  

### This PR does not: 
* Fix other tech debt issues in the web server and orchestrator

### Key places to review: 
* orchestrator/lib.rs
* web_server/lib.rs
* examples/infra/mod.rs

### DIFF
--- a/crates/hotshot/examples/combined/all.rs
+++ b/crates/hotshot/examples/combined/all.rs
@@ -45,7 +45,11 @@ async fn main() {
     async_spawn(async move {
         if let Err(e) = hotshot_web_server::run_web_server::<
             <DemoTypes as hotshot_types::traits::node_implementation::NodeType>::SignatureKey,
-        >(Some(server_shutdown_cdn), 9000)
+        >(
+            Some(server_shutdown_cdn),
+            "http://localhost".to_string(),
+            9000,
+        )
         .await
         {
             error!("Problem starting cdn web server: {:?}", e);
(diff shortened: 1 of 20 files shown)


rob-maron:
I have to update some of the examples to use the URL instead , will make those changes and piggyback on this PR

elliedavidson:
> I have to update some of the examples to use the URL instead , will make those changes and piggyback on this PR

Oh I might have just done this now!  Sorry, I didn't see your comment until now.

elliedavidson:
> Works for me !
> 
> It might be better if we had the URL include the port (and then had the examples default to 0.0.0.0 with the specified port), but this is a step in the right direction

Yes, we should do this!  I'm going to do this now, since it will actually make fewer changes in the sequencer repo.  I'll switch everything to using the `Url` type instead of strings and u16s.

Issue #2088, opened 2023-11-20
[LIBP2P] Parameterize the DHT timeout


rob-maron:
### Describe the task
* What is the reason for this task? 

Right now, the DHT timeout as specified in the Libp2p network is hardcoded and optimal for 10-30 nodes. We want to parameterize this so we can change it based on the number of actual nodes in the run. This is especially helpful for testing purposes.

* What work will need to be done to complete this task?  If this is unknown, describe what information is needed to determine the work needed

To complete this task, we will need to 
1. Add the DHT timeout to the testing metadata.
2. Add the DHT timeout to the example parameters

Both of these just require adding and using a simple parameter. May require working a bit through the codebase to chase down anything.

### Acceptance Criteria
Describe the acceptance criteria needed for this task to be closed
* List any sub-tasks that need to be completed 
* List which tests are applicable to this change.  List any new tests that need to be added for this 

- [ ] Add the DHT timeout to the testing metadata.
- [ ] Add the DHT timeout to the example parameters

### Checklist
- [x] Label this issue appropriately.  If this issue is created by an outside contributor, label it `outside-issue`. Unless a task is labeled `low-priority`, `outside-issue`, or `tech-debt` it must include a `sprint` label and be associated with the `sprint` milestone. 
- [x] If this is a macro task, add [MACRO] to the front of the issue name
- [x] If this is not a macro task, add [<MACRO_TASK_SHORT_NAME>] to the front of the issue name. Then add a link to the macro task it is associated with.  Every non-macro task must be associated with a macro task unless it is labeled `low-priority`,`tech-debt`, or `outside-issue`. 
- [x] Add this issue to the appropriate project boards, unless it is labeled `low-priority`, `outside-issue`, or `tech-debt` 
- [x] Assign this issue to the appropriate person, if applicable.

bfish713:
We set this pretty high during investigations, we should work it down to something more reasonable now that the issues are resolved.

PR #2150, opened 2023-12-01
[Stability] Fix DHT timeout 

### LINKED ISSUE #2132: [Stability] - Investigate and fix DHT lookup failure

This task is to investigate and fix a DHT lookup failure that is causing our Libp2p timeout test to fail.

It was likely introduced in the combined network test PR, which fixed an issue where Libp2p nodes weren't shutting down all the way. Now that they are shutting down all the way, a recursive lookup call that would normally resolve quickly hits the maximum timeout and causes successive views to fail.

- Figure out why the DHT takes so long to return a value when a single node is down
- Resolve the issue

_No response_

- [ ] The test for timing out Libp2p passes without being modified

rob-maron:
Closes #2132

### This PR: 

Fixes the DHT lookup timeout by doing the following:
- Decreasing DHT timeout (still needs to be parameterized anyway: https://github.com/EspressoSystems/HotShot/issues/2088)
- Removing the walk for every PID lookup (to be re-added with https://github.com/EspressoSystems/HotShot/issues/2151)
- Adding some delays, changing some tolerances
- Makes everyone a bootstrap node, making it more deterministic by removing the recursive lookup

It would be better if we could do this without changing the test, but in a real run the timeout time would be much higher, giving us more room. We are also working on auditing the current DHT code for improvements.

### This PR does not: 

- Change any other DHT functionality


bfish713 (approved):

DieracDelta (approved):
In the interest of getting tests to pass asap, LGTM.

Making all nodes bootstrap nodes isn't ideal long term due to the heavy network requirements of the high mesh degree, unless we're assuming all the nodes are well equipped to handle high amounts of traffic (which may be true?). Some other things to try:

- Increase the replication factor, so more nodes have the keys out of the gate
- Increase parallelism, so we query more nodes at once [here](https://github.com/EspressoSystems/HotShot/blob/develop/crates/libp2p-networking/src/network/node.rs#L250)
- Add more nodes to the test. Kademlia is meant to be running with a large network. I remember having strange edge case issues with a smaller # of nodes.

The rest of the changes look great!

PR #2163, opened 2023-12-05
VID share distribution algorithm (2)

### LINKED ISSUE #2104: [T4.0] - VID share distribution

Before we are able to develop custom transmission logic for VID shares, we need to figure out which VID shares go to which storage nodes.

This task is to implement the algorithm for choosing which storage nodes get which shares. A good start is making it deterministic based on the public key, which makes reconstruction easier.

- [x] Come up with VID dispersal algorithm
- [x] Create an abstract way to interface with this algorithm in HotShot
- [x] Implement the algorithm

_No response_

Required reviewers:
@ggutoski @shenkeyao 

New tests:
- [ ] VID shares go to proper nodes, and no others

rob-maron:
> Closes #2104 
> 
> 
> 
> ### This PR: 
> 
> 
> 
> 
> 
> - Changes the number of shares and chunk size to:
>     - Conform to VID changes (power of 2)
>     - Use the quorum membership
> - Adds a constructor for `VidDisperse` that takes a membership for share distribution
> - Implements an algorithm for the constructor to deterministically disperse VID shares based on index
> - Changes the testing function `vid_init` to use the constructor so it's easy to use during tests
> - Edits the tests to conform to this
> 
> ### This PR does not: 
> 
> 
> - Change any other VID functionality
> - Address networking or actual distribution of shares 
> 
> ### Key places to review: 
> 
> 
> `vid.rs` - where we do the actual dispersal
> `data.rs` - where we specify what algorithm to use for dispersal given membership and the shares
> 
> 
> 
> 
> 
> 
> 

This PR augments and fixes #2149 by modifying the `vid_commitment` function. All the relevant (unapproved) changes are in the second commit "second set of changes" and the commit "remove quorum membership from DA".


crates/task-impls/src/da.rs
```diff
@@ -57,6 +57,9 @@ pub struct DATaskState<
     /// Membership for the DA committee
     pub da_membership: Arc<TYPES::Membership>,
 
+    /// Membership for the quorum committee
+    pub quorum_membership: Arc<TYPES::Membership>,
```
shenkeyao, line 61: I feel a bit counter-intuitive to store quorum membership in the DA task. If all we need is the number of quorum nodes for the VID commitment computation, could we pass this info differently?

A potential way is to add `total_nodes` to the `TransactionsSequenced` event in the VID task (which knows the quorum membership), and pass that to `DAProposalSend`, then `DAProposalRecv`, so that the DA task will have `total_nodes` without storing the quorum membership.

ggutoski: Agree that `quorum_membership` should not be part of `DATaskState`.

Checking my understanding: my intuition is that each node knows the number of nodes in the quorum membership, so there should be no need to pass this info around in events. But I guess this info is only available inside a task whose state includes `quorum_membership`, and we don't want to pollute tasks with this data if they need only the node count. Correct?

In that case, yeah, I guess Keyao's solution seems best.

crates/task-impls/src/events.rs
```diff
@@ -39,7 +43,11 @@ pub enum HotShotEvent<TYPES: NodeType> {
     /// Send a quorum vote to the next leader; emitted by a replica in the consensus task after seeing a valid quorum proposal
     QuorumVoteSend(QuorumVote<TYPES>),
     /// Send a DA proposal to the DA committee; emitted by the DA leader (which is the same node as the leader of view v + 1) in the DA task
-    DAProposalSend(Proposal<TYPES, DAProposal<TYPES>>, TYPES::SignatureKey),
+    DAProposalSend(
+        Proposal<TYPES, DAProposal<TYPES>>,
+        TYPES::SignatureKey,
+        usize,
+    ),
```
ggutoski, line 50: House-cleaning note for the future: for complex enum variants we might consider embedded structs to give each item a name. Or, even better, we could make a stand-alone struct for complex variants. https://stackoverflow.com/a/70321820

crates/task-impls/src/network.rs
```diff
@@ -91,9 +91,14 @@ impl<TYPES: NodeType> NetworkMessageTaskState<TYPES> {
...
                         },
                         Either::Right(committee_message) => match committee_message {
-                            CommitteeConsensusMessage::DAProposal(proposal) => {
-                                HotShotEvent::DAProposalRecv(proposal.clone(), sender)
-                            }
+                            CommitteeConsensusMessage::DAProposal(
+                                proposal,
+                                num_quorum_committee,
+                            ) => HotShotEvent::DAProposalRecv(
+                                proposal.clone(),
```
ggutoski, line 98: Nit: I'm surprised clone is needed here and clippy didn't complain.

crates/types/src/traits/block_contents.rs
```diff
@@ -57,12 +57,18 @@ pub trait BlockPayload:
     /// If the transaction length conversion fails.
     fn from_transactions(
         transactions: impl IntoIterator<Item = Self::Transaction>,
+        num_storage_nodes: usize,
```
ggutoski, line 60: Hmm. I don't think `BlockPayload` constructors need `num_storage_nodes`. We should talk about this on the next call, possibly with input from @jbearer . `BlockPayload` needs VID parameters (including `num_storage_nodes` plus some other stuff) but *only* for tx/namespace proofs.

crates/types/src/traits/block_contents.rs
```diff
@@ -83,13 +89,16 @@ pub trait BlockPayload:
 /// # Panics
 /// If the VID computation fails.
 #[must_use]
-pub fn vid_commitment(encoded_transactions: &Vec<u8>) -> <VidScheme as VidSchemeTrait>::Commit {
+pub fn vid_commitment(
+    encoded_transactions: &Vec<u8>,
+    num_storage_nodes: usize,
```
ggutoski, line 94: In the future I expect we'll need more VID params beyond `num_storage_nodes`. There's a good chance this arg will be replaced by a full on `&VidScheme` or whatever.

ggutoski (commented):
There's a couple design questions we need to discuss sometime. Not sure whether we need to resolve these questions before merging this PR.

crates/types/src/traits/block_contents.rs
rob-maron: I agree, I think we can remove this if we get creative. The reason we have it here is because the constructor uses it to compute the VID commitment all of the transactions. Is this technically a _tx_ proof or am I thinking about it wrong?

crates/task-impls/src/da.rs
rob-maron: I like passing around the entire scheme, I think it lends much better to future changes.

I wonder if the scheme in `DATaskState` would have to change - do we expect to have a dynamic number of storage nodes?

crates/types/src/traits/block_contents.rs
ggutoski: > Is this technically a _tx_ proof or am I thinking about it wrong?

I use "tx proof" to refer to something completely different. These proofs are used at the sequencer layer to service queries from light clients about individual txs or namespaces. It's not relevant to hotshot.

> The reason we have it here is because the constructor uses it to compute the VID commitment all of the transactions. 

The constructor does *not* compute the VID commitment of all the txs. The constructor merely marshals the txs into flat bytes, which can then be used by hotshot to compute the VID commitment.

Basically, `BlockPayload` trait allows the downstream user to specify how exactly txs are marshalled into bytes. But computing the actual VID commitment should be done only once, and it must be inside hotshot.

crates/types/src/traits/block_contents.rs
ggutoski: Yeah, for now I guess we could create a static VidScheme on startup.

In the future we need to create a new VidScheme whenever we change the number of storage nodes. I guess right now that's in the Transactions task? That feels less than ideal to me, but I can't think of anything better.

ggutoski: whatever we do in the future, I'm 99% certain we will *not* be constructing a `VidScheme` inside the `vid_commitment` function; these VID params will be needed elsewhere, too.

crates/types/src/traits/block_contents.rs
jbearer: Yeah, we originally had a `commit()` function on the `BlockPayload` trait, mostly for convenience. But we deleted it, because as Gus said it's not the responsibility of the payload itself to deal with VID. But I guess we just forgot to delete the `payload_commitment` field from the test payload type

crates/types/src/traits/block_contents.rs
rob-maron: In order for us to be able to pass the VidScheme around between nodes, it needs to implement `Serialize` and `Deserialize`. I think we need this to do things like include it in the `DAProposal`, wherein replicas need to vote on the commitment.

Can we (I) just derive it for `Advz` and make a PR? Or is it more complex than that? Or maybe we can offload this to later and just use the `num_committee_nodes` for now (or create a static one on startup).

Does this make sense? WDYT?

crates/task-impls/src/da.rs
ggutoski: Relevant: https://github.com/EspressoSystems/HotShot/pull/2163#discussion_r1419461463

Each node has all the info it needs to build the VidScheme. IMO it's merely a matter what's the cleanest way to access this info. It would suck to add the whole quorum exchange to the DA state just to get the quorum size but I don't know about other alternatives. Need some insight from seasoned veterans of the hotshot codebase.

crates/types/src/traits/block_contents.rs
```diff
@@ -102,6 +97,12 @@ pub fn vid_commitment(
     vid.commit_only(encoded_transactions).unwrap()
 }
 
+/// Computes the (empty) genesis VID commitment
+#[must_use]
+pub fn genesis_vid_commitment() -> <VidScheme as VidSchemeTrait>::Commit {
+    vid_commitment(&vec![], 8)
```
ggutoski, line 103: Pretty sure we don't want to compute the genesis commitment like this but there's no need to worry about that now.

ggutoski (approved):
It's a thing of beauty!

crates/task-impls/src/da.rs
shenkeyao: Sounds good!

PR #2166, opened 2023-12-05
Remove ConsensusApi & HotShotType

### LINKED ISSUE #1194: Refactor ConsensusApi to use HotShotConsensusApi directly
~~`SequencingConsensusApi`, `ValidatingConesensusApi`, and `HotShot` traits have functions with very similar implementations. E.g., they all have `send_broadcast_message` functions that call `inner.exchanges.quorum_exchange()`. We should refactor them to use the underlying implementations of `HotShotInner` directly.~~

[The above description is outdated. See comment below.]

QuentinI:
Closes #1194

### This PR:
- Removes `ConsensusApi` and `HotShotType`
Those traits had only one implementor each at this point, implementations moved to `impl`s for corresponding types. 
- Renamed `ConsensusSharedApi` to `ConsensusApi`

### This PR does not: 
Get rid of `ConsensusSharedApi`. Trying to move it's implementation directly into `HotShotConsensusApi` introduces a circular dependency, since some `ConsensusSharedApi`methods  are used in `task-impls`. As far as my understanding of codebase goes, it's fine as it is and can be left alone unless we want to refactor crate structure altogether.
That said, I'm not sure if we should reconsider which methods should be implemented on `HotShotConsensusApi` directly and which should be part of `ConsensusApi`.

### Key places to review: 
`crates/hotshot/src/lib.rs`, `crates/types/src/traits/consensus_api.rs`


crates/hotshot/src/lib.rs
```diff
@@ -802,59 +701,60 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> ConsensusSharedApi<TYPES, I>
...
-                    Message {
-                        sender: inner.public_key.clone(),
-                        kind: MessageKind::from_consensus_message(message),
-                    },
-                    recipient,
-                )
-                .await
-        });
-        Ok(())
+        inner
```
QuentinI, line 716: I don't think we need `async_spawn_local` here anymore, correct me if I'm wrong

crates/hotshot/src/lib.rs
```diff
@@ -802,59 +701,60 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> ConsensusSharedApi<TYPES, I>
...
+    /// # Errors
+    /// - [`NetworkError`] if sending the message failed
+    pub async fn send_direct_message(
         &self,
         recipient: TYPES::SignatureKey,
         message: SequencingMessage<TYPES>,
     ) -> std::result::Result<(), NetworkError> {
         let inner = self.inner.clone();
         debug!(?message, ?recipient, "send_direct_message");
-        async_spawn_local(async move {
```
QuentinI, line 816: I don't think we need `async_spawn_local` here anymore, do we?

rob-maron: Hmmmmm, good question

I think we still want it to be non-blocking (because some of the downstream futures like DMing on libp2p take kind of long), but not through `async_spawn_local`. I think we should only use that for futures that are not `Send`; we can probably get away with `async_spawn` for this.

However I don't think this function's caller is actually used anywhere upstream, maybe we can remove it entirely 😃

rob-maron:
Looks good to me! You're doing god's work by getting rid of all this extra code 😄 

One minor comment - it looks like we still use `HotShotType` in the examples (in `crates/hotshot/examples/infra/mod.rs:RunDA`. I think these are excluded from the lint and build checks because they are only conditionally compiled.

To see what I'm talking about, you can try something like `just async_std example all-webserver --features="libp2p/rsa"`. (I am working on fixing the libp2p/rsa requirement in another PR)

crates/hotshot/src/lib.rs
QuentinI: My reasoning was if the caller wants this not to be blocking, they can just do the same thing on their side. Also, this has an advantage of not just dropping any errors that might arise.

shenkeyao: > However I don't think this function's caller is actually used anywhere upstream, maybe we can remove it entirely 😃

+1, we could probably remove more functions inside this `impl`.

crates/hotshot/src/lib.rs
```diff
@@ -740,7 +639,7 @@ pub struct HotShotConsensusApi<TYPES: NodeType, I: NodeImplementation<TYPES>> {
 }
 
 #[async_trait]
-impl<TYPES: NodeType, I: NodeImplementation<TYPES>> ConsensusSharedApi<TYPES, I>
+impl<TYPES: NodeType, I: NodeImplementation<TYPES>> ConsensusApi<TYPES, I>
```
shenkeyao, line 642: I noticed in the PR description you mentioned that we could not remove `ConsensusApi` directly because some functions were still used--good catch! However, I'm wondering if it's indeed necessary to have this trait. Most of its functions are just getters, which I think makes more sense to be implemented in `HotShotConsensusApi`, especially now that only `HotShotConsensusApi` uses these functions.

Could we move those needed functions to `HotShotConsensusApi` and delete unused ones?

shenkeyao (changes requested):
Thanks for the cleanup! Left a few comments about moving/deleting functions.

shenkeyao (approved):
Looks good! There are conflicts to be resolved though.

shenkeyao (approved):
LGTM

QuentinI:
@rob-maron I'll wait for your approval as well since you had comments

rob-maron (approved):
Looks great !

PR #2168, opened 2023-12-05
[Tech Debt] Allow nodes to rejoin by saving index and config

### LINKED ISSUE #2135: [TECH-DEBT] - Allow nodes to fully stop, lose all memory state, restart, and properly rejoin the network

Nodes currently cannot rejoin the network after they fully shutdown and restart.  They can't properly get the configuration file from the orchestrator, identify themselves to the network, and get enough information to start participating in the network again.   Our current "shutdown" tests do not fully shutdown nodes, but instead only pause nodes, which isn't realistic behavior.  This task adds functionality for nodes to optionally read configuration items from disk when they start up.  This issue is needed to support more resilient testnets. 

Nodes should write their configuration files and other necessary information to disk and optionally read from disk when they start up. 

- [x] Investigate how the sequencer currently handles this.  It's possible the easiest fix is sequencer-side
- [x] Add ability for node to optionally read from a config file at a parametrizable location on disk at startup in the HotShot example code
- [x] Add ability for node to write config file to disk at startup to a parametrizable location in the HotShot example code
- [x] Test by running a small network of nodes and killing / restarting a range of those nodes.  Ensure that the network continues to function. 
- [x] Discuss integration of changes with sequencer team
- [x] Integrate into sequencer code

Ideally we should add the ability to fully shutdown nodes to our testing harness, but that is outside the scope of this issue. 

- [ ] Sequencer tests pass
- [x] HotShot test network with 5 nodes successfully handles shutting down and restarting all 5 nodes at different times (such that only 1 is ever offline at a time).  Note that this will be a manual test.

rob-maron:
Closes #2135 

### This PR: 

Allows nodes to rejoin consensus after stopping and dropping off completely.

- Adds `from_file` and `to_file` to the `NetworkConfig` implementation
- Adds `from_file_or_orchestrator` to the above implementation, which takes an orchestrator client and returns both the configuration and the source (so we don't make the `Ready` call to the orchestrator every time)
- Cleans up a lot of `OrchestratorClient` code

There are also some unrelated changes that make running the examples easier.

### This PR does not: 

- Edit any example functionality
- Add a default config file  for examples that run the orchestrator, webserver, and nodes (all_xxxx)

### Key places to review: 

- `config.rs` - the `NetworkConfig` implementations
- examples/`mod.rs` - for how we use the new `NetworkConfig` API in our examples


rob-maron:
One q here - cc @elliedavidson @jbearer should I move the "wait for ready" functionality in the orchestrator to the new function (config_from_file_or_orchestrator)? This allows us to conditionally wait based on whether or not we loaded from the orchestrator or not, which would allow us to start even if the orchestrator is down.

crates/orchestrator/src/config.rs
```diff
@@ -78,6 +94,151 @@ pub struct NetworkConfig<KEY: SignatureKey, ELECTIONCONFIG: ElectionConfig> {
...
+impl<K: SignatureKey, E: ElectionConfig> NetworkConfig<K, E> {
+    /// Asynchronously retrieves a `NetworkConfig` either from a file or from an orchestrator.
+    ///
+    /// This function takes an `OrchestratorClient`, an identity string, and an optional file path.
+    ///
+    /// If a file path is provided, the function will first attempt to load the `NetworkConfig` from the file.
+    /// If the file does not exist or cannot be read, the function will fall back to retrieving the `NetworkConfig` from the orchestrator.
+    /// The retrieved `NetworkConfig` is then saved back to the file for future use.
+    ///
+    /// If no file path is provided, the function will directly retrieve the `NetworkConfig` from the orchestrator.
```
dailinsubjam, line 111: Love this documentation! Would you like to add an example of `NetworkConfig` file in [`config`](https://github.com/EspressoSystems/HotShot/tree/main/config) folder? This might be helpful if others want to create their own network config.

dailinsubjam (approved):
LGTM!

crates/orchestrator/src/config.rs
dailinsubjam: Sounds good! Thanks!

crates/hotshot/examples/combined/multi-validator.rs
```diff
@@ -58,9 +63,12 @@ async fn main() {
...
                 ThisRun,
             >(ValidatorArgs {
-                url,
+                url: args.url,
                 port: args.port,
                 public_ip: args.public_ip,
+                network_config_file: args
+                    .network_config_file
+                    .map(|s| format!("{}-{}", s, node_index)),
             })
```
shenkeyao, line 72: Nit: we could add a `from_multi_args` constructor to `ValidatorArgs` to simplify the code here.

crates/hotshot/examples/webserver/multi-validator.rs
```diff
@@ -57,9 +61,12 @@ async fn main() {
...
                 ThisRun,
             >(ValidatorArgs {
-                url,
+                url: args.url,
                 port: args.port,
                 public_ip: args.public_ip,
+                network_config_file: args
+                    .network_config_file
+                    .map(|s| format!("{}-{}", s, node_index)),
             })
```
shenkeyao, line 70: Nit: same here, it would be nice to add a constructor to `ValidatorArgs`.

crates/orchestrator/src/config.rs
```diff
@@ -78,6 +94,151 @@ pub struct NetworkConfig<KEY: SignatureKey, ELECTIONCONFIG: ElectionConfig> {
...
+                Ok(config) => (config, NetworkConfigSource::File),
+                Err(e) => {
+                    // fallback to orchestrator
+                    error!("{e}");
+
+                    let config = client.get_config(client.identity.clone()).await;
+
+                    // save to file if we fell back
+                    if let Err(e) = config.to_file(file) {
+                        error!("{e}");
```
shenkeyao, line 147: Should we handle this error either here or in the caller? I think we should do more than just the logging.

crates/hotshot/examples/libp2p/multi-validator.rs
```diff
@@ -58,9 +62,12 @@ async fn main() {
...
                 ThisRun,
             >(ValidatorArgs {
-                url,
+                url: args.url,
                 port: args.port,
                 public_ip: args.public_ip,
+                network_config_file: args
+                    .network_config_file
+                    .map(|s| format!("{}-{}", s, node_index)),
             })
```
shenkeyao, line 71: Nit: same here, it would be nice to add a constructor to `ValidatorArgs`.

shenkeyao (changes requested):
LGTM, just a few minor comments.

crates/orchestrator/src/config.rs
rob-maron: Hmm, I think maybe we should consider leaving it here. This function is supposed to be higher-level for the common use case of "load from either", but if the caller wants to add additional functionality they can call the functions directly and not use this one

crates/hotshot/examples/combined/multi-validator.rs
rob-maron: I like this idea, will do

crates/orchestrator/src/config.rs
shenkeyao: Makes sense! Logging the error should suffice then.

crates/hotshot/examples/combined/multi-validator.rs
rob-maron: Done !

shenkeyao (approved):
LGTM!

dailinsubjam (approved):

PR #2193, opened 2023-12-11
[Metrics] Fix block height metric

dailinsubjam:
### This PR: 
Fix the block height metrics by assigning the most recent leaf (rather than its old ancestor).

### DIFF
--- a/crates/task-impls/src/consensus.rs
+++ b/crates/task-impls/src/consensus.rs
@@ -582,6 +582,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> + 
                 let mut current_chain_length = 0usize;
                 if parent_view + 1 == view {
                     current_chain_length += 1;
+                    let mut is_ancestor = false; // For metrics use: mark whether we're going to the ancestor of the leaf, only assign block_height for the most recent one.
                     if let Err(e) = consensus.visit_leaf_ancestors(
                         parent_view,
                         Terminator::Exclusive(old_anchor_view),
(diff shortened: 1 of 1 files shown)


bfish713 (changes requested):

crates/task-impls/src/consensus.rs
```diff
@@ -609,11 +610,13 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
                             // starting from the first iteration with a three chain, e.g. right after the else if case nested in the if case above
                             if new_decide_reached {
                                 let mut leaf = leaf.clone();
-                                consensus
-                                    .metrics
-                                    .last_synced_block_height
-                                    .set(usize::try_from(leaf.get_height()).unwrap_or(0));
-
+                                if !is_ancestor {
```
nyospe, line 613: You could also just check `if leaf.view_number == new_anchor_view {`?

shenkeyao: +1 I feel checking the view number is better than relying on the `is_ancester` flag.

shenkeyao (changes requested):
Seconding the opinion that we could just check the view number.

crates/task-impls/src/consensus.rs
bfish713: +1 I think this resolves my comment to, I mostly wanted to avoid the extra `is_ancester` flag as well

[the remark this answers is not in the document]
dailinsubjam: I think this makes more sense. But if I update metrics at line 677 the synced block height will be much higher than updating at line 613, why this could be the case?

dailinsubjam: Aha!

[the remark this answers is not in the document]
dailinsubjam: Oh, my understanding is that the `leaf` we called on 677 is the current newest leaf, we made a decide on this view = leaf.view(), but the decided block is the leaf on view - 2 (in most cases) due to the consensus protocol. Therefore we'd better still do the update on line 613. Is this understanding correct?

shenkeyao (approved):

nyospe (approved):
LGTM

PR #2204, opened 2023-12-12
Cancel Timeout Tasks and Fail tests if too many Timeouts

### LINKED ISSUE #2165: STABILITY - Cancel Timeout Tasks properly

When trying to improve our testing I realized that we don't ever emit anything when a view times out.  This prevents us from failing due to to many views timing out.  Because of this our test parameter for number of failures isn't doing very much.  The only way to fix this is to properly emit a timeout external event.  We can't do this because our timeout tasks don't get canceled properly so we end up with extra timeout events.

- [ ] Complete the task whenever we get a view change which invalidates it, or it hits the time threshold
- [ ] Emit a timeout event externally when the time threshold of the task is hit

_No response_

- All tests pass
- I can observe through logging that the test task is receiving the ViewTimeout events 

_No response_

bfish713:
Closes #2165 

### This PR: 

- Cancels the consensus timeout task as soon as we enter a new view (before we start the next timeout)
- Keeps track of the view sync timeout task, and cancels it whenever we send a new view sync vote
- Emits a `ViewTimedout` event to the application when the consensus timeout task expires
- Fixes the test infra to properly account for view timeouts and fixes an issue where views were never counted as failed

### This PR does not: 

- Properly set the num_failed_views parameter for all tests.  I specifically set it very high so the tests will pass.  This is consistent with prior behavior where the test infra never counted any failed views (`num_failed_views` was effectively infinite).

### Key places to review: 

- Timeout cancellation logic in consensus.rs and view_sync.rs
- Changes to OverallSafetyTask to fix bugs in the test harness


shenkeyao (changes requested):
LGTM! I left some minor comments.

shenkeyao (approved):

shenkeyao (approved):

PR #2210, opened 2023-12-12
Propose When if we can When Getting VID Dispersal Info

### LINKED ISSUE #2207: [BUG FIX] - Fix Failure To Propose because of no VID data

We are getting the log "Wasn't able to publish proposal" often in test net.  This would happen when we form a QC but don't yet have the payload commitment from the VID task and therefore cannot propose.  This would be fine but when we do get the data we need we don't actually create another proposal.

- [ ] Add Logging when this specific situation occurs
- [ ] Fix the bug by trying to propose when we get the VID data
- [ ] Confirm that it fixes the bug by seeing the logging we added hit, and then a decided happen for that same view.

_No response_

The last bullet point above:  Confirm that it fixes the bug by seeing the logging we added hit, and then a decided happen for that same view.

_No response_

bfish713:
Closes #2207 

### This PR: 

Fixes a bug where we can't propose because VID disperse is still pending, but then we don't try to propose after it completes.  This adds the logic to propose if we can after the disperse compeltes

### Key places to review: 

The added code in consensus.rs

### How to test this PR: 

I added a short `async_sleep` for 300ms immediately before sending the `SendPayloadCommitmentAndMetadata` event.  This was enough timing disruption to hit the bug in almost every view.  With my fix and warn logs on you should see "Cannot propose because we don't have the VID payload commitment and metadata" but the view won't timeout because the proposer succeeds later when it gets the info it needs

### DIFF
--- a/crates/task-impls/src/consensus.rs
+++ b/crates/task-impls/src/consensus.rs
@@ -994,11 +994,11 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> + 
             }
             HotShotEvent::SendPayloadCommitmentAndMetadata(payload_commitment, metadata) => {
                 self.payload_commitment_and_metadata = Some((payload_commitment, metadata));
-                // let high_qc = self.consensus.read().await.high_qc.clone();
-                // let leader_view = high_qc.get_view_number() + 1;
-                // if self.quorum_membership.get_leader(leader_view) == self.public_key {
-                //     self.publish_proposal_if_able(high_qc, leader_view, None).await;
-                // }
+                let high_qc = self.consensus.read().await.high_qc.clone();
+                let leader_view = high_qc.get_view_number() + 1;
+                if self.quorum_membership.get_leader(leader_view) == self.public_key {
+                    self.publish_proposal_if_able(high_qc, leader_view, None).await;
+                }
             }
             _ => {}
         }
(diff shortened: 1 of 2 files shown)


rob-maron:
A question for my understanding: is there a reason we don't just remove the other call to `propose_if_able` and use this one instead? (If we're saying we can't propose until the VID disperse is done). Or is this a situation where it's now basically `min(vid disperse, payload commitment)`?

bfish713:
> A question for my understanding: is there a reason we don't just remove the other call to `propose_if_able` and use this one instead? (If we're saying we can't propose until the VID disperse is done). Or is this a situation where it's now basically `min(vid disperse, other criterion)`?

It's actually that we can't propose unless all three of the following conditions are met: 

1. We saw the last proposal (or formed a timeout cert)
2. We collected enough votes and formed a QC on that proposal (or formed a timeout cert)
3. We have the VID disperse information (payload commit and metadata).

These 3 things can come in any order so we might propose during the events for any of the 3.

rob-maron:
> > A question for my understanding: is there a reason we don't just remove the other call to `propose_if_able` and use this one instead? (If we're saying we can't propose until the VID disperse is done). Or is this a situation where it's now basically `min(vid disperse, other criterion)`?
> 
> It's actually that we can't propose unless all three of the following conditions are met:
> 
> 1. We saw the last proposal (or formed a timeout cert)
> 2. We collected enough votes and formed a QC on that proposal (or formed a timeout cert)
> 3. We have the VID disperse information (payload commit and metadata).
> 
> These 3 things can come in any order so we might propose during the events for any of the 3.

Got it, thanks !

rob-maron (approved):
Looks good !

shenkeyao (approved):
Looks good!

PR #2220, opened 2023-12-13
Reduce Logging Level in Many Places

bfish713:
### This PR: 

Lower logs to appropriate levels

### This PR does not: 

change any functionality/features

### Key places to review: 

If I'm changing to the correct level and if I missed any logs that should be lowered.


elliedavidson (commented):
Can we also alter some logging in the view sync task?  Specifically, I'd like to error log when view sync in triggered and when it completed, but reduce all the other logs to warn or lower.  For example "Vote sending timed out" and similar logs don't need to be error level.

rob-maron (approved):

elliedavidson (approved):

PR #2223, opened 2023-12-13
[Stability] Cancel polling for old votes

### LINKED ISSUE #2222: [Stability] - Fix poll cancelling for votes

Right now we don't cancel polls properly for votes; this causes us to continue polling for old views 0..n that have already become irrelevant.

Add poll cancelling for votes when we create a certificate. Also add it for timeouts.

_No response_

- [ ] a manual run where no old polls are still present

_No response_

rob-maron:
Closes #2222 

### This PR: 

Fixes a bug wherein we are not cancelling old polls for votes.

Cancels polls for votes in the following places:
- when we receive a timeout in view sync or a finalized view sync cert 
- when we form a QC or DAC 
- when we time out for both the quorum and view sync (DA already had this)

Also cancels a poll for proposals for view 1, as this is covered by the latest proposal mechanism. For a node finishing catchup, it would constantly try to pull the first proposal.

### This PR does not: 

Add any GC improvements/more advanced proposal logic/cancelling as discussed. This will be in another PR.

### Key places to review: 

`da.rs`, `consensus.rs`, `view_sync.rs` to make sure all the cancels are in a good spot.

### DIFF
--- a/crates/task-impls/src/consensus.rs
+++ b/crates/task-impls/src/consensus.rs
@@ -414,7 +414,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> + 
         match event {
             HotShotEvent::QuorumProposalRecv(proposal, sender) => {
                 debug!(
-                    "Receved Quorum Propsoal for view {}",
+                    "Receved Quorum Proposal for view {}",
                     *proposal.data.view_number
                 );
 
@@ -840,6 +840,13 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> + 
                 debug!("QC Formed event happened!");
 
                 if let either::Right(qc) = cert.clone() {
+                    // cancel poll for votes
+                    self.quorum_network
+                        .inject_consensus_info(ConsensusIntentEvent::CancelPollForVotes(
+                            *qc.view_number,
+                        ))
+                        .await;
+
                     debug!(
                         "Attempting to publish proposal after forming a TC for view {}",
                         *qc.view_number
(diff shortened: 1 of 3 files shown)


crates/task-impls/src/view_sync.rs
```diff
@@ -782,6 +787,13 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
                     return (None, self);
                 }
 
+                // cancel poll for votes
+                self.network
```
bfish713, line 791: Same comment we want `CancelPollForViewSyncVotes`

bfish713 (changes requested):

crates/task-impls/src/view_sync.rs
bfish713: Also I think we should `CancelPollForViewSyncCertificate` here as well

rob-maron: You're right 🙈

rob-maron: Should we also cancel the certificate poll in the case of a timeout?

bfish713: Which certificate and which timeout?

rob-maron: In the case of a `Timeout` event in the view sync task, should we cancel polls for the certificate of that view ?

bfish713: `Timeout` is the start of view sync (if it's the second in a row) so probably shouldn't cancel any view sync related things

crates/task-impls/src/view_sync.rs
```diff
@@ -465,6 +465,13 @@ impl<
                     return;
                 }
 
+                // cancel poll for votes
+                self.network
+                    .inject_consensus_info(ConsensusIntentEvent::CancelPollForViewSyncVotes(
```
bfish713, line 470: I think it was correct before, we just timed out consensus for the round and are now maybe entering view sync, we should cancel polling for quorum votes and proposals but that should be handled in the consensus task

crates/task-impls/src/consensus.rs
```diff
@@ -975,6 +993,11 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
                     return;
                 }
```
bfish713, line 995: Cancel poll for proposal here too, I think

bfish713 (approved):
LGTM, Can you wait for CI to pass before trying to merge

PR #2226, opened 2023-12-13
[Stability] Garbage collect polls for all views < latest-2

### LINKED ISSUE #2225: [Stability] - Implement task garbage collection for all views <n-2

We want to garbage collect polls for all views lower than n-2. Right now polls will just be cancelled if they are _exactly_ 2 in the past, which doesn't cover all cases.

_No response_

_No response_

- [ ] a manual run wherein we inject old polls at random times and see them get cancelled

_No response_

rob-maron:
Closes #2225 

### This PR: 

- Adds better garbage collection for all old tasks. Specifically:
    - It changes the `TaskMap` for all data types to be a `BTreeMap` (which is ordered)
    - For every new poll, it uses our new ordered "list" to cancel polls less than or equal to the current poll - 2
- After view sync, adds a poll for the current (most recent) proposal per the webserver.

### This PR does not: 

Address any other webserver-related errors or bugs.

### Key places to review: 

Both files ! (minus the one which just got linted)

### DIFF
--- a/crates/hotshot/src/traits/networking/web_server_network.rs
+++ b/crates/hotshot/src/traits/networking/web_server_network.rs
@@ -331,6 +331,7 @@ impl<TYPES: NodeType> Inner<TYPES> {
                         ConsensusIntentEvent::CancelPollForVotes(event_view)
                         | ConsensusIntentEvent::CancelPollForProposal(event_view)
                         | ConsensusIntentEvent::CancelPollForDAC(event_view)
+                        | ConsensusIntentEvent::CancelPollForViewSyncCertificate(event_view)
                         | ConsensusIntentEvent::CancelPollForVIDDisperse(event_view)
                         | ConsensusIntentEvent::CancelPollForViewSyncVotes(event_view) => {
                             if view_number == event_view {
--- a/crates/hotshot/src/traits/networking/web_server_network.rs
+++ b/crates/hotshot/src/traits/networking/web_server_network.rs
@@ -175,8 +175,6 @@ impl<TYPES: NodeType> Inner<TYPES> {
                 MessagePurpose::VidDisperse => config::get_vid_disperse_route(view_number), // like `Proposal`
             };
 
-            println!("getting {}", endpoint);
-
             if message_purpose == MessagePurpose::Data {
                 let possible_message = self.get_txs_from_web_server(endpoint).await;
                 match possible_message {
@@ -801,9 +799,9 @@ impl<TYPES: NodeType + 'static> ConnectedNetwork<Message<TYPES>, TYPES::Signatur
                 }
 
                 // Remove all entries in the task map that are polling for a view less than or equal to `view_number - 2`.
-                let view_minus_2 = view_number.checked_sub(2).unwrap_or(0);
-                let mut range = task_map.range(..view_minus_2);
-                while let Some((view, task)) = range.next() {
+                let view_minus_2 = view_number.saturating_sub(2);
+                let range = task_map.range(..view_minus_2);
+                for (view, task) in range {
                     // Cancel the old task by sending a message to it. If the task already exited we expect an error
                     let _res = task
                         .send(ConsensusIntentEvent::CancelPollForProposal(*view))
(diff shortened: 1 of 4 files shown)


rob-maron:
I have tested this in multiple runs and it works well. Includes view sync runs, and runs where 3 (view_sync)+1 nodes are down in a row

crates/hotshot/src/traits/networking/web_server_network.rs
```diff
@@ -986,7 +977,17 @@ impl<TYPES: NodeType + 'static> ConnectedNetwork<Message<TYPES>, TYPES::Signatur
...
+                let view_minus_2 = view_number.saturating_sub(2);
+                let range = task_map.range(..view_minus_2);
+                for (view, task) in range {
+                    // Cancel the old task by sending a message to it. If the task already exited we expect an error
+                    let _res = task
+                        .send(ConsensusIntentEvent::CancelPollForViewSyncCertificate(
+                            *view,
+                        ))
+                        .await;
+                }
```
rob-maron, line 990: I need an expert opinion on this one. I ran tests with view sync and it looks good, but maybe there's some weird case I didn't think about. Since we weren't doing any cancelling before for view sync certs, is it probably better to just remove this?

rob-maron: ^^ cc @elliedavidson @bfish713

bfish713: This should be fine, view sync does have a requirement that we help others get out of view sync but that should be complete after 2 views + other nodes in our system will get future proposals to get out of view sync

crates/hotshot/src/traits/networking/web_server_network.rs
```diff
@@ -799,15 +799,13 @@ impl<TYPES: NodeType + 'static> ConnectedNetwork<Message<TYPES>, TYPES::Signatur
...
-                        .send(ConsensusIntentEvent::CancelPollForProposal(
-                            view_number.wrapping_sub(2),
-                        ))
+                // Remove all entries in the task map that are polling for a view less than or equal to `view_number - 2`.
+                let view_minus_2 = view_number.saturating_sub(2);
+                let range = task_map.range(..view_minus_2);
+                for (view, task) in range {
+                    // Cancel the old task by sending a message to it. If the task already exited we expect an error
+                    let _res = task
+                        .send(ConsensusIntentEvent::CancelPollForProposal(*view))
```
bfish713, line 808: nit Can we make this a function that takes the event, maybe as a fn like `event_fn: fn(view: u64) -> ConsensusIntentEvent` and then the function could just do range.map(event_fn).  Mostly just making this a function so this code isn't copy pasted a bunch of times

crates/hotshot/src/traits/networking/web_server_network.rs
```diff
@@ -768,8 +770,6 @@ impl<TYPES: NodeType + 'static> ConnectedNetwork<Message<TYPES>, TYPES::Signatur
         );
 
         // TODO ED Need to handle canceling tasks that don't receive their expected output (such a proposal that never comes)
-        // TODO ED Need to GC all old views, not just singular views, could lead to a network leak
-
         match event {
             ConsensusIntentEvent::PollForProposal(view_number) => {
```
bfish713, line 774: I think we should also GC the view sync stuff here, it's possible we just don't hit view sync for a long time, and we'd be running the view sync polling task for a ton of views.  In fact if maybe we can just cancel all old task regardless of what we are getting the new polling intent for.  Basically every time we start to poll for a new view number we can GC old view task from all the task maps

bfish713 (approved):
I don't see anything functionally wrong with doing the GC this way.  We guarantee that we have at most 2 of each task at any given time.  That seems reasonable.  

I had some comments on how we might simplify this but IMO this is a good improvement that we should get in for the next deployment tomorrow

crates/hotshot/src/traits/networking/web_server_network.rs
bfish713: Also just double checking, you were specifically running the a webserver based test for view sync right?

rob-maron:
> I don't see anything functionally wrong with doing the GC this way. We guarantee that we have at most 2 of each task at any given time. That seems reasonable.
> 
> I had some comments on how we might simplify this but IMO this is a good improvement that we should get in for the next deployment tomorrow

These are good chances, I will have them out today

crates/hotshot/src/traits/networking/web_server_network.rs
rob-maron: We can do this but I have to be careful; certain things like transactions we poll way ahead for. If we start polling for txes in view n+3, polls for other things in view n would get cancelled before we have everything resolved

bfish713: Agree that's a good point we need to be careful.  I was more thinking only for the big things, like proposal and view sync certificates.  On second thought maybe it's best to have a specific event like `CancelOldPolls` which we inject when we update our view that cancels the old view polling.  That way we have one entry point for GCing but don't have to piggy back on the old events which might have weirdness like you mentioned.

PR #2233, opened 2023-12-14
Update view sync for 2 timeouts

elliedavidson:
Closes #<ISSUE_NUMBER>

### This PR: 
* Fixes a bug where view sync waited for 3 views to trigger instead of 2 views.

### DIFF
--- a/crates/task-impls/src/view_sync.rs
+++ b/crates/task-impls/src/view_sync.rs
@@ -500,9 +500,9 @@ impl<
                     let subscribe_view = if self.membership.get_leader(TYPES::Time::new(next_view))
                         == self.public_key
                     {
-                        next_view
-                    } else {
                         next_view + 1
+                    } else {
+                        next_view
                     };
                     // Subscribe to the next view just in case there is progress being made
                     self.network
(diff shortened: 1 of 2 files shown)


crates/task-impls/src/view_sync.rs
```diff
@@ -743,6 +744,12 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
                         .publish(HotShotEvent::ViewSyncFinalizeVoteSend(vote))
                         .await;
                 }
+
+                error!(
+                    "View sync protocol has received view sync evidence to update the view to {}",
+                    *self.next_view
+                );
```
shenkeyao, line 751: I think the logging at lines 475 and 484 should suffice?

shenkeyao (dismissed):
LGTM, a nit comment about logging.

crates/task-impls/src/view_sync.rs
elliedavidson: My thought here was to log when view sync successfully updates the view, as well as when it starts.  So we can see if view sync is updating the view or if the view is updated because a proposal was received.

shenkeyao: Makes sense!

shenkeyao (approved):
Looks good

bfish713 (dismissed):

shenkeyao (approved):

PR #2230, opened 2023-12-14
[Stability] Only use open ports for test runs

rob-maron:
### This PR: 

- Removes spindown time between combined network tests
- Fixes the intermittent combined network issues by using `portpicker` for both generated `libp2p` and `webserver` networks
- Moves the crypto tests to the "build" CI, renames it to "build_and_test"
- _May_ fix the intermittent webserver network bind failure

### This PR does not: 

- Change any other tests
- Implement deterministic testing

### Key places to review: 

`libp2p_network.rs`

PR #2239, opened 2023-12-14
Imrovements for WebServerNetwork

### LINKED ISSUE #2206: [FIX_CI] - Fix web_server_network test

`web_server_network` test sporadically fails in CI:
```
Message:  TEST FAILED! Results: [("Test Overall Safety Task", NotEnoughDecides { expected: 35, got: 0 })]
Location: /opt/actions-runner/_work/HotShot/HotShot/crates/testing/src/test_runner.rs:177
```
with following error making appearance in the logs:
```
Failed to send message from network task: WebServer { source: ClientError }
```

Locate and fix the cause of `web_server_network` failing

_No response_

`web_server_network` passes consistently

_No response_

QuentinI:
Closes #2206
(which was just closed in #2230, this is mostly improvements on top of that)

### This PR:
- Logs an error when webserver task finishes, which isn't something it's supposed to do
  The actual reason for #2206 was that we've been picking ports from the entire u16 value space, which ran into system ports eventually and server failed to start. This issue was fixed by @rob-maron in #2230, this PR just adds logging to appropriate places so when web server fails it doesn't do that silently.
- When creating a `WebServerNetwork`, allows to wait for the web server to come online
  I suspected it as a cause for #2206, but it ultimately wasn't. I think it still makes sense, so I didn't throw the code out.

### This PR does not: 
Do anything else

### Key places to review: 
`crates/hotshot/src/traits/networking/web_server_network.rs`:
- `WebServerNetwork::create`
- `WebServerNetwork::generator`


QuentinI:
Drafted since it seems like I have broken some tests

rob-maron:
> Drafted since it seems like I have broken some tests

If you're running into the combined network failure, those were fixed in a recent commit to `main`. So depending on your base you might want to `git merge origin/main` 😄

rob-maron:
🤔 Hmm, okay. For the combined network tests, they throw a bunch of events at the webserver telling it to shut down, start up, pause, resume, etc.

Since the combined network is both the libp2p and webserver networks (and will work if it can't connect to the webserver network) I think the panic when we can't connect is causing the issue.

QuentinI:
There's something really weird going on with this branch. I'm closing this for now, since whatever it is it doesn't cause any issues for us on main, but I feel like something somewhere doesn't behave like it should, so I'll continue to investigate in my spare time.

PR #2244, opened 2023-12-15
View Sync Fixes for PreCommit Threshold, Timeout Logic, and Relay Vote Accumulation

### LINKED ISSUE #2243: [STABILITY] - Pre-Commit Veiw Sync Certificate Should have f + 1 as Threshold

After refactor the threshold for all certificates is the success threshould of the membership with is `2f+1`.  View Sync pre-commit certs should have `f+1`

- Allow the SimpleCertificate type to have different threshold
- Change the threshold for PreCommit certs

_No response_

- [ ] Threshold is correct
- [ ] View Sync tests are successful

_No response_

bfish713:
Closes #2243 

### This PR: 
- Adds two structures to swap between different thresholds in `SimpleCertificate`
- Makes the ViewSyncPreCommitCertificate type use the threshold of f+1 and the rest us 2f+1
- Use the `Certificate::threshold` function in `is_valid_cert`
- Always send a pre commit vote after a timeout to the next relay, unless we are already finalized
- Fixes a bug where only one vote accumulator is spawned for a relay, now one is spawned for each phase in a view
- Get rid of `ViewSyncRelayTaskState` - it was not in current use and it was just a wrapper around vote accumulation anyways
- Removes phase checks in the replica tasks because we should respond to all certificates with a vote (if relay 1 gets to commit, but we timeout we don't want to ignore the next pre-commit cert)

### This PR does not: 

Fix the issue with duplicate events being sent through the view sync task

### Key places to review: 

The changes to the view_sync.rs file should be looked at closely since some logic does change

### DIFF
--- a/crates/task-impls/src/view_sync.rs
+++ b/crates/task-impls/src/view_sync.rs
@@ -901,7 +901,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> + 
                         .await;
                     self.relay += 1;
                     match self.phase {
-                        ViewSyncPhase::None => {
+                        ViewSyncPhase::None | ViewSyncPhase::PreCommit | ViewSyncPhase::Commit => {
                             let vote = ViewSyncPreCommitVote::<TYPES>::create_signed_vote(
                                 ViewSyncPreCommitData {
                                     relay: self.relay,
@@ -917,44 +917,6 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> + 
                             if let GeneralConsensusMessage::ViewSyncPreCommitVote(vote) = message {
                                 self.event_stream
                                     .publish(HotShotEvent::ViewSyncPreCommitVoteSend(vote))
-                                    .await;
-                            }
-                        }
-                        ViewSyncPhase::PreCommit => {
-                            let vote = ViewSyncCommitVote::<TYPES>::create_signed_vote(
-                                ViewSyncCommitData {
-                                    relay: self.relay,
-                                    round: self.next_view,
-                                },
-                                self.next_view,
-                                &self.public_key,
-                                &self.private_key,
-                            );
-                            let message =
-                                GeneralConsensusMessage::<TYPES>::ViewSyncCommitVote(vote);
-
-                            if let GeneralConsensusMessage::ViewSyncCommitVote(vote) = message {
-                                self.event_stream
-                                    .publish(HotShotEvent::ViewSyncCommitVoteSend(vote))
-                                    .await;
-                            }
-                        }
-                        ViewSyncPhase::Commit => {
-                            let vote = ViewSyncFinalizeVote::<TYPES>::create_signed_vote(
-                                ViewSyncFinalizeData {
-                                    relay: self.relay,
-                                    round: self.next_view,
-                                },
-                                self.next_view,
-                                &self.public_key,
-                                &self.private_key,
-                            );
-                            let message =
-                                GeneralConsensusMessage::<TYPES>::ViewSyncFinalizeVote(vote);
-
-                            if let GeneralConsensusMessage::ViewSyncFinalizeVote(vote) = message {
-                                self.event_stream
-                                    .publish(HotShotEvent::ViewSyncFinalizeVoteSend(vote))
                                     .await;
                             }
                         }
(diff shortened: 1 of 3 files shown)


crates/task-impls/src/view_sync.rs
```diff
@@ -452,8 +409,31 @@ impl<
                                 )
                                 .await;
                         }
-                        if let Some((_key, relay_task_info)) =
-                            self.relay_task_map.remove_entry(&TYPES::Time::new(i))
+                        if let Some((_key, relay_task_info)) = self
```
shenkeyao, line 412: Nit: Is it possible to combine these `if let` checks? Looks like they are sending the direct message in the same way, though associated with different phases. Maybe we could use `||` to send at most one `Shutdown`.

crates/types/src/simple_certificate.rs
```diff
@@ -24,9 +24,35 @@ use crate::{
 
 use serde::{Deserialize, Serialize};
 
+/// Trait which allows use to inject different threshold calculations into a Certificate type
+pub trait Threshhold<TYPES: NodeType> {
```
shenkeyao, line 28: Nit: typo?

crates/types/src/simple_certificate.rs
```diff
@@ -24,9 +24,35 @@ use crate::{
...
+    /// Calculate a threshold based on the membership
+    fn threshold<MEMBERSHIP: Membership<TYPES>>(membership: &MEMBERSHIP) -> u64;
+}
+
+/// Defines a threshold which is 2f + 1 (Amount needed for Quorum)
+#[derive(Serialize, Deserialize, Eq, Hash, PartialEq, Debug, Clone)]
+pub struct SuccessThreshhold {}
```
shenkeyao, line 35: Looks like there are several struct names with the double “h”.

crates/types/src/simple_certificate.rs
```diff
@@ -123,18 +150,20 @@ impl<TYPES: NodeType> QuorumCertificate<TYPES> {
...
 /// Type alias for a Timeout certificate over a view number
-pub type TimeoutCertificate<TYPES> = SimpleCertificate<TYPES, TimeoutData<TYPES>>;
+pub type TimeoutCertificate<TYPES> =
+    SimpleCertificate<TYPES, TimeoutData<TYPES>, SuccessThreshhold>;
 
 // TODO ED Update this to use the correct threshold instead of the default `success_threshold`
 /// Type alias for a `ViewSyncPreCommit` certificate over a view number
 pub type ViewSyncPreCommitCertificate2<TYPES> =
-    SimpleCertificate<TYPES, ViewSyncPreCommitData<TYPES>>;
+    SimpleCertificate<TYPES, ViewSyncPreCommitData<TYPES>, OneHonestThreshhold>;
```
shenkeyao, line 163: Nit: can remove the TODO comment above.

shenkeyao (approved):
Looks great! No major comments.

rob-maron (approved):
The changes look good, and I did some view sync runs 

I am still getting the issue with a timeout for the view directly after view sync. But I think that's a separate issue

crates/task-impls/src/view_sync.rs
```diff
@@ -730,11 +705,6 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
                     return (Some(HotShotTaskCompleted::ShutDown), self);
                 }
 
-                // Ignore if the certificate is for an already seen phase
-                if last_seen_certificate <= self.phase {
-                    return (None, self);
-                }
-
```
elliedavidson, line 737: I'm not sure we should delete this?  I think it is possible that a node gets a Commit cert, and then also somehow gets the PreCommit cert afterwards, say because of network delay.  In this case I think it would incorrectly reset itself to be in the PreCommit phase, which isn't what we want.

bfish713: The phase doesn't really matter that I think.  In this code it only matters if we get to finalize since we don't use the phase anything on it's own.  I think we could probably get rid of self.phase but I was trying to get something done quickly

elliedavidson (approved):

bfish713:
> The changes look good; I also did some view sync runs locally.
> 
> I am still getting the issue with a timeout for the view directly after view sync. But I think that's a separate issue

Yeah I think that issue is happening in the consensus task, so not actually a bug with view sync

PR #2246, opened 2023-12-15
[Stability] Poll for latest view sync cert

### LINKED ISSUE #2245: [Stability] - Add polling for latest view sync cert

We need to add polling for the latest view sync cert. We want to make sure that during view sync we have not only the current cert, but potential certs for the next view. It should stop polling when we've finalized view sync.

_No response_

_No response_

- we poll for the latest view sync cert
- we stop when finalized

_No response_

rob-maron:
Closes #2245 

### This PR: 

Adds polling for the latest view sync cert. 

Some creative liberties to note:
- Since we don't want to miss any certs (and we can have multiple proposals/certs per view), the "latest view sync (proposal/cert)" function on the webserver just returns all certs for the latest view. We could do server side filtering but if we expect these to be sufficiently small and for view sync to be sufficiently infrequent, then I think this is good until our larger CDN changes.

Everything else is about the same as the "PollForRecentProposal" things from @bfish713. 

### This PR does not: 

Change or update the view sync replicas or tasks, only the networking.

### Key places to review: 

Everything in `view_sync.rs`. To me it is, but I want to be sure the logic to start/stop the poll is sufficient.

I have run a bunch of manual view sync tests to make sure that we start/stop this polling when necessary, and also that unique certs it pulls are processed properly. I'm not sure if there are any more advanced tests I should run, please let me know!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2755,6 +2755,7 @@ dependencies = [
  "libp2p-identity",
  "libp2p-networking",
  "local-ip-address",
+ "lru",
  "portpicker",
  "rand 0.8.5",
  "rand_chacha 0.3.1",
--- a/crates/hotshot/Cargo.toml
+++ b/crates/hotshot/Cargo.toml
@@ -116,6 +116,7 @@ local-ip-address = "0.5.6"
 dyn-clone = { git = "https://github.com/dtolnay/dyn-clone", tag = "1.0.16" }
 derive_more = "0.99.17"
 portpicker = "0.1.1"
+lru = "0.12.1"
 
 tracing = { workspace = true }
 typenum = { workspace = true }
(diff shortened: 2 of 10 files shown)


elliedavidson (approved):
In the "viewchange" event handler in view_sync.rs I think we need to add a "ConsensusIntentEvent::CancelPollForLatestViewSyncProposal".  It is possible that a node will go into view sync, but then cancel view sync because they've seen a proposal with a higher view.  In this case I don't think we'll properly cancel polling for the latest view sync certificate.  

But, since it looks like we GC view sync certs properly, I don't think it will cause a memory leak - probably just more network traffic.  So I'm going to approve this PR for now.

elliedavidson (approved):

PR #2247, opened 2023-12-16
[Tech Debt] Key pair generation integration with orchestrator

dailinsubjam:
### This PR: 
- Moved key pair generation for node and its peers to `main_entry_point`.
- Align validator's seed with the seed from `NetworkConfig` for default setting.
- Rename the folder for validator config from `config` to `validator_config`. (See the discussion below)
- Is a followup on https://github.com/EspressoSystems/HotShot/pull/1990.

### This PR does not: 
- Directly input node's key pair through config. (See the discussion below)

### Key places to review: 
`mod.rs` and `config.rs`

### How to test this PR: 

### Comments requested:
- About loading keys from config. Now we have seed and index in config, and generate the real key pair at the entry point. The reason we don't directly loading key pairs is their formats are too complex (see `validator_config/ValidatorConfigExample` for reference) to be in the config, also our HotshotConfig does contain the secret key originally for signning, so it's fine.

- About peer's configuration (including peer's public key and stake value). This is my big concern, now they're all generated at the entry, but I'm not sure whether we have any interface already. I saw [`get_config()`](https://github.com/EspressoSystems/HotShot/blob/main/crates/orchestrator/src/client.rs#L97) in `orchestrator/src/client.rs`, is this a good one to use to get peer's config? If so, do we want to do that now?

- About the folder `validator_config`. The reason we keep this folder is: once a node operator asked us for a script of key pair generation, and they also want to see how key pairs look like, more details [here](https://github.com/EspressoSystems/HotShot/issues/1971#issuecomment-1792062875), therefore we have config file with pure `node_index` and `seed` and examples there. But this might not be the cleanest way to keep them, also I just learnt our general config is at `crates/orchestrator/run-config.toml`, do we want to keep them together under the same folder or not?


rob-maron:
These changes look good, but for some reason I'm having trouble with the examples.

If I run `just async_std example all-webserver ./crates/orchestrator/run-config.toml`, I enter view sync at view 5. Do you see the same thing?

 I tried merging in main locally but it still has the same issue

dailinsubjam:
> If I run `just async_std example all-webserver ./crates/orchestrator/run-config.toml`, I enter view sync at view 5. Do you see the same thing?
> 
> I tried merging in main locally but it still has the same issue

@rob-maron Thanks for the review! Weird, I didn't see any view sync round in my local log and looks like all nodes are connected and running. "enter view sync at view 5" means the first several views are all timeout right? Do you know whether all the nodes are connected?

rob-maron:
> > If I run `just async_std example all-webserver ./crates/orchestrator/run-config.toml`, I enter view sync at view 5. Do you see the same thing?
> > I tried merging in main locally but it still has the same issue
> 
> @rob-maron Thanks for the review! Weird, I didn't see any view sync round in my local log and looks like all nodes are connected and running. "enter view sync at view 5" means the first several views are all timeout right? Are all the nodes connected in your case?

Yeah, they should all be connected; `all-webserver` should run everything we need. One question - how long does it take for the 100 rounds to finish on that example for you locally? When it's done it should print something like "took 30s to complete n rounds"

dailinsubjam:
> > > If I run `just async_std example all-webserver ./crates/orchestrator/run-config.toml`, I enter view sync at view 5. Do you see the same thing?
> > > I tried merging in main locally but it still has the same issue
> > 
> > 
> > @rob-maron Thanks for the review! Weird, I didn't see any view sync round in my local log and looks like all nodes are connected and running. "enter view sync at view 5" means the first several views are all timeout right? Are all the nodes connected in your case?
> 
> Yeah, they should all be connected; `all-webserver` should run everything we need. One question - how long does it take for the 100 rounds to finish on that example for you locally? When it's done it should print something like "took 30s to complete n rounds"

Yeah all of them finish in approx. 33s e.g."[9]: 100 rounds completed in 33.082711917s".

rob-maron (changes requested):
Okay, I finally figured out the behavior. Basically, the orchestrator uses `known_nodes_with_stake` to figure out when to start the network. Ref: `crates/orchestrator/src/lib.rs:170`. If you print the length of `known_nodes_with_stake` on this branch, it seems to always be 0.

Basically, because of (I think) the change to that variable, 2/3 of the nodes were going up to view 7 before the other nodes got a chance to start.

shenkeyao (changes requested):
Looks good. To fix the CI failure on `test_with_failures_f`, maybe just increase `num_failed_views` a bit.

dailinsubjam:
> Okay, I finally figured out the behavior. Basically, the orchestrator uses `known_nodes_with_stake` to figure out when to start the network. Ref: `crates/orchestrator/src/lib.rs:170`. If you print the length of `known_nodes_with_stake` on this branch, it seems to always be 0.
> 
> Basically, because of (I think) the change to that variable, 2/3 of the nodes were going up to view 7 before the other nodes got a chance to start.

Yeah looks like we do need generate key pairs in `load_config_from_file` to ensure the right length. What confused me a little bit is that usually we have to assign `known_nodes_with_stake` and `my_own_validator_config` together, why previously  `known_nodes_with_stake` was assigned in `load_config_from_file` while `my_own_validator_config` was assigned separately at `main_entry_point`? (Now my change is like assign both variables at both locations).

crates/hotshot/examples/infra/mod.rs
```diff
@@ -842,12 +842,16 @@ pub async fn main_entry_point<
...
 
+    // We use input seed to calculate node's own key pair because its type is too complex to directly input through the config file
     run_config.config.my_own_validator_config =
         ValidatorConfig::<<TYPES as NodeType>::SignatureKey>::generated_from_seed_indexed(
             run_config.seed,
             node_index,
             1,
         );
+    // Here is a way to generate peer's public key, in practice we should input peer's stake_value and public key
+    run_config.config.known_nodes_with_stake =
```
nomaxg, line 853: In practice, now that nodes generate their own keys, how will nodes fetch peer public keys + stake? Maybe an example would be helpful

dailinsubjam: @nomaxg  As mentioned in the above `Comments requested`, the second point, peer's configuration (including peer's public key and stake value) is my biggest concern. Now they are all share the same seed and generated at the entry which is exactly where you pointed out. Once I asked @jbearer during testnet meeting about this, he mentioned it's fine. But would be better to double check whether we want to use any interface in `hotshot` to get peers' configuration. Although I suppose the sequencer is performing all updates and HotShot is taking it as an input.

nomaxg: Yes I believe we will need some method of fetching peers' configuration (public key/stake?/etc.), if sequencer nodes are generating their own keys for Cappuccino

rob-maron (approved):

dailinsubjam:
Followed by https://github.com/EspressoSystems/HotShot/issues/2416, and its PR https://github.com/EspressoSystems/HotShot/pull/2525 . Will close this without merge.

PR #2250, opened 2023-12-18
refactor: allow configurable stake table capacity

alxiong:
As part of https://github.com/EspressoSystems/espresso-sequencer/issues/706, we realized that having a hardcoded `STAKE_TABLE_CAPACITY = 200` is a pain as it results in a million gates circuit and slow down unit tests significantly. 

### This PR: 

We introduce configurability of this hyperparameter via generic constant in `state-prover`'s API and `StakeTable::new(capacity)` in `stake-table` crate.

### Key places to review: 

API changes are sensible, as they are breaking changes (which is fine since we are not using in any testnet yet).


mrain (approved):
LGTM

PR #2238, opened 2023-12-14
[Stability] Disable saving of VID shares (for now)

rob-maron:
Closes #2237 

### This PR: 

Disables both the storing and recall of VID shares. They will be re-enabled in #2236

PR #2256, opened 2023-12-19
[TESTS] Add vid unit test and helper functions for testing

### LINKED ISSUE #2148: [VID][TESTS] - Add VID test to test_consensus_task

After updating the vid vote logic: https://github.com/EspressoSystems/HotShot/issues/2067 we want `test_consensus_task` also test the vid flow.

- [x] #2203
- [x] Do a final check to see whether creating valid `justify_qc` for view 2 is doable
- [x] Manual signature aggregation when creating certs
- [x] Assign correct parent leaf for certs
- [x] Add helper function for creating valid cert in unit test
- [x] https://github.com/EspressoSystems/HotShot/pull/2271
- [x] Add VID-related events of input and output in test_consensus_task().
- [x] Add todo for certificate unit tests (DAC, view sync cert, quorum cert): https://github.com/EspressoSystems/HotShot/issues/2255

_No response_

`just async_std test_consensus_task` can also test vid vote logic.

_No response_

dailinsubjam:
Closes #2148 .

### This PR: 
- Add cert creation infrastruction to the testing scheme, enable cert formation (e.g. justify_qc, DAC) with assembled sig creation.
- Enable valid proposal creation with variant view number.
- Add vid test to consensus task.

### This PR does not: 
- Cover unit test for view sync cert or DAC which we want to do in the future
- Do the logger tracking (because I found it's fine to create a valid DAC...).

### Key places to review: 
`consensus_task.rs`, `task_helpers.rs`.

 ### How to test this PR: 
`just async_std test_consensus_task`

For now vid shares are disabled: https://github.com/EspressoSystems/HotShot/pull/2238#event-11254051262, you have to uncomment all the places marked with "re-enable this when HotShot/the sequencer needs the shares for something" to test it.


crates/testing/src/task_helpers.rs
```diff
@@ -148,15 +235,67 @@ async fn build_quorum_proposal_and_signature(
...
+    for cur_view in 2..=view {
+        consensus.state_map.insert(
+            ViewNumber::new(1),
+            View {
+                view_inner: ViewInner::Leaf {
+                    leaf: leaf.commit(),
+                },
+            },
+        );
+        consensus.saved_leaves.insert(leaf.commit(), leaf.clone());
```
bfish713, line 257: Do we need to add to the saved leaves here?

crates/testing/src/task_helpers.rs
```diff
@@ -148,15 +235,67 @@ async fn build_quorum_proposal_and_signature(
...
+        let parent_leaf = leaf.clone();
+        let leaf_new_view = Leaf {
+            view_number: ViewNumber::new(cur_view),
+            justify_qc: created_qc.clone(),
+            parent_commitment: parent_leaf.commit(),
+            block_header: block_header.clone(),
+            block_payload: None,
+            rejected: vec![],
+            timestamp: 0,
+            proposer_id: *api.public_key(),
```
bfish713, line 283: I think this should be the leaders public key, we can get it from the membership by calling `get_leader(curr_view)`

crates/testing/src/task_helpers.rs
```diff
@@ -148,15 +235,67 @@ async fn build_quorum_proposal_and_signature(
...
+    let mut proposal = QuorumProposal::<TestTypes> {
+        block_header: block_header.clone(),
+        view_number: ViewNumber::new(1),
         justify_qc: QuorumCertificate::genesis(),
         timeout_certificate: None,
         proposer_id: leaf.proposer_id,
     };
+    for cur_view in 2..=view {
+        consensus.state_map.insert(
+            ViewNumber::new(1),
```
shenkeyao, line 250: Why don't we use `cur_view` as the view number?

crates/testing/src/task_helpers.rs
```diff
@@ -148,15 +235,67 @@ async fn build_quorum_proposal_and_signature(
...
+            View {
+                view_inner: ViewInner::Leaf {
+                    leaf: leaf.commit(),
```
shenkeyao, line 253: I'm a bit confused about why the leaf is the same each time we insert it into the state map. Shouldn't it change for each view, or is this a simplification implementation for testing?

bfish713: +1, also do we even need to store the anything in the state_map or the leaves at all?

shenkeyao: Looks like we won't retrieve anything from the `state_map` or `saved_leaves` so there's no need to update them here.

dailinsubjam: `leaf` will be re-assigned at the end of the loop every time QAQ, maybe the naming is causing some confusion here, should I change it to `cur_leaf`?

shenkeyao: Oh sorry, I didn't realize this `leaf` was mutable. The renaming sounds good.

However, I'm still confused about why we need to do the `saved_leaves` and `state_map` insertions. In lines 205 and 211, don't we already have the parent state and leaf corresponding to the requested view? Why do we need to build and store them again?

crates/testing/src/task_helpers.rs
```diff
@@ -148,15 +235,67 @@ async fn build_quorum_proposal_and_signature(
...
+            &quorum_membership,
+            ViewNumber::new(cur_view - 1),
+            public_key,
+            private_key,
+        );
+        let parent_leaf = leaf.clone();
+        let leaf_new_view = Leaf {
+            view_number: ViewNumber::new(cur_view),
+            justify_qc: created_qc.clone(),
+            parent_commitment: parent_leaf.commit(),
```
shenkeyao, line 278: Nit: I think we could just use `leaf.commit()` without creating `parent_leaf`. But maybe this whole construction isn't necessary if we could use the leaf from `saved_leaves` directly (see my other comment).

crates/testing/src/task_helpers.rs
dailinsubjam: When creating the vote to trigger `QuorumVoteSend`, `build_vote()` is going to use `saved_leaves`.

crates/testing/src/task_helpers.rs
dailinsubjam: Without updating `state_map`, the test will get an error during `visit_leaf_ancestors`. For `saved_leaves`, see the comment above.

crates/testing/src/task_helpers.rs
dailinsubjam: Good catch! It's a bug T.T. updated to `cur_view - 1`.

crates/testing/src/task_helpers.rs
dailinsubjam: Here I use `leaf` and `parent_leaf` separately, because for every loop it'll do:
```
let parent_leaf = leaf.clone();
let leaf_new_view = Leaf { ..., parent_leaf.commit(), ...};
leaf = leaf_new_view;
```

Maybe the structure is not the best, let me know if there's a better way to do the loop.

crates/testing/src/task_helpers.rs
dailinsubjam: Good catch, never really think about view 3 T.T  I'll add one comment for this.

crates/testing/src/task_helpers.rs
shenkeyao: Hmm, I still don't fully get it. In lines 205 and 211, we already have the parent state and leaf corresponding to the _high QC_ (sorry in my previous reply I said it's to the _requested view_, but it's actually high QC). If we want to update `state_map` and `saved_leaves`, I think it should start from the view after the high QC, instead of view 1. Otherwise, are we assuming the parent view is the genesis view?

crates/testing/src/task_helpers.rs
shenkeyao: I just think it's probably unnecessary to create `parent_leaf`. Maybe something like:

```rust
let leaf_new_view = Leaf { ..., leaf.clone().commit(), ...};
leaf = leaf_new_view;
```

crates/testing/src/task_helpers.rs
dailinsubjam: Aha, sorry I didn't see your earlier comment about this. For line 205 and 211, yeah they're reading the parent state (can I call this view 0? maybe not well-defined but I need a way to separate it from view 1), then we create the leaf for view 1 in line 228, and in the loop we start to create leaves for view 2 and higher. 

And, to validate a proposal at `view x`, we'll need valid `state_map` and `saved_leaves` for `view x-1`. I'm not sure why when `x = 1`, it already had valid `state_map`(I guess due to some initialization), but starting `x = 2`, we do need to save these states for `x - 1` manually to pass all the checks, that's also why I keep these updates at the start of the loop.

Does all these make sense? Let me know if it raised any concerns :)

crates/testing/src/task_helpers.rs
dailinsubjam: I just realized I should add more comments for code read-ability, hope the new comments could help QAQ.
And on the second thought of this, I'm more inclined to keep `let parent_leaf = leaf.clone();` if it will not introduce more latency or storage, as it serves as a change point of views, which could help with readability (I think). Would love to hear different sound if there's a better reason.

shenkeyao (commented):
@dailinsubjam Could we discuss the unresolved comments in the pair programming tomorrow? Might be more efficient and clearer that way.

crates/testing/src/task_helpers.rs
```diff
@@ -101,30 +115,105 @@ pub async fn build_system_handle(
...
-    let consensus_lock = handle.get_consensus();
-    let consensus = consensus_lock.read().await;
+    // build the genesis view
+    let genesis_consensus = handle.get_consensus();
+    let cur_consensus = genesis_consensus.upgradable_read().await;
+    let mut consensus = RwLockUpgradableReadGuard::upgrade(cur_consensus).await;
     let api: HotShotConsensusApi<TestTypes, MemoryImpl> = HotShotConsensusApi {
         inner: handle.hotshot.inner.clone(),
     };
+    // parent_view_number should be equal to 0
```
shenkeyao, line 205: Could we add an assertion here to make sure the view number is as expected?

crates/testing/src/task_helpers.rs
shenkeyao: Sounds good. I have no strong preference for this.

shenkeyao (changes requested):
Looks good. Just one comment re: checking the parent view number.

shenkeyao (approved):

PR #2260, opened 2023-12-19
[Tech Debt] Add logging for when web server task completes

QuentinI:
### This PR: 
Awaits the web server task in `WebServerNetwork` and logs it's completion.
Not having this made #2206 harder to investigate since web server was just failing silently.

### Key places to review: 
`crates/hotshot/src/traits/networking/web_server_network.rs`


rob-maron (approved):
Looks good to me !

PR #2269, opened 2023-12-20
Prevent Duplicate Events To Spawned Tasks

### LINKED ISSUE #2242: [STABILITY] - Fix Duplicate events in View Sync

The way task spawning works means that we have to send events from the spawning task to the spawned task to avoid missing events in the spawned task.  The view sync main task spawns a replica and relay task when view sync is started.  The main task then has to send all it's events to either of the two tasks.  After task start up, new events are then duplicated in the relay and replica tasks (one from main task, and one from the network task).  Further the relay task spawns vote accumulator tasks, which also need to be forward messages.  This means they get 3 vote events for any given vote, one from main task duplicated to replica task, one from replica task, and one from the network.  

Their are two possible fixes for this.  One would be to create a mechanism for coping the queued events to the event stream of the new task when we create it.  After this initial event copy we would not send any more events directly from one task to the other.  This might be very difficult to do with the current task and event infrastructure.

The other fix for this is to change the way we use tasks in the view sync model.  We could just have main task, which instead of spawning replica and relay tasks could create replica and relay states for any view.  These states can be exactly as they are now but instead of being sent events on the event stream, we can just call handle_event on them from the main task.  It's a bit hacky but can would be a slight improvement to the direct_message hack we have now.

_No response_

- [ ] View sync tests pass
- [ ] In view sync we don't see logging indicating duplicate votes/events

_No response_

bfish713:
Closes #2242 

### This PR: 

Prevents duplicate events in our tasks.  Whenever one task directly sends an event to another task, the receiver has to handle it twice (except when it doesn't get it from the event stream).  To change this I altered the task spawning logic in view sync and with vote collectors to not actually spawn a new task.  Instead we keep the task state and invoke the `handle_event` logic from the main task, instead of sending an event on the event stream.

This also incidentally fixes a bug in the consensus task where the `vote_collector` member was accidently being shared by the timeout vote and quorum vote task.  This mean only 1 of those would be spawned at a time.  I am pretty sure it means if a leader got a timeout vote as the first vote it would ignore quorum votes and the round would timeout.

### This PR does not: 

This is not the ideal fix for this.  Eventually we should be able to spawn a task, and know that it will get every event it cares about, even if those events were generated before the task spawned.  Then we can get rid of the log we have all over (before and after this PR) that says `if task is running {send event to task} else { spawn task }` to just `if no task running { spawn task }`.

### Key places to review: 

All the changes to the view sync task, specifically looking at:

- How the replicas are get events now
- How replicas are spawned via the `Timeout` event, and the various CertificateRecv events
- The logic of updating the replica task map and cleaning it up

Also look how we don't spawn a task for vote collection anymore and double check that the logic is correct for creating and updating vote accumulators in view sync, da, and consensus vote recv handlers.

Also note that I altered the flow a little in the DA and Consensus tasks where votes are received.  I think this flow is a little simpler but please double check.  The intended flow is if there is a accumulator and it's for this view have it handle the vote event.  If there is no accumulator or it's for a prior view, create one


crates/task-impls/src/consensus.rs
```diff
@@ -775,19 +775,20 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
                         QuorumCertificate<TYPES>,
-                    >(
-                        &info, vote.clone(), event, name.to_string()
-                    )
+                    >(&info, vote.clone(), event)
                     .await;
-                } else if let Some((_, _, stream_id)) = self.vote_collector {
-                    self.event_stream
-                        .direct_message(stream_id, HotShotEvent::QuorumVoteRecv(vote.clone()))
-                        .await;
```
shenkeyao, line 790: Just wanted to make sure I understand correctly why we remove `direct_message` here. Looks like we are no longer sending `*VoteRecv` in the associated task implementations. Is it because we've already had the network sending `Recv` events for all `Send` events?

It's a bit weird since I used to think it made sense to also have these tasks send `Recv` events again, but I'm not sure now, so wanted to double-check.

shenkeyao (commented):
Looks good! Thanks for the detailed PR description. 👍

crates/task-impls/src/consensus.rs
bfish713: Before this PR both the network task,  these direct message calls would send the `Recv` events.  If just removed `direct_send` then the task would miss some events.  Every event that came in before the task is created would not be seen and it caused some bugs before.  We can't get rid of the network event sending these events because then nobody would receive them

shenkeyao (approved):
LGTM

PR #2276, opened 2023-12-22
Check for Genesis before Clearing `payload_commitment_and_metadata`

bfish713:
### This PR: 

Fixes a race condition where a node could get it's VID payload commit and metadata before it got the first proposal with genesis qc justification and then would clear the correct data and cause it to not propose.

### Key places to review: 
The fix, and in general if the flow of clearing this variable which is set on task creation makes sense

### DIFF
--- a/crates/task-impls/src/consensus.rs
+++ b/crates/task-impls/src/consensus.rs
@@ -224,17 +224,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> + 
                     &self.public_key,
                     &self.private_key,
                 );
-                if let Some((payload_commit, meta)) = &self.payload_commitment_and_metadata {
-                    let (genesis_payload, genesis_meta) =
-                        <TYPES::BlockPayload as BlockPayload>::genesis();
-                    let genesis_commitment = vid_commitment(
-                        &genesis_payload.encode().unwrap().collect(),
-                        self.quorum_membership.total_nodes(),
-                    );
-                    if meta == &genesis_meta && payload_commit == &genesis_commitment {
-                        self.payload_commitment_and_metadata = None;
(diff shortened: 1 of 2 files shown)


rob-maron (dismissed):

crates/task-impls/src/consensus.rs
```diff
@@ -236,6 +235,17 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
                     self.event_stream
                         .publish(HotShotEvent::QuorumVoteSend(vote))
                         .await;
+                    if let Some((payload_commit, meta)) = &self.payload_commitment_and_metadata {
+                        let (genesis_payload, genesis_meta) =
+                            <TYPES::BlockPayload as BlockPayload>::genesis();
+                        let genesis_commitment = vid_commitment(
+                            &genesis_payload.encode().unwrap().collect(),
+                            self.quorum_membership.total_nodes(),
+                        );
```
jbearer, line 244: We are computing an extraneous VID commitment every time we receive a vote? We need to cache this somewhere. Or we could add an explicit `is_genesis` flag to `payload_commitment_and_metadata`, since we should know at the time we set this whether it is the genesis or not

bfish713: This only happens if the proposal is for view 1 and the genesis flag on the justify QC is set.  So it only happens in the first round of voting

bfish713: We could add a flag to remove this check

rob-maron (approved):
Looks good, want to make sure the CI passes first though

bfish713:
> Looks good, want to make sure the CI passes first though

Same, passed on the last commit and this should improve things

jbearer (approved):

PR #2283, opened 2023-12-29
[Tech Debt] Remove async-trait

### LINKED ISSUE #2282: [Tech Debt] - Remove async-trait

Rust 1.75 adds support for async functions in traits. This is supposed to replace the async-trait crate. Remove the async-trait dependency from our code.

Removing all occurrences of `#[async-trait]`

_No response_

We no longer depend on the `async-trait` crate

_No response_

DieracDelta:
Closes #2282 

### This PR: 
Removes `async-trait` from most places in our code.

### This PR does not: 
Remove async-trait from our implementation of libp2p's trait, which is also using async-trait. Since the compiler `async` keyword isn't quite a drop in replacement for `async-trait`, it seems like we should wait for libp2p to upgrade their traits to not use async-trait.

### Key places to review: 
I'm not sure what's idiomatic, and would be interested in what y'all think! By default, clippy warns that `async` should not appear in traits, because then it is not clear what autotraits the future should implement (notably Send, Sync). On the other hand, clippy warns about the return type being `Future<...>` instead of `async` in trait implementations.

What I opted to do (and I'm not sure if this is optimal or idiomatic) is to replace all `async` occurrences in traits with `-> impl Future<..> + Send`, but retain `async fn` in the corresponding trait implementations.

Other options that I can think of:
- disable the trait definition lint and add `async` in trait definitions
- disable the trait implementation lint, so the trait and implementation signatures match

NOTE: This PR should not be merged until packages that depend on it (e.g. espresso-sequencer) have been upgraded to rust 1.75


QuentinI:
As far as I can see, none of the traits are actually supposed to be implemented outside of `HotShot` repo, so I'd say trait definition lint doesn't really apply and can be disabled, since if one needs additional bounds, they can just add them in the same commit.
That said, many of the traits already make use of Send requirement, so we can't have `async fn` in _all_ trait definitions anyway.

DieracDelta:
Decided not to merge this because unsure of how to handle clippy lints

PR #2284, opened 2023-12-29
[TEST] - Add back testing macros

### LINKED ISSUE #2153: TEST - Add back test macros

We are no longer using test macros, but they can be helpful to make the tests more flexible and investigations more efficient.

_No response_

_No response_

Minimal requirement: make the network types adjustable.

(We may add more criteria.)

_No response_

DieracDelta:
Closes #2153 

### This PR: 
Adds the cross product testing macros crate back into hotshot.

### Key places to review: 
TODO


bfish713:
To answer you questions:

1. I think it might be marginally useful, but I think really we just test over different networks.  Since I don't really envision a case where we use different networks for DA and Quorum I don't really see a use for cross product on the associated types in `NodeImplementation`.  I think we want this macro to run our tests on all possible `NodeTypes` which I see expanding as we add tests for the sequencer, with all different network types.  I think this PR is sufficient for that.
2. I think every test in test_basic, test_timeout, and test_catchup should be run with this macro.  Once that's done we can probably clean up and remove some of the libp2p and webserver specific tests (since they should now be fully covered).  Looks to me like the commented out tests are/would be redundant with what we already have in `testing`

bfish713 (approved):
Looks good, I think needs a rebase and it's probably safe to remove the commented out code.

bfish713:
> Commented code has been removed, and I rebased. I think it's probably safe to merge now. I made a separate issue about adding the tests back in. See #2433 . One question: do we want to move the tests under the targets you listed from `hotshot-testing` to `hotshot-macros`? Or copy them and add more networks? If we do the latter, we'll have the same test running twice. Not sure if that's what we want.

I think we should move them so we don't run the tests twice.  Agree this is safe to merge and we can address moving the tests around in another pr.

bfish713 (approved):

PR #2287, opened 2023-12-31
[TESTS] - View Sync Testing Harness Callback

### LINKED ISSUE #2285: [TESTS] - View Sync Testing Harness Callback

We would like to have the testing harness ensure that view sync is triggered when it's supposed to be triggered. We need this to be more confident of the correctness of our protocol.

We'll need to either need to add a custom callback or test task that listens and checks for this. We'll also need to add in some TestMetadata with the view sync expectations.

_No response_

We're able to fail/pass tests based on view sync events.

_No response_

DieracDelta:
Closes #2285 

### This PR: 
This PR adds a view sync check to ensure that certain nodes did have view sync activated during the test run.

### This PR does not: 
Add any substantial tests using this feature. It would be worth discussing what sort of tests should make use of this feature.

### Key places to review: 
Worth confirming that `ViewSyncTaskDescription` matches the sort of API we want to specify nodes hitting view sync

### DIFF
--- a/crates/testing/src/view_sync_task.rs
+++ b/crates/testing/src/view_sync_task.rs
@@ -1,0 +1,114 @@
+use std::{sync::Arc, collections::{HashSet, HashMap}};
+use futures::FutureExt;
+use hotshot_task::task::HotShotTaskTypes;
+use async_compatibility_layer::channel::UnboundedStream;
+use hotshot_task::{task_impls::{TaskBuilder, HSTWithEventAndMessage}, task::{FilterEvent, HandleMessage, HandleEvent, TS}, MergeN, event_stream::ChannelStream};
+use hotshot_task_impls::events::HotShotEvent;
+use hotshot_types::traits::node_implementation::{TestableNodeImplementation, NodeType};
+use snafu::Snafu;
+
+use crate::{test_launcher::TaskGenerator, test_runner::Node, GlobalTestEvent};
+
+/// ViewSync Task error
+#[derive(Snafu, Debug)]
+pub struct ViewSyncTaskErr {}
+
+/// ViewSync task state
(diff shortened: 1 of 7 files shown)


crates/testing/src/view_sync_task.rs
```diff
@@ -0,0 +1,202 @@
...
+}
+
+/// Description for a view sync task.
+#[derive(Clone, Debug)]
+pub enum ViewSyncTaskDescription {
+    /// (min, max) number nodes that may hit view sync, inclusive
+    Threshold(usize, usize),
+    /// node idx -> whether or not the node should hit view sync
+    /// if node not in map, assumed to be `ShouldHItViewSync::DontCare`
+    Precise(HashMap<usize, ShouldHitViewSync>),
```
bfish713, line 63: We can probably simplify this a little bit.  I think either all nodes should hit view sync or none.  The only exception here is the fact that some nodes will be down.  I think this can be just simplified to `Threshold` and then eliminate some of the matching below.

crates/testing/src/view_sync_task.rs
```diff
@@ -0,0 +1,202 @@
...
+                .register_event_handler(event_handler)
+                .register_message_handler(message_handler)
+                .register_message_stream(MergeN::new(streams));
+                let task_id = builder.get_task_id().unwrap();
+                (task_id, ViewSyncTaskTypes::build(builder).launch())
+            }
+            .boxed()
+        })
+
+        // match self {
```
bfish713, line 193: Nit: remove comment

bfish713 (approved):
Really just one small comment about simplifying the task, generally looks good and useful

bfish713 (approved):

PR #2293, opened 2024-01-01
[CI] - Fix docs in CI

### LINKED ISSUE #2171: CI - Check docs in CI

We don't have a CI check for docs at the moment, which leads to proliferation of broken links and such.

Add a CI check that runs `cargo doc` with appropriate flags to fail on broken links and any other lints we deem unacceptable

_No response_

A CI check for docs is implemented

_No response_

DieracDelta:
Closes #2171 

### This PR: 
Hard-fails docs in CI by adding to justfile and fixes doc lints/tests s.t. CI passes. I ended up ignoring most of the doc tests. I suspect these doc tests were intended as illustrative examples and not as tests.

### This PR does not: 
Add more lints. Not sure what further lints to add.

### Key places to review: 
Documentation changes

### DIFF
--- a/crates/orchestrator/src/config.rs
+++ b/crates/orchestrator/src/config.rs
@@ -184,7 +184,7 @@ impl<K: SignatureKey, E: ElectionConfig> NetworkConfig<K, E> {
     ///
     /// # Examples
     ///
-    /// ```no_run
+    /// ```ignore
     /// # use hotshot_orchestrator::config::NetworkConfig;
     /// # use hotshot_signature_key::bn254::BLSPubKey;
     /// // # use hotshot::traits::election::static_committee::StaticElectionConfig;
--- a/crates/orchestrator/src/config.rs
+++ b/crates/orchestrator/src/config.rs
@@ -126,8 +126,8 @@ impl<K: SignatureKey, E: ElectionConfig> NetworkConfig<K, E> {
     /// # Examples
     ///
     /// ```ignore
-    /// use hotshot_orchestrator::config::NetworkConfig;
-    /// use hotshot_orchestrator::client::OrchestratorClient;
+    /// # use hotshot_orchestrator::config::NetworkConfig;
+    /// # use hotshot_orchestrator::client::OrchestratorClient;
     /// let client = OrchestratorClient::new();
     /// let identity = "my_identity".to_string();
     /// let file = Some("/path/to/my/config".to_string());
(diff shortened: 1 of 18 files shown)


crates/types/src/traits/consensus_api.rs
```diff
@@ -13,7 +13,8 @@ use async_trait::async_trait;
 
 use std::{num::NonZeroUsize, time::Duration};
 
-/// The API that [`HotStuff`] needs to talk to the system
+/// The API that tasks use to talk to the system
+/// TODO do we plan to drop this?
```
QuentinI, line 17: I think we discussed this at a team meeting recently. My summary is - we'd like to remove it, but it runs into `task-impls` and `hotshot` becoming dependent on each other. If I'm not missing something, we agreed that we can just demote `task-impls` to a submodule of `hotshot` and nuke this trait, but I didn't actually open an issue then. I went ahead and opened one: #2294

crates/orchestrator/src/config.rs
```diff
@@ -182,7 +184,8 @@ impl<K: SignatureKey, E: ElectionConfig> NetworkConfig<K, E> {
     ///
     /// # Examples
     ///
-    /// ```no_run
+    /// ```ignore
+    /// use hotshot_orchestrator::config::NetworkConfig;
```
QuentinI, line 188: Nit: aren't `use`-statements usually left out of examples?

QuentinI (approved):
Great!

As for doctests, I'd open an issue for fixing those and promoting them back to `no_run`. `# Example` sections that straight up don't compile is a problem IMO.

crates/orchestrator/src/config.rs
DieracDelta: I would love to avoid the use statement, but I'm not sure how else to tell rust about type. If I drop that line (and retain `no_run` instead of `ignore`), the first error I get is:
```
failures:

---- crates/orchestrator/src/config.rs - config::NetworkConfig<K,E>::from_file (line 187) stdout ----
error[E0433]: failed to resolve: use of undeclared type `NetworkConfig`
 --> crates/orchestrator/src/config.rs:189:14
  |
4 | let config = NetworkConfig::from_file(file).unwrap();
  |              ^^^^^^^^^^^^^ use of undeclared type `NetworkConfig`
  |
help: consider importing this struct
  |
2 + use hotshot_orchestrator::config::NetworkConfig;
  |

error: aborting due to previous error
```

QuentinI: It's usually hidden by prefixing it with #:
```suggestion
    /// # use hotshot_orchestrator::config::NetworkConfig;
```
This way it's still included when checking the doctest but isn't rendered in the docs.

crates/types/src/traits/consensus_api.rs
DieracDelta: Awesome, thanks for the summary! I went ahead and linked the issue here instead

DieracDelta:
+1, we should definitely fix those doc tests. I opened a separate issue for this [here](https://github.com/EspressoSystems/HotShot/issues/2297)

crates/orchestrator/src/config.rs
DieracDelta: Woah, awesome! I had no idea that was possible. TIL!

QuentinI (approved):

PR #2363, opened 2024-01-09
[A2A Network] -  Implement A2A Broadcast

### LINKED ISSUE #2303: [A2A Network] - Implement A2A Broadcast

Implement an All-to-All broadcast using direct messaging

_No response_

_No response_

A libp2p test is able to pass using this new communication mode for broadcast

_No response_

DieracDelta:
Closes #2303 

### This PR: 
Implements an All to All broadcast mode and has the all to all broadcast test pass.

### Key places to review: 
The implementation of an all to all broadcast.

### DIFF
--- a/crates/hotshot/examples/infra/mod.rs
+++ b/crates/hotshot/examples/infra/mod.rs
@@ -8,7 +8,7 @@ use hotshot::traits::implementations::{CombinedCommChannel, CombinedNetworks};
 use hotshot::{
     traits::{
         implementations::{
-            Libp2pCommChannel, Libp2pNetwork, MemoryStorage, NetworkingMetricsValue,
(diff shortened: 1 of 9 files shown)


bfish713 (commented):
Sorry I've been holding off on reviewing this because I'm not really sure what the requirements for Timeboost are/would be.  To me it seems that we should be just using gossip to do the all to all, but I can see if the network is small that this wouldn't make sense and direct messages make sense.  

Also in the long run if we're just doing this point to point we should probably just use a simplified libp2p implementation underneath, e.g. without the gossipsub behaviour, and potentially using streams instead because we'll likely be in constant contact with all other nodes.

My conclusion is if we need some point to point only benchmarks then I'm fine with this.  However if we don't care about that I'd rather not add more tech debt because we won't use this for HotShot, and will have to remove it at some point when we split off timeboost (if we ever implement that)

PR #2369, opened 2024-01-09
Spawn VID with `spawn_blocking` WIP

### LINKED ISSUE #2349: [VID] - Parallelize VID disperse calculations

Depends on: #2347 

Once we can calculate VID on smaller parts of the encoded transactions. then we can run each calculation in parellel.  Probably this will mean running each calculation asynchronously and joining all the calc tasks. 

_No response_

We should see a performance increase and no change in anything else in the protocol.

_No response_

bfish713:
Closes #2349 

### This PR: 

Spawn the VID calculation on a different thread from the async executor to not block other tasks.  From this blog https://ryhl.io/blog/async-what-is-blocking/ from the tokio creator we should not be going more than ~100 micro seconds.  VID disperse takes an order of magnitude or longer than this so we should run it on it a dedicated thread for blocking.  The blog recommends rayon for cpu tasks but also says if we don't have a lot of CPU work than `spawn_blocking` is ok 

I also increased the async threads to two in the just file because I believe the CI infra has 2 threads (except for maybe self hosted).  Personally I think we should remove this restriction entirely.  

### This PR does not: 

Look at other places where we block async tasks for cpu heavy work.  We should probably audit the code for other places where we might go a long time between awaits in our code.  Potentially signature aggregation, qc validation, and serialization.

It also doesn't make the VID disperse run on multiple threads.  As mentioned in the task we should not be dividing up the block on the consensus side.  The further parallelization will be done in jellyfish.

### Key places to review: 

vid.rs where we now spawn a task.  Note that the `unwrap` for the tokio version just passes on any panics that happen during the thread, so it is "safe", i.e it's not adding a new place we can panic.

This should improve the consistency of our CI since we should be able to run faster.  I think a key issue with our tests was that all nodes would block each other from running while the leader calculated the vid disperse and blocked the main executor.  In a single node setting it's probably not so bad since a leader would probably not have other work to do while it was blocked on vid calculation.

### DIFF
--- a/crates/task-impls/src/vid.rs
+++ b/crates/task-impls/src/vid.rs
@@ -2,8 +2,6 @@ use crate::events::HotShotEvent;
 use async_lock::RwLock;
 #[cfg(async_executor_impl = "async-std")]
 use async_std::task::spawn_blocking;
-#[cfg(async_executor_impl = "tokio")]
-use tokio::task::spawn_blocking;
 use hotshot_task::{
     event_stream::ChannelStream,
     global_registry::GlobalRegistry,
@@ -26,6 +24,8 @@ use hotshot_types::{
     data::{test_srs, VidScheme, VidSchemeTrait},
     traits::network::ConsensusIntentEvent,
 };
+#[cfg(async_executor_impl = "tokio")]
+use tokio::task::spawn_blocking;
 
 use hotshot_task::event_stream::EventStream;
 use snafu::Snafu;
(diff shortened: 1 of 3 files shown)


crates/task-impls/src/vid.rs
```diff
@@ -90,9 +94,14 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
                 let chunk_size = 1 << num_quorum_committee.ilog2();
 
                 // calculate vid shares
-                let vid = VidScheme::new(chunk_size, num_quorum_committee, &srs).unwrap();
-                let vid_disperse = vid.disperse(encoded_transactions.clone()).unwrap();
-
+                let vid_disperse = spawn_blocking(move || {
+                    let vid = VidScheme::new(chunk_size, num_quorum_committee, &srs).unwrap();
```
shenkeyao, line 98: Can we add the note below (copied from the PR description) as a comment here?
> Note that the unwrap for the tokio version just passes on any panics that happen during the thread, so it is "safe", i.e it's not adding a new place we can panic.

shenkeyao (approved):
LGTM, just a minor comment.

shenkeyao (approved):

PR #2388, opened 2024-01-11
[DOCS] - Document text editor usage

### LINKED ISSUE #2373: [DOCS] - Document text editor usage

When running RA an async executor must be chosen. Document how to do this in the README

Documentation will need to be added

_No response_

README includes documentation

_No response_

DieracDelta:
Closes #2373 

### This PR: 
Adds documentation for how to use with text editors

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,10 +200,8 @@ Choose an async runtime to use before launching a text editor. This may be done 
 ```
 export RUSTFLAGS='--cfg async_executor_impl="tokio" --cfg async_channel_impl="tokio"' # export RUSTFLAGS so the editor is aware of extra flags
 nvim # launch text editor of choice. We choose neovim in this example
-unset RUSTFLAGS # Unset rustflags so we may continue to use the justfile without confusing it
+unset RUSTFLAGS # Unset rustflags so we may continue to use the justfile. The justfile sets these particular config options
 ```
-
-Be sure t
 
 # Debugging
 


README.md
```diff
@@ -195,27 +195,11 @@ For espresso developers we have written up a description of our workflow [here](
...
-            {features =
-              loadfile(".nvim/env.lua") and { loadfile(".nvim/env.lua")() } or "all"
-            }
-          }
-      }
-    }
-  }
-)
+```
+export RUSTFLAGS='--cfg async_executor_impl="tokio" --cfg async_channel_impl="tokio"'
```
QuentinI, line 201: This will have unintended consequence of messing up `just` commands with the other executor ran from the same shell.
E.g. after
```sh
export RUSTFLAGS='--cfg async_executor_impl="tokio" --cfg async_channel_impl="tokio"'
nvim crates/hotshot/src/lib.rs
# enjoys rust-analyzer, exits nvim
just async_std_check # <-- fails, because just includes already-defined RUSTFLAGS
                     # when setting it's own and now we have executor set to both tokio and async_std
```

DieracDelta: Ah, good catch! I'll add a note about that to the readme.

README.md
```diff
@@ -195,28 +195,15 @@ For espresso developers we have written up a description of our workflow [here](
...
-    }
-  }
-)
 ```
+export RUSTFLAGS='--cfg async_executor_impl="tokio" --cfg async_channel_impl="tokio"' # export RUSTFLAGS so the editor is aware of extra flags
+nvim # launch text editor of choice. We choose neovim in this example
+unset RUSTFLAGS # Unset rustflags so we may continue to use the justfile without confusing it
+```
+
+Be sure t
```
QuentinI, line 206: Something got lost it seems

bfish713 (changes requested):
Seems like something got dropped

bfish713 (approved):

PR #2390, opened 2024-01-11
[TECH DEBT] Organize dependencies

QuentinI:
### This PR: 
Fixes a problem where `hotshot-signature-key` used `blake3` feature that it didn't specify in `Cargo.toml` and as a result wouldn't compile on its own.
Removes a bunch of unused dependencies.

### This PR does not: 
Change absolutely anything else, hopefully

### Key places to review: 
All changed `Cargo.toml`s

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -892,19 +892,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "asynchronous-codec"
-version = "0.7.0"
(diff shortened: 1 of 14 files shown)


QuentinI:
@shenkeyao there isn't actually, I just stumbled into the blake3 problem when experimenting with `cargo-semver-checks` and ran a quick scan for unused deps while I was at it.

PR #2360, opened 2024-01-09
[CI] Sequencer tests in CI

QuentinI:
Closes #2337

### This PR: 
Adds a workflow that builds and runs tests for sequencer with hotshot dependencies patched to the PR being tested

### This PR does not: 
Touch any other workflows.
Do anything clever to skip building in cases where API is obviously going to be broken.

### Key places to review: 
`.github/workflows/test-sequencer.yml`

### How to test this PR:
See CI. This PR fails the check on build stage, since sequencer doesn't build with current `main` due to API breakage.
A sample passing run:  #2361 (revert to hotshot version currently used in sequencer + this workflow)
A sample run with successful build, but failing tests: #2364 (revert to hotshot version currently used in sequencer + a sample commit breaking hotshot + this workflow)

PR #2391, opened 2024-01-11
[CI] Add a workflow to run `cargo-semver-checks`

QuentinI:
### This PR: 
Sort of a companion to #2360, adds a manually-triggered workflow to run `cargo-semver-checks` for all crates in the workspace.

### This PR does not: 
Touch other workflows or introduce any conditions on which this workflow will be automatically run.

### Key places to review: 
`.github/workflows/semver-check.yml`

### DIFF
--- a/.github/workflows/semver-check.yml
+++ b/.github/workflows/semver-check.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       matrix:
         async_runtimes:
-          - async-std
+          - async_std
           - tokio
     steps:
       - uses: actions/checkout@v4
@@ -32,6 +32,12 @@ jobs:
         with:
           path: baseline
           ref: ${{ inputs.baseline }}
+
+      - name: Install Just
+        run: |
+          wget https://github.com/casey/just/releases/download/1.14.0/just-1.14.0-x86_64-unknown-linux-musl.tar.gz
+          tar -vxf just-1.14.0-x86_64-unknown-linux-musl.tar.gz just
+          sudo cp just /usr/bin/just
 
       - name: Install Rust Stable
         uses: dtolnay/rust-toolchain@stable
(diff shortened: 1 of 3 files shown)


.github/workflows/semver-check.yml
```diff
@@ -0,0 +1,63 @@
...
+          cd current
+          export RUSTFLAGS='--cfg async_executor_impl="${{matrix.async_runtimes}}" --cfg async_channel_impl="${{matrix.async_runtimes}}"'
+          export RUSTDOCFLAGS="${RUSTFLAGS}"
+          # Checking each crate manually, because cargo-semver-checks
+          # exits early if it fails to check one of the crates
+          while IFS= read -r crate; do
+            cargo semver-checks \
+              --package "${crate}" \
+              --baseline-root ../baseline || true
+          done < <(cargo workspaces list)
```
DieracDelta, line 62: I think the rest of CI calls out to the justfile. Any chance we could keep that idiom here too? That is, we could add a semver-checks target to the justfile, and call that target in this CI pass.

This allows us to keep all our `cargo` commands in one place, and hides the internals from CI. This (in theory) makes it easier to make changes (e.g. add a feature flag or something), since we wouldn't need to grokk through a bunch of CI files to find the `cargo` invocations. Instead, we'd only need to change the justfile.

 WDYT?

QuentinI: Makes a lot of sense, I'll move the semver invocation to justfile and add semver-checks to our nix flake

QuentinI: Done
Apologies for copious force-pushes, can't test the workflow otherwise

elliedavidson (approved):
Such a nice addition to have!  Thank you!

Issue #2339, opened 2024-01-08
[VALIDATED_STATE] - Support custom validation


shenkeyao:
### What is this task and why do we need to work on it?

Add a hook for the application to do custom validation. 

### What work will need to be done to complete this task?

A state should take the previous and current block headers and use them for validation.

### What are the acceptance criteria to close this issue?

The sequencer can use this to validate the state.

PR #2394, opened 2024-01-12
[VALIDATED_STATE] - Add the parent header to state validation

shenkeyao:
Closes https://github.com/EspressoSystems/HotShot/issues/2339.

### This PR: 
* Adds the parent block header as an additional parameter to the state validation function.
* Combines the `validate_block` and `append` functions of `State` into `validate_and_apply_header`.
* Updates the `TestState` implementation.

### This PR does not: 
Change the logic of state update or retrieval in HotShot, which will be done in future PRs.

### Key places to review: 

`state_type.rs`.


crates/types/src/traits/state.rs
```diff
@@ -47,23 +46,23 @@ pub trait State:
...
+    fn validate_and_apply_header(
         &self,
-        block_header: &Self::BlockHeader,
+        proposed_header: &Self::BlockHeader,
+        parent_header: &Self::BlockHeader,
         view_number: &Self::Time,
     ) -> Result<Self, Self::Error>;
 
+    /// Initialize the state.
+    fn initialize() -> Self;
```
jbearer, line 64: This is when we're forced to initialize from scratch because of catchup, right? I think we should at least get a `&Self::BlockHeader` in this case, since we should always have a leaf or proposal or something for a view where we're trying to figure out the state. Right?

We need this in the sequencer so we can use [`from_commitment`](https://jellyfish.docs.espressosys.com/jf_primitives/merkle_tree/trait.ForgetableMerkleTreeScheme.html#tymethod.from_commitment) to build a completely sparse merkle tree from the merkle commitment contained in the header.

crates/testing/src/state_types.rs
```diff
@@ -60,36 +60,32 @@ impl State for TestState {
...
         Ok(TestState {
             block_height: self.block_height + 1,
             view_number: *view_number,
             prev_state_commitment: self.commit(),
         })
     }
 
+    fn initialize() -> Self {
+        let mut state = Self::default();
+        state.block_height += 1;
```
jbearer, line 85: This also seems problematic. We're basically taking a wild guess about the block height if we missed some blocks, but if we're wrong we won't be able to validate or vote on anything. This should take a header and get the correct block height using `BlockHeader::block_number()`

crates/types/src/traits/state.rs
shenkeyao: Oh, I only moved but didn't modify this function. Actually, let me delete this function for now since catchup is out of the scope of this PR. When I work on the state building for catchup, I'll add a function with additional parameter(s) as you commented here.

crates/testing/src/state_types.rs
shenkeyao: As mentioned in the other comment, I'll remove this function for now since it's unused.

jbearer (approved):

Issue #2374, opened 2024-01-10
[TECH DEBT] - Add `TryFrom<&TaggedBase64>` trait bound for signature keys


jbearer:
### What is this task and why do we need to work on it?

We are getting rid of `EncodedPublicKey`. Currently, the query service uses `EncodedPublicKey` (which implements `TryFrom` and `Into` `TaggedBase64`) to receive proposer IDs in URLs. We need to be able to do the same thing now with actual signature keys.

### What work will need to be done to complete this task?

Add `: TryFrom<TaggedBase64>` to whatever trait is implemented by public signature keys.

### What are the acceptance criteria to close this issue?

Query service should work without referring to `EncodedPublicKey` at all

Issue #2328, opened 2024-01-05
[TECH DEBT] - Remove the trait SignatureKey & remove key serialization


dailinsubjam:
### What is this task and why do we need to work on it?

- We do not need `SignatureKey` anymore, it's a wrap of jellyfish key scheme.
- We can remove serialization of public key so that the key size on-chain will be smaller.
- https://github.com/EspressoSystems/HotShot/issues/2327 is blocked by this.

### What work will need to be done to complete this task?

- [x] Remove EncodedPublicKey
~~Remove the trait SignatureKey~~
- [x] Remove the wrap of QCSignKey: BLSPriveKey
- [x] Remove the wrap of QCVerKey: BLSPubKey

### Are there any other details to include?

Followup on this https://github.com/EspressoSystems/HotShot/pull/2112.

### What are the acceptance criteria to close this issue?

Unblock https://github.com/EspressoSystems/HotShot/issues/2327.

mrain:
`SignatureKey` is still there.

dailinsubjam:
@mrain I think we can also close this if it 1. unblock the light client work and 2. reduce the serialization size. Since the motivation of removing `SignatureKey` is these two. Although the wrapping is still here, but as we see it also serves as a bridge so keep it is good. The key question here is that do we want to remove it in the future?

mrain:
Ahh right. Feel free to close it. We can ignore it even if it's kept open :)

Issue #2070, opened 2023-11-15
Transfer from `bit_vector_old` to `bit_vector`


dailinsubjam:
Currently in `hotshot` we're running the [old bit vector](https://github.com/EspressoSystems/HotShot/blob/develop/crates/hotshot-signature-key/src/bn254/bn254_pub.rs#L6) due to some merge conflicts, better transfer to the new one.
This might be a big merge and might need sub issues, since it's related to transfer from old `StakeTableEntry` to new defined `StakeTableScheme`.

Issue #2375, opened 2024-01-10
[TECH DEBT] - Replace `StakeTableScheme` trait bound with stake table snapshot


mrain:
### What is this task and why do we need to work on it?

The current `StakeTableScheme` trait defines a richer set of interfaces beyond HotShot needs. The consensus only needs to take a stake table snapshot from the sequencer and perform read-only query to it. There's no reason to bother with stake table updates and versioning stuff. 

We should replace the `StakeTableScheme` trait bounds in HotShot components, for example the `QuorumCertificateScheme` trait, with a `StakeTableSnapshot` trait or struct.

### What work will need to be done to complete this task?

- [ ] Define a `StakeTableSnapshot` trait or struct
- [ ] For other HotShot components, replace the `StakeTableScheme` trait bound with snapshot.

### What are the acceptance criteria to close this issue?

as described above

PR #2395, opened 2024-01-12
SignatureKey refactor

mrain:
Closes: #2374 
Partially Closes: #2328 

### This PR: 

- Add `TryFrom<TaggedBase64>` and `TryInto<TaggedBase64>` for `SignatureKey` trait.
- Remove the `EncodedPublicKey` type.
- Remove the wraps for BLS keys, we are implementing `SignatureKey` trait directly to the Jellyfish primitive types.
- Completely remove `hotshot-signature-key` crate, implementations are moved inside `types` crate
- Rename `QuorumCertificate` to `QuorumCertificateScheme` to avoid ambiguity
- QC implementation is moved to `types` crate. Because signature key implementation depends on QC impl, and `hotshot-qc` crate depends on `types` crate.
- Remove `hotshot-qc` crate from workspace. But the code is kept because some of them are not moved.

### This PR does not: 

Remove the `SignatureKey` trait.

### Key places to review: 

- Serialization of a `BLSPubKey` now only has 64 bytes.


crates/types/src/signature_key.rs
```diff
@@ -0,0 +1,132 @@
+//! Types and structs for the hotshot signature keys
+
+use crate::{
+    qc::{BitVectorQC, QCParams},
```
dailinsubjam, line 4: Originally we imported this from `hotshot_qc::bit_vector_old`, does this refactor also move things from `hotshot_qc/src/bit_vector.rs` to `types/src/qc.rs`? If so, this PR can also close https://github.com/EspressoSystems/HotShot/issues/2070.

dailinsubjam (approved):
Really clean, love this whole refactor!

crates/types/src/signature_key.rs
mrain: No. I moved `bit_vector_old.rs`. 
I think the current stake table scheme is richer than hotshot needs. All consensus need here is a stake table snapshot, see #2375 . So `bit_vector_old.rs` might be the best choice here.

mrain: `hotshot-qc` crate is no longer in the workspace. So the hotshot need nothing from `hotshot-qc` directory.

I'm just keeping those, because we may need the SNARKed QC implementation in the future.

dailinsubjam (approved):

PR #2397, opened 2024-01-12
[CHORE] Update jellyfish

jbearer:


crates/task-impls/src/vid.rs
```diff
@@ -122,9 +122,10 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
             }
 
             HotShotEvent::BlockReady(vid_disperse, view_number) => {
-                let Ok(signature) =
-                    TYPES::SignatureKey::sign(&self.private_key, &vid_disperse.payload_commitment)
-                else {
+                let Ok(signature) = TYPES::SignatureKey::sign(
+                    &self.private_key,
+                    vid_disperse.payload_commitment.as_ref().as_ref(),
```
ggutoski, line 127: `as_ref().as_ref()` yikes. wish that could be cleaner 😕

ggutoski (approved):

crates/task-impls/src/vid.rs
jbearer: Yeah this could be fixed Jellyfish side. Just add `AsRef<[u8]>` or `Into<Vec<u8>>` or something like that

ggutoski: new issue https://github.com/EspressoSystems/jellyfish/issues/471

PR #2399, opened 2024-01-13
fix: add metadata arg to BlockPayload::transaction_commitments

### LINKED ISSUE #2398: [API] - BlockPayload::transaction_commitments needs a metadata arg

We forgot to add an arg `metadata: &Self::Metadata` to the `transaction_commitments` method of the `BlockPayload` trait.

This new arg is needed downstream in `espresso-sequencer` to enable use of the updated `QueryPayload` trait from the `hotshot-query-service` repo. Example:

https://github.com/EspressoSystems/espresso-sequencer/blob/e177c1e2434193f3905e927a163afbb2aaf8f72c/sequencer/src/block.rs#L345-L347

The new `enumerate` call requires a `metadata` arg.

Add the arg to the trait method, then fix everything that breaks in this repo and downstream.

Also, there are a few downstream build warnings due to `hotshot-testing` feature flag that could be fixed here too.

_No response_

Nothing else.

_No response_

ggutoski:
Closes #2398 

### This PR: 

- Add `metadata` arg to `BlockPayload::transaction_commitments` and fix everything that breaks.
- Fix a few downstream build warnings due to `hotshot-testing` feature flag.

### This PR does not: 

...do anything else

### Key places to review: 

- Core API change in `block_contents.rs`
- All changes in `task-impls/` and `testing/` is an immediate consequence of this API change.
- `lib.rs` and `handle.rs`: fix downstream build warnings and tidy `use` statements.

### DIFF
--- a/crates/hotshot/src/lib.rs
+++ b/crates/hotshot/src/lib.rs
@@ -60,7 +60,7 @@ use hotshot_types::{
     traits::{
         consensus_api::ConsensusApi,
         network::{CommunicationChannel, NetworkError},
-        node_implementation::{ChannelMaps, NodeType, SendToTasks},
+        node_implementation::{NodeType, SendToTasks},
         signature_key::SignatureKey,
         state::ConsensusTime,
         storage::StoredView,
@@ -78,6 +78,10 @@ use std::{
 };
 use tasks::add_vid_task;
 use tracing::{debug, error, info, instrument, trace, warn};
+
+#[cfg(feature = "hotshot-testing")]
+use hotshot_types::traits::node_implementation::ChannelMaps;
+
 // -- Rexports
 // External
 /// Reexport rand crate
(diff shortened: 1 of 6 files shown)


ggutoski:
Forgot to mention: After this PR is merged, in order to use this fix downstream in espresso-sequencer we need:
- a new version tag 0.5.whatever
- update hotshot dependency of hotshot-query-service
- update hotshot, hotshot-query-service dependencies of espresso-sequencer

ggutoski:
cc @akonring fixed the final few build errors with some help from @nomaxg . This PR is the next step.

ggutoski:
libp2p test failure in CI. Everything passes locally. 🤷

crates/hotshot/src/lib.rs
```diff
@@ -256,6 +261,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> SystemContext<TYPES, I> {
 
         let inner: Arc<SystemContextInner<TYPES, I>> = Arc::new(SystemContextInner {
             id: nonce,
+            #[cfg(feature = "hotshot-testing")]
```
elliedavidson, line 264: @shenkeyao Can we make an issue to delete this field?  It looks like we don't use it anymore.

elliedavidson (approved):

crates/hotshot/src/lib.rs
ggutoski: While you're at it... there's another field in this struct called `_metrics` that also appears not to be in use. Maybe we could delete it, too. 🙏

shenkeyao: Sure! Issue created: https://github.com/EspressoSystems/HotShot/issues/2406.

PR #2408, opened 2024-01-15
[BUILDER] Add external events for builder

### LINKED ISSUE #2371: [BUILDER] - Add external events for builder

Builder should be able to consume events corresponding to `DAProposalRecv`, `TransactionRecv`, `QuorumProposalRecv` in order to track the state of the chain and build blocks.

_No response_

_No response_

Applications building on HotShot can consume listed events.
Tests pass.

_No response_

QuentinI:
Added external events for `Transaction`, `DAProposal` and `QuorumProposal`

Closes #2371 

### This PR: 
Adds external events and logic to emit those.

### This PR does not: 
Change consensus logic.

### Key places to review: 
All changed files.

### DIFF
--- a/crates/hotshot/src/lib.rs
+++ b/crates/hotshot/src/lib.rs
@@ -325,8 +325,8 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> SystemContext<TYPES, I> {
     /// Emit an external event
     // A copypasta of `ConsensusApi::send_event`
     // TODO: remove with https://github.com/EspressoSystems/HotShot/issues/2407
-    async fn send_event(&self, event: Event<TYPES>) {
-        debug!(?event, "send_event");
+    async fn send_external_event(&self, event: Event<TYPES>) {
+        debug!(?event, "send_external_event");
         let mut event_sender = self.inner.event_sender.write().await;
         if let Some(sender) = &*event_sender {
             if let Err(e) = sender.send_async(event).await {
@@ -372,7 +372,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> SystemContext<TYPES, I> {
                         da_membership,
                     ),
                 api
-                    .send_event(Event {
+                    .send_external_event(Event {
                         view_number: api.inner.consensus.read().await.cur_view,
                         event: EventType::Transactions {
                             transactions: vec![transaction],
--- a/crates/hotshot/src/lib.rs
+++ b/crates/hotshot/src/lib.rs
@@ -43,6 +43,7 @@ use async_lock::{RwLock, RwLockUpgradableReadGuard, RwLockWriteGuard};
 use async_trait::async_trait;
 use commit::Committable;
 use custom_debug::Debug;
+use futures::join;
 use hotshot_task::{
     event_stream::{ChannelStream, EventStream},
     task_launcher::TaskRunner,
@@ -53,6 +54,7 @@ use hotshot_types::{
     consensus::{Consensus, ConsensusMetricsValue, PayloadStore, View, ViewInner, ViewQueue},
     data::Leaf,
     error::StorageSnafu,
+    event::EventType,
     message::{
         DataMessage, InternalTrigger, Message, MessageKind, ProcessedGeneralConsensusMessage,
     },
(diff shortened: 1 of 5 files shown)


crates/hotshot/src/lib.rs
```diff
@@ -320,6 +322,20 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> SystemContext<TYPES, I> {
         };
     }
 
+    /// Emit an external event
+    // A copypasta of `ConsensusApi::send_event`
+    // TODO: remove with https://github.com/EspressoSystems/HotShot/issues/2407
+    async fn send_event(&self, event: Event<TYPES>) {
```
elliedavidson, line 328: Nit: Can we rename this to `send_external_event` just to differentiate it from sending an internal event.

crates/task-impls/src/consensus.rs
```diff
@@ -1199,12 +1210,22 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
+                    )),
+                self.api
+                    .send_event(Event {
+                        view_number: self.cur_view,
+                        event: EventType::QuorumProposal {
+                            proposal: message,
+                            sender: self.public_key.clone(),
+                        },
+                    })
+            };
```
elliedavidson, line 1228: Personally, I feel it is cleaner to just send the event when we receive the proposal.  If we send the external event here, the application will receive two events, one from here and one when it receives the proposal internally.

crates/task-impls/src/da.rs
```diff
@@ -299,12 +312,20 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
-                        message.clone(),
-                        self.public_key.clone(),
-                    ))
-                    .await;
+                join! {
+                    self.event_stream
+                        .publish(HotShotEvent::DAProposalSend(
+                            message.clone(),
+                            self.public_key.clone(),
+                        )),
```
elliedavidson, line 320: Same here

crates/task-impls/src/transactions.rs
```diff
@@ -92,28 +93,42 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
-                    })
-                    .await;
+                futures::join! {
+                    self.api
+                        .send_event(Event {
+                            view_number: self.cur_view,
+                            event: EventType::Transactions {
+                                transactions: transactions.clone(),
+                            },
+                        }),
```
elliedavidson, line 103: And here

elliedavidson (changes requested):

crates/task-impls/src/consensus.rs
QuentinI: Sorry, I'm not sure I follow. How does the application receive duplicate events in this case?

elliedavidson: The application will receive 2 `EventType::QuorumProposal` external events: 1 from when HotShot sends the proposal to the other nodes (which is the case above), and 1 from when HotShot receives the proposal (from itself).  So if a node is the proposer for a particular view, it will end up sending 2 of these events to the application.

QuentinI: Oh, nevermind. The sender receives the proposal as well, duh.

QuentinI: @elliedavidson right, sorry, GitHub didn't yet show me your comment when I was responding. Thank you for the explanation

crates/task-impls/src/transactions.rs
QuentinI: Okay, I think I remembered why I did this duplication. For transactions I feel it still makes sense, because if the sender isn't in DA committee, it won't receive the transaction back. That was the first case I implemented and just did the same for other events by inertia. I'll fix the other ones obviously, but for transactions it's justified I think.

elliedavidson: okay, this makes sense.  We should really fix the networking so that transactions go to everyone and not the DA committee, but we can do that after the new CDN is finished.

elliedavidson (approved):

PR #2412, opened 2024-01-16
check liveness checkeven if we cannot find the parent in storage

elliedavidson:
Closes #<ISSUE_NUMBER>

### This PR: 
Previously, if we received a proposal for which we didn't have the parent in storage, we would throw away the proposal and not vote.  This PR changes this behavior so that we still vote if we pass the liveness check. 

### This PR does not: 

Properly fix the libp2p clippy lint.  This is already fixed in main. 

### Key places to review: 
consensus.rs

### DIFF
--- a/crates/task-impls/src/consensus.rs
+++ b/crates/task-impls/src/consensus.rs
@@ -509,10 +509,8 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> + 
                         .cloned()
                 };
 
-                //
                 // Justify qc's leaf commitment is not the same as the parent's leaf commitment, but it should be (in this case)
                 let Some(parent) = parent else {
-                    // If no parent then just update our state map and return.  We will not vote.
                     error!(
                         "Proposal's parent missing from storage with commitment: {:?}",
                         justify_qc.get_data().leaf_commit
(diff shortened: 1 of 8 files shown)


shenkeyao (dismissed):

QuentinI:
[This comment](https://github.com/EspressoSystems/HotShot/pull/2412/files#diff-bd3299d85e179494b5d046442fd33a4ea027c838347f5ab393013dfee9e7d302R515) is now misleading

bfish713 (approved):
I'm wondering do we still need to update all the consensus state (high_qc, locked qc etc.) in the case we don't have the proposal from the QC

crates/task-impls/src/consensus.rs
```diff
@@ -539,14 +535,49 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
                     );
                     consensus.saved_leaves.insert(leaf.commit(), leaf.clone());
 
+                    // If we are missing the parent from storage, the safety check will fail.  But we can
+                    // still vote if the liveness check succeeds.
+                    let liveness_check = justify_qc.get_view_number() > consensus.locked_view;
+
+                    if justify_qc.get_view_number() > consensus.high_qc.view_number {
+                        consensus.high_qc = justify_qc;
+                    }
```
shenkeyao, line 544: Should we do this only if the liveness check passes?

elliedavidson: I was thinking about that too.  My thought was that we should do it even if the liveness check fails.  So long as the QC is valid, then it is a high QC (if this check passes).  We may not vote on this proposal if the liveness check fails, but I think we still want to store the high QC.

shenkeyao: Makes sense! Then I think we should also update the high QC in the case of having the parent leaf.

bfish713: Also I think if this qc has a higher view number than our high qc the liveness check must have passed because I don't think it's possible (or shouldn't be possible) for `locked_qc.view > high_qc.view`

crates/task-impls/src/consensus.rs
```diff
@@ -680,7 +712,6 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
                 };
 
                 // promote lock here to add proposal to statemap
-                let mut consensus = RwLockUpgradableReadGuard::upgrade(consensus).await;
```
shenkeyao, line 683: I can't comment on the unchanged line, but I think we could remove the `consensus.high_qc = high_qc` update since it's done in line 501 now?

bfish713 (approved):

shenkeyao (approved):

PR #2442, opened 2024-01-22
[VALIDATED_STATE] - Add app-specific data to validated state and store the state

### LINKED ISSUE #2344: [VALIDATED_STATE] - Allow date storage before the finalization

Add space for the application to store data before the finalization.

* Add application-specific type to the state map (probably inside the `view_inner`).
* For nodes that fall behind, if there's no parent state, support geting the state from the header.

_No response_

The sequencer can use the new type to store data.

_No response_

shenkeyao:
Closes #2344.

### This PR: 
- Stores validated state in the saved view.
- Implements getters for the last decided state.
- Updates the validated state trait.
  - Adds app-specific data to the validated state.
  - Adds functions to build the state from the header and get the app-specific data.
- Removes unused state-related code that might cause misuse.
   - The state getter from `Leaf`.
   - The state map from the safety task.
- Updates tests.

### This PR does not: 
Support state fetching for a given view, which will be done in the next PR.

### Key places to review: 
Major changes are in:
* crates/task-impls/src/consensus.rs
* crates/types/src/traits/state.rs

Changes in the files below need only a quick review. They are for removing the unused state map from the safety test task.
* crates/testing/src/test_builder.rs
* crates/testing/src/overall_safety_task.rs


crates/task-impls/src/consensus.rs
```diff
@@ -590,6 +591,14 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
                 };
                 let parent_commitment = parent.commit();
+                let Ok(state) = consensus.get_decided_state().validate_and_apply_header(
+                    &proposal.data.block_header.clone(),
+                    &parent.block_header.clone(),
+                    &view,
+                ) else {
+                    error!("Block header doesn't extend the proposal",);
+                    return;
+                };
```
elliedavidson, line 601: Why do we validate against the last decided state instead of the parent state?

shenkeyao: Good catch! I mistakenly used the last decided state in a few places and I thought I had fixed all of them. Apparently, I missed this one here.

bfish713 (approved):
This looks good to me based on the discussions in Zulip

PR #2446, opened 2024-01-23
[FIX_CI] Fix test issues

### LINKED ISSUE #2211: [FIX_CI] - Investigate and fix networking-related CI failures

We have nondeterministic test failures for networking-related tests in CI.

Specifically, `test_combined_network` has a problem where all tests try to bind to the same port. This is a bigger problem on the self-hosted runner, which does not virtualize networking like the GitHub runners do.

The other failure is with the DHT in `test_timeout_libp2p`. I think that the test parameters have to be massaged, barring a rewrite/deep dive.

_No response_

_No response_

- [ ] all networking related tests pass deterministically

_No response_

### LINKED ISSUE #2441: [FIX_CI] - Investigate test timeouts

Twice we've seen the self-hosted runner time out while running the tests over the `async-std` executor. 
One looks like it was making slow "progress" and the other looks like it didn't log anything for almost the entire hour.

Both times we encounter the behavior during the catchup tests. They pass fine on Tokio, and as well as with the `async-std` executor and the standard GitHub runner. It is _possible_ it's a runner issue. 

This ticket is to investigate and resolve this issue.

_No response_

_No response_

- [ ] all tests pass deterministically on this configuration

_No response_

rob-maron:
Closes #2441 
Closes #2211 

### This PR: 

Makes a few improvements to reduce the probability of a test failing or hanging.
Namely:
- Adds a way for the late start nodes to force stop. Without it, nodes started late continue running past when they're supposed to be killed. This seems to cause a lot of hanging tests, and could also cause an OOM eventually.

- Changes timings for the catchup tests, like `catchup_test_in_view_sync` to make sure that we more accurately depict what the test should be testing.  

- Fixes the spinning task so that we don't miss spinning up or down any nodes. Before this, if nodes try to start or stop during a timeout or view sync, we don't receive the underlying event. This means the action wouldn't actually execute

- Prevents transaction task deadlock by timing out and using non-blocking operations. This prevents tests from hanging when using the `async_std` executor.

- Tunes some libp2p-specific parameters and re-introduce direct message retries. The "retry forever" issue still exists but that is to be dealt with in a separate issue.

It also switches back the `justfile` to use the debug release profile, which lets us iterate faster. Without it, build times average ~15m instead of 5.

### This PR does not: 

A few improvements I want to make to the testbed, although not sure how time consuming:
- The timing issues can stem from the fact that while we execute spinning events on a specific view, it can sometimes take longer to actually execute on that spin. I wonder if we could get it to a middleware-like layer where all tasks have to finish their operations for a particular view before nodes are allowed to move on. This would make it much more deterministic.

- It does not fix all tests everywhere (like fix the libp2p test always). I'm going to get this one in as an incremental improvement and see where to go from there.

### Key places to review: 

- Want to make sure the `catchup` tests are still configured properly - cc @bfish713

### DIFF
--- a/crates/hotshot/src/lib.rs
+++ b/crates/hotshot/src/lib.rs
@@ -386,8 +386,8 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> SystemContext<TYPES, I> {
         self.inner.consensus.read().await.get_decided_leaf()
     }
 
-    /// Synchronously tries to return a copy of the last decided leaf,
-    /// if it is currently available for reading.
+    /// Instantly returns a copy of the last decided leaf if it is currently 
+    /// available for reading. If not, we will return.
     ///
     /// # Panics
     /// Panics if internal state for consensus is inconsistent
--- a/crates/hotshot/src/lib.rs
+++ b/crates/hotshot/src/lib.rs
@@ -399,6 +399,19 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> SystemContext<TYPES, I> {
         self.inner.consensus.read().await.get_decided_leaf()
     }
 
+    /// Synchronously tries to return a copy of the last decided leaf,
+    /// if it is currently available for reading.
+    ///
+    /// # Panics
+    /// Panics if internal state for consensus is inconsistent
+    #[must_use]
+    pub fn try_get_decided_leaf(&self) -> Option<Leaf<TYPES>> {
+        self.inner
+            .consensus
+            .try_read()
+            .map(|guard| guard.get_decided_leaf())
+    }
+
     /// Initializes a new hotshot and does the work of setting up all the background tasks
     ///
     /// Assumes networking implementation is already primed.
--- a/crates/hotshot/src/lib.rs
+++ b/crates/hotshot/src/lib.rs
@@ -386,8 +386,8 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> SystemContext<TYPES, I> {
         self.inner.consensus.read().await.get_decided_leaf()
     }
 
-    /// Instantly returns a copy of the last decided leaf if it is currently 
-    /// available for reading. If not, we will return.
+    /// [Non-blocking] instantly returns a copy of the last decided leaf if
+    /// it is available to be read. If not, we return `None`.
     ///
     /// # Panics
     /// Panics if internal state for consensus is inconsistent
--- a/crates/hotshot/src/tasks/mod.rs
+++ b/crates/hotshot/src/tasks/mod.rs
@@ -11,6 +11,7 @@ use hotshot_task::{
     task_launcher::TaskRunner,
     GeneratedStream, Merge,
 };
+use tracing::error;
 use hotshot_task_impls::{
     consensus::{
         consensus_event_filter, CommitmentAndMetadata, ConsensusTaskState, ConsensusTaskTypes,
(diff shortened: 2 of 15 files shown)


crates/testing/tests/catchup.rs
```diff
@@ -147,18 +153,23 @@ async fn test_catchup_one_node() {
...
+            updown: UpDown::Up,
+        },
+        ChangeNode {
+            idx: 19,
+            updown: UpDown::Up,
+        },
+    ];
     metadata.timing_data = timing_data;
-    metadata.start_nodes = 19;
+    metadata.start_nodes = 18;
```
bfish713, line 167: This test should just remain being one node that starts down

crates/hotshot/src/tasks/mod.rs
```diff
@@ -67,12 +68,16 @@ pub async fn add_network_message_task<TYPES: NodeType, NET: CommunicationChannel
...
             loop {
-                let msgs = Messages(
-                    network
-                        .recv_msgs(TransmitType::Broadcast)
-                        .await
-                        .expect("Failed to receive broadcast messages"),
-                );
+                let msgs = match network.recv_msgs(TransmitType::Broadcast).await {
+                    Ok(msgs) => Messages(msgs),
+                    Err(err) => {
```
bfish713, line 73: Were we getting this error? Seems like this shouldn't happen on the receive side

crates/hotshot/src/tasks/mod.rs
```diff
@@ -87,12 +92,15 @@ pub async fn add_network_message_task<TYPES: NodeType, NET: CommunicationChannel
...
-                let msgs = Messages(
-                    network
-                        .recv_msgs(TransmitType::Direct)
-                        .await
-                        .expect("Failed to receive direct messages"),
-                );
+                let msgs = match network.recv_msgs(TransmitType::Direct).await {
+                    Ok(msgs) => Messages(msgs),
+                    Err(err) => {
+                        error!("failed to receive broadcast messages: {err}");
```
bfish713, line 98: nit: broadcast should be direct

Also same question, do we hit this case sometimes?

bfish713 (changes requested):
Changes look good, Just need to fix the single node catch up test.  Also can we ignore the libp2p tests while you work on figuring out the issue with them?  I don't think anyone is working anything libp2p related and the test flaky enough we are ignoring it failing anyway

crates/hotshot/src/tasks/mod.rs
rob-maron: We were, but I believe it was from the nodes that were suppoed to be killed. This is actually fixed with the other changes I made, but I still don't think we should be panicking - WDYT?

bfish713: I think it's a good change, just would be quite surprised if we hit this outside of shutdown

crates/hotshot/src/lib.rs
```diff
@@ -386,6 +386,19 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> SystemContext<TYPES, I> {
         self.inner.consensus.read().await.get_decided_leaf()
     }
 
+    /// Synchronously tries to return a copy of the last decided leaf,
```
bfish713, line 389: nit: When I see synchronously I think this is blocking maybe update the comment to say it won't block and fail if it can't get the lock without waiting

bfish713 (dismissed):

bfish713 (approved):

shenkeyao (approved):
Looks good!

PR #2449, opened 2024-01-23
Indexes Payload store by view number

### LINKED ISSUE #2448: [BUG_FIX] - Index Payload Map by View number

We've seen bugs where many nodes aren't getting the payload for a decide event when they should.  This is probably due to the fact that the payload map is ref counted by payload commitment and we get tons of empty block commitments.  

_No response_

_No response_

Payload is indexed by view and payload commitment

_No response_

bfish713:
Closes #2448 

### This PR: 

Stops using our custom `PayloadStore` and just uses BTreeMap.  This means we may have duplicate Payloads in the store, but the store will remain quite small so should not be an issue.  We use `split_off` for GCing the map just like the state map.  This gets rid of all the payloads prior to decide, should be a straight forward way to GC.

### This PR does not: 

This is not the PR which fixes this for the release branch.  Review that one first

### Key places to review: 

Is garbage collection and intialization correct? Am I always indexing with the correct view number

### DIFF
--- a/crates/types/src/consensus.rs
+++ b/crates/types/src/consensus.rs
@@ -7,7 +7,7 @@ pub use crate::{
 use displaydoc::Display;
 
 use crate::{
-    data::{Leaf, VidCommitment},
+    data::Leaf,
     error::HotShotError,
     simple_certificate::{DACertificate, QuorumCertificate},
     traits::{
@@ -17,9 +17,9 @@ use crate::{
     utils::Terminator,
 };
 use commit::Commitment;
-use derivative::Derivative;
+
 use std::{
-    collections::{hash_map::Entry, BTreeMap, HashMap},
+    collections::{BTreeMap, HashMap},
     sync::{Arc, Mutex},
 };
 use tracing::error;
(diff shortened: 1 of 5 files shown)


crates/types/src/consensus.rs
```diff
@@ -55,7 +55,7 @@ pub struct Consensus<TYPES: NodeType> {
     ///
     /// Contains the block payload commitment and encoded transactions for every leaf in
     /// `saved_leaves` if that payload is available.
-    pub saved_payloads: PayloadStore,
+    pub saved_payloads: BTreeMap<TYPES::Time, Vec<u8>>,
```
shenkeyao, line 58: Same as the other PR. Let's update the comment.

shenkeyao (dismissed):
Just a nit comment.

shenkeyao (approved):

jbearer (commented):
Does this branch also need the genesis special case?

jbearer (approved):

jbearer (approved):

PR #2450, opened 2024-01-23
HOT-FIX: Index Payload Store by View

bfish713:
Closes #2449 

### This PR: 

Stops using our custom PayloadStore and just uses BTreeMap. This means we may have duplicate Payloads in the store, but the store will remain quite small so should not be an issue. We use split_off for GCing the map just like the state map. This gets rid of all the payloads prior to decide, should be a straight forward way to GC.

### Key places to review: 

Is garbage collection and intialization correct? Am I always indexing with the correct view number

### DIFF
--- a/crates/types/src/consensus.rs
+++ b/crates/types/src/consensus.rs
@@ -7,7 +7,7 @@ pub use crate::{
 use displaydoc::Display;
 
 use crate::{
-    data::{Leaf, VidCommitment},
+    data::Leaf,
     error::HotShotError,
     simple_certificate::{DACertificate, QuorumCertificate},
     traits::{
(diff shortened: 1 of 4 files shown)


elliedavidson (approved):

shenkeyao (approved):

jbearer (approved):
LGTM. Way simpler!

jbearer:
Oh, the whole payload store implementation is dead code now, right? Can we remove it?

crates/types/src/consensus.rs
```diff
@@ -55,7 +55,7 @@ pub struct Consensus<TYPES: NodeType> {
     ///
     /// Contains the block payload commitment and encoded transactions for every leaf in
     /// `saved_leaves` if that payload is available.
-    pub saved_payloads: PayloadStore,
+    pub saved_payloads: BTreeMap<TYPES::Time, Vec<u8>>,
```
shenkeyao, line 58: Nit: update the comment.

shenkeyao (approved):
LGTM. Depending on the urgency, maybe we could remove the payload store in the other PR as @jbearer commented, and merge this one just as is?

bfish713:
> LGTM. Depending on the urgency, maybe we could remove the payload store in the other PR as @jbearer commented, and merge this one just as is?

I just removed it in both since it was a pretty simple deletion

Issue #2444, opened 2024-01-23
[Upgradability] - Implement logic to detect if the network has been upgraded


ss-es:
### What is this task and why do we need to work on it?

A node should only start using the latest version of the protocol after the network has been upgraded. We should implement logic that detects if this has happened, and switches to the latest protocol version if so.

### What work will need to be done to complete this task?

- [ ] Create a global mutable variable that is updated when an upgrade certificate is encountered.
- [ ] Incoming messages with higher version numbers should be discarded.

### What are the acceptance criteria to close this issue?

Subtasks should be completed. Designing tests is likely to be complicated and should be a separate task.

PR #2451, opened 2024-01-23
Add version handling for incoming messages

### LINKED ISSUE #2409: [Upgradability] - Implement version handling for incoming messages

We should scan incoming messages for a version field (before even deserializing them), in order to be able to handle them appropriately (according to different consensus protocols).

_No response_

_No response_

Every entry-point for inbound messages into HotShot must have a version field check that occurs *before* serialization is attempted.

_No response_

ss-es:
Closes #2409

### This PR: 
Adds version handling for incoming messages. 

The goal is that incoming binary messages should have their version number read before the message is deserialized, and then put in an appropriate queue based on the version number.

### Key places to review:
The logic in `hotshot/src/traits/networking/libp2p_network.rs` and `/web_server_network.rs` should be looked at closely, to make sure it hasn't changed.

In particular,  `web_server_network::poll_web_server` may return early depending on the message type it receives from the web server, but the way this works was slightly rewritten (and should be checked carefully).


crates/utils/src/version.rs
```diff
@@ -0,0 +1,21 @@
...
+
+use hotshot_constants::Version;
+
+/// Read the version number from a message (passed a byte vector),
+/// returning None if there are not enough bytes.
+#[must_use]
+#[allow(clippy::module_name_repetitions)]
+pub fn read_version(message: &[u8]) -> Option<Version> {
+    let maybe_bytes_major: Result<[u8; 2], _> = message[0..2].try_into();
+    let maybe_bytes_minor: Result<[u8; 2], _> = message[2..4].try_into();
```
shenkeyao, line 11: Looks like this will panic if the length of `message` is less than 4. I think we can do a length check before this.

crates/hotshot/src/traits/networking/libp2p_network.rs
```diff
@@ -528,49 +589,39 @@ impl<M: NetworkMsg, K: SignatureKey + 'static> Libp2pNetwork<M, K> {
...
-                        if handle
-                            .inner
-                            .handle
-                            .direct_response(chan, &Empty::Empty)
-                            .await
-                            .is_err()
-                        {
-                            error!("failed to ack!");
-                        };
+                    NetworkEvent::IsBootstrapped => Some(VERSION_0_1),
```
shenkeyao, line 597: Should we add the version number to the `NetworkEvent::IsBootstrapped` variant, as we do for other variants? Otherwise, it looks like we'll handle other message types differently according to the version number but always use `0.1` for `IsBootstrapped`.

crates/hotshot/src/traits/networking/web_server_network.rs
```diff
@@ -192,9 +194,9 @@ struct Inner<TYPES: NodeType> {
     /// Our own key
     _own_key: TYPES::SignatureKey,
     /// Queue for broadcasted messages
-    broadcast_poll_queue: Arc<RwLock<Vec<RecvMsg<Message<TYPES>>>>>,
+    broadcast_poll_queue_0_1: Arc<RwLock<Vec<RecvMsg<Message<TYPES>>>>>,
     /// Queue for direct messages
-    direct_poll_queue: Arc<RwLock<Vec<RecvMsg<Message<TYPES>>>>>,
+    direct_poll_queue_0_1: Arc<RwLock<Vec<RecvMsg<Message<TYPES>>>>>,
```
shenkeyao, line 199: If we are going to use `Inner` for all versions, I think these should have names without `_0_1`. Otherwise, the struct should be `Inner_0_1`. If we need a way to distinguish version numbers, we could maybe add `Version` as a generic type.

crates/hotshot/src/traits/networking/web_server_network.rs
```diff
@@ -231,6 +233,182 @@ struct Inner<TYPES: NodeType> {
 
 impl<TYPES: NodeType> Inner<TYPES> {
     #![allow(clippy::too_many_lines)]
+
+    /// Handle version 0.1 transactions
+    async fn handle_tx_0_1(&self, tx: Vec<u8>, index: u64, tx_index: &mut u64) {
```
shenkeyao, line 238: Could you add a comment for the `index` and `tx_index` arguments? Not so easy to guess what the difference is. 🤔

crates/hotshot/src/traits/networking/web_server_network.rs
```diff
@@ -231,6 +233,182 @@ struct Inner<TYPES: NodeType> {
...
+        let broadcast_poll_queue = &self.broadcast_poll_queue_0_1;
+        if index > *tx_index + 1 {
+            debug!("missed txns from {} to {}", *tx_index + 1, index - 1);
+            *tx_index = index - 1;
+        }
+
+        *tx_index += 1;
```
shenkeyao, line 245: Are we assuming `index > *tx_index`? If so, is this equivalent to assigning `*tx_index = index`?

crates/utils/src/version.rs
ss-es: Oops, yes! I missed that taking the slice can panic.

I rewrote this with `get` instead, and added some unit tests to make sure it really does return `None` (and doesn't panic) if the message is too short.

crates/hotshot/src/traits/networking/web_server_network.rs
ss-es: Added!

I think `index` is supposed to be the first transaction returned by the server, and `tx_index` is the last transaction read.

(But let me know if this seems wrong! I'm mostly guessing from the context where they were used.)

crates/hotshot/src/traits/networking/web_server_network.rs
ss-es: From my understanding, I don't think it's the same thing. 

In the [original code](https://github.com/EspressoSystems/HotShot/blob/c953c1ee546ad2ad9b89f164c2946b04d48b1395/crates/hotshot/src/traits/networking/web_server_network.rs#L278-L289), fetching transactions from the server returns something like `(index: u64, txs: Vec<Vec<u8>>)` (originally the `txs` was also deserialized in this step, but now that happens later).

I believe the `index` is just for the very first `tx` in the block of `txs` you get from the server, and the following transactions are processed as consecutive. So `index` itself will be the same for all of them, but you increment `tx_index` as you process each transaction.

crates/hotshot/src/traits/networking/web_server_network.rs
ss-es: My thought here was that we should separate messages into different queues depending on the version number, but `Inner` would contain shared state needed between the different version numbers.

But I think I also don't fully understand where e.g. `broadcast_poll_queue` is consumed. I'll bring this up during pair programming tomorrow!

crates/hotshot/src/traits/networking/libp2p_network.rs
ss-es: I think `IsBootstrapped` is just an internal signal, so I'm not sure about adding a version number (but as you point out, it's just as strange to give it a version number here).

I'll try to reorganize to handle this event separately, without assigning it a version number? Since the only thing we do (as far as I can tell) is set `is_bootstrapped` in `Inner`.

justfile
```diff
@@ -7,7 +7,7 @@ original_rustdocflags := env_var_or_default('RUSTDOCFLAGS', '--cfg hotshot_examp
 
 run_ci: lint build test
 
-async := "async_std"
```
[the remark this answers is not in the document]
ss-es: This is a variable used in the `cargo` subcommand (line 17), as e.g. `just async=async-std cargo <anything>` or `just async=tokio cargo <anything>`.

crates/hotshot/src/traits/networking/libp2p_network.rs
```diff
@@ -528,49 +588,36 @@ impl<M: NetworkMsg, K: SignatureKey + 'static> Libp2pNetwork<M, K> {
...
-                    }
+            while let Ok(message) = handle.inner.handle.receiver().recv().await {
+                match &message {
                     NetworkEvent::IsBootstrapped => {
                         is_bootstrapped.store(true, Ordering::Relaxed);
                     }
+                    GossipMsg(raw, _) | DirectRequest(raw, _, _) | DirectResponse(raw, _) => {
+                        let message_version = read_version(raw);
+                        match message_version {
+                            Some(VERSION_0_1) => {
```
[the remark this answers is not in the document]
ss-es: Currently, outgoing messages always use `VERSION_0_1` (defined in `constants`) -- I think the current idea is that a single binary will only ever send out messages in a single version.

I think this will come up later in https://github.com/EspressoSystems/HotShot/issues/2444; once we've seen a committed upgrade, the version number on outgoing messages might change (and depending on what we upgrade, we might start sending out new types of messages etc).

elliedavidson (commented):
I'm not going to approve because I only reviewed the web server part, but that part looks good to me!  Let me know if you want me to review the rest :)

crates/hotshot/src/traits/networking/libp2p_network.rs
```diff
@@ -528,49 +588,36 @@ impl<M: NetworkMsg, K: SignatureKey + 'static> Libp2pNetwork<M, K> {
...
-                            .context(FailedToSerializeSnafu);
-                    }
+            while let Ok(message) = handle.inner.handle.receiver().recv().await {
+                match &message {
                     NetworkEvent::IsBootstrapped => {
                         is_bootstrapped.store(true, Ordering::Relaxed);
                     }
+                    GossipMsg(raw, _) | DirectRequest(raw, _, _) | DirectResponse(raw, _) => {
+                        let message_version = read_version(raw);
+                        match message_version {
```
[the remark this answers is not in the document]
ss-es: I'm not sure I completely follow, but I think (at least in the near future) there's probably not much savings here; we're likely to only ever handle two versions at a time, and I don't think we'd have to replicate any of these matches anywhere else.

(But I think I'd probably like to come back to this once I have a clearer picture of what exactly might be versioned in HotShot!)

crates/hotshot/src/traits/networking/libp2p_network.rs
```diff
@@ -518,6 +524,60 @@ impl<M: NetworkMsg, K: SignatureKey + 'static> Libp2pNetwork<M, K> {
             .map_err(Into::<NetworkError>::into)
     }
 
+    /// Handle events for Version 0.1 of the protocol.
+    async fn spawn_events_0_1(
```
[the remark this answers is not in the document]
ss-es: Changed!

crates/hotshot/src/traits/networking/libp2p_network.rs
```diff
@@ -518,6 +524,60 @@ impl<M: NetworkMsg, K: SignatureKey + 'static> Libp2pNetwork<M, K> {
...
+                    error!("failed to ack!");
+                };
+            }
+            DirectResponse(msg, _) => {
+                let _result: Result<M, _> = bincode_opts()
+                    .deserialize(&msg)
+                    .context(FailedToSerializeSnafu);
+            }
+            NetworkEvent::IsBootstrapped => {
+                error!("spawn_events_0_1 received `NetworkEvent::IsBootstrapped`, which should be impossible.");
```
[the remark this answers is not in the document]
ss-es: Changed!

crates/hotshot/src/traits/networking/libp2p_network.rs
```diff
@@ -517,59 +523,100 @@ impl<M: NetworkMsg, K: SignatureKey + 'static> Libp2pNetwork<M, K> {
...
+                    error!("failed to ack!");
+                };
+            }
+            DirectResponse(msg, _) => {
+                let _result: Result<M, _> = bincode_opts()
+                    .deserialize(&msg)
+                    .context(FailedToSerializeSnafu);
+            }
+            NetworkEvent::IsBootstrapped => {
+                error!("handle_recvd_events_0_1 received `NetworkEvent::IsBootstrapped`, which should be impossible.");
```
shenkeyao, line 574: 👍

crates/hotshot/src/traits/networking/libp2p_network.rs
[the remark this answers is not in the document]
shenkeyao: Was thinking the same but then realized we would only need to handle two versions as Salman said, so it's probably fine as is. Restricting to two versions also helps prevent us from forgetting to update the version numbers here.

crates/hotshot/src/traits/networking/web_server_network.rs
shenkeyao: As discussed in the last pair programming, this makes sense!

crates/testing/tests/unit/version.rs
shenkeyao, line 1: Nice that we have these unit tests!

shenkeyao (commented):
There are merge conflicts but this looks good!

crates/hotshot/src/traits/networking/web_server_network.rs
shenkeyao: Makes sense. Thanks for the code comment and the explanation here!

shenkeyao (approved):

Issue #2463, opened 2024-01-25
[VALIDATED_STATE] - Update block header construction and the state


shenkeyao:
### What is this task and why do we need to work on it?

This is a follow-up task for https://github.com/EspressoSystems/HotShot/issues/2344.
* The block header should take the validated state, so the application can pull the necessary data (e.g., the state commitment, the root hash of the block Merkle tree, etc.) and add it to the header.
* Misc: The metadata getter of the validated state is unused.

### What work will need to be done to complete this task?

* Add `&State` to `BlockHeader::new` and update callers and test implementations.
* Remove `metadata()` from `State`.

### What are the acceptance criteria to close this issue?

The sequencer can use the updated header constructor to store state-related info.

PR #2464, opened 2024-01-25
[VALIDATED_STATE] - Add state to header and remove metadata getter

shenkeyao:
Closes https://github.com/EspressoSystems/HotShot/issues/2463.

### This PR: 
* Adds the parent state to the `BlockHeader` construction.
* Removes the metadata getter function from the `State` trait.

### Key places to review: 
* crates/testing/src/block_types.rs

### DIFF
--- a/crates/task-impls/src/consensus.rs
+++ b/crates/task-impls/src/consensus.rs
@@ -1174,7 +1174,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> + 
             return false;
         };
         // Leaf hash in view inner does not match high qc hash - Why?
-        let Some(leaf_commitment) = parent_view.get_leaf_commitment() else {
+        let Some((leaf_commitment, state)) = parent_view.get_leaf() else {
             error!(
                 ?parent_view_number,
                 ?parent_view,
@@ -1223,6 +1223,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> + 
                 commit_and_metadata.commitment,
                 commit_and_metadata.metadata.clone(),
                 &parent_header,
+                state,
             );
             let leaf = Leaf {
                 view_number: view,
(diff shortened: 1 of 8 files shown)


jbearer (approved):
I think this is sufficient.

I did realize that since `Header::new` gets the parent state, not the new state, we will end up calculating the new state twice in the leader node. `Header::new` will effectively have to duplicate some of the `validate_and_apply` logic in order to include the new state commitment in the header. Then the consensus task within the leader node will end up doing the full `validate_and_apply` to check that the state commitment in the proposed header is consistent with applying that proposal to the parent state. This could maybe be optimized by passing the computed state from the leader task to the replica task, but probably not important for now

shenkeyao:
> I did realize that since `Header::new` gets the parent state, not the new state, we will end up calculating the new state twice in the leader node. `Header::new` will effectively have to duplicate some of the `validate_and_apply` logic in order to include the new state commitment in the header. Then the consensus task within the leader node will end up doing the full `validate_and_apply` to check that the state commitment in the proposed header is consistent with applying that proposal to the parent state. This could maybe be optimized by passing the computed state from the leader task to the replica task, but probably not important for now

Makes sense! I made an issue for this: https://github.com/EspressoSystems/HotShot/issues/2465. I labeled it as low priority and could work on it after the major tasks for the validated state are done.

tbro:
I noticed you removed the `fn metadata()`. Should the associated type `Metadata` on the `impl State for TestState` in `state_types.rs` (line 63) also be removed?

shenkeyao:
> I noticed you removed the `fn metadata()`. Should the associated type `Metadata` on the `State` trait and implementation for `TestState ` also be removed?

I think the `Metadata` type is still useful for the sequencer to pass in the app-specific data. It’s an unknown type for HotShot so we need it to be an associated type. We could remove the `metadata` function because we don’t need to retrieve it in HotShot.

jbearer:
Wait but the `Metadata` type is no longer used anywhere in the `State` trait. Wouldn't we just refer to `Header::Metadata` when we need to name this type elsewhere in HotShot?

jbearer (approved):

PR #2469, opened 2024-01-28
[TEST] - Add macro tests

### LINKED ISSUE #2433: [TEST] - Add tests to testing macro

Run every test in the test_basic, test_timeout, and test_catchup targets using the testing-macros crate.

Port over configurations from hotshot-testing

_No response_

Tests are added and tests pass.

_No response_

DieracDelta:
Closes #2433 

### This PR: 
Moves the integration tests in `test_basic` to use the `cross_tests` macro.

### This PR does not: 
Move the unit tests.

### Key places to review: 
It's worth confirming that this is the intended usage (e.g. the tests are cross product-ing the right types). Note that for `test_success`, I disabled libp2p and the combined network since there aren't enough bootstrap nodes.

### DIFF
--- a/crates/hotshot/src/traits/networking/libp2p_network.rs
+++ b/crates/hotshot/src/traits/networking/libp2p_network.rs
@@ -264,7 +264,7 @@ where
                     {
                         Ok(network) => network,
                         Err(err) => {
-                            panic!("Failed to create network: {err:?}");
+                            panic!("Failed to create libp2p network: {err:?}");
                         }
                     }
                 })


crates/hotshot/src/traits/networking/libp2p_network.rs
```diff
@@ -264,7 +264,7 @@ where
                     {
                         Ok(network) => network,
                         Err(err) => {
-                            panic!("Failed to create network: {err}");
+                            panic!("Failed to create network: {err:?}");
```
elliedavidson, line 267: Nit: Can we make this "Failed to create libp2p network" - just to make it easier to understand if it shows up :)

elliedavidson: LGTM otherwise!

bfish713 (approved):

PR #2470, opened 2024-01-28
[HOTFIX] Check Relay Number when Accumulating View Sync Votes

bfish713:
### This PR: 

Keys the relay maps by relay number as well.  This makes it so that relays will accumulate votes separately for each time they are the relay.  

It also changes the vote accumulator to map vote commitment to the signers (sig_list and signers bitvec).  This mirrors how we map vote commitment to the vote outcomes and should make it impossible to form an invalid cert by accident.

### This PR does not: 

No garbage collection for old relay rounds (they will be clean up when the view moves on)

### Key places to review: 

In view_sync.rs make sure the new mapping makes sense.
In the vote accumulator, check the new mapping as well

### DIFF
--- a/crates/task-impls/src/view_sync.rs
+++ b/crates/task-impls/src/view_sync.rs
@@ -41,7 +41,7 @@ use hotshot_types::{
     },
 };
 use snafu::Snafu;
-use std::{collections::HashMap, fmt::Debug, sync::Arc, time::Duration};
+use std::{collections::BTreeMap, collections::HashMap, fmt::Debug, sync::Arc, time::Duration};
 #[cfg(async_executor_impl = "tokio")]
 use tokio::task::JoinHandle;
 use tracing::{debug, error, info, instrument, warn};
@@ -62,9 +62,9 @@ pub enum ViewSyncPhase {
 /// Stub of a view sync error
 pub struct ViewSyncTaskError {}
 
-/// Type alias for a map from View Number to Vote Task
+/// Type alias for a map from View Number to Relay to Vote Task
 type RelayMap<TYPES, VOTE, CERT> =
-    HashMap<<TYPES as NodeType>::Time, VoteCollectionTaskState<TYPES, VOTE, CERT>>;
+    HashMap<<TYPES as NodeType>::Time, BTreeMap<u64, VoteCollectionTaskState<TYPES, VOTE, CERT>>>;
 
 /// Main view sync task state
 pub struct ViewSyncTaskState<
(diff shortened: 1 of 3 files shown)


crates/types/src/vote.rs
```diff
@@ -85,9 +85,12 @@ pub struct VoteAccumulator<
...
     pub vote_outcomes: VoteMap2<Commitment<VOTE::Commitment>>,
     /// A list of valid signatures for certificate aggregation
-    pub sig_lists: Vec<<TYPES::SignatureKey as SignatureKey>::PureAssembledSignatureType>,
+    pub sig_lists: HashMap<
+        Commitment<VOTE::Commitment>,
+        Vec<<TYPES::SignatureKey as SignatureKey>::PureAssembledSignatureType>,
+    >,
     /// A bitvec to indicate which node is active and send out a valid signature for certificate aggregation, this automatically do uniqueness check
-    pub signers: BitVec,
+    pub signers: HashMap<Commitment<VOTE::Commitment>, BitVec>,
```
jbearer, line 93: Does it make sense to combine these all into one map?

jbearer (approved):
This seems to implement exactly what we talked about yesterday. Someone with consensus expertise should also probably review for any unintended side effects.

Also a question about combining the maps indexed by vote commitment.

crates/types/src/vote.rs
bfish713: I was debating this I think it makes sense

crates/task-impls/src/view_sync.rs
```diff
@@ -277,25 +278,22 @@ impl<
...
                     debug!("Forwarding message");
                     let result = relay_task.handle_event(event.clone()).await;
 
                     if result.0 == Some(HotShotTaskCompleted::ShutDown) {
                         // The protocol has finished
                         return;
                     }
-
-                    map.insert(vote_view, result.1);
+                    relay_map.insert(relay, result.1);
```
shenkeyao, line 291: Should we also remove the `vote_view` from the `map` if the `relay_map` is empty? We did that in the previous code, but now it looks like we only clear the inner `relay_map`.

bfish713: I think if remove the whole view, we'd just end up creating it again on the next vote from this view.  We still do the Garbage collection on view change to remove views from the map

shenkeyao: Makes sense!

shenkeyao (approved):

PR #2472, opened 2024-01-28
Check Relay Number when Accumulating View Sync Votes

bfish713:
Same as: https://github.com/EspressoSystems/HotShot/pull/2470 but into main


crates/task-impls/src/vote.rs
```diff
@@ -202,8 +201,8 @@ where
     }
     let new_accumulator = VoteAccumulator {
         vote_outcomes: HashMap::new(),
-        sig_lists: Vec::new(),
-        signers: bitvec![0;info. membership.total_nodes()],
+        sig_lists: HashMap::new(),
+        signers: HashMap::new(),
```
shenkeyao, line 205: Looks like this needs an update to be the same as the one in the hotfix branch.

shenkeyao (changes requested):

shenkeyao (approved):

PR #2473, opened 2024-01-28
[HOTFIX] Poll forever for latest data

rob-maron:
### This PR: 

- Changes terminology from view sync proposal -> view sync cert everywhere to be more consistent

For view sync certificates and quorum proposals:
- Refactors so that we always poll for the latest data
- Starts each "poll for latest" task at the beginning of consensus
- Adds an additional wait to specific polls

### This PR does not: 

Fix the relay number check.

### Key places to review: 

`web_server_network.rs`

### DIFF
--- a/crates/hotshot/src/traits/networking/web_server_network.rs
+++ b/crates/hotshot/src/traits/networking/web_server_network.rs
@@ -223,10 +223,14 @@ struct Inner<TYPES: NodeType> {
     view_sync_vote_task_map: Arc<RwLock<TaskMap<TYPES::SignatureKey>>>,
     /// Task map for transactions
     txn_task_map: Arc<RwLock<TaskMap<TYPES::SignatureKey>>>,
-    /// Whether or not we have a task polling for latest quorum propsal
-    latest_quorum_proposal_poll_running: Arc<AtomicBool>,
-    /// Whether or not we have a task polling for the latest view sync certificate
-    latest_view_sync_certificate_poll_running: Arc<AtomicBool>,
+    /// A handle for cancelling the task polling for latest quorum propsal
+    cancel_latest_quorum_proposal_task: Arc<
+        RwLock<Option<UnboundedSender<ConsensusIntentEvent<<TYPES as NodeType>::SignatureKey>>>>,
+    >,
+    /// A handle for cancelling the task polling for the latest view sync certificate
+    cancel_latest_view_sync_certificate_task: Arc<
+        RwLock<Option<UnboundedSender<ConsensusIntentEvent<<TYPES as NodeType>::SignatureKey>>>>,
+    >,
 }
 
 impl<TYPES: NodeType> Inner<TYPES> {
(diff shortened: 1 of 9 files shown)


crates/hotshot/src/traits/networking/web_server_network.rs
```diff
@@ -962,58 +998,89 @@ impl<TYPES: NodeType + 'static> ConnectedNetwork<Message<TYPES>, TYPES::Signatur
...
+                    let inner = self.inner.clone();
+
+                    // Create sender and receiver for cancelling parent task
                     let (sender, receiver) = unbounded();
-                    *proposal_task = Some(sender);
+                    *cancel_handle = Some(sender);
+
+                    // Create the new task
+                    async_spawn(async move {
+                        loop {
```
bfish713, line 1012: Do we need this loop here or can we just not return from `poll_webserver()` since it's already in a loop there, and uses the `running` atomic bool to do shutdown

crates/hotshot/src/traits/networking/web_server_network.rs
```diff
@@ -962,58 +998,89 @@ impl<TYPES: NodeType + 'static> ConnectedNetwork<Message<TYPES>, TYPES::Signatur
...
+                                return;
+                            }
 
-                    async_spawn({
-                        let inner_clone = self.inner.clone();
-                        async move {
-                            if let Err(e) = inner_clone
+                            let inner = inner.clone();
+                            // Create a new channel that we will not use (as we don't cancel)
+                            let (_, fake_receiver) = unbounded();
```
bfish713, line 1066: I think we can just use the same cancel pattern as `PollForVotes`

crates/hotshot/src/traits/networking/web_server_network.rs
rob-maron: I think the `poll_webserver` loop returns once it gets the data needed

`                            MessagePurpose::Proposal => {
                                // Only pushing the first proposal since we will soon only be allowing 1 proposal per view
                                self.broadcast_poll_queue
                                    .write()
                                    .await
                                    .push(deserialized_messages[0].clone());

                                return Ok(());`

rob-maron: Oh yeah, it doesn't in our case actually. I can do this

rob-maron: Actually , some of the message types return on completion and some don't. What should we go with?

crates/hotshot/src/traits/networking/web_server_network.rs
```diff
@@ -325,7 +332,7 @@ impl<TYPES: NodeType> Inner<TYPES> {
 
                                 return Ok(());
```
bfish713, line 333: If we don't return here we could get away without the outer loop I mentioned in my other comment.  I also think we want to use an `LruCatch` here just like in the `LatestViewSyncCertificate` case

crates/hotshot/src/traits/networking/web_server_network.rs
bfish713: I say don't return from poll for the `Latest_` message purpose.  Follow the pattern of handling votes

jbearer:
@rob-maron can you do this for `release/0.5.5` also?

crates/hotshot/src/traits/networking/web_server_network.rs
rob-maron: Should we also consider actually having the cache a layer up? This way it can be shared for every task instead of just per task. The deduplication in this case was just so we don't pull multiple of the same latest proposal

bfish713 (approved):
Fixes look good to me, lets get this into the release branch as well

crates/hotshot/src/traits/networking/web_server_network.rs
bfish713: Not sure I understand what you mean by per task.  If you mean the task we spawn to poll the webserver then yeah we should probably do a shared cache between the view number based and latest polls.  On the other hand we could probably just get rid of the view based polling for proposals after this.  I say we keep it as is for now

rob-maron: Yeah, the cache is intra-task and not inter-task

bfish713 (approved):

PR #2474, opened 2024-01-28
[TECH_DEBT] - Enable unwrap_used lint

### LINKED ISSUE #2471: [TECH_DEBT] - Enable unwrap_used lint

- What is this task: enable the unwrap_used lint globally in our workspace cargo.toml
- Why do we need to work on it: to be more careful about where we panic

We'll need to enable the lint then fix all the resulting warnings

_No response_

linting passes and the lint is globally enabled

_No response_

DieracDelta:
Closes #2471 

### This PR: 
Enables the unwrap_used lint, and corrects any mismatched places. Turns out we unwrap in quite a few places. I've replaced all our unwraps with `expect`.

### Key places to review: 
Not sure there's anything super key here. This should give us better errors, though.


crates/hotshot-stake-table/src/vec_based.rs
```diff
@@ -255,14 +255,15 @@ where
     #[must_use]
     pub fn new(capacity: usize) -> Self {
         let bls_comm_preimage = vec![F::default(); capacity * <K1 as ToFields<F>>::SIZE];
-        let default_bls_comm =
-            VariableLengthRescueCRHF::<F, 1>::evaluate(&bls_comm_preimage).unwrap()[0];
+        let default_bls_comm = VariableLengthRescueCRHF::<F, 1>::evaluate(&bls_comm_preimage)
+            .expect("Couldn't evaluate bls commitment preimage")[0];
```
shenkeyao, line 259: Is `new` only called by test code? If not, it's probably better to return an error rather than use `expect`.

crates/hotshot/src/tasks/mod.rs
shenkeyao, line 1: Same as my other comment about error vs `expect`, I think it's better to return an error if `get_task_id` fails in this file.

shenkeyao (approved):
I think in general we should do error handling instead of `expect` unless it's test code or something that shouldn't fail. But from the description of this PR, the goal is to replace `unwrap` with `expect` which I agree is already an improvement. Therefore, I'm going to approve the PR, but leave my thoughts about this as well. 😃

jparr721:
Closing for now since realistically fixing these conflicts would be extremely painful.

PR #2477, opened 2024-01-29
[HOTFIX] [RELEASE BRANCH] Forever polling

rob-maron:
The same as #2473 but for the release branch.


crates/web_server/src/lib.rs
```diff
@@ -24,24 +28,19 @@ type Error = ServerError;
...
-    /// view for oldest proposals in memory
-    oldest_proposal: u64,
     /// view for the most recent proposal to help nodes catchup
     latest_quorum_proposal: u64,
     /// view for the most recent view sync proposal
-    latest_view_sync_proposal: u64,
-
-    /// view for teh oldest DA certificate
+    latest_view_sync_certificate: u64,
+    /// view for the oldest DA certificate
```
bfish713, line 42: Whoops missed this in the last one, we can get rid of this too, but let's keep it for now.

bfish713 (approved):

Issue #2376, opened 2024-01-10
[VALIDATED_STATE] - Add instance-level state


jbearer:
### What is this task and why do we need to work on it?

With validated state catchup, the validated state can be missing or default-constructed at any given time. The application's implementation of state validation is expected to fetch missing data dynamically as needed to apply a new block and update the state.

However, in the sequencer, we need some minimum amount of data to even be able to fetch missing data to fill in the validated state. For example, we need an L1 client , and a list of peer URLs for fetching fee state. Since our implementations of the HotShot traits only have access to data that is passed in by HotShot, we will need a way to pass this "instance level state" into HotShot at initialization, and have HotShot keep track of it and pass it back into the validated state callbacks.

For example, the `ValidatedState::validate_and_apply` method might look like this:
```rust
trait ValidatedState {
  fn validate_and_apply(
    &mut self, // _mutable_ reference to the validated state we are trying to update
    &instance: TYPES::InstanceState, // _immutable_ reference to instance level state, which allows us to fetch missing validated state
    parent: Header,
    proposal: Header,
  )
}
```

Here `InstanceState` refers to a new associated type defined in `NodeType`:

```rust
trait NodeType {
  type InstanceState: InstanceState;
 // ...
}
```

The `InstanceState` trait can be very simple since HotShot doesn't actually care about what type this is: all HotShot does is pass it around. `InstanceState` may just be used to collect basic super traits like `Send + Sync`.

### What work will need to be done to complete this task?

* Create the `InstanceState` trait and add it to `NodeTypes`.
* Rename the `State` trait to `ValidatedState` and combine its `validate` and `append` functions into `validate_and_apply` (after confirming they aren't used separately).
* Add an `instance_state` argument to `validate_and_apply`.

### What are the acceptance criteria to close this issue?

The integrations is able to implement `ValidatedState` including catchup, without relying on global static variables.

PR #2486, opened 2024-01-29
[VALIDATED_STATE] - Add instance-level state

shenkeyao:
Closes https://github.com/EspressoSystems/HotShot/issues/2376.

### This PR: 
* Creates the `InstanceState` trait for the instance-level state.
  * Adds it as an associated type of `NodeType`.
  * Adds test implementation.
* Creates and stores the `InstanceState` upon initializing `SystemContext`.
* Renames `State` to `ValidatedState` and the `state` crate to `states`.
* Removes `Committable` bound from `ValidatedState`.

### Key places to review: 
* crates/hotshot/src/lib.rs
* crates/types/src/consensus.rs
* crates/types/src/traits/node_implementation.rs
* crates/types/src/traits/states.rs


crates/types/src/traits/states.rs
```diff
@@ -65,7 +75,7 @@ pub trait State:
     /// This can also be used to rebuild the state for catchup.
     fn from_header(block_header: &Self::BlockHeader) -> Self;
 
-    /// Construct a genesis state.
+    /// Construct a genesis validated state.
     #[must_use]
     fn genesis() -> Self {
```
jbearer, line 80: Probably we should get `instance` here too (example: we might pass in some pre-funded accounts to initialize the state with at genesis, for testing purposes)

jbearer (changes requested):
I think `Header::genesis` and `Header::new` should also get `instance`:
* Genesis would allow us to configure some properties of the genesis block via command line, like network ID, genesis timestamp, etc, see https://github.com/EspressoSystems/espresso-sequencer/issues/1048
* In `new` we would use the L1 client stored in the instance state to fetch deposits and check the finalized L1 block number

crates/hotshot/src/lib.rs
```diff
@@ -755,9 +755,11 @@ impl<TYPES: NodeType> HotShotInitializer<TYPES> {
     /// initialize from genesis
     /// # Errors
     /// If we are unable to apply the genesis block to the default state
-    pub fn from_genesis() -> Result<Self, HotShotError<TYPES>> {
+    pub fn from_genesis(
+        instance_state: &TYPES::InstanceState,
```
jbearer, line 759: Not a big deal but if we're passing in the instance state here, and this initializer ultimately gets passed to `SystemContext::init`, we might as well store the instance state in `HotShotInitializer` and remove it as a param from `init`/`new`

crates/task-impls/src/consensus.rs
```diff
@@ -1220,6 +1220,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
             let block_header = TYPES::BlockHeader::new(
                 commit_and_metadata.commitment,
                 commit_and_metadata.metadata.clone(),
+                &consensus.instance_state,
```
jbearer, line 1223: At some point we may want to collect all these arguments into a struct, since for various implementations a lot of them might be unused anyways. Plus, I believe the arguments to `validate_and_apply_header` should be the same as the arguments to `Header::new`, plus the proposed header?

jbearer (approved):
LGTM

crates/task-impls/src/consensus.rs
shenkeyao: Looks like the only arguments this function and `validate_and_apply_header` have in common are the instance state and parent header. I'm not sure if adding a struct with two fields would be useful. Or do you mean a struct with more fields not necessarily used by `validate_and_apply_header`?

jbearer: Hmm you're right. In *principle*, `validate_and_apply_header` and `Header::new` should get the same exact inputs, because the former needs to check that whatever the latter did was done correctly. But, for `validate_and_apply_header`,
* `state` is given as `self`
* `commitment` and `metadata` are part of `proposed_header`
So we are left with only `instance_state` and `parent_header` explicitly in common.

_Maybe_ to make this analogy more explicit we could make the order of arguments more consistent, e.g.
* Make `state` the first argument to `BlockHeader::new`, since `self` is first in `validate_and_apply_header` (alternative idea: make `new_header` a method of `State`)
* Make `instance` the second argument to both
* Move `parent_header` before `proposed_header` since `parent_header` is an argument to both but `proposed_header` only applies to `validate_and_apply_header`
* Delete `view_number` from `validate_and_apply_header` since we don't get it here, we don't really need it, and it is probably best for it to remain internal to consensus

Anyway, this is just some logic that I am using to check that we have all the necessary arguments in both places, I guess it's not really necessary to make this logic explicit in the code but I think it would be nice to have.

shenkeyao: Makes sense! I think it's nice to have a more consistent and reasonable ordering of arguments. I'm also going to make `commitment` and `metadata` the last arguments to `BlockHeader::new`.

jbearer (approved):
Looks great, thanks for making all those changes!

PR #2488, opened 2024-01-30
Ed/diagrams part1

elliedavidson:
Closes #<ISSUE_NUMBER>

### This PR: 

This PR is part 1 of several forthcoming PRs containing diagrams of HotShot's flow.  This PR only contains 1 diagram: QuorumProposalRecv.  This is the most complex diagram and the one most different from the current HotShot logic.  

### This PR does not: 

Contain the rest of the HotShot diagrams.  Those will come in a future PR

### Key places to review: 

`QuorumProposalRecv.md` in `/docs/diagrams` and `QuorumProposalRecv.png` in `docs/diagrams/images`
The main thing is to check that the logic of the diagram is correct!


.gitignore
```diff
@@ -10,3 +10,4 @@
 **/.DS_Store
 *.cache
 **/.direnv
+**/*.bkp
```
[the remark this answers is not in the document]
elliedavidson: Yes, good idea!

docs/diagrams/images/HotShotFlow-QuorumProposalRecv.drawio.png
shenkeyao, line 1: 1. **Global Storage Accessed**
Shouldn't we include `saved_leaves` as well?
2. **View Change Evidence**
I feel it might be better to have a consistent annotation for the view number associated with the TC and view sync commit cert. Looks like here we want the TC to be for the previous view, but the view sync commit cert for the current view--could we adjust what the "for" means to make them consistent and probably clearer?
3. **Is proposal's QC valid?**
I think this should point to the **Proposal Validation** box, which will then point to **ViewChange** in the "yes" case, and **Invalid QC** in the "no" case.
4. **Store proposal in state map**
We should also update `saved_leaves`.

docs/diagrams/QuorumProposalRecv.md
```diff
@@ -0,0 +1,9 @@
+## QuorumProposalRecv
+
+![QuorumProposalRecv](/docs/diagrams/images/HotShotFlow-QuorumProposalRecv.drawio.png "QuorumProposalRecv")
+
+#### Notes
+* We emit a `ViewChange` event if we've seen evidence for a view change *and* if the quorum certificate on the proposal is valid.  We could optionally emit a `ViewChange` event if we've seen evidence but the quorum certificate on the proposal is invalid, but that is not the choice we make here.  Either choice is correct, but the former is simpler.  
+* We include a view sync certificate as view change evidence.  This isn't strictly required, but it makes the view change logic more understandable and avoids an attack where a view sync relay only sends a commit certificate to the next leader, resulting in other nodes rejecting the leader's proposal because it doesn't have proper view change evidence. It also signals nodes to handle view sync certificate evidence, which is handled differently than timeout certificate evidence or quorum certificate evidence.
```
shenkeyao, line 7: Based on the diagram, the view sync commit cert isn't required as the view change evidence, unless the view doesn't match and there's no TC, but here it seems that the view sync cert is always included. Is the view sync cert different from the commit cert (because the former represents the "finalize" stage rather than "commit")? If so, shouldn't they all be included in the diagram?

I guess I'm confused about this, and so don't quite understand the attack described here as well.

shenkeyao (changes requested):
Thanks for the diagrams and documentation! Left some comments about `QuorumProposalRecv`.

jbearer (commented):
This is awesome.

Is it possible for two "determine committed blocks" tasks spawned from proposal validation for two different views to be running at the same time? If so, what would prevent them both from generating a decide for the same block?

docs/diagrams/images/HotShotFlow-QuorumProposalRecv.drawio.png
dailinsubjam, line 1: What's the definition of purple blocks? I feel for the fetching missing block one it's like network task, and for the determine committed blocks one it needs some updates in the storage (correct me if I'm wrong) which is similar to green blocks.

elliedavidson: 1. Technically yes, though I feel like saved_leaves should be moved to be inside state_map?  Here I used state_map to represent basically all state we need to keep track of concerning leaves, proposals, DA proposals, etc.  I think it would be useful to add a section that gives some pseudocode of what each global variable is supposed to be.  I will work on this.  

2. Yes, I can adjust the wording to something like "TC's view equals proposal's view". I think that is more clear than "for"

3. I don't quite understand :). My thinking was at this point in the diagram, we've seen evidence to change views.  We can change views even if the proposal itself is invalid. 

4. Yes, (see comment 1)

docs/diagrams/QuorumProposalRecv.md
elliedavidson: I will work on rewording this!  Thank you for the feedback!

elliedavidson:
> This is awesome.
> 
> Is it possible for two "determine committed blocks" tasks spawned from proposal validation for two different views to be running at the same time? If so, what would prevent them both from generating a decide for the same block?

Good question.  Each task would need to obtain a write lock on the state to mark things as Decided and GC.  So only one of them could do this at once.  Does this sound alright?  I will include a diagram of what that function does at a high level.  (In the diagram I assume any access to global state is through some sort of lock, but I don't explicitly say where we lock stuff).

docs/diagrams/images/HotShotFlow-QuorumProposalRecv.drawio.png
elliedavidson: The purple blocks are supposed to be async functions that we spawn, but don't await.  (I should have included my template diagram, sorry!).  Since they aren't spawned via an event, I wanted to represent them differently.

shenkeyao: Makes sense to consider the saved leaves part of the state map in the diagram!

> 3. I don't quite understand :). My thinking was at this point in the diagram, we've seen evidence to change views. We can change views even if the proposal itself is invalid.

Hmm, if so, why do we only do `ViewChange` if `Is proposal's QC valid?` is `yes`? IIUC, when we reach the `Is proposal's QC valid?` step, we will do the `Proposal Validation`, and if it passes, we update the view. So the arrow between `ViewChange` and the big `Proposal Validation` box (as the arrow pointing down from `ViewChange` in the screenshot) should be in the other direction.
Screenshot 2024-02-01 at 9 05 52 AM

elliedavidson: I should probably add what being "valid" means for a certificate.  By valid I just mean that the signatures are valid.  So in this case, if the QCs signatures are valid over the QCs data, we can enter a new view.  And then we will do all the other logical checks in the Proposal Validation box.

shenkeyao: Oh okay, this makes sense!

bfish713, line 1: When we spawn the task to fetch missing leaves it seems like it will always leave us with an invalid proposal.  Shouldn't this task wait for those results and continue if it gets the leaves it needs?

bfish713, line 1: I'm assuming that we'll have some task which waits for a `QuorumProposalValidated` and other things (dependency task) and then Votes.

bfish713, line 1: Why do we check the QC is valid in the case of TC or View Sync certificate?  If a proposal has a TC/View sync cert isn't that evidence enough that we can enter the next view?

elliedavidson: Yes!

elliedavidson: Yes, you are right.  I added a note about this in `QuorumProposalRecv.md`. We could change the view if the QC was invalid, but I felt like for simplicity / readability, it was better if we don't.  Either way is safe.

elliedavidson: I did have it the way you described in a previous version, but felt like it overcomplicated things.

bfish713 (dismissed):

shenkeyao (approved):

PR #2491, opened 2024-01-31
feat(light-client): remove state-prover crate

mrain:
### This PR: 

- `hotshot-state-prover` crate is moved to sequencer repo.
- Types and structs are moved inside `hotshot-types` crate.


crates/types/src/light_client.rs
```diff
@@ -3,9 +3,12 @@
 use ark_ed_on_bn254::EdwardsConfig as Config;
 use ark_ff::PrimeField;
 use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
+use ethereum_types::U256;
```
alxiong, line 6: we should use `ethers_core::types::U256` instead (crate [here](https://crates.io/crates/ethers-core))

crates/types/src/light_client.rs
```diff
@@ -62,6 +65,13 @@ impl<F: PrimeField> From<&LightClientState<F>> for [F; 7] {
     }
 }
 
+/// Concrete type for light client state
+pub type LightClientState = GenericLightClientState<ark_ed_on_bn254::Fq>;
```
alxiong, line 69:
```suggestion
pub type LightClientState = GenericLightClientState<CircuitField>;
```

crates/types/src/light_client.rs
```diff
@@ -109,3 +119,100 @@ impl From<schnorr::KeyPair<Config>> for StateKeyPair {
...
+    pub accumulated_weight: U256,
+}
+
+/// Public input to the light client state prover service
+#[derive(Clone, Debug)]
+pub struct GenericPublicInput<F: PrimeField>(Vec<F>);
+
+/// Concrete public input type
+pub type PublicInput = GenericPublicInput<CircuitField>;
+
```
alxiong, line 151: can we move up this, so that all concrete types are in the same place, maybe better yet just move concrete type alias to a separate file?

alxiong (commented):
LGTM, left some minor comments

alxiong (approved):

PR #2500, opened 2024-02-01
[CX_CLEANUP] Create new Task Abstraction `async_broadcast` Channels

### LINKED ISSUE #2403: [CX-CLEANUP] - Create a Task for Collecting Dependencies

A task to collect dependencies is one which only executes some code after it receives all it's dependencies.  The main example is running logic after a collection of events is received.  This pattern comes up a lot in the consensus code.  Primarily with voting and proposing.  The logic is something like once we receive a QC, VID info and previous proposal you propose.  

The goal of the task is to have a generic implementation of this where you can define the dependencies, await them, then execute some logic.  The logic should also support having OR semantics.  I.e. collect this set of things, or some other set of things before running the logic.

- [x] Create any async primitives for awaiting the dependencies
- [x] Create a spawnable async task which will collect event dependencies
- [x] Add the logic to the task to run some logic when dependency collection completes

_No response_

We should write some tests using some test hotshot events and the event stream and see that we can trigger the logic just by publishing all the dependent events to the stream.

_No response_

bfish713:
Closes #2403 

### This PR: 

Creates a few async primitives for use in HotShot.  The next PR integrates them.

The first set of things are Task primatives.  This is meant to replace the "hotshot-task" crate currently in the code ("crates/task").  The tasks are based on `async_broadcast` channels.  These channels provide us with very nice set of properties.  First they are MPMC (Multi Producer, Multi Consumer) which we need, and second they have the nice property that clones of the `Receiver` will receive all messages the original receiver receives (I've gotten very good at spelling receiver haha).  This means that spawning a sub task is now possible.  If the spawned task takes a `Receiver` clone from the parent it will only "miss" the event which triggered the spawning, which can easily be accounted.  

The tasks work with a simple wait for events loop.  In the loop we check for shutdown conditions, do filtering and handle the event.  Using just channels and awaiting means no more need for tasks to sleep/yeild to allow other tasks to run.  The async executor and the channel itself will properly awake tasks to run when things are ready.

The main task primitives are: 
- `Task` - This is a simple version of the old task.  It simply loops waiting for events and handles them until it reaches a shutdown condition
- `TestTask` - Similar to `Task` but it receives messages from a bunch of receivers.  The idea is to use this for testing tasks, like the `OverallSafetyTask` to receive messages from all of the nodes in the test.  It contains within it a `Task` as well which is simply to hold the event receiver, registry, and state.  The event receiver should be for receiving internal events, so for testing tasks this will be `GlobalTestEvent`s.
- `TaskState` trait - This trait is for defining the behavior of the tasks.  In our code the main tasks are pretty simple, they handle events until they shutdown.  We allow for event filtering and there is a hook to cleanup after a shutdown condition.  This is for saving some special state if a Shutdown event is received. All the main logic task will implement this trait rather than passing in a `EventHandler` to some registry.
- `TestTaskState` trait - extends the `TaskState` trait with the addition of handling messages.  Messages are events but from outside tasks (outside isn't the best word but I mean: from not other testing things).  In our case this would be events from the main tasks.  

I also created some "Dependency" abstractions.  These abstractions are for encapsulating the idea of waiting for some preconditions to be met, then executing something which uses the results of the preconditions.  And example of preconditions might be waiting for a `DACRecv`, `QCFormed`, and vid data received events, and the execution would be to use all 3 of those events to form and send a `QuorumProposal`
The main primitives are:
- `Dependency` trait - This just provides an interface for something that can be awaited and return some data of type `T`.  It allows dependencies to be compostable.
- `AndDependency` - This is a dependency which collects many futures and is complete when all the futures complete.  It's intended to combine multiple other dependencies.  Essentially `futures::future::join_all` but for specifically dependencies
- `OrDependency` - Similar to `AndDependency`, but completes when any of it's futures succeeds and returns just the completed dependency result.  Essentially `futures::future::select_all` but for dependencies.
- `CombinedDependency` trait - is a nice way to combine dependencies into and or or versions. See the tests for examples
- `EventDependency` - A dependency that completes when it gets a specific event.  Takes a closure which returns whether an event is a match.  This is similar to `TaskState::filter_event` but is a closure so you can do things like look for a proposal for a given view (which will dynamically change each view).
- `DependencyTask` - A task which awaits on some dependency then runs so code which takes the results of it's dependency as input.

In my next PR I'm using the Task part of this extensively so I'm somewhat confident that it will provide us with what we need.  The dependency stuff can probably be improved and made more ergonomic as well.  I'm happy to here any feedback on the interface, and I'm sure we'll have to iterate on this as we actually use it in implementation.

### This PR does not: 

This PR does not add or change any functionality, it's simply to allow review for the new types.  

### Key places to review: 

All the added coded.  For examples see my next PR https://github.com/EspressoSystems/HotShot/pull/2493


crates/new_task/src/dependency.rs
```diff
@@ -0,0 +1,227 @@
...
+    }
+    pub fn add_dep(&mut self, dep: impl Dependency<T> + Send + 'static) {
+        self.deps.push(dep.completed().boxed());
+    }
+}
+
+pub struct EventDependency<T: Clone + Send + Sync> {
+    pub(crate) event_rx: Receiver<T>,
+    pub(crate) match_fn: Box<dyn Fn(&T) -> bool + Send>,
+}
```
elliedavidson, line 85: This will be so useful to be able to match events based on their view numbers!

crates/new_task/src/dependency.rs
```diff
@@ -0,0 +1,227 @@
...
+        tx.broadcast(5).await.unwrap();
+        tx.broadcast(6).await.unwrap();
+
+        let mut and1 = eq_dep(rx.clone(), 4).and(eq_dep(rx.clone(), 6));
+        let and2 = eq_dep(rx.clone(), 4).and(eq_dep(rx.clone(), 5));
+        and1.add_deps(and2);
+        let result = and1.completed().await;
+        assert_eq!(result, vec![4, 6, 4, 5]);
+    }
+}
```
elliedavidson, line 227: Very nice testing!

crates/new_task/src/dependency_task.rs
```diff
@@ -0,0 +1,129 @@
...
+use crate::dependency::Dependency;
+
+pub trait HandleDepResult: Send + Sized + Sync + 'static {
+    type Result: Send + Sync + 'static;
+
+    /// Called once when the Dependency completes handles the results
+    fn handle_dep_result(self, res: Self::Result) -> impl Future<Output = ()> + Send;
+}
+
+pub struct DependencyTask<D: Dependency<H::Result> + Send, H: HandleDepResult + Send> {
```
elliedavidson, line 17: Nit: did we include this crate in the lints?  I feel like clippy should be warning for no documentation.

crates/new_task/src/task.rs
```diff
@@ -0,0 +1,376 @@
...
+        event: Self::Event,
+        task: &mut Task<Self>,
+    ) -> impl Future<Output = Option<Self::Result>> + Send
+    where
+        Self: Sized;
+
+    /// Return true if the event should be filtered
+    fn filter(&self, _event: &Self::Event) -> bool {
+        // default doesn't filter
+        false
```
elliedavidson, line 45: Nice!  I think false is a better default than true like we had before.  That way we specifically handle the events we intend.

elliedavidson (changes requested):
Not approving because I think we need to enable lints.  Otherwise I approve.

crates/new_task/src/dependency_task.rs
bfish713: Hmm it must not be enabled, I'll clean up

crates/new_task/src/task.rs
bfish713: This actually tripped me up in debugging haha, I copied the filters from before they were the oppsoite

crates/new_task/src/dependency.rs
```diff
@@ -0,0 +1,251 @@
...
+use std::future::Future;
+
+/// Type which describes the idea of waiting for a dependency to complete
+pub trait Dependency<T> {
+    /// Complete will wait until it gets some value `T` then return the value
+    fn completed(self) -> impl Future<Output = T> + Send;
+}
+
+/// Used to combine dependencies to create `AndDependency`s or `OrDependency`s
+trait CombineDependencies<T: Clone + Send + Sync + 'static>:
```
QuentinI, line 15: Should this be a separate trait? I understand that this has stricter requirements for T, but I don't think there's much utility in using `Dependency` without combinators anyway.

crates/new_task/src/dependency.rs
```diff
@@ -0,0 +1,251 @@
...
+            event_rx: receiver,
+            match_fn: Box::new(match_fn),
+        }
+    }
+}
+
+impl<T: Clone + Send + Sync + 'static> Dependency<T> for EventDependency<T> {
+    async fn completed(mut self) -> T {
+        loop {
+            let next = self.event_rx.recv().await.unwrap();
```
QuentinI, line 123: This looks scary. Can we make this into `Dependency<Option<T>>` instead?

crates/new_task/src/dependency_task.rs
```diff
@@ -0,0 +1,140 @@
...
+use tokio::task::{spawn, JoinHandle};
+
+use futures::Future;
+
+use crate::dependency::Dependency;
+
+/// Defines a type that can handle the result of a dependency
+pub trait HandleDepResult: Send + Sized + Sync + 'static {
+    /// Type we expect from completed dependency
+    type Result: Send + Sync + 'static;
```
QuentinI, line 13: Nit: I think Output terminology would be more appropriate here

crates/new_task/src/task.rs
```diff
@@ -0,0 +1,445 @@
...
+    dependency::Dependency,
+    dependency_task::{DependencyTask, HandleDepResult},
+};
+
+/// Type for mutable task state that can be used as the state for a `Task`
+pub trait TaskState: Send {
+    /// Type of event sent and received by the task
+    type Event: Clone + Send + Sync + 'static;
+    /// The result returned when this task compeltes
+    type Result: Send;
```
QuentinI, line 35: Nit: again, Output would be more appropriate IMO

crates/new_task/src/dependency.rs
bfish713: I made this a trait so that I could implement it for all of our dependency types, we had, but this is a good point we can promote to the `Dependency` trait.

crates/new_task/src/dependency.rs
bfish713: Good Catch, yes this should be `Option<T>` or `Result<T>` and handle the errors appropriately

crates/new_task/src/dependency_task.rs
```diff
@@ -0,0 +1,140 @@
...
+use async_std::task::{spawn, JoinHandle};
+#[cfg(async_executor_impl = "tokio")]
+use tokio::task::{spawn, JoinHandle};
+
+use futures::Future;
+
+use crate::dependency::Dependency;
+
+/// Defines a type that can handle the result of a dependency
+pub trait HandleDepResult: Send + Sized + Sync + 'static {
```
QuentinI, line 11: Nit: Can we switch this to something like `HandleDepOutput` and `HandleDepOutput::handle_output` as well? Sorry for not wording this more explicitly in my original comment.
My reasons for requesting this is that reading "result" in Rust code makes me - and I assume most other Rust users - think that this abstraction deals explicitly with fallible computation, which isn't the case. I don't much care about exact names here, my main concern is reserving "result" for something that explicitly deals with errors

bfish713: Yeah this makes sense, I had forgotten that I named this "result"

QuentinI (commented):
No more comments from my end, looks awesome
I'll leave the approval to Ellie

bfish713:
Closed because #2493 incorporates these changes

PR #2522, opened 2024-02-04
refactor: move NUM_REPLICATED_TO_TRUST to mod

### LINKED ISSUE #2434: [TECH_DEBT] - Two NUM_REPLICATED_TO_TRUST

We have two NUM_REPLICATED_TO_TRUST. We should either document why we have two, or remove one of them.

_No response_

_No response_

Either additional docs or removal of one of the constants

_No response_

maschad:
Seems like this was an oversight when https://github.com/EspressoSystems/HotShot/pull/302 was merged.

Perhaps the most effective way to avoid a recurrence is to move all constants to a file within the folder structure of a crate as it seems non-trivial to accomplish using [clippy](https://github.com/rust-lang/rust-clippy)

Closes #2434


elliedavidson (approved):
Thank you for this clean up!

PR #2525, opened 2024-02-05
[Key Management] Get peer's public key from orchestrator

### LINKED ISSUE #2416: [KEY MANAGEMENT] - Post public keys to orchestrator

What we want is having user posted their public keys to the orchestrator, rather than the orchestrator generating the keys.

- [x] When the node post to orchestrator, the node also post its public key.
- [x] The orchestrator keep track of how many nodes have post their keys.
- [x] Once all the nodes have given the orchestrator with their public keys, the orchestrator put their keys into the config.
- [x] Once all the nodes have requested to the config, the orchestrator will send the start command.

A follow up on https://github.com/EspressoSystems/HotShot/pull/2247.

Unblock https://github.com/EspressoSystems/espresso-sequencer/issues/934.

_No response_

dailinsubjam:
Closes #2416

### This PR: 
- Add one more round for orchestrator to collect all validator's public keys.
- Correct and move some key generation part.
- Is a follow up on https://github.com/EspressoSystems/HotShot/pull/2247, some changes are renewed here, I'll check the old one to see whether anything there still needed and probably close it without the merge. 
- Might be related to https://github.com/EspressoSystems/espresso-sequencer/issues/934.

### This PR does not: 
- Directly input node's key pair through config. The reason is that 1. their formats are too complex (see config/ValidatorConfigExample for reference) to be in the config 2. Node can only generate key pairs once they got `node_index` if initializing their config from orchestrator 3. HotshotConfig does contain the secret key originally for signing, so it's fine.

### Key places to review: 
All changes.

### How to test this PR: 


crates/orchestrator/src/lib.rs
```diff
@@ -179,6 +203,50 @@ where
         Ok(self.config.clone())
     }
 
+    #[allow(clippy::cast_possible_truncation)]
+    fn post_my_public_key(
+        &mut self,
```
nomaxg, line 208: Should we throw an error if the public key is already assigned?

crates/orchestrator/src/lib.rs
```diff
@@ -179,6 +203,50 @@ where
         Ok(self.config.clone())
     }
 
+    #[allow(clippy::cast_possible_truncation)]
+    fn post_my_public_key(
```
nomaxg, line 207: I think "my" is a little confusing because this is on the orchestrator side, which doesn't have its own key. Maybe register_public_key would be better.

crates/orchestrator/api.toml
```diff
@@ -23,6 +23,29 @@ This must be a POST request so we can update the OrchestratorState in the server
...
+
+# GET whether or not the config with all peers' public keys / configs are ready
+[route.peer_pubconfig_ready]
+PATH = ["peer_pub_ready"]
+DOC = """
+Get whether the node can collect the final config which includs all peer's public config/info like public keys, returns a boolean.
+"""
+
+# GET the updated config with all peers' public keys / configs
+[route.config_after_peer_collected]
```
mrain, line 43: This looks like something that we could use for light client stuff as it contains `known_nodes_with_stake`. But the problem is, it doesn't contain the Schnorr verification keys.
However, Schnorr key shouldn't be in the consensus logic.

Before the stake table integration, if all keys are generated from the indexed seed, it might be good to generate the stake table in the same way.

crates/hotshot/examples/infra/mod.rs
```diff
@@ -872,20 +877,30 @@ pub async fn main_entry_point<
...
-        ValidatorConfig::<<TYPES as NodeType>::SignatureKey>::generated_from_seed_indexed(
-            run_config.seed,
-            node_index,
-            1,
-        );
     //run_config.libp2p_config.as_mut().unwrap().public_ip = args.public_ip.unwrap();
 
+    // one more round of orchestrator here to get peer's public key/config
+    orchestrator_client
+        .post_and_wait_all_public_keys::<TYPES::SignatureKey, TYPES::ElectionConfigType>(
```
nomaxg, line 893: Why cant we return the updated peer config here?

crates/orchestrator/src/lib.rs
dailinsubjam: Added. Also added a hashset to record who has posted.

crates/hotshot/examples/infra/mod.rs
dailinsubjam: Previously I separated them due to a weird bug which turned out not to be a bug at all lol, now they're merged to one function.

crates/orchestrator/api.toml
dailinsubjam: For Schnorr verification keys, is there any action item we want to take now for easier light client integration?
For stake table, do you mean we also want the stake be inputed as the same place as indexed seed?

mrain: Not now. If we are not changing the way to generate keys before stake table integration, we are all good.

nomaxg (approved):
LGTM, someone from consensus should probably review as well

crates/hotshot/examples/infra/mod.rs
nomaxg: great!

crates/hotshot/examples/infra/mod.rs
```diff
@@ -101,20 +108,18 @@ pub fn load_config_from_file<TYPES: NodeType>(
...
+    config.config.my_own_validator_config =
+        ValidatorConfig::generated_from_seed_indexed(config.seed, config.node_index, 1);
+    let my_own_validator_config_with_stake = config
+        .config
+        .my_own_validator_config
+        .public_key
+        .get_stake_table_entry(1u64);
+    // initialize it with size for better assignment of other peers' config
+    config.config.known_nodes_with_stake =
+        vec![my_own_validator_config_with_stake; config.config.total_nodes.get() as usize];
```
shenkeyao, line 122: In the previous version, we generated different keys for peers, but now it looks like they are all identical. Is this intentional?

shenkeyao: From the comment on this function, this is indeed intentional. Is it because we can only generate the key for ourselves, and wait for the orchestrator to send peers' keys to us?

crates/orchestrator/src/client.rs
```diff
@@ -136,12 +139,52 @@ impl OrchestratorClient {
         };
 
         let mut config = self.wait_for_fn_from_orchestrator(f).await;
-
         config.node_index = From::<u16>::from(node_index);
+        // The orchestrator will generate keys for validator if it doesn't load keys from file
+        config.config.my_own_validator_config =
+            ValidatorConfig::<K>::generated_from_seed_indexed(config.seed, config.node_index, 1);
```
shenkeyao, line 145: If the comment is accurate, maybe we should do an "if loading keys from file" check before calling `generated_from_seed_indexed`.

crates/orchestrator/src/client.rs
```diff
@@ -136,12 +139,52 @@ impl OrchestratorClient {
...
+        // get the newest updated config
+        let get_newest_config = |client: Client<ClientError>| {
+            async move {
+                let config: Result<NetworkConfig<K, E>, ClientError> =
+                    client.get("api/config_after_peer_collected").send().await;
+                config
+            }
+            .boxed()
+        };
+        self.wait_for_fn_from_orchestrator(get_newest_config).await
```
shenkeyao, line 185: Can we just call `send().await` to get the updated config? Is it possible that we need `wait_for_fn_from_orchestrator` though we know peers' keys are ready?

crates/orchestrator/src/lib.rs
```diff
@@ -179,6 +207,57 @@ where
...
+        if self.pub_posted.contains(&node_index) {
+            return Err(ServerError {
+                status: tide_disco::StatusCode::BadRequest,
+                message: "Node has already posted public key".to_string(),
+            });
+        }
+        self.pub_posted.insert(node_index);
+
+        // Sishan NOTE: let me know if there's a better way to remove the first extra 8 bytes
+        pubkey.drain(..8);
```
shenkeyao, line 225: Makes sense to use `drain`! Can you add a comment about why we want to delete these bytes?

crates/orchestrator/src/lib.rs
```diff
@@ -179,6 +207,57 @@ where
...
+
+    fn get_config_after_peer_collected(&self) -> Result<NetworkConfig<KEY, ELECTION>, ServerError> {
+        if !self.peer_pub_ready {
+            return Err(ServerError {
+                status: tide_disco::StatusCode::BadRequest,
+                message: "Peer's public configs are not ready".to_string(),
+            });
+        }
+        Ok(self.config.clone())
+    }
```
shenkeyao, line 259: I'm thinking maybe we can have a `get_peers_and_config` endpoint to wait for the peers to be ready and then get the config, in case we forget to call `peer_pub_ready` first. But I'm not sure how useful this will be, so curious about your thoughts. 😄

shenkeyao (commented):
Looks good. Left some questions and thoughts.

crates/hotshot/examples/infra/mod.rs
dailinsubjam: Yes this is intentional, basically for initializing a vector with fixed type and size! So that when we assign peer's public key to the config, we can directly do `vec[index] = ...` which is easier.

crates/orchestrator/src/client.rs
dailinsubjam: This `get_config()` is called in a function `from_file_or_orchestrator` only when the file loading is failed. I'll add a comment there.

crates/orchestrator/src/client.rs
dailinsubjam: Yeah that's redundant.

crates/orchestrator/src/lib.rs
dailinsubjam: Sure, added! Actually I'm not 100% sure about why, my guess is orchestrator did their own serialization, therefore it adds 8 more bytes for the type configuration.

crates/orchestrator/src/lib.rs
dailinsubjam: Aha, this is exactly what `post_and_wait_all_public_keys()` in `config.rs` does. Here for the function directly updating or posting to orchestrator state, I'm more inclined to keep them atomic and do not use async there.

crates/orchestrator/src/client.rs
shenkeyao: I see. Sorry to be nit-picking, but I still don't think the comment here is accurate. According to the comment, I would think this function determines whether to load keys from the file, but it's actually the caller's job to determine this. At least the "if" should be "because".

crates/orchestrator/src/lib.rs
```diff
@@ -110,6 +181,58 @@ where
...
+    ) -> Result<(), ServerError> {
+        if self.pub_posted.contains(&node_index) {
+            return Err(ServerError {
+                status: tide_disco::StatusCode::BadRequest,
+                message: "Node has already posted public key".to_string(),
+            });
+        }
+        self.pub_posted.insert(node_index);
+
+        // Sishan NOTE: let me know if there's a better way to remove the first extra 8 bytes
```
shenkeyao, line 198: Nit: Can we remove this comment, unless there's still a question about this?

shenkeyao (approved):
Just nit comments.

bfish713 (approved):

rob-maron (approved):

PR #2527, opened 2024-02-05
Ed/diagrams part2

elliedavidson:
Closes #<ISSUE_NUMBER>

### This PR: 

* Adds diagrams and notes for the following HotShot tasks: OptimisticDACertificateRecv, OptimisticDAProposalRecv, VIDShareRecv, and VoteOnQuorumProposal

### This PR does not: 

* Add other task diagrams.  Those will come in a future PR

### Key places to review: 

* Each .md file of the newly added tasks.  The other changes were reviewed in my last PR.


docs/diagrams/OptimisticDACertificateRecv.md
```diff
@@ -0,0 +1,8 @@
+# OptimisticDACertificateRecv
```
shenkeyao, line 1: Are we going to distinguish the optimistic DAC from the normal (or other types of) DAC? We were previously, but not in the current code, so I'm curious about the terminology change here.

shenkeyao (approved):
Left a question about the optimistic DAC, but the PR looks good!

docs/diagrams/OptimisticDACertificateRecv.md
elliedavidson: I figured this was a more descriptive name for the DAC.  (Actually you were the one who suggested it a while ago!).  I think it is what we call it in the HotShot paper too, so I figured it would be good to match that.

PR #2530, opened 2024-02-05
[VALIDATED_STATE] - Send Decide event for the genesis block

### LINKED ISSUE #2529: [VALIDATED_STATE] - Send Decide event for the genesis block

HotShot currently only sends the `Decide` event for non-genesis blocks, which makes it necessary for the query service code to [insert the genesis block](https://github.com/EspressoSystems/hotshot-query-service/blob/main/src/data_source/fetching.rs#L124). We should do it in HotShot, then the query service doesn't have to use HotShot test types outside of the test code.

Emit a `Decide` event for the genesis block in the consensus task.

_No response_

Query service doesn't need to insert the genesis block in non-test code.

_No response_

shenkeyao:
Closes #2529.

### This PR: 
Emits the `Decide` event for the genesis block in the consensus task when getting a genesis justify QC.

### DIFF
--- a/crates/task-impls/src/consensus.rs
+++ b/crates/task-impls/src/consensus.rs
@@ -507,9 +507,30 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> + 
 
                 let consensus = self.consensus.upgradable_read().await;
 
-                // Construct the leaf.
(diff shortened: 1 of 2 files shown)


jbearer (dismissed):
LGTM

bfish713 (approved):

PR #2557, opened 2024-02-09
Add handling for UpgradeCertificate

### LINKED ISSUE #2444: [Upgradability] - Implement logic to detect if the network has been upgraded

### LINKED ISSUE #2445: [Upgradability] - Implement mechanism to initiate a network upgrade

In our current vision for the upgrade process, the leader for a round will begin collecting signatures on an upgrade certificate a few rounds in advance (which all nodes will then vote on). 

_No response_

_No response_

There should be a mechanism for:

- [ ] Initiating the upgrade process as an upcoming leader for a round.
- [ ] Voting on an upgrade certificate

_No response_

### LINKED ISSUE #2556: [Upgradability] - Handle upgrade certificate in quorum proposal

Leaders collect signatures for an upgrade in an `UpgradeCertificate`, which should be attached to and handled with any `QuorumProposal` that goes out.

_No response_

_No response_

The leader for a view should be able to attach an upgrade certificate to a quorum proposal to initiate an upgrade. Nodes voting on an upgrade should validate this certificate, if a quorum proposal has one attached.

_No response_

ss-es:
Closes #2556
Closes #2444
Closes #2445 

### This PR: 
Adds `UpgradeCertificate` handling, implementing the remaining upgradability tasks.

- Leaders can collect signatures on an `UpgradeCertificate` (done in [PR#2532](https://github.com/EspressoSystems/HotShot/pull/2532)), which is published as an `UpgradeCertificateFormed` event to the consensus task.
- If an `UpgradeCertificate` has been formed by the time the leader proposes, it is attached to the quorum proposal.
- Nodes validate any attached `UpgradeCertificate` before voting.
- Any new decided leaf with an `UpgradeCertificate` attached updates the consensus state with that certificate.

### Key places to review: 
The `upgrade` and `consensus` tasks in `crates/task-impls`.


crates/task-impls/src/consensus.rs
```diff
@@ -496,6 +507,28 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
+                //
+                // We should just make sure we don't *sign* an UpgradeCertificate for an upgrade
+                // that we do not support.
+                if let Some(ref upgrade_cert) = proposal.data.upgrade_certificate {
+                    if !upgrade_cert.is_valid_cert(self.quorum_membership.as_ref()) {
+                        error!("Invalid upgrade_cert in proposal for view {}", *view);
+                        return;
+                    }
+                }
+
```
bfish713, line 531: I think this is the correct logic as an honest node.  An honest node should not try to subvert the system by using a proposal as as second round of voting

crates/task-impls/src/consensus.rs
```diff
@@ -727,6 +760,9 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
                                         .last_synced_block_height
                                         .set(usize::try_from(leaf.get_height()).unwrap_or(0));
                                 }
+                                if let Some(ref upgrade_cert) = proposal.data.upgrade_certificate {
+                                    self.decided_upgrade_cert = Some(upgrade_cert.clone());
+                                }
```
bfish713, line 765: Nit: I think we should at least log this event at info level with the details of the upgrade cert

crates/task-impls/src/consensus.rs
```diff
@@ -1274,12 +1331,27 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
+            if let Some(cert) = self.upgrade_cert.clone() {
+                if cert.view_number == view {
+                    upgrade_cert = Some(cert);
+                } else {
+                    upgrade_cert = None;
+                }
+            } else {
+                upgrade_cert = None;
+            }
+
```
bfish713, line 1347: nit can probably remove the clone with something like this:
```suggestion

            let upgrade_cert = self.upgrade_cert.is_some_and(|cert| { cert.view_number == view }) {
                self.upgrade_cert.take()
            } else {
               None
            };

```

Also will leave self.upgrade_cert == None after we use it

bfish713 (approved):
All makes sense besides the few minor comments

crates/task-impls/src/consensus.rs
ss-es: Oh, yes -- this is definitely worth logging. Added!

crates/task-impls/src/consensus.rs
ss-es: Changed!

crates/task-impls/src/consensus.rs
```diff
@@ -1274,12 +1332,29 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
+            let upgrade_cert = if self
+                .upgrade_cert
+                .as_ref()
+                .is_some_and(|cert| cert.view_number == view)
+            {
+                // If the cert view number matches, set upgrade_cert to self.upgrade_cert
+                // and set self.upgrade_cert to None.
+                //
+                // Note: the certificate is discarded, regardless of whether the vote on the proposal succeeds or not.
```
bfish713, line 1344: I assumed this was the desired behavior, otherwise we can simply switch take() to clone()

bfish713 (approved):

crates/task-impls/src/consensus.rs
ss-es: Yes, I think so! Just wanted to make it explicit.

Issue #2427, opened 2024-01-19
[HOTSHOT_DIAGRAMS] - Diagram HotShot flow and state transitions for core consensus logic


elliedavidson:
### What is this task and why do we need to work on it?

HotShot is a complex system and it would be helpful to have diagrams showing the proper HotShot state flow.  E.g., why are the conditions for voting on a proposal?  What if one of those conditions isn't met?  These diagrams should help us improve our code correctness and simplicity. 

### What work will need to be done to complete this task?

Create diagrams that detail the state flow for each task in HotShot. 

### What are the acceptance criteria to close this issue?

Consensus team plus other key reviewers agree the diagram is correct

PR #2564, opened 2024-02-12
Ed/diagrams part3

elliedavidson:
Closes https://github.com/EspressoSystems/HotShot/issues/2427

### This PR: 
* Adds the remaining diagrams and notes to the `docs` section

### This PR does not: 
* Include the upgradability logic: that will come in a future PR. 
* Have as much detail in the descriptions as I'd like.  I will work on that in the future concurrently with refactoring the consensus logic. 

### Key places to review: 
* The entire `docs/diagrams/event_descriptions` directory. 
* @move47 Can you specifically review the `BlockFromBuilderRecv.md` and `VIDDataFromBuilder.md` files?  I tried to capture the current builder API behavior where the payload and vid commitment are sent separately.  I also assumed for now we are trusting the builder, and therefore we are not doing any validation on what the builder returns to us.  We can update this in the future.  

Some major changes from the current HotShot event model: 
* No `TransactionsRecv` events, since that is handled by the builder module (even if a node is running the builder locally)
* BlockPayload and VID commitments come from the builder module
* In the case the network skips a view due to the view sync protocol, the logic now allows nodes to send DA proposals, VID shares, etc. in the view that they are the leader (if they haven't sent it in the previous view).  This should prevent views from failing unnecessarily.


docs/diagrams/images/HotShotFlow-VIDDataFromBuilderRecv.drawio.png
shenkeyao, line 1: Nit:
* `Is the data correctly...`: missing the `No` case.
* `Do we trust...`: missing the `Yes` label.

docs/diagrams/images/HotShotFlow-ViewChange.drawio.png
shenkeyao, line 1: Nit: Some tasks should be added to `Possible Spawning Tasks`.

docs/diagrams/images/HotShotFlow-ViewSyncCommitCertificateRecv.drawio.png
shenkeyao, line 1: Same comments for the three `ViewSync...CertificateRecv` diagrams: Should we also add `ViewSyncTimeout` spawning to the diagram?

docs/diagrams/images/HotShotFlow-BlockFromBuilderRecv.drawio.png
move47, line 1: LGTM!

shenkeyao (commented):
I have some questions that I'd like to double-check before approving. The rest are just nit comments.

docs/diagrams/images/HotShotFlow-BlockFromBuilderRecv.drawio.png
move47: But I think the `No` case w.r.t Signature Validation might have something to do also.

docs/diagrams/images/HotShotFlow-QuorumProposalSend.drawio.png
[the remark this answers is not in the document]
elliedavidson: I agree that it'd be nice to be consistent across all the diagrams, but I feel like QuorumProposalSend is a good summary of what the task is doing, which is the point of the title.  I'd personally like to keep it, but let me know if you think otherwise!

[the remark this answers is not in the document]
shenkeyao: I feel it may cause some confusion in the future if there are more diagrams and we want to search for this one. But I'm also not satisfied with the alternative I came up with, so I don't have a strong opinion. We could keep it as is and modify it later if needed!

docs/diagrams/images/HotShotFlow-Timeout.drawio.png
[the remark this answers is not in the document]
elliedavidson: 1. In my mind the view in the timeout view would be the view we should have entered.  E.g. View 11 timing out means that we are in view 10 and haven't seen evidence for view 11 in time.  But, you make a good point that that logic seems confusing.  Usually we want to say "View 10 timed out" instead of the above.  So I made the change you suggested!  

2. If we make the above change in logic, then I think you're right that we don't need the latest_voted_view.  I will make this fix in a future PR so we can get the rest of the diagrams merged!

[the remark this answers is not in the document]
elliedavidson: Thank you so much for this feedback!  It will definitely simplify the diagrams!

docs/diagrams/images/HotShotFlow-BlockPayloadFromBuilderRecv.drawio.png
shenkeyao, line 1: Nit: Missing "No" label for the invalid signature.

shenkeyao (approved):
Looks good! Just a very minor comment.

PR #2571, opened 2024-02-13
Enable testing macros for libp2p

### LINKED ISSUE #2536: [LIBP2P] - Mirror Testing Macros For Libp2p

All of the tests in `crates/macro_tests` should either include the lip2p network, or there should be equivalent tests in `crates/testing/tests/libp2p.rs`.   The main reason to split would be that we need a longer timeout for libp2p due to DHT lookups being slow when there are node failures.  See: https://github.com/EspressoSystems/HotShot/issues/2534

- Add the appropriate libp2p tests such that they mirror all other network types
- Fix any issues that arise when running these tests

_No response_

libp2p tests are no different then other network types. 

_No response_

lukaszrzasik:
Closes #2536 

### This PR: 
Enables Libp2p implementation in test macros along side MemoryImpl and WebImpl.
Increases number of nodes to 6 for Libp2pImpl to work.

### Key places to review: 

crates/testing-macros/tests/tests.rs
`Default for TestMetadata` in crates/testing/src/test_builder.rs

### How to test this PR:

`just async_std run_test libp2pimpl_::test_with_failures`


justfile
```diff
@@ -52,6 +52,22 @@ test_success:
...
+  cargo test --lib --bins --tests --benches --workspace --no-fail-fast test_with_failures_half_f -- --test-threads=1 --nocapture
+
+test_with_failures_f:
+  echo Testing f nodes leaving the network
+  cargo test --lib --bins --tests --benches --workspace --no-fail-fast test_with_failures_f -- --test-threads=1 --nocapture
+
+test_with_failures_2:
+  echo Testing that a good leader can succeed in the view directly after view sync
+  cargo test --lib --bins --tests --benches --workspace --no-fail-fast test_with_failures_2 -- --test-threads=1 --nocapture
+
```
bfish713, line 70: I think these are captured in `test_basic`.  We probably don't want to grow the justfile anymore.

To run specific tests you can run `just async_std run_test with_failures` or similar

crates/testing/src/test_builder.rs
```diff
@@ -152,14 +152,14 @@ impl TestMetadata {
             completion_task_description: CompletionTaskDescription::TimeBasedCompletionTaskBuilder(
                 TimeBasedCompletionTaskDescription {
                     // Increase the duration to get the expected number of successful views.
-                    duration: Duration::new(340, 0),
+                    duration: Duration::new(3400, 0),
```
bfish713, line 155: I think we probably should keep these as is.  If the test is taking more than 5 minutes it's probably some kind of deadlock anyway.

crates/testing/src/test_builder.rs
```diff
@@ -152,14 +152,14 @@ impl TestMetadata {
...
                 },
             ),
             overall_safety_properties: OverallSafetyPropertiesDescription {
                 ..Default::default()
             },
             timing_data: TimingData {
-                next_view_timeout: 5000,
+                next_view_timeout: 50000,
```
bfish713, line 162: Same deal here, it's probably fine to keep as is.  If libp2p is specifically slow then we can change the timeout just for those tests, and move them out of macro tests.  E.g. adding the tests to `crates/testing/tests/libp2p.rs` and overriding the default `TimingData` for those tests

bfish713 (approved):

bfish713:
I merged: https://github.com/EspressoSystems/HotShot/pull/2589 with includes these changes so we can close this

PR #2578, opened 2024-02-14
[TECH_DEBT] Remove `CommunicationChannel` trait

### LINKED ISSUE #2572: [TECH_DEBT] - Remove or Refactor `CommunicationChannel`

Context from zulip:
"
What are peoples thoughts on the future or the CommunicationChannel trait? Right now I'm looking at how we should properly handle DA and VID with libp2p and they will need to be handled in a specific way. For DA we probably want to do direct messages to everyone in the comittee for DA proposals. And for VID a first pass will probably be direct message based as well. I think neither of these fit into the orginal CommunicationChannel idea. It is supposed to be an abstraction of a subset of a network where broadcast would send to all members of the subset. For libp2p I don't think that abastraction works nicely because Gossiping to a subset only works if well if you are always a part of that subset. With small DA committees it will mean leaders will have to join the DA committee for one view then leave (except of the rarer cases where the leader is on the comittee). It's even less viable when the comitte starts rotating

Right now all the implementations just wrap ConnectedNetwork, more or less. The one exception is CombinedNetwork which wraps the 2 different ConnectedNetworks. What we end up doing anyway is passing the membership in to broadcast anyway and using that list to decide who to broadcast two. With that pattern the CommunicationChannel does literally nothing. The webserver as well ignores the membership anyway as it's just posting all it's outgoing broadcast to whichever server it's connected to.

The reason this is coming up now is because we need to add an interface for VID share distribution at some point soon, and I don't want to have to change CommunicationChannel trait and add a new function to every impl that is just a wrapper to ConnectedNetwork.

I propose we do either:
1) Remove the trait
2) Somehow incorporate the membership into the channel so that it at least abstracts that part of broadcasting from us.

I prefer removing it, but I'm curious on other thoughts or if it's maybe useful for something I'm missing
"

Decide to remove or refactor the trait and then complete the work.

_No response_

The refactor is done without effecting test behavior

_No response_

bfish713:
Closes #2572 

### This PR: 

- Gets rid of the `CommunicationChannel` and `TestableCommunicationChannel` traits
- adjust the test generator functions such that we don't clone the libp2p layer
- Removes so unused types/parameters for networks which we don't use (vid/viewsync) these just use the quorum network

### This PR does not: 

Again should not alter anything functionally as `CommunicationChannel` was mostly a wrapper that did nothing

### Key places to review: 

The updates to the 4 networking impls in `crates/hotshot/src/traits/networking`


rob-maron (approved):
Looks great, nice work

PR #2588, opened 2024-02-15
[Auto Benchmarks] Add stat collection

### LINKED ISSUE #2345: [Auto Benchmarks] Add stat collection

After the new run metrics/stats are defined in #2343, we need to implement them. We need these to enable more detailed benchmarking runs. 

This involves going through the code and adding counters for the data we want to look at. It does not involve publishing anything to the orchestrator, as that will be solved later.

_No response_

The new decided metrics all have values that can be printed at the end of a benchmarking run.

_No response_

dailinsubjam:
Closes #2345 

### This PR: 
- Tracks a list of stats for benchmarking. For more details:  #2345  .
- Adds arguments  to parameterize benchmarking.
- Adds scripts for benchmarking runs.
- Fixes `readme` for examples.

### This PR does not: 
- Involve publishing anything to the orchestrator, as that will be solved later.
- Include neat output for benchmarking results, it's in the format of error messages now, later we might move to CSV file.
- Track `commit` that benchmarks running, will be added once integrated to CI.
- Track `notes field`, also to be investigated after CI integration.
- Carefully design the parameter we want for each run. The script includes roughly designed runs based on my local results. Also local machines only support limited threads, number of nodes cannot be too large (20 is fine, 50 is not). 
- Consider any VID-related factor.

### Key places to review: 
`crates/examples/infra/mod.rs`

### How to test this PR:
`./scripts/benchmarks.sh`

### Local results example:
With `just async_std example all-webserver -- --config_file ./crates/orchestrator/run-config.toml --total_nodes 20 --da_committee_size 5 --transactions_per_round 10 --transaction_size 512 --rounds 100` I'll get:
```
Failed views: 0, Total views: 100, num_successful_commits: 100
100 rounds completed in 44.871671042s
Avg latency: 15 sec, Minimum latency: 1 sec, Maximum latency: 45 sec, Throughput: 292067 bytes/sec
```


crates/examples/infra/mod.rs
```diff
@@ -830,6 +970,7 @@ pub async fn main_entry_point<
         hotshot,
         &mut transactions,
         transactions_to_send_per_round as u64,
+        (transaction_size + 8) as u64,
```
jparr721, line 973: In the changeset above I see you added a comment to explain the `+8`. I presume that its use here is for the same purpose, however, since it's >300 lines below the initial comment, do you think it'd be worth adding a comment here? Or perhaps making a named constant like (`TIMESTAMP_BYTES_SIZE`)?

jparr721 (commented):
Howdy! I'm the _new guy_. Just had one thing that I was curious about. Feel free to ignore it though if it isn't relevant.

crates/examples/infra/mod.rs
dailinsubjam: Welcome to the team! I should add a comment here, the reason is that `transaction_size` in the config means appending this size of bytes to a default transaction with 8 bytes (see `create_random_transaction`).

bfish713 (commented):
This looks good to me, would be good if someone a bit more familiar with the examples/orchestrator could take a look

scripts/benchmarks.sh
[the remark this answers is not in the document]
dailinsubjam: Yes for certain! We have a script to run locally now, but I'll need to coordinate with integration team to see how scripts should be like to run in AWS.

crates/examples/combined/orchestrator.rs
```diff
@@ -24,7 +24,13 @@ pub mod infra;
 async fn main() {
     setup_logging();
     setup_backtrace();
-    let args = OrchestratorArgs::parse();
-
-    run_orchestrator::<TestTypes, DANetwork, QuorumNetwork, NodeImpl>(args).await;
+    let config = read_orchestrator_init_config::<TestTypes>();
+    let orchestrator_url = Url::parse("http://localhost:4444").unwrap();
```
elliedavidson, line 28: Can we make it so we set the orchestrator url to something we pass in as arguments?  We may need to change the port sometimes when we run on AWS.  We can have it default to localhost:4444 if no arguments are provided though, since having a default will make it more ergonomic.  This is the only change request I have for the PR.

crates/examples/infra/mod.rs
```diff
@@ -85,6 +84,95 @@ pub struct ConfigArgs {
     pub config_file: String,
 }
 
+/// Reads the orchestrator initialization config from the command line
+/// # Panics
+/// If unable to read the config file from the command line
+pub fn read_orchestrator_init_config<TYPES: NodeType>(
```
elliedavidson, line 90: Love this function!

crates/examples/infra/mod.rs
```diff
@@ -460,20 +582,38 @@ pub trait RunDA<
                             // when we make progress, submit new events
                         }
                         EventType::ReplicaViewTimeout { view_number } => {
+                            failed_num_views += 1;
                             warn!("Timed out as a replicas in view {:?}", view_number);
                         }
                         EventType::NextLeaderViewTimeout { view_number } => {
+                            failed_num_views += 1;
```
elliedavidson, line 589: Do we actually ever emit these events?  I don't think we do?  I think they are left over from a long time ago before the event-based model.  (Though that isn't related to this PR)

crates/examples/infra/mod.rs
```diff
@@ -460,20 +582,38 @@ pub trait RunDA<
...
+        let consensus_lock = context.hotshot.get_consensus();
+        let consensus = consensus_lock.read().await;
+        let total_num_views = usize::try_from(consensus.locked_view.get_u64()).unwrap();
+        // When posting to the orchestrator, note that the total number of views also include un-finalized views.
+        error!("Failed views: {failed_num_views}, Total views: {total_num_views}, num_successful_commits: {num_successful_commits}");
+        // +2 is for uncommitted views
+        assert!(total_num_views <= (failed_num_views + num_successful_commits + 2));
         // Output run results
-        let total_time_elapsed = start.elapsed();
+        let total_time_elapsed = start.elapsed(); // in seconds
```
elliedavidson, line 609: Thank you for the "in seconds" comment!

crates/orchestrator/README.md
```diff
@@ -2,6 +2,6 @@
 
 This crate implements an orchestrator that coordinates starting the network with a particular configuration.  It is useful for testing and benchmarking.  Like the web server, the orchestrator is built using [Tide Disco](https://github.com/EspressoSystems/tide-disco).  
 
-To run the orchestrator for a libp2p network: `just async_std example orchestrator-libp2p 0.0.0.0 3333 ./crates/orchestrator/run-config`
+To run the orchestrator for a libp2p network: `just async_std example orchestrator-libp2p http://0.0.0.0:3333 ./crates/orchestrator/run-config.toml`
```
elliedavidson, line 5: Thank you for updating the docs!

scripts/benchmarks.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
elliedavidson, line 1: Love this!

elliedavidson (changes requested):
I'm only requesting one change - that we allow the orchestrator to run on a parametrizable port.  All the other comments are optional changes or just comments.

jparr721 (commented):
A few questions. Feel free to ignore whatever isn't helpful.

crates/example-types/src/block_types.rs
```diff
@@ -169,6 +169,10 @@ impl BlockPayload for TestBlockPayload {
         }
         BuilderCommitment::from_raw_digest(digest.finalize())
     }
+
+    fn get_transactions(&self) -> &Vec<Self::Transaction> {
```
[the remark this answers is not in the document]
elliedavidson: No, we should really standardize it :) We're open to suggestions - personally I like including the "get" prefix.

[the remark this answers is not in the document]
dailinsubjam: Thanks for pointing this out!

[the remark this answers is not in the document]
dailinsubjam: This part is mostly for `BlockPayload`, I only added a single function, so I tend to not change it too much (which is out of scope of this PR). But I'll clean up the name for posting / getting things from orchestrator according to this principle.

crates/examples/infra/mod.rs
```diff
@@ -85,6 +84,95 @@ pub struct ConfigArgs {
...
+            Arg::new("rounds")
+                .short('r')
+                .long("rounds")
+                .value_name("NUM")
+                .help("Sets the number of rounds to run")
+                .required(false),
+        )
+        .get_matches();
+    let mut args = ConfigArgs {
+        config_file: "./crates/orchestrator/run-config.toml".to_string(),
```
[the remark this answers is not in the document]
dailinsubjam: Oh this is for default setting, I'd better remove this.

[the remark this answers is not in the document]
dailinsubjam: BTW, they're all very helpful questions!

crates/examples/infra/mod.rs
dailinsubjam: Yeah I've deleted one EventType, but after taking a close look I also deleted `NextLeaderViewTimeout` since it's never emitted, for others I just keep them. This is also the reason I added `EventType::ViewTimeout` to track failed views, together with `assert!(total_num_views <= (failed_num_views + num_successful_commits + 2))`.

jparr721 (approved):
I'll defer to @elliedavidson before you merge this, but this looks good to me!

bfish713 (approved):
I've been using this locally and having no issues.  I think it's good to go!

elliedavidson (approved):

PR #2589, opened 2024-02-15
Fixes to Not Block Proposing/Fix Test Failures

bfish713:
Closes #https://github.com/EspressoSystems/HotShot/issues/2216

### This PR: 

- Don't block the network task when sending messages.  Libp2p specifically could block for a while on direct message and thus prevent us sending messages on that particular network task
- Sequence transactions even if we don't have the parent leaf.  If a leader misses a view it's possible they have what they need to propose but not a parent block/leaf.  We never actually used this in the transactions task, but if we fail to retreive the parent we still fail for no reason
- Sometimes nodes are slow at the start and get the view 3 proposal before the view 2 stuff, then ignore the old view.  This PR allows that not to cause a failure when they decide a different chain than the others. (just move leaves because they had a gap)
- Send a shutdown event when we get an error in the safety task so it doesn't just keep running until the completion task tells everyone to shutdown

### Key places to review: 

We should expect 100% CI success


crates/testing/src/overall_safety_task.rs
```diff
@@ -159,9 +159,7 @@ impl<TYPES: NodeType, I: TestableNodeImplementation<TYPES>> TestTaskState
                 block_size: maybe_block_size,
             } => {
                 // Skip the genesis leaf.
-                if leaf_chain.len() == 1
-                    && leaf_chain[0].0.get_view_number() == TYPES::Time::genesis()
-                {
+                if leaf_chain.last().unwrap().0.get_view_number() == TYPES::Time::genesis() {
```
shenkeyao, line 162: I'm wondering whether we should still check the length in case somehow the leaf chain contains a non-genesis leaf at the second index.

shenkeyao: Oops just saw that you changed `[0]` to `last()`. Makes sense now!

shenkeyao (approved):

PR #2591, opened 2024-02-15
[Tech Debt] - fix nix flake

DieracDelta:
### This PR: 
Disables checks so we can get the semver binary on darwin (these aren't necessary to get the working binary), and moves the semver to a separate shell so we don't have to build it each time

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -205,6 +205,7 @@
             inherit CARGO_TARGET_DIR;
             buildInputs = [
               (pkgs.cargo-semver-checks.overrideAttrs (final: prev: {doCheck = false;}))
+              fenixStable
             ] ++ buildDeps;
           };
 


flake.nix
```diff
@@ -202,6 +201,13 @@
 
           };
 
+          semverShell = pkgs.mkShell {
+            inherit CARGO_TARGET_DIR;
+            buildInputs = [
```
ss-es, line 206: I think `fenixStable` needs to be added here, otherwise `cargo-semver-checks semver-checks` fails to find the compiler.

```suggestion
            buildInputs = [
              fenixStable
              (pkgs.cargo-semver-checks.overrideAttrs (final: prev: {doCheck = false;}))
            ] ++ buildDeps;
```

ss-es (approved):

PR #2452, opened 2024-01-23
[Tech Debt] Move `task-impls` to hotshot

QuentinI:
Closes #2294

### This PR: 
Demotes `task-impls` from a crate to a part of `hotshot`

### This PR does not: 
include any additional refactorings

### Key places to review: 
`hotshot::tasks` submodule, which is where `task-impls` was moved.

PR #2601, opened 2024-02-16
Unify HotShotConsensusApi into SystemContext

### LINKED ISSUE #2407: [TECH DEBT] - Remove `HotShotConsensusApi`

We have two separate wrappers for `SystemContextInner`: `HotShotConsensusApi` and `SystemContext`. They should be unified.

_No response_

From Zulip:
> Artemii Gerasimovich: QQ: why do we have two different types `SystemContext` and `HotShotConsensusApi` that are structurally the same?
>
> Ellie Davidson: Hotshot consensus api is old and we should get rid of it.

`HotShotConsensusApi` is removed, implementations moved to `SystemContext`.
Tests pass.

_No response_

### LINKED ISSUE #2600: [Upgradability] - Sequencer API

We'd like to allow the sequencer to make decisions about whether HotShot should perform an upgrade. We should add an API to allow it to signal this in `SystemContext`. 

_No response_

_No response_

The sequencer is able to signal to HotShot to initiate an upgrade as a leader.

_No response_

ss-es:
Closes #2407

### This PR: 
`HotShotConsensusApi` is now merged into `SystemContextHandle`. `SystemContext` is removed, and `SystemContextHandle` now contains an `Arc<SystemContextInner>` without an intermediate wrapper.

A `CreateTaskState` trait is introduced for initialization of most HotShot tasks.

### This PR does not: 

`ConsensusApi` is not removed in this PR. This will be done at later date, after #2452 is merged.

### Key places to review:
Tests now use `*::create_from(handle)` to initialize task states; it's probably worth double-checking that this doesn't change any of the tests.


crates/hotshot/src/lib.rs
```diff
@@ -564,16 +563,9 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> SystemContext<TYPES, I> {
...
-#[derive(Clone, Debug)]
-pub struct HotShotConsensusApi<TYPES: NodeType, I: NodeImplementation<TYPES>> {
-    /// Reference to the [`SystemContextInner`]
-    pub inner: Arc<SystemContextInner<TYPES, I>>,
-}
-
 #[async_trait]
 impl<TYPES: NodeType, I: NodeImplementation<TYPES>> ConsensusApi<TYPES, I>
-    for HotShotConsensusApi<TYPES, I>
+    for SystemContext<TYPES, I>
```
bfish713, line 568: I know this is still a draft, but I think we want the handle to impl this trait so we can avoid the whole `c_api = ...` thing.  

Also maybe we want to get rid of the api trait altogether.  Would we ever have another implementation of `ConsensusApi`?

bfish713 (dismissed):
Looks good overall to me!  Small change, but I think I'd prefer `SystemContextInner` to just be `SystemContext`.  Feel free to merge though, no need to hold up over naming

bfish713 (approved):

PR #2603, opened 2024-02-18
Clean py

hkey0:
There is no issue linked to this PR.

### This PR: 
- This PR makes it easier to review the code.
- Applies pep8 improvements to scripts.

### DIFF
--- a/scripts/gen_launch.py
+++ b/scripts/gen_launch.py
@@ -3,103 +3,109 @@
 # USAGE: after changing tests, generate a launch.json configuration with this script
 # ./scripts/gen_launch.py > .vscode/launch.json
 
-import json;
-import copy;
-import subprocess;
+import json
+import copy
+import subprocess
 from typing import Final
 
-def get_launch_json_examples(input: str, crate: str) -> list[dict]:
-    test_names = [];
-    for a_line in input.splitlines():
-        test_names.append(a_line.replace("\"", ""));
 
-    json_lists = [];
+def get_launch_json_examples(input: str, crate: str) -> list[dict]:
+    test_names = [a_line.replace("\"", "") for a_line in input.splitlines()]
+
+    json_lists = []
     base_json = {
-            "type": "lldb",
-            "request": "launch",
-            "cargo" : {
-                "args" : [
+        "type": "lldb",
+        "request": "launch",
+        "cargo": {
+                "args": [
                     "build",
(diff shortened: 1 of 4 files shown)


bfish713:
@DieracDelta Do we use these scripts still, I'm not familiar with what it's for.

scripts/gen_launch.py
```diff
@@ -3,103 +3,109 @@
...
-import json;
-import copy;
-import subprocess;
+import json
+import copy
+import subprocess
 from typing import Final
 
+
 def get_launch_json_examples(input: str, crate: str) -> list[dict]:
```
jparr721, line 12: What are your thoughts on updating the type information for the containers as well? in this case, the type `list[dict]` would use the more conformant `List[Dict[A, B]]` where `A` and `B` are placeholders for what the real types are (looks like `str, str` would suffice?)

scripts/split_test_output.py
```diff
@@ -8,18 +8,19 @@
...
 import re
 
-id_regex = re.compile("id: (\d+)")
+
+id_regex = re.compile("id: (\\d+)")
 ansi_escape = re.compile(r'(\x9B|\x1B\[)[0-?]*[ -\/]*[@-~]')
 
-def split_input(input):
+
+def split_input(input_stream):
```
jparr721, line 18: Thoughts on adding a type to `input_stream` here?

jparr721 (commented):
Howdy! I'm the _new guy_. Normally I don't like to call out code that is not touched in a change set, but it _looks_ like the intention here is to improve the python code overall. Regardless, feel free to ignore these comments if they're irrelevant to the changes here.

scripts/split_test_output.py
hkey0: input is a built-in python function, it is not correct to use built-in python keywords as variable names

scripts/gen_launch.py
hkey0: I changed it to `List[Dict[str, Any]]`. I think this is the right one. Thanks for the review!

scripts/split_test_output.py
jparr721: In this case, I mean adding a type as `input_stream: SomeTypeHere`, not the name itself.

hkey0: Ah sorry, fixed it.

Issue #2573, opened 2024-02-13
[LIBP2P] - DA Broadcast for Small Committes


bfish713:
### What is this task and why do we need to work on it?

Libp2p gossipsub only supports sending to topics which we are subscribed to.  In our code only the committee should always be subscribed to the DA topic.  This leaves us with two choices for sending DA proposals.  The first is a leader can join the topic around the view of their leadership and broadcast to the topic after they've joined.  Then after broadcasting, leave the topic (or else they will get all DA proposals).  It's even more complex and untenable if the comittee rotates often as the gossip network will constantly being reformed.  I thinks this is antithetical to the gossipsub pattern.

The other option is to simply direct message each person of the committee.  This is much simpler and will work with any committee pattern we choose.  It has the drawback for larger committees that the leader must use a lot of bandwidth to get the same thing out to every committee member.  

We should favor direct messages for small committees, and broadcasting with gossipsub for larger comittes if possible.  We can also use a hybrid approach where we use gossipsub if we are already in the committee, and direct messages if not.

In the further future we can look at a protocol where the leader direct messages some in the committee and they gossip to the rest of the committee. This could be complex and should only be an optimization if other approaches are holding us back.

### What are the acceptance criteria to close this issue?

We can run libp2p tests with 20 nodes and 5 in the da committe without issues.

PR #2611, opened 2024-02-19
[LIBP2P] Use Regular Gossip Behaviour

bfish713:
### This PR: 

Gets rid of our own implementation of the `GossipBehaviour`.  This was mostly just a pure wrapper around the libp2p one.  The only thing it provided was retries, but Gossip only fails immediately (on the call to `publish`) and only for reason we either should not retry, require some intervention, or are indicative of a much greater issues.
The Error conditions are:

- `Duplicate` -Retrying is bad, it will just keep failing
- `SigningError(SigningError)` - Same as above
- `InsufficientPeers` - We really should be connected and only retry once we get peers.
- `MessageTooLarge` - Retries will 
- `TransformFailed(std::io::Error)` - Message is bad, retrying won't help, or there is a bug in our transform logic

For these reasons I'm comfortable saying that for the majority of cases we should not even be retrying gossips and am proposing to just use the regular behaviour.  If we want to add the retry logic back I think it should at the very least check the error type.

### This PR does not: 

Besides removing some retry logic, should not functionally change much

### Key places to review: 

Changes to where we use Gossip
Are we ok getting rid of this


maschad:
You may still want to include a retry mechanism for the case  of `InsufficientPeers` in this PR  - akin to what is being done with the [`drain_publish_gossips`](https://github.com/EspressoSystems/HotShot/pull/2611/files#diff-ca9cf2bb601a1cbf33787229164ba4b8ab22026c1fcf4b55fcc12aae1fea9810L256) method during [polling](https://github.com/EspressoSystems/HotShot/pull/2611/files#diff-ca9cf2bb601a1cbf33787229164ba4b8ab22026c1fcf4b55fcc12aae1fea9810L76-L79) because the default gossip behaviour's when publishing [would drop the message](https://github.com/libp2p/rust-libp2p/blob/master/protocols/gossipsub/src/behaviour.rs#L696-L698) otherwise.

That notion is precedented on the idea that one should be monitoring its peers and ensuring it has enough _topic peers_ for its important topics to avoid that error all together - I am not as familiar on the workings of how the topic map is constructed to comment on this

bfish713:
> You may still want to include a retry mechanism for the case of `InsufficientPeers` in this PR - akin to what is being done with the [`drain_publish_gossips`](https://github.com/EspressoSystems/HotShot/pull/2611/files#diff-ca9cf2bb601a1cbf33787229164ba4b8ab22026c1fcf4b55fcc12aae1fea9810L256) method during [polling](https://github.com/EspressoSystems/HotShot/pull/2611/files#diff-ca9cf2bb601a1cbf33787229164ba4b8ab22026c1fcf4b55fcc12aae1fea9810L76-L79) because the default gossip behaviour's when publishing [would drop the message](https://github.com/libp2p/rust-libp2p/blob/master/protocols/gossipsub/src/behaviour.rs#L696-L698) otherwise.
> 
> That notion is precedented on the idea that one should be monitoring its peers and ensuring it has enough _topic peers_ for its important topics to avoid the `InsufficuentPeers` error all together - I am not as familiar with the workings of how the _topic map_ is constructed to comment on this
Thanks for the feedback 

There are only 2 topics currently, and soon to be 1 after https://github.com/EspressoSystems/HotShot/issues/2573.  The one topic is global and all peers should be subscribed, if not they cannot participate in anyway.  If we aren't connected to enough peers then it's likely there is some other problem at play.  Also the things we are gossiping are time sensitive, there is no use sending a proposal after everyone has moved onto the next leader.  

That's not to say we should not retry, just that retrying everything as we do now is probably worse than not retrying at all.  Maybe I didn't look closely enough but I think all those messages will retry forever and the queue will grow infinitely because most of those errors are irreconcilable. 

I think the retry logic should be handled separately somewhere higher in the stack so we can make a better choice about when/what to retry.  The other alternative is to only deal with the error if it's `InsufficientPeers` and retry after some backoff. Open to opinions from others.

maschad:
Thanks for the clarification.

My original proposal was intended to address the fact that the current state of the PR  - if merged, would lead to messages being dropped by just resorting to the default Gossip behaviour, which may have unintended side effects.

> I think the retry logic should be handled separately somewhere higher in the stack so we can make a better choice about when/what to retry. The other alternative is to only deal with the error if it's InsufficientPeers and retry after some backoff. Open to opinions from others.

Good point, but I think you can do both, based on the fact that messages are time-sensitive, my recommendation would be to modify the [ExponentialBackoff struct](https://github.com/EspressoSystems/HotShot/blob/main/crates/libp2p-networking/src/network/behaviours/exponential_backoff.rs#L6) to have an expiry window and flush the `in_progress_gossip` queue if that window has expired. That window would be dictated by the length of the leader's term (which would be retrieved once this is refactored to be higher in the stack).

A retry mechanism of this nature would be particularly important if https://github.com/EspressoSystems/HotShot/issues/2573 is implemented considering you _may_  have a scenario where a leader direct messages every node in a committee.

bfish713:
> Good point, but I think you can do both, based on the fact that messages are time-sensitive, my recommendation would be to modify the [ExponentialBackoff struct](https://github.com/EspressoSystems/HotShot/blob/main/crates/libp2p-networking/src/network/behaviours/exponential_backoff.rs#L6) to have an expiry window and flush the `in_progress_gossip` queue if that window has expired. That window would be dictated by the length of the leader's term (which would be retrieved once this is refactored to be higher in the stack).

We could do that, but I think it's probably better to not have to implement the entire `NetworkBehaviour` trait just to get retries in one specific case.  I agree it makes sense to retry in the `InsufficientPeers` case and to set the time before retrying parameter to make sense with the leader timeout.  I can add that to to this PR.

> A retry mechanism of this nature would be particularly important if https://github.com/EspressoSystems/HotShot/issues/2573 is implemented considering you may have a scenario where a leader direct messages every node in a committee.

Why would this make retrying more important, direct messaging is handled separately from gossipsub and we already handle in our direct message behaviour.

bfish713:
> I'm getting a bit confused about the context for retries. Could someone help me understand _under which case does it make sense to ever retry_?

The only case we'd really want to actually retry is `InsuffcientPeers` error, and where it's important to get the gossip out (CDN is down).  My opinion is that if we start seeing this error pop up in testnet or staging we'll probably need to figure out what is going on and why, seems like simply just trying the request again won't solve much.  

> My (possibly flawed) understanding so far: The gossiped data remains relevant for very little time. On the happy path with a functioning CDN (which should happen most of the time) we won't need retries since we won't be using gossip. And even under the sad path when the CDN is down, if we miss a couple of views while the DHT fixes itself this (1) isn't harmful since it just delays consensus slightly and (2) the CDN shouldn't be down for very long anyway.

It's not entirely true we won't be using gossip, We assume some nodes might be getting censored by the CDN and thus rely on gossip to see Proposals, even when everyone else is using the CDN.  This would allow them to still be able to propose.  However it is really not a big deal if they miss 1 proposal because a leader couldn't gossip.  I do agree that on the sad path if the p2p network gets disconnected it would force view sync and give time for everyone to become connected and reform the gossip mesh.

> My 2 cents: I'm in favor of nuking the NetworkBehaviour implementation, and reintroducing it if we ever need to do something fancy. I like how it isolates the behaviour from the other behaviours. But in this case I'm missing the application and thereby the need for it.

I agree, the question is do we keep my very simple retry logic or just scrap that?

Issue #2504, opened 2024-02-01
[CATCHUP] - Support restarting without losing state


shenkeyao:
### What is this task and why do we need to work on it?

[Paraphrased from Jeb's comment.]

When we have to restart all nodes and they lose their in-memory state, it would be nice to have the restart using saved data from the query service, rather than just restarting in the genesis state. If we could do this we could be much more confident that we won't lose data in decaf, as opposed to trying to ensure we have covered every case where the network might get into a bad state and make sure catchup works if we restart one by one.

We do have [HotShotInitializer::from_reload](https://hotshot.docs.espressosys.com/hotshot/struct.HotShotInitializer.html#method.from_reload), which takes an anchor leaf. It would be easy enough to patch the sequencer node to, e.g., take a "last decided leaf hash" and fetch the full leaf from storage or a query service, and initialize HotShot with that. But it's not used and probably doesn't work.

We would need the last several views from before the shutdown. Potential scenarios:
* We could get into a situation where we have several consecutive successful views, across a shutdown, but no Decide, which could throw off nodes replaying the chain.
* We could also have a scenario where we had some views before the shutdown without a Decide, but where some nodes sent quorum votes or view sync votes. If we started back up in an earlier view, some nodes who lost their state from the undecided views might end up participating in those views again, but then there could be multiple votes floating around the network that could result in a fork or a node getting slashed.

### What work will need to be done to complete this task?

Possible workaround: Let `HotShotInitializer::from_reload` take both anchor leaf and a "current view" parameter, which each node could set to a high enough view that they are confident would not lead to a double vote after the restart. Then view sync would kick in after we restarted and finally, we would resume consensus from a future view, but at least without losing any decided state.

### Are there any other details to include?

Zulip thread: https://espresso.zulipchat.com/#narrow/stream/311646-Consensus/topic/Restarting.20without.20losing.20state/near/418530563.

### What are the acceptance criteria to close this issue?
Either of the following:
* The test should pass on the sequencer side: `cargo test --release --all-features --all-targets -- --nocapture sql::generic_tests::test_restart`.
  * It currently fails probably because the nodes start with parent view 0 even though they were given a leaf from view >0, so they can't find the parent leaf and none of them can propose.
* https://github.com/EspressoSystems/HotShot/issues/2506 is done.

PR #2612, opened 2024-02-20
[CATCHUP] - Support and test restarting without losing state

shenkeyao:
Closes https://github.com/EspressoSystems/HotShot/issues/2504.

### This PR: 
* Supports restarting without losing state.
  * Adds `start_view` to `HotShotInitializer` for restarting.
    * Updates `SystemContext` creation and `from_reload ` accordingly.
  * Fixes `publish_transaction_async` and task state constructions to take the starting view rather than the genesis view.
  * Fixes the subtraction overflow in the consensus task due to restarting.
* Tests restarting.
  * Refactors `TestRunner`.
    * Adds a `skip_late` parameter to indicate whether to skip initializing nodes that will start late.
    * Refactors node adding functions to work for both cases.
  * Refactors `SpinningTask`.
    * Adds `last_decided_leaf` to `SpinningTask` as the potential anchor leaf for catchup.
    * Initializes skipped late-starting nodes when the preset view approaches.
  * Adds a catchup test for reloading rather than restarting from the genesis leaf.
* ~~Renamings.~~
  * ~~Adds `initial` to transaction submission functions to make it clear they are only for the initial view.~~
  * ~~Renames `cur_view` in `Consensus` to `start_view`.~~
* _Update_: Fixes the view numbers for `Consensus` and relevant functions to make transaction submission work for any view.

### Key places to review: 
* `crates/hotshot/src/lib.rs`
* `crates/testing/src/spinning_task.rs`
* `crates/testing/src/test_runner.rs`
* `crates/testing/tests/catchup.rs`


crates/testing/src/test_runner.rs
```diff
@@ -110,10 +123,15 @@ where
...
+    ///
+    /// # Arguments
+    /// *  `skip_late` - Whether to skip initializing nodes that will start late, which will catch
+    /// up later with `HotShotInitializer::from_reload` in the spinning task.
+    ///
     /// # Panics
     /// if the test fails
     #[allow(clippy::too_many_lines)]
-    pub async fn run_test(mut self) {
+    pub async fn run_test(mut self, skip_late: bool) {
```
bfish713, line 134: I think this should be a parameter in the test metadata instead of an argument here

crates/hotshot/src/lib.rs
```diff
@@ -288,22 +292,25 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> SystemContext<TYPES, I> {
...
+    /// Publishes an initial transaction asynchronously to the network.
+    ///
+    /// The transaction corresponds to either the genesis view or the view to restart on.
     ///
     /// # Errors
     ///
     /// Always returns Ok; does not return an error if the transaction couldn't be published to the network
     #[instrument(skip(self), err)]
-    pub async fn publish_transaction_async(
+    pub async fn publish_initial_transaction_async(
```
bfish713, line 303: I think adding initial to the name makes it feel like this is something you should only call once for a first transaction.  It's still the main transaction submission path.  I think what we should do is fix `consensus.curr_view`.  We can updated it in the consensus task and read it instead of `start_view` here.

I actually noticed during some libp2p stuff that transactions gossips get rejected in tests because they are duplicate messages.  It's because the test transaction payload is always the same, and now seeing that view 0 was hard coded, the view is clearly also always the same.  Would be nice if it updated roughly with consensus to  avoid this issue.

crates/hotshot/src/lib.rs
```diff
@@ -641,14 +651,24 @@ impl<TYPES: NodeType> HotShotInitializer<TYPES> {
...
-    pub fn from_reload(anchor_leaf: Leaf<TYPES>, instance_state: TYPES::InstanceState) -> Self {
+    /// Reload previous state based on most recent leaf and the instance-level state.
+    ///
+    /// # Arguments
+    /// *  `start_view` - The minimum view number that we are confident won't lead to a double vote
+    /// after restart.
+    pub fn from_reload(
+        anchor_leaf: Leaf<TYPES>,
+        instance_state: TYPES::InstanceState,
+        start_view: TYPES::Time,
```
bfish713, line 666: Would we ever set `start_view` to anything but `anchor_leaf.get_view_number()` or `anchor_leaf.get_view_number() + 1`?

crates/testing/tests/catchup.rs
```diff
@@ -233,6 +233,70 @@ async fn test_catchup_in_view_sync() {
...
+    metadata.completion_task_description =
+        CompletionTaskDescription::TimeBasedCompletionTaskBuilder(
+            TimeBasedCompletionTaskDescription {
+                duration: Duration::from_secs(60),
+            },
+        );
+    metadata.overall_safety_properties = OverallSafetyPropertiesDescription {
+        // Make sure we keep commiting rounds after the catchup, but not the full 50.
+        num_successful_views: 22,
+        num_failed_views: 5,
```
bfish713, line 292: Shouldn't we expect no failures in this test?

crates/testing/tests/catchup.rs
```diff
@@ -233,6 +233,70 @@ async fn test_catchup_in_view_sync() {
...
+
+// Almost the same as `test_catchup`, but with catchup nodes reloaded from anchor leaf rather than
+// initialized from genesis.
+#[cfg(test)]
+#[cfg_attr(
+    async_executor_impl = "tokio",
+    tokio::test(flavor = "multi_thread", worker_threads = 2)
+)]
+#[cfg_attr(async_executor_impl = "async-std", async_std::test)]
+async fn test_catchup_reload() {
```
bfish713, line 248: Love that we can test this so easily!

crates/testing/tests/catchup.rs
bfish713, line 1: As a follow up can we try to somehow add one more test where more than a quorum goes down, then the all restart from the last decided view and progress continues?

crates/hotshot/src/lib.rs
jbearer: Yes. Suppose anchor_leaf has view 100, but when your node shut down the latest view was 200: there were many views where you voted but never reached decide. If you restart in view 100 or 101, you may end up voting again in views 101-200, while there were already votes with your signature floating around for those views. This can end up with your node getting slashed.

So I think it is important that consensus works even if `start_view` is much greater than `anchor_leaf.get_view_number()` (as long as a quorum of nodes agree that nothing has been committed between `anchor_leaf` and `start_view`

crates/hotshot/src/lib.rs
```diff
@@ -641,14 +651,24 @@ impl<TYPES: NodeType> HotShotInitializer<TYPES> {
         Ok(Self {
```
jbearer, line 651: I can't comment on the line but while we're here, can we make `from_genesis` ^ take `instance_state` by value to avoid unnecessary clone and be consistent with `from_reload`?

crates/hotshot/src/lib.rs
shenkeyao: The transaction submission functions are currently only used for the first view, so I thought it was okay to add `initial` to the names. But I do think it's a better idea to fix `cur_view` instead. Will do that!

crates/testing/tests/catchup.rs
shenkeyao: Oh, this construction was copied from `test_catchup` and I didn't modify the parameters. I'll change it to `0`.

crates/testing/tests/catchup.rs
shenkeyao: Sounds good. Made an issue: https://github.com/EspressoSystems/HotShot/issues/2617.

shenkeyao:
@bfish713 @jbearer No need to re-review now. Two tests failed in the self-hosted CI. I'll investigate!

shenkeyao:
> @bfish713 @jbearer No need to re-review now. Two tests failed in the self-hosted CI. I'll investigate!

Ok, they are now passed (hopefully deterministically)!

jbearer (approved):

PR #2616, opened 2024-02-20
Adds DA committee broadcast TransmitType

### LINKED ISSUE #2573: [LIBP2P] - DA Broadcast for Small Committes

lukaszrzasik:
Closes #2573

### This PR: 

Adds new TransmitType::DACommitteeBroadcast. Its purpose is broadcasting a message to DA committee only.
For memory and web implementation this can be achieved using normal broadcast_message method.
Libp2p now internally uses direct messaging to send the message to the DA committee members.

### Key places to review: 

Files:
crates/task-impls/src/network.rs
crates/hotshot/src/traits/networking/libp2p_network.rs

### DIFF
--- a/crates/hotshot/src/traits/networking/libp2p_network.rs
+++ b/crates/hotshot/src/traits/networking/libp2p_network.rs
@@ -748,8 +748,8 @@ impl<M: NetworkMsg, K: SignatureKey + 'static> ConnectedNetwork<M, K> for Libp2p
         recipients: BTreeSet<K>,
     ) -> Result<(), NetworkError> {
         let future_results = recipients
-            .iter()
-            .map(|r| self.direct_message(message.clone(), r.clone()));
+            .into_iter()
+            .map(|r| self.direct_message(message.clone(), r));
         let results = join_all(future_results).await;
 
         let errors: Vec<_> = results


crates/hotshot/src/traits/networking/libp2p_network.rs
```diff
@@ -740,6 +741,32 @@ impl<M: NetworkMsg, K: SignatureKey + 'static> ConnectedNetwork<M, K> for Libp2p
...
     }
 
+    #[instrument(name = "Libp2pNetwork::da_broadcast_message", skip_all)]
+    async fn da_broadcast_message(
+        &self,
+        message: M,
+        recipients: BTreeSet<K>,
+    ) -> Result<(), NetworkError> {
+        let future_results = recipients
+            .iter()
```
bfish713, line 751: small nit: we can also into_iter this so we don't have to clone r

bfish713 (approved):
Looks good, just one tiny change

bfish713 (approved):

PR #2640, opened 2024-02-23
[CHORE] Update Jellyfish

jbearer:
* Pin to version (0.4.0)
* Update VID construction
* Brings in persistent merkle tree for use in sequencer


bfish713 (approved):

bfish713:
Thank you for doing this!

PR #2642, opened 2024-02-23
Add tests for the upgrade task

### LINKED ISSUE #2628: [Upgradability] - Test that upgrades correctly update the internal consensus state

As a first test for the upgrade process, we would like to test that the internal consensus state gets updated correctly.

_No response_

_No response_

There should be tests that ensure that

- an invalid `UpgradeCertificate` attached to a `QuorumProposal` does not cause `ConsensusTaskState` to update,
- a valid `UpgradeCertificate` attached to a `QuorumProposal` correctly updates `ConsensusTaskState`.

and these tests should pass.

_No response_

ss-es:
Closes #2628

### This PR:
- Adds a stronger testing harness for task logic in `testing/src/script.rs`, with some utilities in `testing/src/predicates.rs`.
- Adds an iterator to generate test `QuourmProposal`s, `DACertificate`s and `VidProposal`s for multiple views in `testing/src/view_generator.rs`.
- Adds a test to validate that the consensus task correctly updates its internal state when receiving (and later deciding) a `QuorumProposal` with a valid `UpgradeCertificate` attached in `testing/tests/upgrade_task.rs`.
- Corrects previously faulty logic in the consensus task that used the `UpgradeCertificate` attached to the newly received `QuorumProposal`, rather than the decided leaf.

### This PR does not: 
There is an open question on whether `UpgradeCertificate`s should be included in the leaf (and therefore go into the leaf commitment), or not. Right now, they are not captured by the commitment, but we should probably discuss this during the next consensus meeting (and possibly fix in the future).


crates/testing/src/view_generator.rs
```diff
@@ -0,0 +1,304 @@
...
+        <TestTypes as NodeType>::SignatureKey,
+    ),
+    pub leader_public_key: <TestTypes as NodeType>::SignatureKey,
+    pub da_certificate: DACertificate<TestTypes>,
+    pub transactions: Vec<TestTransaction>,
+    upgrade_data: Option<UpgradeProposalData<TestTypes>>,
+}
+
+impl TestView {
+    pub fn genesis(quorum_membership: &<TestTypes as NodeType>::Membership) -> Self {
```
bfish713, line 47: This assumes DA = Quorum, I think for unit tests that's fine, but worth noting that if we ever want to test some scenario where the committee is smaller than the quorum we'll need both memberships here.

bfish713 (commented):
This looks fantastic.  Excited to use this for other tests!

Issue #2618, opened 2024-02-20
[KEY MANAGEMENT] - Combine `get_config` with `post_and_wait_all_public_keys`


jbearer:
### What is this task and why do we need to work on it?

The current `OrchestratorClient` interface is awkward: we get a config, then use that config to call `post_and_wait_all_public_keys` to get a new config, and then we have to update the new config with the `node_index` from the first config because of a bug in `post_and_wait_all_public_keys`.

It would be much simpler to have a single interface which takes a public key, posts it to the orchestrator, and returns a config with all public keys and node index set correctly.

### What work will need to be done to complete this task?

We don't need to change anything with the orchestrator itself. This can all be done as a new interface on the client side.

### What are the acceptance criteria to close this issue?

Orchestrator set up including public key distribution requires a single `OrchestratorClient` call, and the resulting config does not need to be modified at all.

PR #2656, opened 2024-02-26
[KEY MANAGEMENT] Add state pub keys to HotshotConfig

### LINKED ISSUE #2599: [KEY MANAGEMENT] - Add state pub keys to HotShotConfig

Currently sequencer nodes need the set of public staking keys _and_ public state keys, to compute the initial stake table commitment. The orchestrator distributes staking keys by having all nodes post their public key using `post_and_wait_all_public_keys` and returning an updated config with `known_nodes_with_stake` set appropriately. We need something similar for state signing keys.

Add `state_keys` to `HotShotConfig` _or_ add a state schnorr key to the stake table entry, so that the state keys are available from `known_nodes_with_stake`. Add a schnorr ver key as an input to `post_and_wait_all_public_keys`. Have the orchestrator populate the state keys in the config from the inputs posted via `post_and_wait_all_public_keys`.

_No response_

All nodes receive from the orchestrator a config listing state keys that the nodes generated themselves.

_No response_

dailinsubjam:
Closes #2599 , #2618 .

### This PR: 
- Add state pub keys to HotShotConfig.
- Post serialization of state pub keys to orchestrator.
- Refactor functions of getting config from orchestrator. The one for getting complete config with given ValidatorConfig is called `get_complete_config`.
- Rename old `get_config` for orchestrator to `get_config_without_peer`.

### Key places to review: 
- @mrain Check out `crates/types/src/lib.rs/PeerConfig`. It's using `bincode_opts().serialize()` rather than Canonical Serialization for both state keys and signature keys. Is this okay? Do we want to also use canonical serialization for state keys?
- @jbearer @imabdulbasit Check out `crates/examples/infra/mod.rs`. The function `get_complete_config` will take in a ValidatorConfig, which includes public & private signature keys, stake value and state public & private keys. The private part will be saved in local HotshotConfig, and the public part will be posted to orchestrator.

### How to test this PR: 


crates/orchestrator/src/config.rs
```diff
@@ -208,27 +208,75 @@ impl<K: SignatureKey, E: ElectionConfig> NetworkConfig<K, E> {
...
+        let cur_node_index = client.get_node_index_for_init_validator_config().await;
+        ValidatorConfig::generated_from_seed_indexed([0u8; 32], cur_node_index.into(), 1)
+    }
+
+    /// Asynchronously retrieves a `NetworkConfig` from an orchestrator.
+    /// The retrieved one includes correct `node_index` and peer's public config.
+    pub async fn get_complete_config(
+        client: &OrchestratorClient,
+        my_own_validator_config: ValidatorConfig<K>,
+        file: Option<String>,
```
jbearer, line 249: Just FYI, I think the sequencer is _always_ going to use this with `file: None`, loading the config from the orchestrator. We manually check if the config already exists before calling this function, because depending on the sequencer config, we may check either a file or Postgres, which Hotshot doesn't support.

I think this is fine though, looking at the logic of this function. Setting `file` to `None` basically just means "get this config from the orchestrator" which is exactly what we want

jbearer (approved):
Interface to sequencer looks good

Issue #2654, opened 2024-02-23
[CATCHUP] - Fix state initialization upon restarting


shenkeyao:
### What is this task and why do we need to work on it?

We use `from_header` to construct the validated state when creating the `SystemContext`, which means we always start with an incomplete state. We should replace it with a complete state whenever we can.

### What work will need to be done to complete this task?

* Add an `Option<ValidatedState>` to `HotShotInitializer` and `from_reload`.
* Set the state to `Some(ValidatedState::genesis())` in `from_genesis`.
* In `SystemContext::new`, use the genesis state or the application-specified state if given. Otherwise, use `from_header`.

### What are the acceptance criteria to close this issue?

The application can set a validated state and we have a complete state is restarting from genesis.

PR #2663, opened 2024-02-26
[CATCHUP] - Fix state initialization upon restarting

shenkeyao:
Closes https://github.com/EspressoSystems/HotShot/issues/2654.

### This PR: 
* Add an optional validated state to `HotShotInitializer`.
  * Sets this to the genesis state if restarting from genesis.
  * Supports an app-specific state if reloading.
* Creates the state via `from_header` only if it's not given.

### Key places to review: 

* `crates/hotshot/src/lib.rs`.


crates/hotshot/src/lib.rs
```diff
@@ -620,10 +632,13 @@ impl<TYPES: NodeType> HotShotInitializer<TYPES> {
     /// initialize from genesis
     /// # Errors
     /// If we are unable to apply the genesis block to the default state
-    pub fn from_genesis(instance_state: TYPES::InstanceState) -> Result<Self, HotShotError<TYPES>> {
+    pub fn from_genesis(
+        instance_state: &TYPES::InstanceState,
```
jbearer, line 636: Why switch this back to a reference? We're still consuming it in this function. Just need to use it to construct the validated state *before* consuming it.

jbearer (approved):

crates/hotshot/src/lib.rs
shenkeyao: Yeah, you are right! I should have just constructed the state first.

jbearer (approved):

PR #2666, opened 2024-02-27
Adding vbs support

nyospe:
Closes https://github.com/EspressoSystems/versioned-binary-serialization/issues/6

### This PR: 
- Enables [`versioned-binary-serialization`](https://github.com/EspressoSystems/versioned-binary-serialization) in HotShot, including support through tide-disco and surf-disco.
- Sets up a path for making HotShot generic over updatability versioning in the application layer.

### This PR does not: 
- replace the hardcoded version 0.1 that was added for Updatability

### Key places to review: 
- all implementators of `ConnectedNetwork`; note that the versions of `WebServerNetwork` and `CombinedNetworks` are specific to that network, and not the same as the versions of payloads (`broadcast_message`, `direct_message`) over those networks.
- network control messages for `Libp2pNetwork` - I'm not sure I got that right, and it might need its own unique version for those.

 https://github.com/EspressoSystems/versioned-binary-serialization/issues/6


crates/hotshot-qc/src/bit_vector_old.rs
```diff
@@ -263,12 +263,12 @@ mod tests {
             // Check the QC and the QCParams can be serialized / deserialized
             assert_eq!(
                 qc,
-                bincode::deserialize(&bincode::serialize(&qc).unwrap()).unwrap()
+                Serializer::<0, 1>::deserialize(&Serializer::<0, 1>::serialize(&qc).unwrap()).unwrap()
```
ss-es, line 266: Since this a round-trip tests that should really work in any version, should this (and L271 in the same file) be `Serializer::CURRENT_STATIC_VERSION`? (Should we have a `CURRENT_STATIC_VERSION`? We could alias it to `STATIC_VERSION_0_1`, but then we wouldn't have to update it here.)

crates/hotshot-qc/src/bit_vector.rs
```diff
@@ -274,14 +274,14 @@ mod tests {
             // Check the QC and the QCParams can be serialized / deserialized
             assert_eq!(
                 qc,
-                bincode::deserialize(&bincode::serialize(&qc).unwrap()).unwrap()
+                Serializer::<0, 1>::deserialize(&Serializer::<0, 1>::serialize(&qc).unwrap()).unwrap()
```
ss-es, line 277: Same question here and L284 about using current version.

crates/hotshot-qc/src/bit_vector_old.rs
ss-es, line 266: Same question here and L271 about using current version

crates/hotshot/src/traits/networking/libp2p_network.rs
```diff
@@ -418,7 +426,11 @@ impl<M: NetworkMsg, K: SignatureKey + 'static> Libp2pNetwork<M, K> {
     }
 
     /// Initiates connection to the outside world
-    fn spawn_connect(&mut self, id: usize) {
+    fn spawn_connect<const MAJOR: u16, const MINOR: u16>(
+        &mut self,
+        id: usize,
+        bind_version: StaticVersion<MAJOR, MINOR>,
```
ss-es, line 434: question to check my understanding: we're just using `bind_version` just to propagate the `MAJOR` and `MINOR` types down the chain right?

crates/hotshot/src/traits/networking/memory_network.rs
```diff
@@ -147,7 +146,7 @@ impl<M: NetworkMsg, K: SignatureKey> MemoryNetwork<M, K> {
                         Combo::Direct(vec) => {
                             trace!(?vec, "Incoming direct message");
                             // Attempt to decode message
-                            let x = bincode_opts().deserialize(&vec);
+                            let x = Serializer::<0, 1>::deserialize(&vec);
```
ss-es, line 149: I think this (and L168) hard-codes the version, though the function isn't explicitly versioned.

I think either `Serializer::CURRENT_STATIC_VERSION`  or `new_0_1` (or similar) just to indicate that the function is explicitly versioned.

crates/libp2p-networking/tests/counter.rs
```diff
@@ -64,7 +64,7 @@ pub async fn counter_handle_network_event(
     match event {
         IsBootstrapped => {}
         GossipMsg(m, _) | DirectResponse(m, _) => {
-            if let Ok(msg) = bincode_opts().deserialize::<CounterMessage>(&m) {
+            if let Ok(msg) = Serializer::<0, 1>::deserialize::<CounterMessage>(&m) {
```
ss-es, line 67: I think this (+ the other `<0,1>` and `STATIC_V_0_1` in this file) are hardcoded to a version, even though the function they're called in are not explicitly versioned.

crates/orchestrator/src/lib.rs
```diff
@@ -30,6 +30,13 @@ use libp2p::identity::{
     ed25519::{Keypair as EdKeypair, SecretKey},
     Keypair,
 };
+
+/// Orchestrator is not, strictly speaking, bound to the network; it can have its own versioning.
+/// Orchestrator Version (major)
+pub const ORCHESTRATOR_MAJOR: u16 = 0;
```
ss-es, line 36: nit/minor (feel free to ignore): I think this should be `ORCHESTRATOR_MAJOR_VERSION`

crates/testing/tests/memory_network.rs
```diff
@@ -184,7 +187,7 @@ async fn memory_network_direct_queue() {
     // Send messages
     for sent_message in first_messages {
         network1
-            .direct_message(sent_message.clone(), pub_key_2)
+            .direct_message(sent_message.clone(), pub_key_2, STATIC_V_0_1)
```
ss-es, line 186: Same comments re: hard-coding versions without versioning function

crates/testing/tests/unit/message.rs
```diff
@@ -46,17 +43,16 @@ fn version_number_at_start_of_serialization() {
...
     };
-    let serialized_message: Vec<u8> = bincode_opts().serialize(&message).unwrap();
+    let serialized_message: Vec<u8> = Serializer::<MAJOR, MINOR>::serialize(&message).unwrap();
     // The versions we've read from the message
-    let major_version_read = u16::from_le_bytes(serialized_message[..2].try_into().unwrap());
-    let minor_version_read = u16::from_le_bytes(serialized_message[2..4].try_into().unwrap());
 
-    assert_eq!(version.major, major_version_read);
-    assert_eq!(version.minor, minor_version_read);
+    let version_read = Version::deserialize(&serialized_message).unwrap().0;
```
ss-es, line 54: I think this actually changes the test -- it's now testing a property of the `Serializer` (round-trip version de/serialize = identity), where before it was testing that the binary representation actually had the version in the right place.

I think the previous test was good to have somewhere, though you probably already have these tests in `vbs`. I think we could just delete this file?

crates/types/src/message.rs
```diff
@@ -34,9 +33,6 @@ use std::{fmt::Debug, marker::PhantomData};
 #[derive(Serialize, Deserialize, Clone, Debug, Derivative, PartialEq, Eq, Hash)]
 #[serde(bound(deserialize = "", serialize = ""))]
 pub struct Message<TYPES: NodeType> {
```
ss-es, line 35: Just checking for my own understanding: `NodeType` is versioned right? (So `Message` is implicitly versioned through that?)

ss-es (commented):
I think this makes sense to me overall -- left some minor comments + questions to check my understanding!

crates/constants/src/lib.rs
```diff
@@ -1,6 +1,6 @@
 //! configurable constants for hotshot
 
-use serde::{Deserialize, Serialize};
+use versioned_binary_serialization::version::{StaticVersion, Version};
```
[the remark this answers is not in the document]
nyospe: Yes. We need the type-level version in a number of places... if Rust's const generics were more mature, I'd be using it as a binding in place of the MAJOR/MINOR thing as well... but having a value Version is also useful in a number of places.

crates/constants/src/lib.rs
```diff
@@ -17,17 +17,18 @@ pub const COMBINED_NETWORK_MIN_PRIMARY_FAILURES: u64 = 5;
...
-    pub major: u16,
-    /// minor version number
-    pub minor: u16,
-}
-
 /// Constant for protocol version 0.1.
 pub const VERSION_0_1: Version = Version { major: 0, minor: 1 };
 
+/// Constant for protocol static version 0.1.
+pub const STATIC_V_0_1: StaticVersion<0, 1> = StaticVersion;
```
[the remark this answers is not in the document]
nyospe: It got really verbose in places, I wanted a shorter name. But the `_0_1` thing is going to go away as soon as I finish working out a few more things, in favor of generic `_BASE` and `_UPDATE` containers, and there will be a `::StaticVersion` on those.

crates/example-types/src/node_types.rs
```diff
@@ -44,6 +47,10 @@ impl NodeType for TestTypes {
     type ValidatedState = TestValidatedState;
     type InstanceState = TestInstanceState;
     type Membership = GeneralStaticCommittee<TestTypes, Self::SignatureKey>;
+    const MAJOR_V_OLD: u16 = 0;
+    const MINOR_V_OLD: u16 = 0;
```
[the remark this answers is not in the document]
nyospe: That's a valid point... though this particular field is no longer present. It was a leftover from an attempt to hoist generic parameterization from associated constants, which is, it turns out, only available with unstable features.

crates/hotshot-qc/src/bit_vector_old.rs
nyospe: There does need to be a `CURRENT_BASE_VERSION`, and a `CURRENT_UPGRADE_VERSION`, and a compile-time check to determine if they are the same (e.g. no upgrade available), but aliasing won't work. I'm working on a solution using a generic type to impl on the core types, but I haven't worked it out yet. I left this (and a few other places) as-is because they're not actually hooked up to the real versioning.

crates/libp2p-networking/tests/counter.rs
nyospe: ... this one, specifically, might be a problem. I was less stringent about using real versions in tests, but I'm not sure this is uniform in all directions. I'll take a closer look.

crates/testing/tests/unit/message.rs
nyospe: Yes, there's tests in vbs... This probably should be deleted.

crates/types/src/message.rs
nyospe: Not right now, no. I'm working on a way to do that...

nyospe: Sorry, a bit of clarification. Nothing implements the trait `Versioned`, which has been added in `vbs`; it was meant to be added alongside Serialize and Deserialize as a feature macro, e.g.

```rust
#[derive(Serialize, Deserialize, Versioned)]
#[base_version(1, 3)]
#[max_version(1, 7)]
struct MyStruct {
}
```

but I've come to realize that we usually don't really know the actual version in crates that have generic types...

We may end up needing a crate-level version and a mechanism for mapping the versions of crates to the actual version, but that's not trivial. Another solution might be a tagged, but never versioned, repository containing a single "version" crate that gets bound in the executable target's `Cargo.lock`, and injects versions into our library crates.

nyospe: `Message` is not inherently bound to a specific protocol version, though a change to `Message` will force versioning once `Versioned` is implemented. Not by way of `NodeType`, though.

I'm still trying to come up with a nice way to inject versions into the consensus by way of something coupled to `NodeType`, but giving it generic parameters is catastrophically contagious.

crates/hotshot/src/traits/networking/memory_network.rs
nyospe: On re-examination, this is, effectively, a `MEMORY_NET_VERSION_MAJOR`/`MEMORY_NET_VERSION_MINOR` situation...

crates/constants/src/lib.rs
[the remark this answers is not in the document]
nyospe: Is this better?

bfish713 (approved):

Issue #2662, opened 2024-02-26
[VALIDATED_STATE] - Give proposal and validation functions access to parent leaf


jbearer:
### What is this task and why do we need to work on it?

Sequencer state catchup needs access to the parent view number so we know for which view to fetch missing state from a peer. The leaf contains this information, as well as the parent header which we are already receiving.

### What work will need to be done to complete this task?

Replace the `parent_header` argument of `Header::new` and `validate_and_apply_header` with `parent_leaf`.

### What are the acceptance criteria to close this issue?

The task is done

PR #2674, opened 2024-02-28
[VALIDATED_STATE] - Give proposal and validation functions access to parent leaf

shenkeyao:
Closes https://github.com/EspressoSystems/HotShot/issues/2662.

### This PR: 
* Replaces the `parent_header` argument with `parent_leaf` in `Header::new` and `validate_and_apply_header`.
* Adds `NodeType` generic type to the block header and validated state traits and testing structs.

### Key places to review: 
* `crates/types/src/traits/block_contents.rs`.
* `crates/types/src/traits/states.rs`.

### DIFF
--- a/crates/example-types/src/block_types.rs
+++ b/crates/example-types/src/block_types.rs
@@ -3,19 +3,19 @@ use std::{
     mem::size_of,
 };
 
+use crate::{node_types::TestTypes, state_types::TestValidatedState};
 use commit::{Commitment, Committable, RawCommitmentBuilder};
 use hotshot_types::{
-    data::{BlockError, VidCommitment, VidScheme, VidSchemeTrait},
+    data::{BlockError, Leaf, VidCommitment, VidScheme, VidSchemeTrait},
     traits::{
         block_contents::{vid_commitment, BlockHeader, TestableBlock, Transaction},
+        node_implementation::NodeType,
         BlockPayload, ValidatedState,
     },
     utils::BuilderCommitment,
 };
 use serde::{Deserialize, Serialize};
 use sha3::{Digest, Keccak256};
-
-use crate::state_types::TestValidatedState;
 
 /// The transaction in a [`TestBlockPayload`].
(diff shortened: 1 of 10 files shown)


crates/example-types/src/block_types.rs
```diff
@@ -180,25 +180,25 @@ pub struct TestBlockHeader {
     pub payload_commitment: VidCommitment,
 }
 
-impl BlockHeader for TestBlockHeader {
+impl<TYPES: NodeType> BlockHeader<TYPES> for TestBlockHeader {
     type Payload = TestBlockPayload;
     type State = TestValidatedState;
```
jbearer, line 185: I wonder if we can remove either of these associated types and refer to them as `TYPES::Something` now

crates/task-impls/src/consensus.rs
```diff
@@ -618,7 +620,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
                 };
                 let Ok(state) = parent_state.validate_and_apply_header(
                     &consensus.instance_state,
-                    &parent_leaf.block_header.clone(),
+                    &parent_leaf.clone(),
```
jbearer, line 623: Why are we cloning this if we're passing it in by reference?

jbearer (approved):

crates/example-types/src/block_types.rs
shenkeyao: I think we can! Let me try this.

crates/task-impls/src/consensus.rs
shenkeyao: I don't the cloning is necessary. I'll try removing it.

crates/example-types/src/block_types.rs
shenkeyao: Also removed these types from `ValidatedState`.

jbearer (approved):

PR #2675, opened 2024-02-28
Cache async-std and tokio builds separately

ss-es:
No linked issue.

### This PR: 
Makes `just *` commands use different directories for `async-std` and `tokio`. `cargo` should now cache these separately, avoiding wiping the cache when switching async executors.

### DIFF
--- a/.github/workflows/build-and-test-self-hosted.yml
+++ b/.github/workflows/build-and-test-self-hosted.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       matrix:
         just_variants:
-          - async_std
+          - async-std
           #- tokio
     runs-on: [self-hosted]
     steps:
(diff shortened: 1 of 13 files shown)


bfish713 (changes requested):
I think this did break the build dockers workflow.  It's expecting the build to just be in target/

ss-es:
@bfish713 I think it's working now. I replaced every `target/` with `target/${{ matrix.just_variant }}`; I think some of these might not have been necessary, but I don't think it hurts.

Since the async_std executor is actually called `async-std` (hyphen not underscore), I added a just variant `async-std` as a copy of `async_std`. The workflow now uses the version with the hyphen. Unfortunately, I don't think there's a way to avoid this without complicated logic either in the `just cargo` command or the workflow files, but this shouldn't affect anyone's local workflow.

bfish713 (approved):

PR #2677, opened 2024-02-29
Prevent CI getting cancelled on main (again)

ss-es:
No linked issue.

### This PR: 
Makes the concurrency group by `run_number` on main (which is unique per run), and by `ref` on other branches.

This means that every commit on main should get its own unique set of CI runs, which should be completely independent from every other run.

### Key places to review: 
`.github/workflows`


ss-es:
Redoing this because I think the last PR didn't quite fix the issue. See e.g. c35466ad1e138b173b178c68e9ff587020cadca3 on main, where only 6 CI tests ran.

I think what happened was: there was a PR merged at 10:05, then this one merged at 10:12 and another one merged right after 10:12. so the one in the middle got skipped over.

This is still a draft -- will request review when ready.

bfish713 (approved):

PR #2678, opened 2024-02-29
Add update_view in ConnectedNetwork trait

### LINKED ISSUE #2645: [LIBP2P] - Add Hook Update view to `ConnectedNetwork`

We need some way to know when to stop trying to send messages for a view.  This is specifically useful for libp2p but also could apply to other networks where requesting data is relevant.  The steps for this are pretty simple:

- [ ] Add a function called `update_view` to `ConnectedNetwork` trait 
- [ ] `NetworkEventTask` should handle `ViewChange` events and call `update_view` on the `ConnectedNetwork` it's using
- [ ] Implement for each network (probably just an empty fucntion

_No response_

_No response_

Check that the view number updates are hitting the new function with reasonable views, probably by logging the view in tests.

_No response_

### LINKED ISSUE #2646: [LIBP2P] - Delay Sending over libp2p in `CombinedNetwork`

In the combined network we don't want to have to send every message over the fallback.  We should wait a configurable timeout before sending the following messages:

- `DAProposal`
- `DACertificate`
- `VIDShares`
- `DAVote`
- `QuorumVote`

As well should store these delayed sends by view number.  When the `update_view` function we added in #2645 is called we can cancel all the send tasks for views less than the new view.

I imagine the could would be something like (probably won't compile but this is the gist of it):

```
let view = message.get_view_number();
primary_network.broadcast(message).await;
let delayed = async_spawn( async move {
    async_sleep(delay).await;
    self.fallback_network.broadcast(message).await;
});
let entry = self.delayed_map.entry(view).or_default();
*entry.push_back(delayed);
```

_No response_

_No response_

We can pass all tests with the combined network, even when the CDN is down.  We should also check that this improves benchmarks with the combined network.  

_No response_

lukaszrzasik:
Closes #2645
Closes #2646

### This PR: 
* Adds a `update_view` method to the `ConnectedNetwork` trait and implements a default behaviour
* Delays sending messages through a secondary network in `CombinedNetworks` if they match the specified message type
* Cancels sending messages if a view update for a newer view has been received

### Key places to review: 

* `update_view` implementation in `CombinedNetworks`
* `should_delay` in `CombinedNetworks`
* `send_both_networks` in `CombinedNetworks`
* the default delay duration value: 1000 ms

### How to test this PR:

* `test_combined_network_webserver_crash`

### DIFF
--- a/crates/hotshot/src/traits/networking/combined_network.rs
+++ b/crates/hotshot/src/traits/networking/combined_network.rs
@@ -118,7 +118,7 @@ pub fn calculate_hash_of<T: Hash>(t: &T) -> u64 {
 }
 
 /// thread-safe ref counted lock to a map of delayed tasks
-type DelayedTasksLockedMap = Arc<RwLock<BTreeMap<u64, JoinHandle<Result<(), NetworkError>>>>>;
+type DelayedTasksLockedMap = Arc<RwLock<BTreeMap<u64, Vec<JoinHandle<Result<(), NetworkError>>>>>>;
 
 /// A communication channel with 2 networks, where we can fall back to the slower network if the
 /// primary fails
@@ -206,13 +206,15 @@ impl<TYPES: NodeType> CombinedNetworks<TYPES> {
 
         if !primary_failed && Self::should_delay(&message) {
             let duration = *self.delay_duration.read().await;
-            self.delayed_tasks.write().await.insert(
-                message.kind.get_view_number().get_u64(),
-                async_spawn(async move {
+            self.delayed_tasks
+                .write()
+                .await
+                .entry(message.kind.get_view_number().get_u64())
+                .or_default()
+                .push(async_spawn(async move {
                     async_sleep(duration).await;
                     secondary_future.await
-                }),
-            );
+                }));
             Ok(())
         } else {
             secondary_future.await
@@ -465,10 +467,11 @@ impl<TYPES: NodeType> ConnectedNetwork<Message<TYPES>, TYPES::SignatureKey>
         let mut cancel_tasks = Vec::new();
         {
             let mut map_lock = self.delayed_tasks.write().await;
-            while let Some((first_view, _task)) = map_lock.first_key_value() {
+            while let Some((first_view, _tasks)) = map_lock.first_key_value() {
                 if first_view < view {
-                    if let Some((_view, task)) = map_lock.pop_first() {
-                        cancel_tasks.push(cancel_task(task));
+                    if let Some((_view, tasks)) = map_lock.pop_first() {
+                        let mut ctasks = tasks.into_iter().map(cancel_task).collect();
+                        cancel_tasks.append(&mut ctasks);
                     } else {
                         break;
                     }


crates/hotshot/src/traits/networking/combined_network.rs
```diff
@@ -381,6 +456,31 @@ impl<TYPES: NodeType> ConnectedNetwork<Message<TYPES>, TYPES::SignatureKey>
         <Libp2pNetwork<_, _> as ConnectedNetwork<Message<TYPES>,TYPES::SignatureKey>>::
             inject_consensus_info(self.secondary(), event).await;
     }
+
+    async fn update_view(&self, view: &u64) {
+        let mut map_lock = self.delayed_tasks.write().await;
+        while let Some((first_view, _task)) = map_lock.first_key_value() {
+            if first_view < view {
+                if let Some((_view, task)) = map_lock.pop_first() {
+                    cancel_task(task).await;
```
bfish713, line 465: I think we do not want to do other async actions while holding the write lock.  It can lead to a deadlock.  I think it's ok here.  It's probably safer to aggregate the tasks to cancel and then cancel them.

crates/hotshot/src/traits/networking/combined_network.rs
```diff
@@ -139,6 +162,105 @@ impl<TYPES: NodeType> CombinedNetworks<TYPES> {
...
+        }
+    }
+
+    /// a helper function to send messages through both networks (possibly delayed)
+    async fn send_both_networks(
+        &self,
+        message: Message<TYPES>,
+        recipients: BTreeSet<TYPES::SignatureKey>,
+        message_type: MessageType,
+    ) -> Result<(), NetworkError> {
```
bfish713, line 185: I wonder if instead of adding this `MessageType` enum and having match here, you could pass this function two futures called `primary` and `backup`. 
maybe
```rust
async fn send_both_networks(
    &self, primary: impl Future<Output = Result<(), NetworkError>>, 
    backup: impl Future<Output = Result<(), NetworkError>>)
```

Then the call would be like:
```rust
self.send_both_networks(
    self.primary().broadcast_message(message.clone(), recipients.clone(), 
    self.secondary().broadcast_message(...)).await
```

bfish713: Not 100% sure this works, but something like that should

crates/hotshot/src/traits/networking/combined_network.rs
```diff
@@ -381,6 +463,27 @@ impl<TYPES: NodeType> ConnectedNetwork<Message<TYPES>, TYPES::SignatureKey>
...
+                    }
+                } else {
+                    break;
+                }
+            }
+        }
+        for task in cancel_tasks {
+            cancel_task(task).await;
+        }
+    }
```
bfish713, line 486: let's do `futures::future::join_all(cancel_tasks).await` to minimize blocking

crates/types/src/traits/network.rs
```diff
@@ -299,6 +300,12 @@ pub trait ConnectedNetwork<M: NetworkMsg, K: SignatureKey + 'static>:
     /// blocking
     /// Ideally we would pass in the `Time` type, but that requires making the entire trait generic over NodeType
     async fn inject_consensus_info(&self, _event: ConsensusIntentEvent<K>) {}
+
+    /// handles view update
+    #[instrument(name = "ConnectedNetwork::update_view", skip(self))]
+    async fn update_view(&self, _view: &u64) {
+        info!("Entered update_view");
```
bfish713, line 307: nit: lets lower this to debug! or remove it.

bfish713 (commented):
Just 2 very small comments, then we can merge

bfish713 (approved):

bfish713 (approved):

PR #2679, opened 2024-02-29
[Libp2p] Add Request Response for Data to networking layer

### LINKED ISSUE #2647: [LIBP2P] - RequestResponse Protocol for VID Shares

Create a new behaviour for requesting VID shares.  We want this so nodes can signal that they need information rather than having the leader broadcast to everyone on the broadcast channel.  This should save us a ton of bandwidth/messages on the happy path.

The behaviour should have requests which contain a VID Commitment and HotShot Public key and responses which return contain the VID share for that commitment and public key.

Just create the behaviour and hook it up with the swarm.  The follow up is to send requests/responses

_No response_

We have a testable behaviour for getting VID shares by request.

_No response_

bfish713:
Closes #2647 

### This PR: 

Goal of this PR is to add a request response mechanism to the libp2p layer and exposed it via `ConnectedNetwork`.  The tricky part of this was figuring out a way for the network layer and consensus to interact nicely.  

My first pass at this I wanted the network layer to completely contain the tasks for doing requesting and responding.  I thought maybe we could inject some consensus info into the network layer however I didn't see a way to make this work because the `ConnectedNetwork` interface is completely unaware of consensus.  It might make sense to keep the responding to requests internal by injecting the consensus state map (wrapped in `Arc<RwLock>`), I think that's just a small interface change.  It also might be possible to have the task which requests data also live in the libp2p network layer.  This one would require the network layer either getting events from Hotshot, or validating messages on it's own.  I didn't see a great path forward for either.

This PR does the following:

- Adds to the `Message` structure a new data messages for requesting data and sending responses to those requests
- Adds 2 new functions to `ConnectedNetwork` for requesting data, and for receiving requests to respond to.
- Adds new behaviour to libp2p stack.  This is the standard request response behaviour, with no retries
- Adds handling of all the events from the new behaviour

One thing to note is that `request_data` returns results directly to the caller.  It's not fire and forget.  This is done so it's easy to implement a retry mechanism where we iterate through different peers

### This PR does not: 

We don't yet use the request response.  On top of this we can build 2 new modules.  One will a new task that handles incoming requests for data by looking for the it in our consensus data and then responding.  The other will be a task which requests missing data based on the consensus messages we are receiving.  

### Key places to review: 

The new behaviour
The flow of events through the libp2p stack
The interface additions to ConnectedNetwork.


rob-maron (commented):
The changes look good, but some general questions (since I'm a bit out of date with the plan for VID share dissemination):
- How will nodes know which VID share to request and when? Will we do something similar to the polling model we have currently?
- Won't we be sending direct messages with the share data to all VID participants? Would we save bandwidth by switching to this request/response model?

bfish713:
> The changes look good, but some general questions (since I'm a bit out of date with the plan for VID share dissemination):
> 
> * How will nodes know which VID share to request and when? Will we do something similar to the polling model we have currently?
> * Won't we be sending direct messages with the share data to all VID participants? Would we save bandwidth by switching to this request/response model?

Good questions, The request response plan is detailed here: https://www.notion.so/espressosys/Libp2p-Fallback-cea28d2cf806466da0699a658ed312eb.

You can roughly how the requests will work in this new draft PR: https://github.com/EspressoSystems/HotShot/pull/2698.  

To more specifically answer both questsions:

- Nodes will spawn a request task just like other network tasks, this task will receive `QuorumProposalValidated` events (a new event emitted after QC verification on new proposals).  Then the task will wait some delay before sending the request.  Delay will be a configuration option.  If the data is available in our state at after the delay we'll cancel sending the request.  Same thing for trying other peers, we will stop if the data is in our state.  So we don't need any special injection of consensus events.
- After this is merged we actually want to remove the direct sending of shares to everyone.  Combined with a tuned delay and CDN it should be the case that the majority of times no VID shares are sent over request response.  In the cases where many nodes do need to ask for the shares the load will be spread to the committee (nodes will request the leader or random committee member) which should help our scaling by 1 order of magnitude (assuming 10 nodes in a DA committee).

Side note:  The DA committee will have to calculate the VID shares from the block data for this to work.

rob-maron (approved):

Issue #2648, opened 2024-02-23
[LIBP2P] - Request VID Shares after Receiving a Proposal on Libp2p


bfish713:
### What is this task and why do we need to work on it?

After: #2647 we want to stop sending VID shares except via request.  This task is to use the new behaviour to retrieve VID shares when we need them.   We can use receiving the proposal on the fallback path as a trigger to send, or some delay after getting the proposal from the CDN.  We should never start a new request for a stale view, or for a share we already have.  Assuming neither of those two conditions get met we should wait a long delay after receiving the proposal from the CDN and a shorter delay after receiving from libp2p, whichever comes sooner, then request.

### What work will need to be done to complete this task?

1. Add logic to trigger a delay request for VID shares in the libp2p `ConnectedNetwork` impl.
2. Add logic to cancel those delayed sends if the view changes, or we get a share.  This will involve mapping the delays by view and vid commitment
3. Add logic to get the response (VID Share) to the hotshot layer.
4. Tune the timeouts such that we have enough time to get the VID share (it takes a while because the leader must calculated it then send it) but no so long that we slow views down too much under the sad path.

### Are there any other details to include?

When this task is complete we still won't have a response mechanism, so we can't remove the delayed sending of VID shares from the leader.  I think the best way to test this is to log when the request gets trigger, and when we get the VID share from the CDN.  This way we can see if we are requesting too eagerly, or to conservatively.  In other words, if the request is getting triggered alot, and the CDN share comes in not long after we know it's requesting to aggressively.

### What are the acceptance criteria to close this issue?

We can show that we request the right shares when the CDN is down, but don't request when the CDN has good latency for us.

PR #2691, opened 2024-03-04
Add Event for Validated Quorum Proposal

bfish713:
### This PR: 

Adding a new event before I do: #2648.  I want to use this event to drive the request task.  We'll wait some time after getting the `QuorumProposalValidated` event and then try to request VID if we don't get it sooner.

I'm making the assumption that if we couldn't vote on the proposal we don't deem it valid.  A proposal that doesn't have a parent in our state map would fit this criteria, but it could otherwise be valid.  This fits my purpose because we only need would request the data to be able to vote.  This assumption might not be correct if others use this event.

I also think this event will be useful for the task cleanup work.

### This PR does not: 

We don't yet use the event. 

### Key places to review: 

New event, and the place it's emitted.

### DIFF
--- a/crates/task-impls/src/consensus.rs
+++ b/crates/task-impls/src/consensus.rs
@@ -698,6 +698,12 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> + 
                         },
                     })
                     .await;
+                // Notify other tasks
+                broadcast_event(
+                    HotShotEvent::QuorumProposalValidated(proposal.data.clone()),
+                    &event_stream,
+                )
+                .await;
 
                 let mut new_anchor_view = consensus.last_decided_view;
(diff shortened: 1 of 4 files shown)


shenkeyao (approved):
Looks good! The following change mentioned in the description isn't made yet, but I'm assuming this will be done in the next PR?
> We'll wait some time after getting the QuorumProposalValidated event and then try to request VID if we don't get it sooner.

bfish713:
> Looks good! The following change mentioned in the description isn't made yet, but I'm assuming this will be done in the next PR?
> 
> > We'll wait some time after getting the QuorumProposalValidated event and then try to request VID if we don't get it sooner.

Yeah that's what I was meaning to say.  In the future I'll use this to trigger requests on the network

ss-es (approved):
No comments - looks good to me!

PR #2696, opened 2024-03-05
Add timestamp to `TestBlockHeader`

jbearer:
This is extremely useful for downstream testing. For example, many tests in the query service rely on a timestamp. Without this change, we would have to implement our own test types or implement something very unrealistic (like always getting the current time every time we get the timestamp for the same header).

This is also pretty realistic, as any real application will probably timestamp its blocks, and it demonstrates getting inputs in `Header::new` from the outside world (ie the hardware clock) in addition to just the inputs which are passed in, which is good for example purposes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2820,6 +2820,7 @@ dependencies = [
  "sha2 0.10.8",
  "sha3",
  "snafu",
+ "time 0.3.34",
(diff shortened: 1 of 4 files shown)


shenkeyao (approved):
LGTM

PR #2697, opened 2024-03-05
[DRAFT] Implement Testing for Regression in Consensus Event Ordering

### LINKED ISSUE #2309: [STABILITY] - Test Leader Proposes with all Event Orderings

We've seen bugs in testnets where we don't propose if the consensus tasks gets events in the wrong order (where any order should be fine).  Currently there are 2 or 3 events that need to be received for the leader to propose: `SendPayloadCommitmentAndMetadata`, `QCFormed` (either QC or TC), and `QuorumProposalRecv` for the previous view if a QC was formed by us for it.  We had a bug where `SendPayloadCommitmentAndMetadata` was the last of theses events to be received and we didn't propose.  We need regressions tests agains such scenarios

Create a unit test for all possible orderings and be sure that the consensus task proposes when it's the leader.

_No response_

All those unit tests pass.

_No response_

jparr721:
Closes #2309 

### This PR: 
From the issue:

>We've seen bugs in testnets where we don't propose if the consensus tasks gets events in the wrong order (where any order should be fine).  Currently there are 2 or 3 events that need to be received for the leader to propose:
`SendPayloadCommitmentAndMetadata`, `QCFormed` (either QC or TC), and `QuorumProposalRecv` for the previous view if a QC was formed by us for it.  We had a bug where `SendPayloadCommitmentAndMetadata` was the last of theses >events to be received and we didn't propose.  We need regressions tests against such scenarios.

This PR implements testing to detect regressions when such events occur out of some pre-defined order. In the working case, the code should be able to handle events in any order and successfully propose no matter what. 

### This PR does not: 
This PR does not attempt to fix any existing issues within the core logic, any such changes will be subject to later patches.

### Key places to review: 
All test cases.


crates/testing/tests/proposal_ordering.rs
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    // We need a couple of events to occur for the leader to propose:
+    // `SendPayloadCommitmentAndMetadata`, `QCFormed` (either QC or TC), and `QuorumProposalRecv`
+    // for the previous view if a QC was formed by us for it.
+
+    // First, receive a proposal and move to view 1, this should correspond with the node id.
+    let view_1 = TestScriptStage {
+        inputs: vec![
+            QCFormed(either::Left(qc.clone())),
```
jparr721, line 66: The idea here was to send the initial `QCFormed` message to start consensus up.

crates/testing/tests/proposal_ordering.rs
```diff
@@ -0,0 +1,87 @@
...
+            QuorumProposalRecv(proposals[0].clone(), leaders[0]),
+        ],
+        outputs: vec![
+            exact(QuorumProposalSend(proposals[0].clone(), public_key)),
```
jparr721, line 70: It seemed as though the next step, as the node starting consensus, is to then have emitted a new proposal to the network.

crates/testing/tests/proposal_ordering.rs
```diff
@@ -0,0 +1,87 @@
...
+            exact(ViewChange(ViewNumber::new(node_id))),
```
jparr721, line 71: Following Receiving the proposal (is it the same one we just sent above?), the view changes to 1. In this setup, would the view change also imply which node is the leader here?

crates/testing/tests/proposal_ordering.rs
```diff
@@ -0,0 +1,110 @@
...
+        ],
+        asserts: vec![is_at_view_number(1)],
+    };
+
+    // Node 2 is the leader up next, so we form the QC for it.
+    let cert = proposals[1].data.justify_qc.clone();
+    let view_1_inputs = if qc_formed_first {
+        vec![
+            QuorumProposalRecv(proposals[1].clone(), leaders[1]),
+            QCFormed(either::Left(cert)),
```
jparr721, line 63: Not sure how to do this without first receiving a proposal and getting the view change. I _could_ do this in a separate stage, but it would still create the case where QuorumProposalRecv needs to be in a consistent place in the vec (namely, the first).

crates/testing/src/predicates.rs
```diff
@@ -76,3 +89,10 @@ pub fn decided_upgrade_cert() -> Predicate<ConsensusTaskTestState> {
         "expected decided_upgrade_cert to be Some(_)",
     )
 }
+
+pub fn is_at_view_number(n: u64) -> Predicate<ConsensusTaskTestState> {
```
jparr721, line 93: This allows us to easily verify that the view is at a particular node leader for a given step, which is helpful for debugging but also knowing when a proposal happens and enforcing who the leader was during the proposal.

crates/testing/tests/proposal_ordering.rs
```diff
@@ -0,0 +1,107 @@
...
+    let mut view_1_inputs = vec![
+        QuorumProposalRecv(proposals[1].clone(), leaders[1]),
+        QCFormed(either::Left(cert)),
+        SendPayloadCommitmentAndMetadata(payload_commitment, (), ViewNumber::new(node_id)),
+    ];
+
+    // Swap QCFormed and SendPayloadCommitmentAndMetadata
+    if !qc_formed_first {
+        view_1_inputs.swap(1, 2);
+    }
```
bfish713, line 69: I think the most interesting test is actually swapping the `QuorumProposalRecv` event to be received second or third out of all the events.

crates/testing/tests/proposal_ordering.rs
```diff
@@ -0,0 +1,107 @@
...
+    tokio::test(flavor = "multi_thread", worker_threads = 2)
+)]
+#[cfg_attr(async_executor_impl = "async-std", async_std::test)]
+/// A leader node may receive one of a couple of possible events which can trigger a proposal. This
+/// test ensures that, no matter what order these events are received in, the node will still
+/// trigger the proposal event regardless. This is to catch a regression in which
+/// `SendPayloadCommitmentAndMetadata`, when received last, resulted in no proposal occurring.
+async fn test_proposal_ordering() {
+    test_ordering_with_specific_order(true /* qc_formed_first */).await;
+    test_ordering_with_specific_order(false /* qc_formed_first */).await;
```
bfish713, line 106: Like I mentioned before I think there should be a test for every possible ordering of the 3 events.  So 6 tests here.

bfish713 (commented):
Looks good! just need to add a few more permutations to the test

crates/testing/tests/proposal_ordering.rs
jparr721: Done!

crates/testing/tests/proposal_ordering.rs
jparr721: Done!

crates/testing/tests/proposal_ordering.rs
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    // Prevent a double free. Rust's borrow checker will eventually drop `input` since it is
+    // consumed by this call. However, since we pulled the memory out from underneath the data
+    // structure, we will get an error if we don't explicitly blow the memory away before it goes
+    // out of scope. This ensures that the borrow checker won't get mad at us when it finds out how
+    // we've left the place.
+    unsafe {
+        input.set_len(0);
+    }
```
bfish713, line 36: I don't think we need or want unsafe code here.  Do we need to permute the outputs?

jparr721: This appears to indeed be required, but I would be open to discussing this further. So for these inputs and outputs:
```rust
    let inputs = vec![
        QuorumProposalRecv(proposals[1].clone(), leaders[1]),
        QCFormed(either::Left(cert)),
        SendPayloadCommitmentAndMetadata(payload_commitment, (), ViewNumber::new(node_id)),
    ];
    let view_1_outputs = vec![
        exact(ViewChange(ViewNumber::new(2))),
        exact(QuorumProposalValidated(proposals[1].data.clone())),
        quorum_proposal_send(),
    ];
```
When the QuorumProposal comes last, the outputs must be in this order:
```rust
    let view_1_outputs = vec![
        quorum_proposal_send(),
        exact(ViewChange(ViewNumber::new(2))),
        exact(QuorumProposalValidated(proposals[1].data.clone())),
    ];
```
And this is the required order for any inputs where `QuorumProposalRecv` comes last. I'm trying to trace through the task, but I'll be honest that I do not know the root cause here. Is this indicative of a problem? Or am I just misunderstanding what it takes to get this to work properly?

bfish713: I think my confusion is coming from the fact that the outputs are ordered.  I was thinking they could be in any order for a stage.

jparr721: Spoke offline, opted to hard-code the two annoying cases instead of making a bespoke solution.

bfish713 (approved):

PR #2698, opened 2024-03-05
[NETWORK] Add Request Task

### LINKED ISSUE #2650: [LIBP2P] - VID shares can be Requested from any DA committee member

We don't want to put too much burden on the leader when the CDN is down.  We should instead spread the distribution load around the network a little.  Instead of requesting just the leader when we need a VID share the replica should randomly select from the set of DA Committee + Leader to request, until it gets the share it wants.

- [ ] Add random selecting of a peer to request from the DA Committee + leader set
- [ ] Add timeouts to the request and goto the next random peer if a request times out or we receive `NotFound`

_No response_

DA members get requested and return VID shares, not just the leader.

_No response_

bfish713:
Closes #2650, #2648
Depends on #https://github.com/EspressoSystems/HotShot/pull/2689

### This PR: 

- Adds a new task type for requesting data from the network.
  - Requests are sent after a delay, iteratively, to all the DA members in a random order 
  - Requests timeout after 500ms if nothing is received.
  - We keep trying new peers until our state has what we need, the view has moved on, or we tried everyone
  - Responses are sent back to 
- Removes the sender from the `VIDDisperseRecv` event because we already verify that the disperse is from the leader by verifying the signature.

### This PR does not: 

The next PR will integrate the request and response tasks, and test the work

### Key places to review: 

The new task


crates/task-impls/src/request.rs
```diff
@@ -0,0 +1,261 @@
...
+        request: RequestKind<TYPES>,
+        sender: Sender<HotShotEvent<TYPES>>,
+        view: TYPES::Time,
+    ) {
+        let mut recipients: Vec<_> = self
+            .da_membership
+            .get_whole_committee(view)
+            .into_iter()
+            .collect();
+        recipients.shuffle(&mut thread_rng());
```
jparr721, line 138: Any change we could get a comment explaining what's going on here? It's not immediately obvious to me why we shuffle the recipients.

crates/task-impls/src/request.rs
```diff
@@ -0,0 +1,261 @@
...
+        let mut reqs = Vec::new();
+        if !self.state.read().await.vid_shares.contains_key(&view) {
+            reqs.push(RequestKind::VID(view, self.public_key.clone()));
+        }
+        // TODO request other things
+        reqs
+    }
+
+    /// run a delayed request task for a request.  The first response
+    /// recieved will be send over `sender`
```
jparr721, line 126: Is this a typo?
```suggestion
    /// recieved will be sent over `sender`
```

crates/task-impls/src/request.rs
```diff
@@ -0,0 +1,261 @@
...
+                        ResponseMessage::Denied => {
+                            error!("Request for data was denied by the receiver");
+                        }
+                    }
+                }
+                Ok(Err(e)) => {
+                    warn!("Error Sending request.  Error: {:?}", e);
+                }
+                Err(_) => {
+                    warn!("Request to other node timed out");
```
jparr721, line 221: Should these be error logs?

jparr721 (commented):
A couple of questions

crates/task-impls/src/request.rs
bfish713: I think in the spirit of reducing our error logs warn is ok here.  These errors are hit when we can't reach one specific peer.  It's abnormal, but expected.  The above `error!` log for Denied is error because we know we have stake and a valid key so something is wrong if we get that.

jparr721 (approved):
Great work

PR #2699, opened 2024-03-05
[Auto Benchmarks] Post benchmark stats to orchestrator

### LINKED ISSUE #2346: [Auto benchmarks] Publish stats to orchestrator

For automated benchmarks, at the end of a run, we want to post stats we care about to the orchestrator. This allows us to use a centralized repository for metrics and metric comparisons. It will also help in the case we want to automatically create something like a CSV from them.

We need to post stats to the orchestrator. This will likely involve:
- [x] adding a metrics endpoint to the orchestrator
- [x] adding a post at the end of the examples for the stats
- [x] format output into a csv file

_No response_

metrics for a run can all be printed and aggregated from the orchestrator

_No response_

dailinsubjam:
Closes #2346 .

### This PR: 
- adds a metrics endpoint to the orchestrator.
- adds a post at the end of the examples for the stats.
- formats output into a csv file `scripts/benchmarks_results/results.csv`.
- adds an end point for run commit, it will `GITHUB_SHA` later.

### This PR does not: 
- consider any VID-related factor yet.
- start CI integration yet, as this will be the next step.
- consider saving past benchmark results yet, as this will be solved after integration.

### Key places to review: 
All changes.

### How to test this PR:
`just async_std example all-webserver -- --config_file ./crates/orchestrator/run-config.toml --commit_sha [ANY STRING]` and check out `scripts/benchmarks_results/results.csv`.

### DIFF
--- a/crates/orchestrator/src/lib.rs
+++ b/crates/orchestrator/src/lib.rs
@@ -293,16 +293,34 @@ where
             } else {
                 // Deal with the bench results from different nodes
                 let cur_metrics = self.bench_results.clone();
-                self.bench_results.avg_latency_in_sec = (metrics.avg_latency_in_sec * metrics.num_latency + cur_metrics.avg_latency_in_sec * cur_metrics.num_latency) / (metrics.num_latency + cur_metrics.num_latency);
+                self.bench_results.avg_latency_in_sec = (metrics.avg_latency_in_sec
+                    * metrics.num_latency
+                    + cur_metrics.avg_latency_in_sec * cur_metrics.num_latency)
+                    / (metrics.num_latency + cur_metrics.num_latency);
                 self.bench_results.num_latency += metrics.num_latency;
-                self.bench_results.minimum_latency_in_sec = metrics.minimum_latency_in_sec.min(cur_metrics.minimum_latency_in_sec);
-                self.bench_results.maximum_latency_in_sec = metrics.maximum_latency_in_sec.max(cur_metrics.maximum_latency_in_sec);
-                self.bench_results.throughput_bytes_per_sec = metrics.throughput_bytes_per_sec.max(cur_metrics.throughput_bytes_per_sec);
-                self.bench_results.total_transactions_committed = metrics.total_transactions_committed.max(cur_metrics.total_transactions_committed);
-                assert_eq!(metrics.transaction_size_in_bytes, cur_metrics.transaction_size_in_bytes);
-                self.bench_results.total_time_elapsed_in_sec = metrics.total_time_elapsed_in_sec.max(cur_metrics.total_time_elapsed_in_sec);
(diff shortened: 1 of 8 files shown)


crates/orchestrator/src/lib.rs
```diff
@@ -82,8 +88,36 @@ impl<KEY: SignatureKey + 'static, ELECTION: ElectionConfig + 'static>
...
     }
+
+    /// Output the results to a csv file according to orchestrator state
+    pub fn output_to_csv(&self) {
+        let output_csv = BenchResultsDownloadConfig {
+            commit_sha: self.config.commit_sha.clone(),
+            total_nodes: self.config.config.total_nodes.into(),
+            da_committee_size: self.config.config.da_committee_size,
+            transactions_per_round: self.config.transactions_per_round,
+            transaction_size: self.config.transaction_size,
```
elliedavidson, line 103: Do we need to add 8 bytes here for the timestamp too?

elliedavidson (approved):
Just one question, but if it doesn't need a change then I approve.  This looks great!

scripts/benchmarks_results/results.csv
```diff
@@ -0,0 +1,2 @@
+commit_sha,total_nodes,da_committee_size,transactions_per_round,transaction_size,rounds,leader_election_type,avg_latency_in_sec,minimum_latency_in_sec,maximum_latency_in_sec,throughput_bytes_per_sec,total_transactions_committed,total_time_elapsed_in_sec,total_num_views,failed_num_views
```
jparr721, line 1: Is this intended to be checked in?

crates/orchestrator/src/lib.rs
dailinsubjam: Oh good catch! Let me do some changes.

scripts/benchmarks_results/results.csv
dailinsubjam: Yes! If you want to do test, you can run the command `just async_std example all-webserver -- --config_file ./crates/orchestrator/run-config.toml --commit_sha [ANY STRING]` and then check this file to see whether it reflects the right config setting.

crates/orchestrator/src/lib.rs
jparr721: Seems like having a lot of `+ 8`'s everywhere could potentially introduce some issues down the road. For example, if I were asked to edit this code, and had not been in this conversation, I may have missed the transaction_size change and introduced a bug here. Is there perhaps a way to signify that transaction_size must ALWAYS have that + 8 for the timestamp? This is a bit of a meta-discussion, so I'm happy to punt this to later, but I thought it might be worth at least documenting (however ephemeral review comments are).

dailinsubjam: Yes! So I did an update on simply changing to `self.bench_results.transaction_size_in_bytes`, and we don't need `+8` elsewhere.

PR #2700, opened 2024-03-05
[Cleanup] Remove BoxedSync `RecvMsgs`

rob-maron:
### This PR: 

Changes `recv_msgs` in the network trait from using `BoxedSync` to being an async function returning the value directly. Since we never use it in sync code, this is okay. This is just one of the changes made in the Push CDN branch, but put into a smaller PR to make more sense.

### This PR does not: 

Change any message receiving logic

### Key places to review: 

`network.rs`


bfish713 (approved):
This looks good to me, but @DieracDelta do you have any objections?  I wasn't sure why the trait returned the box future anyway

PR #2714, opened 2024-03-06
Test upgrade and consensus task together

### LINKED ISSUE #2693: [Upgradability] - Add tests for the upgrade and consensus task together

We should test that the upgrade and consensus task work together correctly.

Minimally, there should be tests that simulate receiving signed votes from all other nodes in the network and ensure an `UpgradeCertificate` is formed and attached to the next quorum proposal.

Part of this task will involve extending the newly-added test script so that it run multiple tasks at once.

_No response_

Tests added and pass.

_No response_

ss-es:
Closes #2693

### This PR: 
- Adds a macro to script and test multiple tasks together (`macros/src/lib.rs`)
- Adds a test for the upgrade and consensus task, where `UpgradeVote`s are received to form an `UpgradeCertificate`, which should be attached to the next `QuorumProposal` sent.

### Key places to review: 
`macros/src/lib.rs` and `testing/tests/upgrade_task.rs`.

You can run the added test with: `RUST_LOG=debug just cargo test test_upgrade_and_consensus_task -- --nocapture`


crates/task-impls/src/vote.rs
```diff
@@ -73,6 +73,7 @@ impl<
         event_stream: &Sender<HotShotEvent<TYPES>>,
     ) -> Option<HotShotTaskCompleted> {
         if vote.get_leader(&self.membership) != self.public_key {
+            error!("Received vote for a view in which we were not the leader.");
```
ss-es, line 76: I don't think this should be `error!`, but I'm just matching the other `error!` in this function (which I also don't think should be `error!`).

It might be nice if we had some rules for when to `error!`, `info!`, or `debug!`.

crates/testing/src/script.rs
```diff
@@ -66,41 +106,86 @@ pub async fn run_test_script<TYPES, S: TaskState<Event = HotShotEvent<TYPES>>>(
...
-                assert!(
-                    (expected.function)(&received_output),
-                    "Stage {} | Output failed to satisfy {:?}",
-                    stage_number,
-                    expected
-                );
+                validate_output_or_panic(stage_number, &received_output, assert);
             } else {
-                panic!("{}", output_missing_error);
+                panic_missing_output(stage_number, assert);
```
jparr721, line 114: This is nice, thanks for adding these as named panics.

jparr721 (dismissed):
Really great work. Just one small comment here.

crates/macros/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+            let task_state_asserts = &#task_expectations[stage_number].task_state_asserts;
+
+            for assert in task_state_asserts {
+                validate_task_state_or_panic_in_script(stage_number, #script_names.to_string(), #task_names.state(), assert);
+            }
+        )*
+    } }
+
+    };
+    // println!("{expanded}");
```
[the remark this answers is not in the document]
ss-es: Oops -- yes, leftover from debugging!

jparr721 (approved):

ss-es:
@jparr721 Sorry for all the re-requests -- there were merge conflicts when CI passed!

bfish713 (approved):

PR #2716, opened 2024-03-07
Jparr721/refactor consensus task tests

jparr721:
Depends on: https://github.com/EspressoSystems/HotShot/pull/2697

Does not Close an Issue

### This PR: 
I have some other Consensus Task tests to add, so I refactored this file to use the new testing paradigm as an in-between PR. I took the new test harness and generator and swapped them in for the `input` `output` manual construction that was there before. I copied the commands almost verbatim.

### Key places to review: 
The changes to the test, I want to make sure that I didn't accidentally wipe away specific things that this test was checking for.

 ### How to test this PR:
`cargo test -p hotshot-testing --test consensus_task`


ss-es:
I think this looks fine to me.

My only concern is: we did have a failure (today) in `test_consensus_task`: https://github.com/EspressoSystems/HotShot/actions/runs/8192042140/job/22402665222#step:7:1863 and this was non-deterministic (and very rare). 

I don't know if the failure is exposing a timing issue in the old test harness (which is fine) or the consensus logic (bad), but I want to suggest we keep the old tests around for a while (especially since we'll be adjusting the log level soon, which should give us a much better idea of why it failed). 

I think we could have e.g. `test_consensus_task_old`, with a comment to delete it once we know why that failure happened, or if the consensus task changes and the test needs to be rewritten or anyway, or a few months have passed without the failure showing up again.

ss-es (approved):

PR #2717, opened 2024-03-07
[NETWORKING] Integrate Request and Response Tasks

### LINKED ISSUE #2650: [LIBP2P] - VID shares can be Requested from any DA committee member

bfish713:
Closes #2650 

### This PR: 

- Adds a request delay to the HotShotConfig
- Creates a request task during init
- Creates a response task if the networking layer supports it. 

### This PR does not: 

Remove VID sending over any networks.  Likely the request protocol will not be invoke just yet.  

### Key places to review: 

Config changes - did I change everywhere I need to to have this config option be usable?
Does the task creation make sense?


crates/task-impls/src/response.rs
```diff
@@ -23,19 +25,21 @@ use hotshot_types::{
...
+use tokio::task::JoinHandle;
 
 use crate::events::HotShotEvent;
 
 /// Type alias for consensus state wrapped in a lock.
 type LockedConsensusState<TYPES> = Arc<RwLock<Consensus<TYPES>>>;
 
 /// Type alias for the channel that we receive requests from the network on.
-type ReqestReceiver<TYPES> = mpsc::Receiver<(Message<TYPES>, ResponseChannel<Message<TYPES>>)>;
+pub type ReqestReceiver<TYPES> = mpsc::Receiver<(Message<TYPES>, ResponseChannel<Message<TYPES>>)>;
```
jparr721, line 37: typo in `Reqest`

jparr721 (commented):
One thing then I think this is good to go. I know the typo wasn't introduced by you, and we can ignore if it's a big ask.

crates/task-impls/src/response.rs
bfish713: This was my typo

jparr721 (dismissed):

crates/task-impls/src/request.rs
```diff
@@ -236,7 +238,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> DelayedRequester<TYPES, I> {
     async fn handle_response_message(&self, message: SequencingMessage<TYPES>) {
```
shenkeyao, line 238: Am I understanding correctly that the request task also handles the response, i.e., if we don't have a response task, this request task can still do the handling in the `handle_response_message` function?

Would it be better to rename this task to something like `RequestResponse` to make it explicit that it's handling both requests and responses?

shenkeyao (commented):
A nit comment and a question about the distinction between the two tasks.

crates/task-impls/src/request.rs
bfish713: Hmm I agree with the naming being confusing.  The problem is what I called the `Request` task is responsible for sending out requests for data that the node wants, and then getting the response.

What I call the `Response` task, is responsible for for taking incoming requests, creating a response and responding to the node asking for the data.

So in a way both tasks to request and response.  Would it make sense to call the `Request` task the `Requester` task, and the `Response` task the `Responder` task or something of that nature?

shenkeyao (approved):

crates/task-impls/src/request.rs
shenkeyao: Oh, I missed your reply before giving the PR approval. This makes sense. It's probably not necessary to do the renaming. I'll leave it to your discretion then!

Issue #2720, opened 2024-03-07
[Builder] Enable internal hotshot events for non-staked participants


move47:
### What is this task and why do we need to work on it?

The builder needs to receive events from hotshot for its progress. Currently, It needs to be an active participant to achieve it. However, a builder might want to be a passive participant and doesn't want to participate in internal voting. To support this participation, this task explores the possibility of having non-staking participants for HotShot.

It might have some co-relation with the following:
- [ ] #2635 
- [ ] #2636

### What work will need to be done to complete this task?

TODO!

### What are the acceptance criteria to close this issue?

All of the existing tests passes with related changes.

### Branch work will be merged to (if not the default branch)

hg/support_non_staking_nodes

Issue #2636, opened 2024-02-22
[BUILDER] - Allow builders to listen for DA streams but not vote on them


elliedavidson:
### What is this task and why do we need to work on it?

Builders (and perhaps other nodes in the future) need to listen to DA proposals, and therefore be sent DA proposals as if they were on the DA committee.  However, they won't vote, so they should not be counted towards the DA committee quorum.  

### What work will need to be done to complete this task?

Proposed Design: 
* At startup, allow nodes to signal what they want to listen for.  If they want to listen for DA proposals but they are not on the DA committee, they can signal this through some set up configuration.  

### Are there any other details to include?

In the future we'll need to somehow prioritize DA traffic to those that need it most, but that is out of scope for this sprint. 

cc @nyospe and @move47 for visibility

### What are the acceptance criteria to close this issue?

Successfully run a test with a builder node receiving DA data but not voting.

move47:
Some work was done in #2721

bfish713:
I think we proved this is completed with the tests

Issue #2635, opened 2024-02-22
[BUILDER] - Orchestrator supports builder


elliedavidson:
### What is this task and why do we need to work on it?

The orchestrator needs to wait for the builder(s) to start before signaling the entire network to start.  But the builder is a non-staked node and therefore should not be given a node index. 

### What work will need to be done to complete this task?

Proposed design: 
* Add a "builders" field to the HotShot config and have the orchestrator wait for all nodes to signal start as well as all builders.  

Other proposed design: 
* Give the builder a node index, but clarify it as a "listening only" node. 

May need more discussion on how we want to support non-voting nodes in general. 

### Are there any other details to include?

See discussion [here](https://espresso.zulipchat.com/#narrow/stream/311646-Consensus/topic/Builder.20support.20issues/near/422680920)

cc @nyospe and @move47 for visibility 

### What are the acceptance criteria to close this issue?

Start a small network with at least 1 builder present.  Confirm that the builder is not assigned a node index.

move47:
Some work was done in #2721.

PR #2721, opened 2024-03-07
[Builder] Hotshot events for non-staking nodes

move47:
In-progress for: 
- [x] #2720 
Also might have some co-relation for:
- [ ] #2635 
- [x] #2636

### DIFF
--- a/crates/examples/infra/mod.rs
+++ b/crates/examples/infra/mod.rs
@@ -181,13 +181,16 @@ pub fn read_orchestrator_init_config<TYPES: NodeType>() -> (
         load_config_from_file::<TYPES>(&args.config_file);
 
     if let Some(total_nodes_string) = matches.get_one::<String>("total_nodes") {
-        config.config.total_nodes = total_nodes_string.parse::<NonZeroUsize>().unwrap();
+        config.config.num_nodes_with_stake = total_nodes_string.parse::<NonZeroUsize>().unwrap();
         config.config.known_nodes_with_stake =
-            vec![PeerConfig::default(); config.config.total_nodes.get() as usize];
-        error!("config.config.total_nodes: {:?}", config.config.total_nodes);
+            vec![PeerConfig::default(); config.config.num_nodes_with_stake.get() as usize];
+        error!(
+            "config.config.total_nodes: {:?}",
+            config.config.num_nodes_with_stake
+        );
     }
     if let Some(da_committee_size_string) = matches.get_one::<String>("da_committee_size") {
-        config.config.da_committee_size = da_committee_size_string.parse::<usize>().unwrap();
+        config.config.da_staked_committee_size = da_committee_size_string.parse::<usize>().unwrap();
     }
     if let Some(transactions_per_round_string) = matches.get_one::<String>("transactions_per_round")
     {
--- a/crates/orchestrator/src/lib.rs
+++ b/crates/orchestrator/src/lib.rs
@@ -97,8 +97,8 @@ impl<KEY: SignatureKey + 'static, ELECTION: ElectionConfig + 'static>
     pub fn output_to_csv(&self) {
         let output_csv = BenchResultsDownloadConfig {
             commit_sha: self.config.commit_sha.clone(),
-            total_nodes: self.config.config.total_nodes.into(),
-            da_committee_size: self.config.config.da_committee_size,
+            total_nodes: self.config.config.num_nodes_with_stake.into(),
+            da_committee_size: self.config.config.da_staked_committee_size,
             transactions_per_round: self.config.transactions_per_round,
             transaction_size: self.bench_results.transaction_size_in_bytes,
             rounds: self.config.rounds,
@@ -176,7 +176,7 @@ where
         let node_index = self.latest_index;
         self.latest_index += 1;
 
-        if usize::from(node_index) >= self.config.config.total_nodes.get() {
+        if usize::from(node_index) >= self.config.config.num_nodes_with_stake.get() {
             return Err(ServerError {
                 status: tide_disco::StatusCode::BadRequest,
                 message: "Network has reached capacity".to_string(),
@@ -351,7 +351,7 @@ where
             }
         }
         self.nodes_post_results += 1;
-        if self.nodes_post_results >= (self.config.config.total_nodes.get() as u64) {
+        if self.nodes_post_results >= (self.config.config.num_nodes_with_stake.get() as u64) {
             self.bench_results.printout();
             self.output_to_csv();
         }


crates/types/src/lib.rs
```diff
@@ -148,13 +148,18 @@ pub struct HotShotConfig<KEY: SignatureKey, ELECTIONCONFIG: ElectionConfig> {
     /// Whether to run one view or continuous views
     pub execution_type: ExecutionType,
     /// Total number of nodes in the network
-    pub total_nodes: NonZeroUsize,
+    // Earlier it was total_nodes
+    pub num_nodes_with_stake: NonZeroUsize,
+    /// Number of nodes without stake
+    pub num_nodes_without_stake: Option<usize>,
```
bfish713, line 154: nit: remove `Option`

crates/hotshot/src/traits/election/static_committee.rs
```diff
@@ -32,10 +34,15 @@ pub type StaticCommittee<T> = GeneralStaticCommittee<T, BLSPubKey>;
 impl<T, PUBKEY: SignatureKey> GeneralStaticCommittee<T, PUBKEY> {
     /// Creates a new dummy elector
     #[must_use]
-    pub fn new(_nodes: &[PUBKEY], nodes_with_stake: Vec<PUBKEY::StakeTableEntry>) -> Self {
+    pub fn new(
+        _nodes: &[PUBKEY],
+        nodes_with_stake: Vec<PUBKEY::StakeTableEntry>,
+        nodes_without_stake: Option<Vec<PUBKEY>>,
```
jparr721, line 40: I have no context of course, but I'm curious about the use of `Option` here. If it's a `Vec`, would `vec![]` or something similar not be a good default case?

crates/types/src/traits/election.rs
```diff
@@ -38,7 +38,10 @@ pub trait Membership<TYPES: NodeType>:
     Clone + Debug + Eq + PartialEq + Send + Sync + Hash + 'static
 {
     /// generate a default election configuration
-    fn default_election_config(num_nodes: u64) -> TYPES::ElectionConfigType;
+    fn default_election_config(
+        num_nodes_with_stake: u64,
+        num_nodes_without_stake: u64,
```
bfish713, line 43: I'd prefer if we could not change the trait, but it's not a big deal if it's difficult to work around

crates/hotshot/src/traits/election/static_committee.rs
```diff
@@ -147,3 +172,18 @@ where
...
     }
 }
+
+impl<TYPES, PUBKEY: SignatureKey + 'static> GeneralStaticCommittee<TYPES, PUBKEY>
+where
+    TYPES: NodeType<SignatureKey = PUBKEY, ElectionConfigType = StaticElectionConfig>,
+{
+    /// get the non-staked builder nodes
+    pub fn non_staked_nodes_count(&self) -> usize {
+        self.committee_nodes_without_stake.as_ref().unwrap().len()
```
jparr721, line 182: Same question as above. In this case, it seems like a good default would be `0` perhaps, and if you decided to remove the `Option` from the vec above it would give a clean fail case. If we decide to keep this method, it might be worth documenting that this function can panic.

crates/hotshot/src/traits/election/static_committee.rs
```diff
@@ -95,24 +104,40 @@ where
...
             .iter()
             .map(|x| x.stake_table_entry.clone())
             .collect();
+        // make committee_nodes_with_stake to be nodes_with_stake
         let mut committee_nodes_with_stake: Vec<PUBKEY::StakeTableEntry> = nodes_with_stake.clone();
         debug!("Election Membership Size: {}", config.num_nodes);
-        committee_nodes_with_stake.truncate(config.num_nodes.try_into().unwrap());
+        // truncate committee_nodes_with_stake to only `num_nodes`
+        // since the `num_nodes_without_stake` are not part of the committee,
+        committee_nodes_with_stake.truncate(config.num_nodes_with_stake.try_into().unwrap());
```
jparr721, line 131: Is it safe to unwrap here?

crates/types/src/lib.rs
move47: Done

crates/hotshot/src/traits/election/static_committee.rs
```diff
@@ -147,3 +172,18 @@ where
...
+where
+    TYPES: NodeType<SignatureKey = PUBKEY, ElectionConfigType = StaticElectionConfig>,
+{
+    /// get the non-staked builder nodes
+    pub fn non_staked_nodes_count(&self) -> usize {
+        self.committee_nodes_without_stake.as_ref().unwrap().len()
+    }
+
+    /// get all the non-staked nodes
+    pub fn get_non_staked_nodes(&self) -> Vec<PUBKEY> {
```
jparr721, line 186: Is this function being used?

jparr721 (commented):
Couple of high-level comments here about the unwrapping behavior and the use of Optionals.

crates/hotshot/src/traits/election/static_committee.rs
jparr721: Looks like this message got put here out of order. Please see the below `is it safe to unwrap here?` comment that I am referencing when I say "above"

crates/hotshot/src/traits/election/static_committee.rs
move47: Sorry I think you reviewed the old version.

crates/hotshot/src/traits/election/static_committee.rs
move47: Not currently, but would be used later.

crates/hotshot/src/traits/election/static_committee.rs
jparr721: Hmm, when I refresh I still see the Option type being used. Are you saying you have new commits coming?

move47: It is already there.

jparr721: Sorry, let me ask a different way. My original question was asking if the `Option` type was necessary here, since an empty Vec type might be sufficient in this case to imply a value is not present. I was curious about your thought process for adding it in this case.

crates/hotshot/src/traits/election/static_committee.rs
bfish713: Forgot to mention, we'll have to use this function in hotshot shot here: https://github.com/EspressoSystems/HotShot/blob/main/crates/task-impls/src/network.rs#L375

crates/hotshot/src/traits/election/static_committee.rs
move47: Yeah, I think empty vec covers that case, I just went with the flow.

crates/hotshot/src/traits/election/static_committee.rs
```diff
@@ -95,24 +104,40 @@ where
...
         let mut committee_nodes_with_stake: Vec<PUBKEY::StakeTableEntry> = nodes_with_stake.clone();
         debug!("Election Membership Size: {}", config.num_nodes);
-        committee_nodes_with_stake.truncate(config.num_nodes.try_into().unwrap());
+        // truncate committee_nodes_with_stake to only `num_nodes`
+        // since the `num_nodes_without_stake` are not part of the committee,
+        committee_nodes_with_stake.truncate(config.num_nodes_with_stake.try_into().unwrap());
+
+        // get the non-staked pub keys from the config's `known_nodes_without_stake`
+        // it is not option anymore
+        committee_nodes_without_stake = config.known_nodes_without_stake.clone();
```
move47, line 135: @bfish713 How do we go about it? I think this is wrong. config is ElectionConfigType, which has num_nodes_with_stake/without stake, so how to get the `committee_nodes_without_stake` from it, which is Vec<PUBKEY>, can we do a similar call i.e `x.get_stake_table_entry()` as above?

move47: It is because we don't to change the function signatures

bfish713: I think you want to filter the entries by stake to split the list into those with/without stake.  Entries should have both entries with 1 and 0 stake.

move47: I have tried, let me know now

bfish713:
In general it looks good, I think you need to audit all the places we call get_committee on an election config and confirm we specifically want nodes with stake.  I know for certain in the network task: https://github.com/EspressoSystems/HotShot/blob/main/crates/task-impls/src/network.rs#L375 you want to be getting all nodes (with and without stake) or else the no stake nodes will not get the info.

move47:
@bfish713 I have made changes as per suggestions. Please let me know next steps.

crates/hotshot/src/traits/election/static_committee.rs
bfish713: I think so, we were already doing this unwrap.  If we can't construct the election then we can't really run HotShot

bfish713 (approved):

PR #2722, opened 2024-03-07
Voting for all Possible Orderings Regression Test

### LINKED ISSUE #2306: [STABILITY] - Unit Tests for Voting for all Event Ordering

We've seen multiple bugs with not sending a vote after receiving events in an order we don't expect.  First we saw this when a replica got a DAC after the proposal.  Another example could be receiving a VID before/after the proposal.

Create two unit tests, one which tests a we vote for a proposal (not genesis) where we get a DAC after Quorum proposal and one where we the proposal before the DAC

_No response_

Both tests pass.  We should also verify that if we remove the vote path from one of propsal recv or dac recv event handlers that one test fails.

_No response_

jparr721:
Depends on: https://github.com/EspressoSystems/HotShot/pull/2697 and https://github.com/EspressoSystems/HotShot/pull/2716.
Closes #2306 

This PR creates two unit tests where we vote for a (non-genesis) proposal, receiving a DAC:
1. After Quorum proposal and 
2. Before the Quorum proposal

### This PR does not: 
This PR does not address the underlying logic issues that my have precipitated this issue, it only adds a test to alert if it happens again.

### Key places to review: 
All test code, in particular, to verify that this does actually test the described case.

### How to test this PR:
cargo test -p hotshot-testing --test consensus_task

### DIFF
--- a/crates/testing/tests/consensus_task.rs
+++ b/crates/testing/tests/consensus_task.rs
@@ -3,23 +3,13 @@ use hotshot::types::SystemContextHandle;
 use hotshot_example_types::node_types::{MemoryImpl, TestTypes};
 use hotshot_task_impls::events::HotShotEvent;
 use hotshot_testing::task_helpers::{build_quorum_proposal, build_vote, key_pair_for_id};
+use hotshot_testing::test_helpers::permute_input_with_index_order;
 use hotshot_types::traits::{consensus_api::ConsensusApi, election::Membership};
 use hotshot_types::{
     data::ViewNumber, message::GeneralConsensusMessage, traits::node_implementation::ConsensusTime,
 };
 use jf_primitives::vid::VidScheme;
 use std::collections::HashMap;
-
-fn permute<T>(inputs: Vec<T>, order: Vec<usize>) -> Vec<T>
-where
-    T: Clone,
-{
-    let mut ordered_inputs = Vec::with_capacity(inputs.len());
-    for &index in &order {
-        ordered_inputs.push(inputs[index].clone());
-    }
-    ordered_inputs
-}
 
 #[cfg(test)]
 #[cfg_attr(async_executor_impl = "tokio", tokio::test(flavor = "multi_thread"))]
@@ -210,7 +200,7 @@ async fn test_vote_with_specific_order(input_permutation: Vec<usize>, force_pani
         DACRecv(dacs[1].clone()),
         QuorumProposalRecv(proposals[1].clone(), leaders[1]),
     ];
-    let view_2_inputs = permute(inputs, input_permutation);
+    let view_2_inputs = permute_input_with_index_order(inputs, input_permutation);
 
     let mut view_2_outputs = vec![
         exact(ViewChange(ViewNumber::new(2))),
--- a/crates/testing/tests/consensus_task.rs
+++ b/crates/testing/tests/consensus_task.rs
@@ -10,6 +10,17 @@ use hotshot_types::{
 use jf_primitives::vid::VidScheme;
 use std::collections::HashMap;
 
+fn permute<T>(inputs: Vec<T>, order: Vec<usize>) -> Vec<T>
+where
+    T: Clone,
+{
+    let mut ordered_inputs = Vec::with_capacity(inputs.len());
+    for &index in &order {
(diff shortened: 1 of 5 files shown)


crates/testing/tests/consensus_task.rs
```diff
@@ -120,8 +220,115 @@ async fn test_consensus_vote() {
...
+    if !dac_first {
+        view_2_inputs.swap(1, 2);
+    }
+
+    // Now, for view 2, receive a DAC before the quorum proposal.
+    let view_2 = TestScriptStage {
+        inputs: view_2_inputs,
+        outputs: vec![
+            exact(ViewChange(ViewNumber::new(2))),
+            exact(QuorumProposalValidated(proposals[1].data.clone())),
```
bfish713, line 300: Shouldn't we expect a VoteSend?

crates/testing/tests/consensus_task.rs
```diff
@@ -120,8 +220,115 @@ async fn test_consensus_vote() {
...
+
+    let mut view_2_inputs = vec![
+        // We need a VID share for view 2 otherwise we cannot vote at view 2 (as node 2).
+        VidDisperseRecv(vids[1].0.clone(), vids[1].1),
+        DACRecv(dacs[1].clone()),
+        QuorumProposalRecv(proposals[1].clone(), leaders[1]),
+    ];
+    if !dac_first {
+        view_2_inputs.swap(1, 2);
+    }
```
bfish713, line 293: Just like the other can we permute all orderings

crates/testing/tests/consensus_task.rs
jparr721: So this was the alternative case mentioned in the [ticket](https://github.com/EspressoSystems/HotShot/issues/2306)
>We should also verify that if we remove the vote path from one of propsal recv or dac recv event handlers that one test fails.

So this test panics because the VoteSend is missing. Let me know if I misunderstood what you meant here.

jparr721: I'm gonna remove the copy paste here and add a param instead to make this more clear.

jparr721: Done, I'll leave this unresolved to make sure that our understanding of this is the same.

crates/testing/tests/consensus_task.rs
```diff
@@ -10,13 +10,283 @@ use hotshot_types::{
 use jf_primitives::vid::VidScheme;
 use std::collections::HashMap;
 
+fn permute<T>(inputs: Vec<T>, order: Vec<usize>) -> Vec<T>
```
jparr721, line 13: This function is going to be deleted once [this](https://github.com/EspressoSystems/HotShot/pull/2697) merges. The branches diverged due to an in-between PR with this one, so I'll move this out to a central location once that goes in.

crates/testing/tests/consensus_task.rs
```diff
@@ -146,6 +147,125 @@ async fn test_consensus_vote() {
...
+    // valid) order allows the system to proceed as it normally would. We force the panic here by
+    // intentionally removing the vote send event from the output.
+    test_vote_with_specific_order(vec![0, 1, 2], true /* force_panic */).await;
+    test_vote_with_specific_order(vec![0, 2, 1], true /* force_panic */).await;
+    test_vote_with_specific_order(vec![1, 0, 2], true /* force_panic */).await;
+    test_vote_with_specific_order(vec![2, 0, 1], true /* force_panic */).await;
+    test_vote_with_specific_order(vec![1, 2, 0], true /* force_panic */).await;
+    test_vote_with_specific_order(vec![2, 1, 0], true /* force_panic */).await;
+}
+
```
bfish713, line 268: Do these tests actually test anything new? Can we remove the `force_panic` parameter?

bfish713 (commented):
Looks good, I just don't think we need the `#[should_panic]` tests because they seem to be testing the exact same thing as the not panicing version

crates/testing/tests/consensus_task.rs
jparr721: Sure I'll remove it.

bfish713 (approved):

PR #2734, opened 2024-03-08
Reorganize build/test in CI

ss-es:
No linked issue.

### This PR: 
This is a PR to reorganize CI. Changes:

- The self-hosted `build`, `build-release` and `lint` jobs are removed. It is unnecessary to duplicate these on GitHub runners and on the self-hosted runners. The self-hosted `test` and `build-arm` jobs are retained.
- The `build` job now depends on `test` passing. This means that tests run first, and that we never build if tests fail. In particular, we still do not build or push dockers if the tests fail.
- The `build-release` job runs in parallel to tests, since the artifacts from this job were not used to build dockers.
- We no longer skip `test_crypto` in the regular tests, so it no longer runs as a separate job.
- We no longer cancel the `async-std` build or test if the `tokio` variant fails, and vice-versa.
- The self-hosted `test` no longer runs with `--no-fail-fast`. This means test failures are reported (in the form of CI failing) immediately, which should be helpful for development. This doesn't affect the regular GitHub runners, which will still continue to run all tests even after encountering a failing test, and may be helpful for debugging. (The choice of which one to run with `--no-fail-fast` was completely arbitrary.)

Unrelated to the primary changes, this PR also adds `DEBUG` logging for `libp2p-networking`, which is helpful when the tests in `libp2p-networking/tests/counter.rs` fail -- currently, these tests don't produce any useful logs when they fail.

Notes: if we expect `build-release` artifacts to be consumed in the future, it should depend on passing tests.

### Key places to review: 
Overall test workflow, e.g. here: https://github.com/EspressoSystems/HotShot/actions/runs/8348230729?pr=2734


jparr721 (dismissed):
🥳

jparr721 (approved):

Issue #2308, opened 2024-01-04
[STABILITY] - Unit Test for Proposing after View Sync


bfish713:
### What is this task and why do we need to work on it?

In Gibraltar we saw a bug where we failed to propose after view sync.  We need a regression test to make sure we don't hit this again.

### What work will need to be done to complete this task?

Create a unit test where view sync is triggered, then completed (ViewSyncCommitCertRecv event).  This will involve creating a test where both consensus and view sync are running

### Are there any other details to include?

There are other scenarios where the other tasks don't start doing some step necessary for the consensus task to form a proposal.  E.g. the Transaction tasks never sends `TransactionsReady`.

### What are the acceptance criteria to close this issue?

Test passes

PR #2739, opened 2024-03-11
[STABILITY] - Remove Hack Fix in Vote Timeout Trigger

### LINKED ISSUE #2732: [STABILITY] - Remove Hack Fix in Vote Timeout Trigger

This PR: https://github.com/EspressoSystems/HotShot/pull/2598 implemented a temporary fix to a bug where:

>We do not propose based on a view sync certificate. A proposal must either have a justify qc which is equal to the prior view, or a timeout certificate for the correct view. There is a race condition which comes up if there is a good leader after 2 bad ones. If view sync completes very fast it's possible that some nodes don't even timeout the second bad view because they get evidence to move ahead before their timer expires! In practice this is extremely unlikely to happen because network latency makes view sync take long enough that each node would timeout on their own. In tests there is effectively no latency and each node is competing with each other for time to execute so they get slightly out of sync but are able to process events vary fast. 

To implement the fix, we'd need to send the `ViewSyncFinalizeCertificate2Recv` event to consensus and allow it to be attached to a proposal, then we'd have to add a check during proposal verification for this cert as well. The proposal would be updated as:

```diff
/// Proposal to append a block.
#[derive(custom_debug::Debug, Serialize, Deserialize, Clone, Eq, PartialEq, Hash)]
#[serde(bound(deserialize = ""))]
pub struct QuorumProposal<TYPES: NodeType> {
    /// The block header to append
    pub block_header: TYPES::BlockHeader,

    /// CurView from leader when proposing leaf
    pub view_number: TYPES::Time,

    /// Per spec, justification
    pub justify_qc: QuorumCertificate<TYPES>,

    /// Possible timeout certificate.  Only present if the justify_qc is not for the preceding view
    pub timeout_certificate: Option<TimeoutCertificate<TYPES>>,

    /// Possible upgrade certificate, which the leader may optionally attach.
    pub upgrade_certificate: Option<UpgradeCertificate<TYPES>>,

+   /// The view sync finalize certificate, an additional validation step that allows us to propose when out of sync.
+   pub view_sync_finalize_certificate: Option<ViewSyncFinalizeCertificate2<TYPES>>, 

    /// the propser id
    pub proposer_id: TYPES::SignatureKey,
}
```

We then need to be able to update the validation logic for the `QuorumProposalValidated` to check that the view sync cert is for the same view as the proposal, this is not the easiest check, and there may be an off-by-one here. This should be verified.

_No response_

From the PR:
>The precise sequence I observed was:
>
>- 12 nodes total, 2 of them completely offline, the two offline are consecutive leaders
>- Everyone times out because they can't send to the first bad leader.  
>- 5-8 nodes are slightly ahead of the remaining 2-5 online nodes.
>- The 5-8 nodes ahead timeout a little before the rest.
>- The 5-8 nodes timeout and send their votes to the next leader (This is not enough for a quorum)
>- The ahead nodes enter view sync and quickly form precommit
>- At this point the remaining nodes (still waiting to for their timer to expire) send votes to form a commit
>- Everyone sees this commit and moves to the next view which has a good leader.
>- The lagging nodes cancel their timeout task for the previous view, thus never sending a timeout cert
>- The honest leader of the view everyone just entered can't form a valid proposal because they don't have a QC or TC from the last view.

Ideally we'd have a way to check this exact case, or some simplified version.

_No response_

jparr721:
Closes #2732, #2308

### This PR: 
- Adds support for the `ViewSyncFinalizeCertificate2Recv` event in the consensus algorithm.
- Adds the implementation of the view sync certificate in consensus. This implementation initiates a proposal when the event is received.
- Adds support for checking the validity of the cert (if it exists) within the voting logic and failing to vote if the certificate has any invalid properties.
- Tests that a proposal is indeed sent when such an event is received.

### This PR does not: 
- Does not explicitly check the vote failure case where the `is_valid_cert` function on a cert fails.

### Key places to review: 
- All logic for the new handler body. 
    - Specifically, this code acts as though it received a timeout certificate, and follows a similar flow to that in `QCFormed`.
- All tests in the `consensus_task` file.
    - These tests cover the happy path case where the event is captured and the proposal is sent, and when the event is captured, but the node is not able to propose and votes instead. It also validates that the node will _not_ propose when the cert is invalid.

Specifically:
- https://github.com/EspressoSystems/HotShot/pull/2739/files#diff-bd3299d85e179494b5d046442fd33a4ea027c838347f5ab393013dfee9e7d302R463
- https://github.com/EspressoSystems/HotShot/pull/2739/files#diff-73f1110b8809596a4a2ff5b3c71bf170534ea3a59b55c1b875d6623481bcfa03R428

### How to test this PR:
cargo test -p hotshot-testing --test consensus_task


crates/task-impls/src/consensus.rs
```diff
@@ -237,6 +242,23 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
+                // Do not vote for the proposal if the cert is valid for an incorrect view
+                if cert.view_number != proposal.view_number {
+                    debug!(
+                        "Cert view number {:?} does not match proposal view number {:?}",
+                        cert.view_number, proposal.view_number
+                    );
+                    return false;
+                }
+            }
+
```
bfish713, line 261: I think we want to move this logic earlier.  Specifically it should be part of proposal validation where we do the similar checks for `timeout_cert` on the `QuorumProposal`.

Smaller nit, we should do the view number check before validating the cert because the validation is a more expensive operation.

crates/task-impls/src/consensus.rs
```diff
@@ -1237,6 +1259,38 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
+                debug!(
+                    "Attempting to publish proposal after forming a View Sync Finalized Cert for view {}",
+                    *certificate.view_number
+                );
+
+                let view = certificate.view_number + 1;
+
+                if !self
+                    .publish_proposal_if_able(view, self.timeout_cert.clone(), &event_stream)
```
bfish713, line 1288: timeout_cert should be `None` here or else we can attach an old timeout cert, which would create an invalid proposal

bfish713: It's possible that both view sync and timeout certs are valid (but really replicas should not be voting for both certs and shouldn't happen) but we only need one for a valid proposal.  So it's ok never include a timeout_cert from this code path.

It probably works just fine because we clear the `self.timeout_cert` after proposing, but might as well make sure a bug can't slip in.

crates/testing/tests/consensus_task.rs
bfish713, line 1: Nice tests, great we can have these specific tests

bfish713 (changes requested):

crates/task-impls/src/consensus.rs
jparr721: Done on the second thing. As for the first thing, moving this logic to the Quorum proposal does not prevent an invalid cert from eliciting a vote in this case:
```diff
HotShotEvent::QuorumProposalRecv(proposal, sender) => {
    ...
                // Verify a timeout certificate exists and is valid
                if proposal.data.justify_qc.get_view_number() != view - 1 {
                    let Some(timeout_cert) = proposal.data.timeout_certificate.clone() else {
                        warn!(
                            "Quorum proposal for view {} needed a timeout certificate but did not have one",
                            *view);
                        return;
                    };

                    if timeout_cert.get_data().view != view - 1 {
                        warn!("Timeout certificate for view {} was not for the immediately preceding view", *view);
                        return;
                    }

                    if !timeout_cert.is_valid_cert(self.timeout_membership.as_ref()) {
                        warn!("Timeout certificate for view {} was invalid", *view);
                        return;
                    }

                    // If we have a ViewSyncFinalize cert, only vote if it is valid.
+                   if let Some(cert) = &self.view_sync_cert {
+                       // Do not vote for the proposal if the cert is valid for an incorrect view
+                      if cert.view_number != view {
+                           debug!(
+                               "Cert view number {:?} does not match proposal view number {:?}",
+                               cert.view_number, view
+                           );
+                          return;
+                       }
+                        if !cert.is_valid_cert(self.quorum_membership.as_ref()) {
+                            debug!("Invalid ViewSyncFinalize cert provided");
+                            return;
+                        }
+                    }
+                }
}
```

Adding it here results in my negative-case test failing.

jparr721: Done.

shenkeyao (approved):
LGTM!

crates/task-impls/src/consensus.rs
```diff
@@ -473,6 +478,23 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
+                            cert.view_number, view
+                        );
+                        return;
+                    }
+
+                    if !cert.is_valid_cert(self.quorum_membership.as_ref()) {
+                        debug!("Invalid ViewSyncFinalize cert provided");
+                        return;
+                    }
+                }
```
bfish713, line 496: Hmm I don't see how the vote tests are passing with this.  This code is only reachable in the case where `proposal.data.justify_qc.get_view_number() == view - 1` but in that case were would not have been in view sync.  This logic needs work alongside the timeout logic.  

The logic needs to be if `proposal.data.justify_qc.get_view_number() == view - 1`  we're good else:  
- check if we have a timeout cert of view sync cert for the right view

crates/testing/tests/consensus_task.rs
```diff
@@ -242,6 +242,319 @@ async fn test_consensus_vote_with_permuted_dac() {
...
+            DACRecv(dacs[2].clone()),
+        ],
+        outputs: vec![
+            exact(ViewChange(ViewNumber::new(3))),
+            exact(QuorumProposalValidated(proposals[2].data.clone())),
+            leaf_decided(),
+            quorum_vote_send(),
+        ],
+        asserts: vec![is_at_view_number(3)],
+    };
```
bfish713, line 442: Sorry I missed this on first pass, this doesn't really test what we want.  We're voting here because the `QuorumProposal` has a QC for the previous view.  That's enough to vote without finalized cert.  Also I think we should test this works if the node doesn't receive `ViewSyncFinalizeCertificate2Recv`

crates/testing/tests/consensus_task.rs
```diff
@@ -242,6 +242,319 @@ async fn test_consensus_vote_with_permuted_dac() {
...
+        inputs: vec![QuorumProposalRecv(proposals[0].clone(), leaders[0])],
+        outputs: vec![
+            exact(ViewChange(ViewNumber::new(1))),
+            exact(QuorumProposalValidated(proposals[0].data.clone())),
+            exact(QuorumVoteSend(votes[0].clone())),
+        ],
+        asserts: vec![is_at_view_number(1)],
+    };
+
+    let view_2 = TestScriptStage {
```
bfish713, line 412: To fix the comments below, this view should probably just except a `Timeout` event, or we just skip this stage all together (as if the node never saw anything in view 2)

crates/testing/src/view_generator.rs
```diff
@@ -289,6 +315,20 @@ impl TestViewGenerator {
             tracing::error!("Cannot attach transactions to the genesis view.");
         }
     }
+
+    pub fn add_view_sync_finalize(
```
bfish713, line 319: I think we need a similar function that makes it so a quorum proposal for a view does not reference a QC for a prior view

bfish713 (changes requested):
I think there might be some confusion on the interplay between the 2 certificate types for timeout, quorum and view sync.  

Every Proposal must have some QuorumCertificate referenced.  If this QC is for the prior view no other certificate should be attatched.  This is the happy path case.

If the QC referenced in a proposal is not for the previous view then either a Timeout or ViewSyncFinalized certificate should be attatched. Only one or the other should be attached. In this case the proposal is saying I'm extending the chain from some older view (skipping some views) and my justification that I shouldn't extending the last view is a Timeout or ViewSync certificate.

crates/testing/tests/consensus_task.rs
jparr721: In this case, I will skip the view so that the node is in a prior view (`v-1`), but received no information that would allow it to move to the next view (`v`), and thus would reasonably trigger a view sync in view 3, which should then proceed as expected.

crates/testing/tests/consensus_task.rs
jparr721: Instead, we need the QuorumProposal to have a ViewSync cert, correct? Then we can go through the proper flow?

To make this work without receiving a view sync cert, I'd need to have another cert available to be able to have the proposal be valid so I (the node) can vote on it. So I could just drop the view sync and this would likely run just fine. I'll do that in a separate test.

crates/testing/src/view_generator.rs
jparr721: The proposals can be overridden from within the generated values, so we can just change whatever we want. Is this sufficient?

Cargo.toml
```diff
@@ -58,7 +58,7 @@ futures = "0.3.30"
 # TODO generic-array should not be a direct dependency
 # https://github.com/EspressoSystems/HotShot/issues/1850
 generic-array = { version = "0.14.7", features = ["serde"] }
-hotshot-types = { git = "https://github.com/EspressoSystems/hotshot-types" }
+hotshot-types = { git = "https://github.com/EspressoSystems/hotshot-types", branch = "jparr721/issue-2732" }
```
jparr721, line 61: This needs a fast-follow when this merges: https://github.com/EspressoSystems/hotshot-types/pull/1. Tracking: https://github.com/EspressoSystems/HotShot/issues/2759

crates/task-impls/src/consensus.rs
```diff
@@ -453,24 +458,42 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
                     return;
                 }
 
-                // Verify a timeout certificate exists and is valid
+                // Verify a timeout certificate OR a view sync certificate exists and is valid.
                 if proposal.data.justify_qc.get_view_number() != view - 1 {
```
jparr721, line 462: Please advise if this is okay. We explicitly first check if there exists a timeout cert, THEN we check if there is a view_sync cert. But the checks are independent. If we have neither, then we exit as normal.

crates/testing/src/view_generator.rs
jparr721: Done.

crates/task-impls/src/consensus.rs
bfish713: Yes this is ok.  We should codify this by making the propsal have and `Option<Either<TimeoutCert, ViewSyncFinalizedCert>>` instead of two options

crates/testing/tests/consensus_task.rs
```diff
@@ -242,6 +242,353 @@ async fn test_consensus_vote_with_permuted_dac() {
...
+    let stages = vec![view_1, view_2_3, view_4];
+
+    inject_consensus_polls(&consensus_state).await;
+    run_test_script(stages, consensus_state).await;
+}
+
+#[cfg(test)]
+#[cfg_attr(
+    async_executor_impl = "tokio",
+    tokio::test(flavor = "multi_thread", worker_threads = 2)
```
bfish713, line 386: nit, we can get rid of `worker_threads = 2` it just limits tokio executor to 2 threads arbitrarily

crates/task-impls/src/consensus.rs
bfish713: This can be a follow up task like we discussed in the team call

bfish713 (commented):
Updated logic + test look good to me now.  @ss-es might want to look at some of the testing harness changes

crates/testing/tests/consensus_task.rs
jparr721: Done!

bfish713 (approved):

bfish713 (dismissed):

bfish713 (approved):

PR #2741, opened 2024-03-11
[CI] [EASY] Update async compat layer for audit

rob-maron:
This PR just updates what version of the `async-compatibility-layer` we pull in. A transitive dependency in it was giving us an audit error.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2245,9 +2245,9 @@ dependencies = [
 
 [[package]]
 name = "eyre"
-version = "0.6.11"
+version = "0.6.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


ss-es:
`cargo audit` is still failing for me on this branch; I think I was misunderstanding something and `async-compitibility-layer` wasn't the culprit.

But it looks like `cargo update` fixes the audit error.

rob-maron:
> `cargo audit` is still failing for me on this branch; I think I was misunderstanding something and `async-compitibility-layer` wasn't the culprit.
> 
> But it looks like `cargo update` fixes the audit error.

Try now, I updated just that one failing dep

ss-es (approved):
Looks like it works -- audit no longer fails!

PR #2742, opened 2024-03-11
[Crypto] Allow signing and verifying data of arbitrary length

rob-maron:
### This PR: 

~~Adds two functions, `sign_arbitrary` and `verify_arbitrary`. Instead of casting to a `GenericArray` (locking us to length defined in `BitVectorQC`), they will allow signing of arbitrary data. This is needed for the Push CDN, where the messages are a different size.~~

Changes the two functions, `sign` and `verify` to operate over arbitrary data, removing the need for a `GenericArray` cast. This is needed for the Push CDN, where the messages are a different size.

### This PR does not: 

~~- Use the `sign_arbitrary` and `verify_arbitrary` functions~~
~~- Change the existing sign and validate functions~~

~~### Question for crypto folk:~~
~~Do we need this? It looks like `BitVectorQC`'s sign just proxies to the underlying scheme anyway. But maybe there is a reason we cast to the array of specific length. If not, I can just update the existing functions to work for arbitrary data.~~

### DIFF
--- a/crates/hotshot-qc/src/bit_vector.rs
+++ b/crates/hotshot-qc/src/bit_vector.rs
@@ -57,13 +57,14 @@ where
     type MessageLength = U32;
     type QuorumSize = U256;
 
-    fn sign<R: CryptoRng + RngCore>(
-        agg_sig_pp: &A::PublicParameter,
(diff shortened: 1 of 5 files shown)


mrain (approved):
LGTM

PR #2743, opened 2024-03-11
[NETWORKING] Fixes for Combined Network

bfish713:
When debugging some issues with the combined network we found a few issues.  This PR fixes these issues.  They are:

- Network task was blocked on collecting and cancelling delayed tasks in the combined network when updating view
  - Fixed by spawning a task 
  - Make `update_view` non-async so we know it's not a blocking call
- Recv was blocking until a message was received on both the primary and fallback, this can slow things down greatly if all messages for view come in from the primary, then after the delay expires we get the first non fallback.
  - fixed by adding a `select` between the two networks
- We implemented our our LRU cache, this just replace it with the LruCache crate.

### DIFF
--- a/crates/hotshot/src/traits.rs
+++ b/crates/hotshot/src/traits.rs
@@ -13,9 +13,7 @@ pub use storage::{Result as StorageResult, Storage};
 pub mod implementations {
     pub use super::{
         networking::{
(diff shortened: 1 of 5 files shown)


rob-maron (approved):
Nice work!

PR #2752, opened 2024-03-11
[CX_CLEANUP] Remove Special Case for View 1

### LINKED ISSUE #2751: [TECH-DEBT] - Remove Special Casing for View 1

View 0 is the genesis view.  In view 0 nobody proposes and nobody votes.  Each node starts with a QC for view 0, which is the genesis QC.  From there the leader of view 1 should do a normal proposal, with DA, and VID, using the genesis QC as the certificate.

_No response_

_No response_

VID and DA Cert is required to vote on the first non genesis propsal

_No response_

bfish713:
Closes #2751 

### This PR: 

- Checks for DA and VID in view 1.  Voting is no longer special cased for the first view
- Starts consensus by changing to view 0 before sending the genesis QC
  - This is need to make sure all polling is started, and the view 1 leader is triggered to send a block for DA
 - Adds special cases for view 0 view change to not return early.  This ensures the `ViewChange(0) event does what it should
 - Removes the special case to add an empty block payload for view 1 (now it's just genesis).

### This PR does not: 

Doesn't change any logic for views after view 1

### Key places to review: 

Changes to consensus task
Are there any places that also should be updated that I missed


crates/task-impls/src/consensus.rs
```diff
@@ -164,69 +164,6 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
         if let Some(proposal) = &self.current_proposal {
             let consensus = self.consensus.read().await;
 
-            // ED Need to account for the genesis DA cert
-            // No need to check vid share nor da cert for genesis
-            if proposal.justify_qc.is_genesis && proposal.view_number == TYPES::Time::new(1) {
-                info!("Proposal is genesis!");
```
dailinsubjam, line 170: It's nice that we removed this.

dailinsubjam (commented):
LGTM! Will wait for other's review.

crates/task-impls/src/consensus.rs
```diff
@@ -1111,11 +1046,31 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
+                        .await;
+
+                    if self.quorum_membership.get_leader(TYPES::Time::new(1)) == self.public_key {
+                        debug!("Polling for quorum votes for view {}", *self.cur_view);
+                        self.quorum_network
+                            .inject_consensus_info(ConsensusIntentEvent::PollForVotes(0))
+                            .await;
+                    }
+                }
+
```
ss-es, line 1073: I don't think this really matters either way, but would it make sense to move these to the `ConsensusTaskState::create_from` function in `hotshot/src/tasks/task_state.rs`?

bfish713: yeah it would make sense I can move it there

bfish713: There was aleardy a function for this so I added it there

rob-maron (dismissed):
Looks good to me !

shenkeyao (dismissed):
LGTM

dailinsubjam (approved):

PR #2762, opened 2024-03-12
[Push CDN] Full integration

rob-maron:
Closes https://github.com/EspressoSystems/Push-CDN/issues/16

### This PR: 

Integrates the Push CDN fully into HotShot, including tests and examples.

### This PR does not: 

Most of the auxiliary changes were made in separate PRs, so this one does not change much else.

### Key places to review: 

`push_cdn_network`. The examples were tested and working.


crates/hotshot/src/traits/networking/push_cdn_network.rs
```diff
@@ -0,0 +1,424 @@
...
+            return Ok(vec![result]);
+        }
+
+        if let PushCdnMessage::Broadcast(data) = message {
+            // Deserialize it
+            let result: Message<TYPES> = bincode_opts()
+                .deserialize(&data.message)
+                .map_err(|e| NetworkError::FailedToSerialize { source: e })?;
+            return Ok(vec![result]);
+        }
```
bfish713, line 408: Can this be combined?  We no longer need to split direct and broadcast on the receiver side in Hotshot.  I'm guessing this was a result of their existing `TransmitType` when you were implementing this.

crates/hotshot/src/traits/networking/push_cdn_network.rs
```diff
@@ -0,0 +1,424 @@
...
+    async fn recv_msgs(&self) -> Result<Vec<Message<TYPES>>, NetworkError> {
+        // Receive a message
+        let message = self.0.receive_message().await;
+
+        // If it was an error, wait a bit and retry
+        let message = match message {
+            Ok(message) => message,
+            Err(error) => {
+                async_sleep(Duration::from_millis(100)).await;
+                error!("failed to receive message: {error}");
```
bfish713, line 384: What's the reason for sleeping after an error?  I tried to dig into the CDN code but didn't really find any cases where sleeping would fix an error case, but I very easily could be missing something.

crates/hotshot/src/traits/networking/push_cdn_network.rs
```diff
@@ -0,0 +1,424 @@
...
+    }
+
+    /// TODO: shut down the networks. Unneeded for testing.
+    fn shut_down<'a, 'b>(&'a self) -> BoxSyncFuture<'b, ()>
+    where
+        'a: 'b,
+        Self: 'b,
+    {
+        // TODO: figure out if drop is sufficient
+        boxed_sync(async move {})
```
bfish713, line 309: I didn't look into the implementation of the client too much, but I assume it's spawning tasks that won't get shutdown properly unless we implement here.  Is that the case, or is it actually possible that dropping the client is enough.  

I'm mostly worried because in CI all the tasks we don't shutdown will continue for the whole test run and it could pollute other tests and cause some flakiness.  It might also be fine.  If it's difficult to implement no reason to block the PR.  Mostly just curious what will happen now during shutdown.

bfish713 (commented):
Overall this looks great, excited to get this merged!  Just had a few questions in the `ConnectedNetwork` implementation that's hopefully some easy cleanup.

crates/hotshot/src/traits/networking/push_cdn_network.rs
rob-maron: Yep, totally correct. Think you had removed that after I wrote this

crates/hotshot/src/traits/networking/push_cdn_network.rs
rob-maron: The initial reason I used a sleep here is because a failed message will always lead to a a reconnect under the hood, which will take some time. I wanted to avoid spinning cycles if, for some reason, we were failing the connection immediately. 

But removing this, as I just checked and we have a sleep in the actual client retry code.

crates/hotshot/src/traits/networking/push_cdn_network.rs
rob-maron: The drop inside the client should be sufficient. I will do some testing tomorrow, but we have `Drop` implemented for the underlying connection so that when a reference there gets dropped, the underlying channels get terminated and the underlying task gets stopped as a result.

crates/hotshot/src/traits/networking/push_cdn_network.rs
rob-maron: Okay, on another run through: `data` is actually a direct or broadcast-specific message (needed to add recipient or topic metadata embedded in the message), which means they cannot be thrown in a match arm or anything. I can do some deduplication though. 

The other option is to have the broker remove the metadata and just send to the user as a `Data` message directly. I was originally doing this but wanted to reduce broker (re)allocations and cycles.

rob-maron:
> Requesting changes because of CI failures, otherwise this looks ready to go

Inadvertently created an issue while adding better timeout logic, almost ready with the fixes

bfish713 (approved):

PR #2766, opened 2024-03-13
Remove excessive logging

ss-es:
No linked issue.

### This PR: 
- Removes logging from adding bootstrap addresses in `hotshot/src/traits/libp2p_network.rs`. This is probably the biggest polluter of log files in our integration tests, creating about ~20 MB of logs (over 50% of the total log) for no reason. We now only log when bootstrapping is completed.
- Removes logging from failing to receive messages in a loop in `hotshot/src/traits/web_server_network.rs`. This is also a significant polluter in logs, and is not logging anything helpful.

### DIFF
--- a/crates/hotshot/src/traits/networking/web_server_network.rs
+++ b/crates/hotshot/src/traits/networking/web_server_network.rs
@@ -503,45 +503,38 @@ impl<TYPES: NodeType> Inner<TYPES> {
                 }
             }
 
-            let maybe_event = receiver.try_recv();
-            match maybe_event {
-                Ok(event) => {
-                    match event {
-                        // TODO ED Should add extra error checking here to make sure we are intending to cancel a task
-                        ConsensusIntentEvent::CancelPollForVotes(event_view)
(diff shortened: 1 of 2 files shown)


crates/hotshot/src/traits/networking/web_server_network.rs
```diff
@@ -503,47 +503,38 @@ impl<TYPES: NodeType> Inner<TYPES> {
...
-                        | ConsensusIntentEvent::CancelPollForLatestProposal(event_view)
-                        | ConsensusIntentEvent::CancelPollForLatestViewSyncCertificate(
-                            event_view,
-                        )
-                        | ConsensusIntentEvent::CancelPollForViewSyncVotes(event_view) => {
-                            if view_number == event_view {
-                                debug!("Shutting down polling task for view {}", event_view);
-                                return Ok(());
-                            }
+            if let Ok(event) = receiver.try_recv() {
```
ss-es, line 506: Most of this is a clippy-mandated change to delete the `match` after removing the log on `Err(_)`. The logic should be exactly the same.

PR #2769, opened 2024-03-13
Send each recipient one corresponding VID share

### LINKED ISSUE #1696: Custom message transmission logic for VID dispersal
From https://github.com/EspressoSystems/HotShot/pull/1684#discussion_r1319231209 :
VID dispersal sends a different message to everyone. It is neither a `Broadcast` nor a `Direct` message, so it does not fit into the existing `TransmitType` format. We need a custom solution.

lukaszrzasik:
Closes #1696

### This PR: 

Modifies `VidDisperseSend` to use a new struct `VidDisperseShare`.
`VidDisperseShare` substitutes `VidDisperse`.
`VidDisperseShare` consists of only one VID share and keeps information about its dedicated recipient.
`VidDisperseShare` is sent as a direct message to its recipient only.

### Key places to review: 

Changes in crates/task-impls/src/consensus.rs
Changes in crates/task-impls/src/network.rs
Changes in crates/task-impls/src/vid.rs

### How to test this PR:

All the tests should pass.

### DIFF
--- a/crates/testing/src/task_helpers.rs
+++ b/crates/testing/src/task_helpers.rs
@@ -398,7 +398,7 @@ pub fn da_payload_commitment(
     vid_commitment(&encoded_transactions, quorum_membership.total_nodes())
 }
 
-/// TODO: https://github.com/EspressoSystems/HotShot/issues/2821
+/// TODO: <https://github.com/EspressoSystems/HotShot/issues/2821>
 pub fn build_vid_proposal(
     quorum_membership: &<TestTypes as NodeType>::Membership,
     view_number: ViewNumber,


crates/task-impls/src/vid.rs
```diff
@@ -105,26 +105,19 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
+                    broadcast_tasks.push(broadcast_event(
+                        Arc::new(HotShotEvent::VidDisperseSend(
+                            vid_disperse_share.to_proposal(&self.private_key)?,
+                            self.public_key.clone(),
+                        )),
+                        &event_stream,
+                    ));
+                }
+                debug!("publishing VID disperse shares for view {}", *view_number);
+                join_all(broadcast_tasks).await;
```
bfish713, line 120: This is just creating a bunch of tasks which write to the broadcast queue,  It maybe makes more sense to just send all the shares + signature in the event and handle the splitting and sending on the network side.

crates/task-impls/src/consensus.rs
```diff
@@ -1062,12 +1071,18 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
+                {
+                    let mut consensus_lock = self.consensus.write().await;
+                    // Add to the storage that we have received the VID disperse for a specific view
+                    if let Some(vid_disperse) = consensus_lock.vid_shares.get_mut(&view) {
+                        vid_disperse.insert(disperse.data.recipient_key.clone(), disperse.clone());
+                    } else {
+                        let mut proposals_map = HashMap::new();
+                        proposals_map.insert(disperse.data.recipient_key.clone(), disperse.clone());
+                        consensus_lock.vid_shares.insert(view, proposals_map);
+                    }
```
bfish713, line 1083: nit: Can be simplified with `.entry.or_default` instead of get_mut.

https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html#method.or_default

bfish713 (commented):
I think it might be better to create a VID broadcast in the network layer similar to the DA version of broadcast.  This would allow each network to easily determine how they want to handle the bundle.  E.g. we night want to just send everything together to the CDN eventually.

bfish713:
> > `test_with_failures_f` with Libp2p implementation fails consistently. I've check it with `next_view_timeout: 10000` (doubled) and then it doesn't fail. I'm trying to analyse it further to find out what's causing the slowdown.
> 
> Unfortunately this is still true after reworking `VidDisperseSend` not to be split into many separate events with one share.



> > `test_with_failures_f` with Libp2p implementation fails consistently. I've check it with `next_view_timeout: 10000` (doubled) and then it doesn't fail. I'm trying to analyse it further to find out what's causing the slowdown.
> 
> Unfortunately this is still true after reworking `VidDisperseSend` not to be split into many separate events with one share.

Not 100% certain but my intuition is that it's taking a long time because we get slowed down on the direct messages where can't send.  I think those request end up timing out and it's possible that some part of the code is blocked from sending more messages while this happens.  

More specifically, when you join on all the direct send futures in `vid_broadcast_message` that's essentially just sending a bunch of things on a channel to the libp2p task and awaiting results.  I wouldn't be surprised to find out that the direct_messages get queued up and sent sequentially in the libp2p implementation.

crates/libp2p-networking/src/network/behaviours/direct_message.rs
```diff
@@ -132,13 +132,16 @@ impl NetworkBehaviour for DMBehaviour {
...
-                self.add_direct_request(req);
-            } else {
-                self.failed_rr.push_back(req);
+                retry_req_indices.push(idx);
             }
         }
+        let _ = retry_req_indices.into_iter().map(|idx| {
+            let req = self.failed_rr.remove(idx).unwrap();
+            self.add_direct_request(req);
+        });
```
bfish713, line 144: Nice this looks will fix things for now

crates/task-impls/src/network.rs
```diff
@@ -446,4 +439,74 @@ impl<
 
         None
     }
+
+    /// handle `VidDisperseSend`
+    fn handle_vid_disperse_proposal(
+        &self,
+        vid_proposal: Proposal<TYPES, VidDisperse<TYPES>>,
+        sender: &<TYPES as NodeType>::SignatureKey,
+        maybe_action: Option<HotShotAction>,
```
bfish713, line 448: nit: I don't think we need this argument, we can just use `Some(HotShotAction::VidDisperse)` in function

crates/testing/src/task_helpers.rs
```diff
@@ -402,27 +403,20 @@ pub fn build_vid_proposal(
...
         &quorum_membership.clone().into(),
     );
 
-    Proposal {
-        data: vid_disperse.clone(),
-        signature: vid_signature,
-        _pd: PhantomData,
-    }
+    VidDisperseShare::from_vid_disperse(vid_disperse)
+        .swap_remove(0)
```
bfish713, line 417: This is probably going to need some updating in the future.  Either we want to return all the proposals as a map `public_key -> Proposal` or return for a specific key.  TODO comment + issue is fine if the tests are working.

bfish713 (dismissed):
Overall looks good just some really minor comments.  If you change them I'll approve again

bfish713 (approved):

PR #2770, opened 2024-03-14
[TECH_DEBT] Pass Events as `Arc`s

bfish713:
### This PR: 

This is an experiment with changing the internal events to be `Arc`s.  There are some pros and cons that I see but am wondering what others think.

The main reason I see for considering this is that the broadcast channel we use will clone the event for each receiver.  Every receiver must get every event.  As we add more tasks that means more clones.  Also I think each task receives every message and then filters most messages (i.e most `HotshotEvents` are just dropped).  This means that as we add more tasks we'll end up with event clones.  Some events can be large as well (e.g. `DAProposalRecv`).  

Pros:
- Events no longer must be cloned for each receiver.
- We won't have to worry about extra clones when we add extra tasks

Cons:
- This probably can be considered a premature optimization
- We have to allocate every time we create an event, when it's created.
- There might be a better overall solution that filters events before sending
- We still have to clone part or all of the events in some tasks during handling.

### This PR does not: 

Everything should work exactly as before.

### Key places to review: 

Generally review if this is a good/bad idea
`task_impls`


ss-es:
+1 from me; seems simple and I agree completely with the reasoning.

I don't see any significant negatives; at worst, this introduces some mild overhead from `Arc`, but we probably wouldn't even be able to measure it. Avoiding cloning every message into every receiver seems worth it. 

For filtering before sending to a task: I think we could do this fairly easily by just adding an extra event collector task to do the filtering and giving every other task an independent channel to/from it. But I don't think this has much of an advantage over wrapping everything in an `Arc` (and we'd probably still want to do that anyway).

rob-maron:
I'm in favor of `Arc`ing.

bfish713:
> For filtering before sending to a task: I think we could do this fairly easily by just adding an extra event collector task to do the filtering and giving every other task an independent channel to/from it. But I don't think this has much of an advantage over wrapping everything in an `Arc` (and we'd probably still want to do that anyway).

I've considered this but it is really tricky with new tasks being spawned.  We would have to ensure they get the same stream of events as the task which spawned them.  It might be as simple as cloning the parent's receiver, adding it to the event filtering task, and still filtering in the child task.  Would involve a bit of synchronization to spawn a task which we currently don't have to worry about now, so I agree with your conclusion.

shenkeyao (approved):
+1 to use `Arc`s. All the listed pros make sense, and the cons seem to be things we could work on later (e.g., the filtering optimization as @ss-es mentioned) that won't get worse with `Arc`s.

The changes also look good besides the reasoning.

Issue #2708, opened 2024-03-06
[CX_CLEANUP] - Implement `QuorumVoteTask` with dummy dependencies


shenkeyao:
### What is this task and why do we need to work on it?

* Add `QuorumVoteTaskState` with `next_vote_view` and `vote_dependencies`.
* Add an event-handling function.
  * If receiving `ViewChange`: update the state.
  * If receiving other events: spawn the `AddDependency` which will be `completed` once receiving all three dependent events.

### Are there any other details to include?

Design doc: https://www.notion.so/espressosys/Refactor-Voting-Logic-9afa14dd027d453aa9b4a80a52de5e44.

### What are the acceptance criteria to close this issue?

Code compiles, PR approved, test to be done next.

Issue #2710, opened 2024-03-06
[CX_CLEANUP] - Complete the dependency implementation


shenkeyao:
### What is this task and why do we need to work on it?

- Fill in implementations (i.e., the validation part) for the three `EventDependency`s.
- Store the VID and leaf commitment for field matches among events.

### Are there any other details to include?

Design doc: https://www.notion.so/espressosys/Refactor-Voting-Logic-9afa14dd027d453aa9b4a80a52de5e44.

### What are the acceptance criteria to close this issue?

Code compiles, PR approved, test to be done next.

PR #2776, opened 2024-03-14
[CX_CLEANUP] - Implement `QuorumVoteTask` with dummy dependencies

shenkeyao:
Closes https://github.com/EspressoSystems/HotShot/issues/2708.

### This PR: 
- Adds basic structures for `QuorumVoteTask` with dummy event verification logic.
- Renames `vote.rs` to `vote_collection.rs`.

### This PR does not: 
- Update `consensus.rs`, i.e., the new vote logic isn't integrated with other code yet.
- Fully verify an event dependency besides the view number check.
- Store and update all task state info, which will be added when completing the dependency verification logic.

### Key places to review: 
- `crates/task-impls/src/quorum_vote.rs`.
- [Comment](https://github.com/EspressoSystems/HotShot/pull/2776#discussion_r1525444652) about state field update.


crates/task-impls/src/quorum_vote.rs
```diff
@@ -0,0 +1,335 @@
...
+    #[instrument(skip_all, fields(id = self.id, next_vote_view = *self.next_vote_view), name = "Quorum vote validate VID", level = "error")]
+    fn create_event_dependency(
+        &self,
+        dependency_type: VoteDependency,
+        view_number: TYPES::Time,
+        event_sender: Sender<HotShotEvent<TYPES>>,
+    ) -> EventDependency<HotShotEvent<TYPES>> {
+        EventDependency::new(
+            self.event_receiver.clone(),
+            Box::new(move |event| match dependency_type {
```
shenkeyao, line 100: A potential problem for the next PR:

There might be lifetime issues if passing `&self` inside this `Box`, but we need to update some fields of `self` when verifying the event dependencies.

shenkeyao: This may work if we update the fields in `handle` where we have `&mut self`, rather than in `create_event_dependency` where accessing `&self` will cause `borrowed data escapes outside of method`.

I can test this idea in the next PR when adding dependency verification logic, but would like to avoid obvious flaws. Curious if anyone has insights on this!

crates/task-impls/src/quorum_vote.rs
```diff
@@ -0,0 +1,359 @@
...
+                        return validate_vid(disperse.clone(), event_sender.clone());
+                    }
+                    false
+                }
+            }),
+        )
+    }
+
+    /// Create an [`AndDependency`] combining three [`EventDependency`]s.
+    fn create_vote_dependency(
```
bfish713, line 145: How will the vote dependency (`AndDependency`) get access to the event which triggered this function to be called?

crates/task-impls/src/quorum_vote.rs
```diff
@@ -0,0 +1,359 @@
...
+                    );
+                    self.vote_dependencies.insert(
+                        view,
+                        async_spawn(async move {
+                            dependency.completed().await;
+                            broadcast_event(HotShotEvent::ViewChange(view + 1), &event_sender)
+                                .await;
+                            broadcast_event(HotShotEvent::DummyQuorumVoteSend(view), &event_sender)
+                                .await;
+                        }),
```
bfish713, line 242: I think the idea of the `DependencyTask` was created to avoid this duplication.

crates/task-impls/src/quorum_vote.rs
jparr721: Can we pass internal state through a locked arc pointer to avoid the escape issue?

jparr721 (commented):
Would love a quick unit test for this. Also, do we need to handle view sync certs now that they're merged into the consensus task?

crates/task-impls/src/quorum_vote.rs
```diff
@@ -0,0 +1,359 @@
...
+                    return;
+                }
+            }
+            _ => {}
+        }
+    }
+}
+
+impl<TYPES: NodeType, I: NodeImplementation<TYPES>> TaskState for QuorumVoteTaskState<TYPES, I> {
+    type Event = HotShotEvent<TYPES>;
```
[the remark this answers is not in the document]
shenkeyao: Checking my understanding -- Is the `Arc` needed for memory efficiency or thread safety? Since you mentioned "have to" I assume it's the latter, but I'm not sure why it's a problem to not use `Arc`.

crates/task-impls/src/quorum_vote.rs
shenkeyao: I think clone/copy makes sense here. According to the current consensus task implementation where we access `self`, it seems unnecessary to use a lock.

shenkeyao:
@jparr721 --

> Would love a quick unit test for this.

As discussed in the meeting today, I'll add unit tests in the next PR.

> Also, do we need to handle view sync certs now that they're merged into the consensus task?

Yes! I think we also need to handle the upgrade certs. Will add these to the future PR with real validation logic. Would like to keep this PR simple to focus on the basic interface.

crates/task-impls/src/quorum_vote.rs
shenkeyao: I've replaced the `create_vote_dependency` function with `create_dependency_task_if_new`, and it will be called every time we receive an event and want to create a new `DependencyTask`.

crates/task-impls/src/quorum_vote.rs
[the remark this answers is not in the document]
shenkeyao: Oh just realized events were all wrapped inside `Arc` in other tasks. I'll update this.

crates/task-impls/src/quorum_vote.rs
```diff
@@ -0,0 +1,383 @@
...
+        self.vote_dependencies.insert(
+            view_number,
+            async_spawn(async move {
+                #[cfg(async_executor_impl = "async-std")]
+                dependency_task.run().await;
+                // Allow `unwrap()` for tokio.
+                #[cfg(async_executor_impl = "tokio")]
+                dependency_task.run().await.unwrap();
+            }),
+        );
```
bfish713, line 233: I think this can just call run on the task.  It internally does the spawning.
```suggestion
        self.vote_dependencies.insert(
            view_number,
            dependency_task.run(),
        );
```

crates/task-impls/src/quorum_vote.rs
```diff
@@ -0,0 +1,383 @@
...
+                debug!("DAC Received for view {}!", *cert.view_number);
+                let view = cert.view_number;
+
+                self.quorum_network
+                    .inject_consensus_info(ConsensusIntentEvent::CancelPollForDAC(*view))
+                    .await;
+
+                self.committee_network
+                    .inject_consensus_info(ConsensusIntentEvent::CancelPollForVotes(*view))
+                    .await;
```
bfish713, line 300: It'll be nice when half of this function disappears once we can get rid of the Polling.

crates/task-impls/src/quorum_vote.rs
```diff
@@ -0,0 +1,383 @@
...
+                    return;
+                }
+                broadcast_event(
+                    Arc::new(HotShotEvent::DACValidated(cert.clone())),
+                    &event_sender.clone(),
+                )
+                .await;
+                self.create_dependency_task_if_new(view, event_receiver, event_sender);
+            }
+            HotShotEvent::VidDisperseRecv(disperse) => {
```
bfish713, line 312: I think you also need to check last voted view here.

crates/task-impls/src/quorum_vote.rs
```diff
@@ -0,0 +1,383 @@
...
+                )
+                .await;
+                self.create_dependency_task_if_new(view, event_receiver, event_sender);
+            }
+            HotShotEvent::QuorumVoteDependenciesValidated(view) => {
+                if !self.update_latest_voted_view(*view).await {
+                    debug!("view not updated");
+                    return;
+                }
+            }
```
bfish713, line 337: I think this is a good way to handle the lasted voted view.

crates/task-impls/src/quorum_vote.rs
```diff
@@ -0,0 +1,383 @@
...
+                    return;
+                }
+                broadcast_event(
+                    Arc::new(HotShotEvent::QuorumProposalValidated(proposal.data.clone())),
+                    &event_sender.clone(),
+                )
+                .await;
+                self.create_dependency_task_if_new(view, event_receiver, event_sender);
+            }
+            HotShotEvent::DACRecv(cert) => {
```
bfish713, line 290: I think we also need to check last voted view here or else we could spawn a task for old views here.

crates/task-impls/src/quorum_vote.rs
```diff
@@ -0,0 +1,383 @@
...
+                    let cert_payload_comm = cert.get_data().payload_commit;
+                    if let Some(comm) = payload_commitment {
+                        if cert_payload_comm != comm {
+                            return;
+                        }
+                    } else {
+                        payload_commitment = Some(cert_payload_comm);
+                    }
+                }
+                _ => {}
```
bfish713, line 114: I think we also need to check that VID Share commitment matches the `payload_commitment`.  We don't do that in the current code either but I think we should

bfish713 (commented):
Structure of this looks good, just a minor logic improvements

crates/task-impls/src/quorum_vote.rs
shenkeyao: Sounds good! I added a comment to line 87 without the implementation. This PR includes one commitment check just to test if the framework works, and I'll add more verifications when completing #2710.

bfish713 (approved):

PR #2777, opened 2024-03-14
[TECH-DEBT] Add View Sync Timeout to Config

### LINKED ISSUE #1472: Add Timeout Config to View Sync Task
Consensus uses the timeout from the config for it's timer.  The view sync task should also have have a configurable timeout

bfish713:
Closes #1472 

### This PR: 

Adds a view sync timeout duration to the config and makes the default a less time than the view timeout.

### This PR does not: 

Shouldn't change anything else

### Key places to review: 

Are the defaults reasonable

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,17 +12,17 @@ repository = "https://github.com/EspressoSystems/HotShot"
 # when implementing traits externally
 [workspace]
 members = [
-        "crates/hotshot",
-        "crates/hotshot-stake-table",
-        "crates/libp2p-networking",
-        "crates/macros",
-        "crates/testing-macros",
-        "crates/task",
-        "crates/task-impls",
-        "crates/testing",
-        "crates/utils",
-        "crates/examples",
-        "crates/example-types",
+  "crates/hotshot",
+  "crates/hotshot-stake-table",
+  "crates/libp2p-networking",
+  "crates/macros",
+  "crates/testing-macros",
+  "crates/task",
+  "crates/task-impls",
+  "crates/testing",
+  "crates/utils",
+  "crates/examples",
+  "crates/example-types",
 ]
 resolver = "2"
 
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -84,45 +84,45 @@ toml = "0.8.11"
...
-        "relay",
-        "request-response",
-        "rendezvous",
-        "secp256k1",
-        "serde",
-        "tcp",
-        "uds",
-        "websocket",
-        "yamux",
+  "macros",
```
jparr721, line 87: What did you run to format this? Is it a vscode command?

crates/hotshot/src/traits/election/static_committee.rs
```diff
@@ -80,9 +80,9 @@ where
     #[cfg(feature = "randomized-leader-election")]
     /// Index the vector of public keys with a random number generated using the current view number as a seed
     fn get_leader(&self, view_number: TYPES::Time) -> PUBKEY {
-        let mut rng: StdRng = rand::SeedableRng::seed_from_u64(*view_number as u64);
-        let randomized_view_number: u64 = rng.gen();
-        let index = (randomized_view_number % self.nodes_with_stake.len() as u64) as usize;
+        let mut rng: StdRng = rand::SeedableRng::seed_from_u64(*view_number);
+        let randomized_view_number: usize = rng.gen();
```
jparr721, line 84: Nice.

jparr721 (approved):
This looks great. Though I think I'll need to re-approve after your types PR goes in.

Cargo.toml
bfish713: Actually been trying out neovim and it was doing the toml formatting

bfish713:
> This looks great. Though I think I'll need to re-approve after your types PR goes in.

Mind taking a look at the other PR?

Cargo.toml
jparr721: Oooh I might bug you for your config later

jparr721 (approved):

PR #2784, opened 2024-03-15
[Draft][CX-CLEANUP] - DA Integrated Storage

### LINKED ISSUE #2621: [CX-CLEANUP] - DA Integrated Storage
We want to integrate a persistent storage interface within HotShot to enable the application to manage data storage effectively. This feature aims to address potential data storage issues (e.g., full disk space) that can prevent nodes from storing DA blobs or VID shares.

See details here: 
https://www.notion.so/espressosys/DA-integrated-storage-27cba2adb0f744fc8ac99591140627c2

Also Discussion on Zulip:

> 
> Jeb Bearer: A while ago we discussed an idea of having a persistent storage interface in HotShot implemented by the application. This would allow the application to store data before HotShot signed a DA proposal. We decided this was not a big enough improvement to bring into decaf because the odds of a node crashing between signing a DA proposal and decide, and thus losing data, is so small.
> 
> I want to revisit this decision because there is another failure mode. What if the node doesn't crash, remains online but unable to store data (for example, its disk is full). This condition could persist for some time and thus make it more likely that it happens in the critical window. This becomes more likely when we implement [pruning](https://github.com/EspressoSystems/hotshot-query-service/issues/376) for DA-but-not-archival nodes, if we are having higher volume than expected, the storage requirement for DA retention could exceed our minimum hardware recommendations.
> 
> Having an application persistence interface integrated with the DA replica logic would allow the application to stop its node from voting on DA (and similarly VID) if it is unable to store its DA blob (or VID share) for any reason.
> 
> Jeb Bearer: Does this sound valuable enough to integrate for decaf?
> 
> Brendon Fish: To me this seems reasonable to add, I think. Does the sequencer currently use/impl the Storage trait in Hotshot? Maybe it calls store_leaf on ConsensusAPI? I'm guessing it doesn't but if it does we can probably work from that.
> 
> Let me know if my understanding of what Hotshot would need to add is correct:
> 
> Create a trait for block storage which the application implements, just like the other Block types
> Take/create a storage when initializing hotshot
> When we receive a DA proposal, call some append or store function from this trait.
> If storing succeeds vote on the DA proposal
> If it fails don't vote
> Same logic for VID shares except don't vote on the Quorum proposal
> Jeb Bearer: Yeah exactly
> 
> Jeb Bearer: We could also then remove the in-memory payload/VID stores in HotShot, and remove them from the Decide event. The application could be responsible for finding the relevant data and archiving it when it sees a decide, and garbage collecting it
> 
> Jeb Bearer: And on the application side, this is a pretty straightforward key-value store, should be easy to implement in postgres
> 

1. **Create `BlockStorage` Trait**
2. **Initialize `BlockStorage` in HotShot's init**
3. **Add Storage Operations on DA Proposals**
    - If this fails, do not vote on the proposal and exit the event handler.
4. **Add Storage Operations on DA Proposals VID Shares**
   - If this fails, do not vote on the VID disperse, otherwise, use the old logic to vote.
5. **Refactor In-memory Payload/VID Stores**:
   - Remove in-memory storage for payloads and VID shares from HotShot. This involves removing the `self.consensus.write().await.vid_shares.insert(view, disperse.clone());` call from the code.

_No response_

- Provide an interface to store DA data, and properly store it using that interface.
- Write a unit test interface for `BlockStorage` called `TestBlockStorage` which uses some simple data type for testing purposes.
- Write tests for both the `DA` and `Consensus` tasks to validate that data is indeed stored properly.

_No response_

jparr721:
Closes #2621 

We want to integrate a persistent storage interface within HotShot to enable the application to manage data storage effectively. This feature aims to address potential data storage issues (e.g., full disk space) that can prevent nodes from storing DA blobs or VID shares.

### This PR: 
1. **Creates `BlockStorage` Trait**
2. **Initializes `BlockStorage` in HotShot's init**
3. **Add Storage Operations on DA Proposals**
    - If this fails, do not vote on the proposal and exit the event handler.
4. **Add Storage Operations on DA Proposals VID Shares**
   - If this fails, do not vote on the VID disperse, otherwise, use the old logic to vote.
5. **Refactor In-memory Payload/VID Stores**:
   - Remove in-memory storage for payloads and VID shares from HotShot. This involves removing the `self.consensus.write().await.vid_shares.insert(view, disperse.clone());` call from the code for VID disperse events.

This code also implements a number of changes to introduce a `TestBlockStorage` type for use in the various testing components.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,7 +59,7 @@ futures = "0.3.30"
 # TODO generic-array should not be a direct dependency
 # https://github.com/EspressoSystems/HotShot/issues/1850
 generic-array = { version = "0.14.7", features = ["serde"] }
-hotshot-types = { git = "https://github.com/EspressoSystems/hotshot-types", tag = "0.1.4" }
+hotshot-types = { git = "https://github.com/EspressoSystems/hotshot-types", branch = "jparr721/issue-2621" }
 
 jf-primitives = { git = "https://github.com/EspressoSystems/jellyfish", tag = "0.4.1" }
 jf-plonk = { git = "https://github.com/EspressoSystems/jellyfish", tag = "0.4.1" }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3425,7 +3425,7 @@ dependencies = [
 [[package]]
 name = "hotshot-types"
 version = "0.1.5"
-source = "git+https://github.com/EspressoSystems/hotshot-types?branch=jparr721/issue-2621#f75c8da63d461685e6b7c5370a58bfdc3f345a25"
+source = "git+https://github.com/EspressoSystems/hotshot-types?branch=jparr721/issue-2621#cb4419edb6fc28c1e55a783c575ef40c93282c2f"
 dependencies = [
  "ark-bls12-381",
  "ark-bn254",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1451,9 +1451,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.5.2"
+version = "4.5.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b230ab84b0ffdf890d5a10abdbc8b83ae1c4918275daea1ab8801f71536b2651"
+checksum = "949626d00e063efc93b6dca932419ceb5432f99769911c0b995f7e884c778813"
 dependencies = [
  "clap_builder",
  "clap_derive",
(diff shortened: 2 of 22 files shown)


Cargo.toml
```diff
@@ -59,7 +59,7 @@ futures = "0.3.30"
 # TODO generic-array should not be a direct dependency
 # https://github.com/EspressoSystems/HotShot/issues/1850
 generic-array = { version = "0.14.7", features = ["serde"] }
-hotshot-types = { git = "https://github.com/EspressoSystems/hotshot-types", tag = "0.1.4" }
+hotshot-types = { git = "https://github.com/EspressoSystems/hotshot-types", branch = "jparr721/issue-2621" }
```
jparr721, line 62: Tracking: https://github.com/EspressoSystems/hotshot-types/pull/15

PR #2790, opened 2024-03-15
[CX_CLEANUP] - Deprecate and remove Storage trait and usages

### LINKED ISSUE #2788: [CX_CLEANUP] - Deprecate and remove `Storage` trait and usages

We have a lot of old dead code because conensus used be completely state driven, i.e. blocks were state updates, it used to be very explicitly one view at a time, not event driven. One of these artifacts exists as the `Storage` trait, found [here](https://github.com/EspressoSystems/hotshot-types/blob/main/src/traits/storage.rs). 

As a result of the need to support storing DA, VID diseperals, and later work by sequencer (tracking: #2621), we need to deprecate and remove this type from the code and modernize its name. This issue is preliminarily responsible for _only_ the removal of this type from the code.

- Remove the type from `HotShot-Types`
- Remove `MemoryStorage`
- Remove `AtomicStorage`

_No response_

Remove all instances of the `Storage` trait and derivative implementations. All existing tests should still pass without needing to change anything in a substantial way.

_No response_

jparr721:
Closes #2788

We have a lot of old dead code because conensus used be completely state driven, i.e. blocks were state updates, it used to be very explicitly one view at a time, not event driven. One of these artifacts exists as the Storage trait, found [here](https://github.com/EspressoSystems/hotshot-types/blob/main/src/traits/storage.rs).

As a result of the need to support storing DA, VID diseperals, and later work by sequencer (tracking: https://github.com/EspressoSystems/HotShot/issues/2621), we need to deprecate and remove this type from the code and modernize its name. This issue is preliminarily responsible for only the removal of this type from the code.

### This PR: 
- Removes the type from `HotShot-Types`
- Removes `MemoryStorage`
- Remove s`AtomicStorage`
- Updates `node_types.rs`, `crates/hotshot/lib.rs`, and the associated test generator files to remove all usages of the storage type.

### This PR does not: 
While the intention is to pave the way for the new storage type, we shrink the scope here by first removing all of the cruft of the old code.

### Key places to review: 
All logic where code was touched that affects production environments, especially `node_types.rs` and `lib.rs`


crates/hotshot/src/types/handle.rs
```diff
@@ -53,6 +50,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES> + 'static> SystemContextHandl
     /// there are two cleaner solutions:
     /// - make the stream generic and in nodetypes or nodeimpelmentation
     /// - type wrapper
+    #[must_use]
```
jparr721, line 53: Clippy went buck wild over this. Not sure why.

bfish713 (dismissed):

bfish713 (approved):

PR #2792, opened 2024-03-15
Example upgrade

### LINKED ISSUE #2811: [Upgradability] - Add example upgrade script

We want some certainty that we can actually perform an upgrade. We should have an example upgrade script that simulates the process with a small number of nodes.

_No response_

_No response_

Script that simulates a working upgrade is added.

_No response_

ss-es:
Closes #2811 

Broadly, the goal was to have some evidence that upgrading is possible and works as intended. 

### This PR:
- Adds a feature flag `example-upgrade`, which makes hotshot a) vote affirmatively on any upgrade proposal it receives and b) has the leader for view 10 issue an upgrade proposal in view 5.
- Fixes some logical errors in the previous upgrade process, including omissions in the network task.
- As a minor simplification, we no longer check whether an `UpgradeProposal` came from the view leader or not. This is an unnecessary check, since we are primarily voting on the upgrade hash, not the proposer.

Minor code changes not related to the upgrade:
- `SequencingMessage` is now an enum, rather than an `Either`. This was originally done to accommodate an additional raw message type; I eventually scrapped the message type, but left this change in since it seemed to be an improvement.
- `TransmitType::Direct` is now `TransmitType::Direct(TYPES::SignatureKey)`. The network task previously had an unnecessary possible panic, and this seemed a quick fix to make while working on it. While it was currently unreachable, it would've been easy to trigger in the future by simply adding a match arm with mismatched `TransmitType` and recipient.

### Key places to review:
The changes to `web_server` and `web_server_network` can be ignored, since these will be removed soon.

You can test the example upgrade process by running:

```
just cargo run --features example-upgrade --example all-push-cdn -- --config_file ./crates/orchestrator/run-config.toml
```
```
just cargo run --features example-upgrade --example all-libp2p -- --config_file ./crates/orchestrator/run-config.toml
```
```
just cargo run --features example-upgrade --example all-webserver -- --config_file ./crates/orchestrator/run-config.toml
```
```
just cargo run --features example-upgrade --example all-combined -- --config_file ./crates/orchestrator/run-config.toml
```

The expectation is that the leader for view 10 (node 0) sends out an `UpgradeProposal` in view 5 (scheduled for view 18), forms an `UpgradeCertificate` and attaches it to their next proposal. Once the leaf containing the `UpgradeCertificate` is decided, nodes update their internal consensus state (this is logged at `WARN`, so will not show up without `RUST_LOG=warn`). At view 18, the consensus task should emit a `VersionUpgrade` event, which the network task will use to update its own internal state. Since there is currently no implementation of this version, this functionally bricks the network task with an error.

(Alternatively, we could emit `HotShotEvent::Shutdown` for an unimplemented upgrade. I chose not to do this, to avoid any risk of shutting down all tasks if the code path is reachable in some unintended way.)

### Remaining work: 
The upgrade example stops once the internal state is updated and the network task can no longer send messages. It will be good to have an example where messages are being correctly sent (and handled) with the new version, but this may take some nontrivial work with the recent versioning changes. This is tracked as a stretch goal in #2842.


crates/task-impls/src/upgrade.rs
```diff
@@ -72,48 +72,34 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
     ) -> Option<HotShotTaskCompleted> {
         match event.as_ref() {
             HotShotEvent::UpgradeProposalRecv(proposal, sender) => {
-                let should_vote = self.should_vote;
+                info!("Received upgrade proposal: {:?}", proposal);
+
                 // If the proposal does not match our upgrade target, we immediately exit.
-                if !should_vote(proposal.data.upgrade_proposal.clone()) {
-                    warn!("Received unexpected upgrade proposal:\n{:?}", proposal.data);
+                if !(self.should_vote)(proposal.data.upgrade_proposal.clone()) {
```
jparr721, line 78: Is `(self.should_vote)` correct here? Does it return a function somehow?

crates/task-impls/src/upgrade.rs
```diff
@@ -139,17 +125,21 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
                 broadcast_event(Arc::new(HotShotEvent::UpgradeVoteSend(vote)), &tx).await;
             }
             HotShotEvent::UpgradeVoteRecv(ref vote) => {
-                debug!("Upgrade vote recv, Main Task {:?}", vote.get_view_number());
+                error!("Upgrade vote recv, Main Task {:?}", vote.get_view_number());
```
jparr721, line 128: Does receiving an upgrade vote imply an error?

jparr721 (changes requested):
Curious about a few things, not sure how much of these are my own misunderstanding or not, just let me know if I've erroneously highlighted anything.

crates/hotshot/src/lib.rs
```diff
@@ -135,6 +136,9 @@ pub struct SystemContext<TYPES: NodeType, I: NodeImplementation<TYPES>> {
     /// The hotstuff implementation
     consensus: Arc<RwLock<Consensus<TYPES>>>,
 
+    /// The network version
+    version: Arc<RwLock<Version>>,
```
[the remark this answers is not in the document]
ss-es: It's not actually static -- it gets updated by the consensus task if an upgrade is reached. 

There might be another channel for this kind somewhere, but I think you need something like this to communicate with the `SystemContextHandle` for stuff that might not be listening on the event stream (I think the sequencer might be an example?).

crates/task-impls/src/consensus.rs
```diff
@@ -1140,23 +1141,36 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
-                    debug!("view not updated");
-                    return;
-                }
-
                 // If we have a decided upgrade certificate,
                 // we may need to upgrade the protocol version on a view change.
                 if let Some(ref cert) = self.decided_upgrade_cert {
                     if new_view >= cert.data.new_version_first_block {
-                        self.current_network_version = cert.data.new_version;
+                        warn!(
```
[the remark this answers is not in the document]
ss-es: I think it should be at least a warn: if we hit a decided upgrade certificate, it means the network has reached consensus on an upgrade. This should happen extremely rarely, we'll always want to be aware when it happens and if it's unexpected manual intervention will always be required (HotShot would effectively shutdown).

crates/task-impls/src/upgrade.rs
ss-es: Oh, oops -- good catch! Back to `debug`!

crates/task-impls/src/upgrade.rs
ss-es: Yeah, `self.should_vote` has type `fn(UpgradeProposalData<TYPES>) -> bool`. There are probably a few ways of doing it; this just gives some flexibility in specifying exactly what proposals you want to accept, without having to change it directly in the consensus logic (you might want to vote yes on a specific upgrade but only within a certain range of view numbers, for example).

crates/hotshot/src/traits/networking/web_server_network.rs
```diff
@@ -929,6 +952,70 @@ impl<TYPES: NodeType + 'static, NetworkVersion: StaticVersionType + 'static>
...
+                                Duration::from_millis(500),
+                            )
+                            .await
+                        {
+                            warn!(
+                                "Background receive latest upgrade proposal polling encountered an error: {:?}",
+                                e
+                            );
+                        } else {
+                            error!("Poll successful!");
```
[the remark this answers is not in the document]
ss-es: Oops, another leftover from debugging. Removed.

crates/task-impls/src/response.rs
```diff
@@ -136,10 +134,10 @@ impl<TYPES: NodeType> NetworkResponseState<TYPES> {
     ) -> Message<TYPES> {
         if !proposals_map.contains_key(key) {
             return self.make_msg(ResponseMessage::NotFound);
-        }
-        let seq_msg = SequencingMessage(Right(CommitteeConsensusMessage::VidDisperseMsg(
+        };
```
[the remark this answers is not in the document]
ss-es: Weird, not sure where that came from (or why `rustfmt` didn't catch it, actually) -- removed.

crates/hotshot/src/traits/networking/web_server_network.rs
```diff
@@ -929,6 +952,70 @@ impl<TYPES: NodeType + 'static, NetworkVersion: StaticVersionType + 'static>
...
+                    *cancel_handle = Some(sender);
+
+                    // Create the new task
+                    async_spawn(async move {
+                        if let Err(e) = inner
+                            .poll_web_server(
+                                receiver,
+                                MessagePurpose::UpgradeProposal,
+                                view_number,
+                                Duration::from_millis(500),
```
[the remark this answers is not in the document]
ss-es: Yeah, I think this was just copied from the other branches. I didn't spend much time cleaning this up because I think we're planning on completely deleting the `web_server` anyway: #2825

crates/hotshot/src/lib.rs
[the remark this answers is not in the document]
jparr721: Makes sense!

crates/hotshot/src/traits/networking/web_server_network.rs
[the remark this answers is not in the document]
jparr721: Sounds good!

crates/task-impls/src/upgrade.rs
jparr721: Sounds good!

crates/task-impls/src/consensus.rs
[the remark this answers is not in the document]
jparr721: Sounds good!

crates/task-impls/src/vote_collection.rs
```diff
@@ -272,6 +272,7 @@ impl<TYPES: NodeType> AggregatableVote<TYPES, UpgradeVote<TYPES>, UpgradeCertifi
         certificate: UpgradeCertificate<TYPES>,
         _key: &TYPES::SignatureKey,
     ) -> HotShotEvent<TYPES> {
+        tracing::error!("Upgrade certificate formed!");
```
jparr721, line 275: Sorry if I already commented about this, is this a debugging message left behind? Or do we want to notify this as an error?

crates/testing/tests/consensus_task.rs
```diff
@@ -78,8 +78,8 @@ async fn test_consensus_task() {
         ],
         outputs: vec![
             exact(ViewChange(ViewNumber::new(2))),
-            quorum_proposal_send(),
             exact(QuorumProposalValidated(proposals[1].data.clone())),
+            quorum_proposal_send(),
```
jparr721, line 82: Nice.

jparr721 (approved):
Tentative approval with the caveat that, while everything looks fine, I'm less familiar with the network task so I'll defer to someone who knows it better to give the green light.

crates/task-impls/src/vote_collection.rs
ss-es: Oops, yes -- another leftover. Thanks for catching it -- removed!

crates/task-impls/src/consensus.rs
```diff
@@ -1542,6 +1557,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
             };
             debug!("Sending proposal for view {:?}", view);
 
+            async_sleep(Duration::from_millis(self.round_start_delay)).await;
```
bfish713, line 1560: We should probably just remove this from the config and not sleep here

bfish713 (approved):

Issue #2354, opened 2024-01-08
[Auto benchmarks] Store previous benchmark runs


rob-maron:
### What is this task and why do we need to work on it?

We want a repository for previous benchmark runs. This makes it easy to compare multiple runs between each other and notice any regressions. Likely relies on https://github.com/EspressoSystems/HotShot/issues/2346 .

### What work will need to be done to complete this task?

- [x] Figure out what format to store the benchmark run outputs in 
- [ ] Make the orchestrator save these somewhere that's easy to look at

### What are the acceptance criteria to close this issue?

Orchestrator saves the benchmark outputs to a centralized repository. Either in a human readable format or with a way to easily read it into something like a chart.

dailinsubjam:
Results are now saved in csv file with commit sha, and re-running benchmarks will not cover the previous results.

PR #2793, opened 2024-03-15
[Auto Bench] Scripts of Running Benchmarks in AWS

### LINKED ISSUE #2723: [Auto Benchmarks] - Scripts to Run Benchmarks in AWS

Goal of this task is to automate the flow we have to do to get benchmarks in the cloud.  The ideal outcome would be to run a script which configures the parameters we are testing (block size, txn size, propose min/max time, vid parameters, num nodes, committee size, etc).  

The benchmarking guide lists most of the steps we'd want to script: https://www.notion.so/espressosys/Running-Benchmarks-in-AWS-as-of-Feb-2024-fa680676053044aa8a04d5bccea0b1b4

The scripts should be run on the EC2 machines to launch/scale to ECS and spin up the CDN.

*The step of logging into github might still be manual

_No response_

_No response_

We can run the benchmark and see results without having to do all of the steps manually.

_No response_

dailinsubjam:
Closes #2723 #2354 .

### This PR: 
- automates the flow we have to do to get benchmarks in the cloud.
- includes benchmark results for init run. See `scripts/benchmarks_results/results_init_run.csv`.
- fixes `run-config.toml` and `post_pub_key` for examples after Builder and vbs get merged in.
- cleans up some files by adding to gitignore.

### This PR does not (but going to do next): 
- consider any vid-related factor.
- add runs for push-based CDN, libp2p and combined network yet (as now it only include runs for webserver).
- use nginx while this is needed for `total_nodes >= 500`.
- tune the parameters according to the results.
- finish ci integration.
- add runs for `tokio` as it only support `async_std` now.

### Key places to review: 
`scripts/aws_ecs_benchmarks.sh`

### How to test this PR:
- Log into Github on the EC2 machine.
- Get things to run `HotShot` installed (like Rust) and clone `HotShot`.
- Run `./scripts/aws_ecs_benchmarks.sh`.
- Three ways to gather the results
    - (recommended) check out`scripts/benchmarks_results/results.csv` on EC2, where you should have organized overall results, 
    - check out Datadog under `host:/hotshot/` where you only have metrics for each individual validator and it's hard to track,
    - wait for the output in terminal where the results are not that organized, also hard to track.


scripts/ci_ecs_benchmarks.sh
```diff
@@ -0,0 +1,28 @@
...
+
+# start orchestrator
+just async_std example orchestrator-webserver -- --config_file ./crates/orchestrator/run-config.toml --orchestrator_url http://0.0.0.0:4444 --webserver_url ${webserver_url} --da_webserver_url ${da_webserver_url} &
+sleep 1m
+
+# start validators
+ecs deploy --region us-east-2 hotshot hotshot_centralized -i centralized ghcr.io/espressosystems/hotshot/validator-webserver:main-async-std
+ecs deploy --region us-east-2 hotshot hotshot_centralized -c centralized ${orchestrator_url}
+ecs scale --region us-east-2 hotshot hotshot_centralized 10 --timeout -1
+sleep 2m #rethink about this
```
shenkeyao, line 22: Looks like in `scripts/aws_ecs_benchmarks.sh`, this is set to be proportional to the number of rounds and nodes. Curious why we don't do the same here.

shenkeyao (commented):
Changes look good, though I didn't try the steps in "How to test this PR". Left some minor comments.

scripts/ci_ecs_benchmarks.sh
dailinsubjam: Yeah we should do the same here, but just ignore this file, it's mostly for reference and not in use now :( I'll have to re-think each step of this file when we do real CI integration.

crates/examples/infra/mod.rs
```diff
@@ -647,19 +677,21 @@ pub trait RunDA<
         let consensus = consensus_lock.read().await;
         let total_num_views = usize::try_from(consensus.locked_view.get_u64()).unwrap();
         // When posting to the orchestrator, note that the total number of views also include un-finalized views.
-        error!("Failed views: {failed_num_views}, Total views: {total_num_views}, num_successful_commits: {num_successful_commits}");
+        println!("[{node_index}]: Total views: {total_num_views}, Failed views: {failed_num_views}, num_successful_commits: {num_successful_commits}");
```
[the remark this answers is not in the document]
dailinsubjam: Yeah it's intentional, so that users can get the results directly from terminal.

scripts/ci_ecs_benchmarks.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+# assign local ip
+ip=`curl http://169.254.169.254/latest/meta-data/local-ipv4`
+webserver_url=http://"$ip":9000
+da_webserver_url=http://"$ip":9001
+orchestrator_url=http://"$ip":4444
```
rob-maron, line 7: I like the use of the AWS metadata server

scripts/ci_ecs_benchmarks.sh
```diff
@@ -0,0 +1,28 @@
...
+
+# start webserver
+just async_std example webserver -- http://0.0.0.0:9000 &
+just async_std example webserver -- http://0.0.0.0:9001 &
+sleep 10m
```
rob-maron, line 12: Are we sleeping 10m after we start the webservers? Do the webservers take long to start up? I don't think the validators start up until the orchestrator is ready after the sleep

scripts/aws_ecs_benchmarks.sh
```diff
@@ -0,0 +1,65 @@
...
+                                                                    --da_committee_size ${da_committee_size} \
+                                                                    --transactions_per_round ${transactions_per_round} \
+                                                                    --transaction_size ${transaction_size} \
+                                                                    --rounds ${rounds} \
+                                                                    --commit_sha test &
+                    sleep 30
+
+                    # start validators
+                    ecs scale --region us-east-2 hotshot hotshot_centralized ${total_nodes} --timeout -1
+                    sleep $(($rounds + $total_nodes))
```
rob-maron, line 54: Maybe instead of sleeping: can we make the orchestrator quit after the results have been received?

This way we run the orchestrator in the foreground and just clean up everything after that process exits. What do you think?

scripts/aws_ecs_benchmarks.sh
```diff
@@ -0,0 +1,65 @@
...
+
+                    # kill them
+                    ecs scale --region us-east-2 hotshot hotshot_centralized 0 --timeout -1
+                    sleep 1m
+                    for pid in $(ps -ef | grep "orchestrator" | awk '{print $2}'); do kill -9 $pid; done
```
rob-maron, line 59: I think it's possible to get the PID when you send the progress to the `bg`: 

https://stackoverflow.com/a/1911387

Maybe we can do that (or for future improvement) by already having the PIDs ready!

scripts/ci_ecs_benchmarks.sh
dailinsubjam: Oh I set this previously for CI, since it re-compile all the things for these two commands. But I should re-think about this. Let me remove this ci script first.

scripts/aws_ecs_benchmarks.sh
dailinsubjam: I tried this before, but it's weird that the corresponding PID is not correct :( And the real PID is not captured

scripts/aws_ecs_benchmarks.sh
dailinsubjam: Sleeping is mainly for waiting some time before scaling ecs back to 0. Looks like we need something in the script to mark the point that the results have been received 🤔, do you have any idea on this?

scripts/aws_ecs_benchmarks.sh
rob-maron: Ah okay, all set here then !

scripts/aws_ecs_benchmarks.sh
rob-maron: I was unsure on what the orchestrator does after a run is finished. If it quits, we could maybe just run that command and wait for it to finish, and then scale down after.

But I just checked and it doesn't do that, so this is probably a future improvement and sleep should be okay for now !

scripts/aws_ecs_benchmarks.sh
```diff
@@ -0,0 +1,65 @@
...
+
+for total_nodes in 10 50 100
+do
+    for da_committee_size in 10 50 100
+    do
+        if [ $da_committee_size -le $total_nodes ]
+        then
+            for transactions_per_round in 1 10 50 100
+            do
+                for transaction_size in 512 4096 #1000000 20000000
```
bfish713, line 31: I think it would be nice to be able to run with all these as parameters instead of looping through.  The way it is now looks good for doing a comprehensive run, but while we iterate, it would be good to be able to run with set committee, total nodes, and txn size

bfish713 (approved):

bfish713:
Overall it looks good, Just one comment about allowing the script (or adding a new script) to do a single benchmark run at a time.  Maybe we can add it as a quick follow up.

scripts/aws_ecs_benchmarks.sh
dailinsubjam: Yeah we can do it with multiple variable assignment in script, I'll add this in my next PR.

PR #2794, opened 2024-03-16
Enable import grouping in rustfmt

ss-es:
No linked issue.

### This PR: 
A lot of our imports are very messy, and by default `rustfmt`/`cargo fmt` does nothing to them. This PR adds nightly `rustfmt` to the stable toolchain provided by Nix (required to configure `rustfmt` for imports), and adds `rustfmt.toml` with the relevant settings.


ss-es:
We agreed internally to do this, but I think it might be better to hold off to avoid creating unnecessary merge conflicts for other branches (between the recent `hotshot-types` and `hotshot-builder-api` move, and the fact that this PR doesn't get any harder to merge later).

Leaving as a draft; will revisit in a week.

shenkeyao (approved):

PR #2799, opened 2024-03-18
[CX-CLEANUP] - DA Integrated Storage

### LINKED ISSUE #2621: [CX-CLEANUP] - DA Integrated Storage

jparr721:
Closes #2621 

We want to integrate a persistent storage interface within HotShot to enable the application to manage data storage effectively. This feature aims to address potential data storage issues (e.g., full disk space) that can prevent nodes from storing DA blobs or VID shares.

### This PR: 
1. **Creates `Storage` Trait**
2. **Initializes `Storage` in `SystemTask`'s init**
3. **Add Storage Operations on DA Proposals**
    - If this fails, do not vote on the proposal and exit the event handler.
4. **Add Storage Operations on DA Proposals VID Shares**
   - If this fails, do not vote on the VID disperse, otherwise, use the old logic to vote.
5. **Refactor In-memory Payload/VID Stores**: (Tracking #2808)

This code also implements a number of changes to introduce a `TestBlockStorage` type for use in the various testing components.

### Key places to review: 

All of the tests and test interface changes. I had to update the `view_generator` to support the various DA related things that we wanted to add.

### DIFF
--- a/crates/example-types/src/storage_types.rs
+++ b/crates/example-types/src/storage_types.rs
@@ -1,0 +1,56 @@
+use anyhow::Result;
+use async_lock::RwLock;
+use async_trait::async_trait;
+use hotshot_types::{
+    data::{DAProposal, VidDisperse},
+    message::Proposal,
+    traits::{node_implementation::NodeType, storage::Storage},
+};
+use std::collections::HashMap;
+use std::sync::Arc;
+
+#[derive(Clone, Debug)]
+pub struct TestStorageState<TYPES: NodeType> {
+    vids: HashMap<TYPES::Time, Proposal<TYPES, VidDisperse<TYPES>>>,
+    das: HashMap<TYPES::Time, Proposal<TYPES, DAProposal<TYPES>>>,
+}
+
+impl<TYPES: NodeType> Default for TestStorageState<TYPES> {
+    fn default() -> Self {
+        Self {
+            vids: HashMap::new(),
+            das: HashMap::new(),
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct TestStorage<TYPES: NodeType> {
+    inner: Arc<RwLock<TestStorageState<TYPES>>>,
+}
+
+impl<TYPES: NodeType> Default for TestStorage<TYPES> {
+    fn default() -> Self {
+        Self {
+            inner: Arc::new(RwLock::new(TestStorageState::default())),
+        }
+    }
+}
+
+#[async_trait]
+impl<TYPES: NodeType> Storage<TYPES> for TestStorage<TYPES> {
+    async fn append_vid(&self, proposal: &Proposal<TYPES, DAProposal<TYPES>>) -> Result<()> {
+        let mut inner = self.inner.write().await;
+        inner
+            .das
+            .insert(proposal.data.view_number, proposal.clone());
+        Ok(())
+    }
+    async fn append_da(&self, proposal: &Proposal<TYPES, VidDisperse<TYPES>>) -> Result<()> {
+        let mut inner = self.inner.write().await;
+        inner
+            .vids
+            .insert(proposal.data.view_number, proposal.clone());
+        Ok(())
+    }
+}
--- a/crates/example-types/src/storage_types.rs
+++ b/crates/example-types/src/storage_types.rs
@@ -39,17 +39,17 @@ impl<TYPES: NodeType> Default for TestStorage<TYPES> {
 
 #[async_trait]
 impl<TYPES: NodeType> Storage<TYPES> for TestStorage<TYPES> {
-    async fn append_vid(&self, proposal: &Proposal<TYPES, DAProposal<TYPES>>) -> Result<()> {
-        let mut inner = self.inner.write().await;
-        inner
-            .das
-            .insert(proposal.data.view_number, proposal.clone());
-        Ok(())
-    }
-    async fn append_da(&self, proposal: &Proposal<TYPES, VidDisperse<TYPES>>) -> Result<()> {
+    async fn append_vid(&self, proposal: &Proposal<TYPES, VidDisperse<TYPES>>) -> Result<()> {
         let mut inner = self.inner.write().await;
         inner
             .vids
+            .insert(proposal.data.view_number, proposal.clone());
+        Ok(())
+    }
+    async fn append_da(&self, proposal: &Proposal<TYPES, DAProposal<TYPES>>) -> Result<()> {
+        let mut inner = self.inner.write().await;
+        inner
+            .das
             .insert(proposal.data.view_number, proposal.clone());
         Ok(())
     }
--- a/crates/example-types/src/storage_types.rs
+++ b/crates/example-types/src/storage_types.rs
@@ -1,4 +1,4 @@
-use anyhow::Result;
+use anyhow::{bail, Result};
 use async_lock::RwLock;
 use async_trait::async_trait;
 use hotshot_types::{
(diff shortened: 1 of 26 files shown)


crates/example-types/src/storage_types.rs
```diff
@@ -0,0 +1,56 @@
...
+#[async_trait]
+impl<TYPES: NodeType> Storage<TYPES> for TestStorage<TYPES> {
+    async fn append_vid(&self, proposal: &Proposal<TYPES, DAProposal<TYPES>>) -> Result<()> {
+        let mut inner = self.inner.write().await;
+        inner
+            .das
+            .insert(proposal.data.view_number, proposal.clone());
+        Ok(())
+    }
+    async fn append_da(&self, proposal: &Proposal<TYPES, VidDisperse<TYPES>>) -> Result<()> {
```
bfish713, line 49: I just noticed this in the types change, the types for this are switched.  `append_da` takes a VID proposal, `append_vid` takes a DA Proposal

jparr721: PRs are up 🤦🏽

crates/task-impls/src/consensus.rs
```diff
@@ -1063,11 +1067,20 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
+                        e
+                    );
+                    return;
+                }
+
                 self.consensus
                     .write()
                     .await
                     .vid_shares
                     .insert(view, disperse.clone());
```
jparr721, line 1082: The gist of this work is to add a trait for persistent storage with a result to allow us to detect things like disk capacity being exceeded, among other things. The initial phase of work is to only store DA and VID proposals, as we see here. But I noticed that the `Consensus` type actually *also* stores these things, but in memory. The question I'm getting at is: should we store both? Right now the type only supports `append`, so my thought it yes, but I wanted to highlight this to make sure I'm not missing something here. cc @jbearer @bfish713

bfish713: I think we want both, consensus needs wants a quick in memory store for it's state which we can garbage collect and modify without worrying about what effect it might have on the application.

bfish713: The inverse is also true, the application might want to mess with the data in ways that would break what consensus wants for voting and it should be free to do so.  That's my 2 cents at least

bfish713 (approved):
Changes look good to me, but lets give @jbearer some time to answer your question

crates/task-impls/src/consensus.rs
jbearer: Totally agree.

The one thing I think we _can_ get rid of is the `PayloadStore` (is it still called that?) where we store the full payloads as a DA replica. I believe this is only used in consensus to populate the `Option<Payload>` field of the leaf in decide events. We could potentially remove this field and the logic to populate it on Decide, since the application is no capable of retrieving the payload itself

jparr721: Tracking: https://github.com/EspressoSystems/HotShot/issues/2808

crates/example-types/src/storage_types.rs
```diff
@@ -0,0 +1,67 @@
...
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct TestStorage<TYPES: NodeType> {
+    inner: Arc<RwLock<TestStorageState<TYPES>>>,
+
+    /// `should_return_err` is a testing utility to validate negative cases.
+    pub should_return_err: bool,
```
jparr721, line 32: Open to thoughts here. See the tests below for examples. I just need a cheesy "fail immediately" system that does not change the interface. The idea here is that I just want to make sure that any logical changes can catch a regression in the failure case for an `append`.

crates/testing/tests/da_task.rs
```diff
@@ -1,90 +1,140 @@
...
+        leaders.push(view.leader_public_key);
+        votes.push(view.create_da_vote(DAData { payload_commit }, &handle));
+        dacs.push(view.da_certificate.clone());
+        vids.push(view.vid_proposal.clone());
+    }
+
+    // Run view 1 (the genesis stage).
+    let view_1 = TestScriptStage {
+        inputs: vec![
+            ViewChange(ViewNumber::new(1)),
```
jparr721, line 59: This is copied direct from the test as it was just to maintain parity.

crates/example-types/src/storage_types.rs
bfish713: Yeah this is fine, it's just a dummy type for our tests anyway

bfish713 (approved):

PR #2807, opened 2024-03-19
[Catchup] Store High QC When it is Updated

bfish713:
### This PR: 

- Adds the High QC to the initializer so a restarting node has a correct high qc when it restarts.  This will prevent it from voting in previous rounds and having a more state.
- Adds high QC to storage.  This will allow the sequencer correctly set the high qc in the initializer on restart.
- Adds to the cathup test so that we actually keep an accurate high_qc and last decided leaf for restart.
- Adds undecided leafs to storage
- Adds undecided states to storage.

The undecided leaves and state will allow a node to vote and propose as soon as it comes back up assuming it didn't miss more views while it was down.

I'm wondering what is worth storing or not.  It may be quite rare that we restart fast enough that we don't miss any views and our undecided state has any value to reload.  We might just waste time storing information on every view that we don't ever use.  I went ahead and added everything we would need to restart and be able to vote/propose if we don't miss more views but I can pretty easily roll that back.  Or maybe there is some other mechanism we should use?


crates/hotshot/src/lib.rs
```diff
@@ -637,6 +643,15 @@ pub struct HotShotInitializer<TYPES: NodeType> {
...
     start_view: TYPES::Time,
+    /// Highest QC that was seen, for genesis it's the genesis QC.  It should be for a view greater
+    /// than `inner`s view number for the non genesis case because we must have seen higher QCs
+    /// to decide on the leaf.
+    high_qc: QuorumCertificate<TYPES>,
+    /// Undecided leafs that were seen, but not yet decided on.  These allow a restarting node
+    /// to vote and propose right away if they didn't miss anything while down.
+    undecided_leafs: Vec<Leaf<TYPES>>,
+    /// Not yet decided state
+    undecided_state: BTreeMap<TYPES::Time, View<TYPES>>,
```
jparr721, line 654: Just for my own clarity: The BTree map here is to ensure that the highest view is always at the top?

crates/hotshot/src/lib.rs
```diff
@@ -651,6 +666,9 @@ impl<TYPES: NodeType> HotShotInitializer<TYPES> {
             validated_state: Some(Arc::new(validated_state)),
             state_delta: Some(Arc::new(state_delta)),
             start_view: TYPES::Time::new(0),
+            high_qc: QuorumCertificate::genesis(),
+            undecided_leafs: Vec::new(),
```
jparr721, line 670: Should it technically be "leaves"?

crates/task-impls/src/consensus.rs
```diff
@@ -515,6 +515,19 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
+                if justify_qc.get_view_number() > consensus.high_qc.view_number {
+                    if let Err(e) = self
+                        .storage
+                        .write()
+                        .await
+                        .update_high_qc(justify_qc.clone())
+                        .await
+                    {
+                        warn!("Failed to store High QC not voting. Error: {:?}", e);
+                        return;
```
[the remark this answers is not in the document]
jparr721: Seems like better-safe-than-sorry would be reasonable, especially if it's an error anyway.

crates/task-impls/src/consensus.rs
```diff
@@ -553,6 +566,19 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
                     );
                     consensus.saved_leaves.insert(leaf.commit(), leaf.clone());
 
+                    if let Err(e) = self
+                        .storage
+                        .write()
+                        .await
+                        .update_undecided_state(
```
jparr721, line 573: Two questions here:
1. What happens when these start to error constantly? Like, if a write fails or an update, does that signify that the service has degraded?
2. Do we garbage collect old data? For instance if we've moved a few hundred steps, does it make sense to hold onto all the prior state?

jparr721 (approved):
Great work. Just a few questions to inform my understanding.

crates/hotshot/src/lib.rs
bfish713: For the most part it just makes the garbage collection easier.  In this case it's the same because it's what hotshot already has available.

We store it in memory as a `BTreeMap` because we only want to keep the undecided views.  When we get decide we can easily remove all views <= the decided view from the map thanks to the ordering.

crates/hotshot/src/lib.rs
bfish713: It should be

crates/task-impls/src/consensus.rs
bfish713: 1. It would have no effect on anything besides being able to recover this state.  It would probably hint at a larger problem with storage but that would also show up when we try to store DA and VID info.
2. Consensus does garbage collect this.  What the storage implementation chooses to do is up for them to decide.  The recommendation would be that every call to `update_X` would rewrite whatever "X" was previously.  In other words, only keep the most recent information

PR #2809, opened 2024-03-19
[PushCDN] Add dockerfiles

rob-maron:
Closes https://github.com/EspressoSystems/Push-CDN/issues/21

## This PR: 

- Adds Dockerfiles for the push CDN [broker/marshal]
- Adds examples for pCDN [broker/marshal]

## This PR does not: 

Change any application or push CDN logic

## Key places to review: 

Trying to build and run the Dockerfiles


crates/hotshot/Cargo.toml
```diff
@@ -54,7 +54,7 @@ snafu = { workspace = true }
 surf-disco = { workspace = true }
 time = { workspace = true }
 tracing = { workspace = true }
-anyhow = { workspace = true }
+anyhow.workspace = true
```
jparr721, line 57: Curious what this change here introduces.

jparr721 (commented):
Just one question.

crates/hotshot/Cargo.toml
rob-maron: Hahaha whoops, this was totally some weird merge conflict behavior. It looks like both the merge and I added `anyhow`

jparr721: Ah yeah that was me. Sorry about that!

ss-es:
I don't actually know why we generate the other dockers in CI/where they go, but should these be added to the `build-dockers` workflow in `.github/workflows/build-and-test.yml`?

rob-maron:
> I don't actually know why we generate the other dockers in CI/where they go, but should these be added to the `build-dockers` workflow in `.github/workflows/build-and-test.yml`?

Good point. I think I'm going to withhold on that until the next PR. One of the reasons it's not in here is because the binaries are built inside of the Dockerfiles (for the broker and marshal) instead of just copying the pre-built binaries over. So it would make CI take _much_ longer to build .

crates/examples/push-cdn/broker.rs
```diff
@@ -0,0 +1,93 @@
...
+    #[arg(short, long)]
+    discovery_endpoint: String,
+
+    /// Whether or not metric collection and serving is enabled
+    #[arg(long, default_value_t = true)]
+    metrics_enabled: bool,
+
+    /// The port to bind to for externalizing metrics
+    #[arg(long, default_value = "127.0.0.1")]
+    metrics_ip: String,
```
jparr721, line 28: I don't know if it matters, but these ip addresses aren't being validated, at least not where I can see. Not sure if that's something we care about.

crates/examples/push-cdn/marshal.rs
```diff
@@ -0,0 +1,53 @@
...
+//!
+use anyhow::{Context, Result};
+use cdn_broker::reexports::connection::protocols::Quic;
+use cdn_marshal::{ConfigBuilder, Marshal};
+use clap::Parser;
+use hotshot::traits::implementations::WrappedSignatureKey;
+use hotshot_example_types::node_types::TestTypes;
+use hotshot_types::traits::node_implementation::NodeType;
+
+//TODO: for both client and marshal, clean up and comment `main.rs`
```
jparr721, line 12: Does it make sense for these to be cleanup tasks?

crates/examples/push-cdn/marshal.rs
```diff
@@ -0,0 +1,53 @@
...
+// TODO: forall, add logging where we need it
+
+#[derive(Parser, Debug)]
+#[command(author, version, about, long_about = None)]
+/// The main component of the push CDN.
+struct Args {
+    /// The discovery client endpoint (including scheme) to connect to
+    #[arg(short, long)]
+    discovery_endpoint: String,
```
jparr721, line 21: Is this intended to be something like an `http://` type value? The comment seems to hint at that, but perhaps an example in the doc comment or help text could clarify.

crates/examples/push-cdn/broker.rs
rob-maron: The broker will format these IPs on startup, and the error would be printed when we call `.start?`. Also just noticed the comment there is wrong

crates/examples/push-cdn/marshal.rs
rob-maron: Totally forgot to remove this, thanks. I did the cleanup already

crates/examples/push-cdn/marshal.rs
rob-maron: It is an `IP:port` pairing, but this makes a lot of sense. I will provide an example in the comment

jparr721 (approved):

PR #2810, opened 2024-03-19
[CX_CLEANUP] - Change Consensus Task State to Use An Enum Type for Timeout and View Sync Certs

### LINKED ISSUE #2767: [CX_CLEANUP] - Change Consensus Task State to Use `Either` for Timeout and View Sync Certs

The consensus task now supports view sync and timeout certificates, but it is _not_ correct for both of them to exist at the same time. We can use an `Either` type to give predictable and consistent handling logic in all the places that use it within `consensus.rs`

- Change the type in `ConsensusTask`
- Update the logic at all spots that check for the TC or View Sync Cert.
- Decide if we want to change the type within [`QuorumProposal`](https://github.com/EspressoSystems/hotshot-types/blob/main/src/data.rs#L177) as well.
- Run and validate tests (they should pass without needing to make any changes)

_No response_

Tests still pass.

_No response_

jparr721:
Closes #2767 

The consensus task now supports view sync and timeout certificates, but it is _not_ correct for both of them to exist at the same time. We can use an `Either` type to give predictable and consistent handling logic in all the places that use it within `consensus.rs`

### This PR: 
- Changes the type in `ConsensusTask`
- Updates the logic at all spots that check for the TC or View Sync Cert.
- Run and validate tests (they should pass without needing to make any changes)
- Changes the type within `QuorumProposal`


crates/task-impls/src/consensus.rs
```diff
@@ -403,20 +401,27 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
+                                error!("Timeout cert found during proposal recv");
+                            }
+                            either::Right(view_sync) => {
+                                // View sync view_syncs _must_ be for the current view.
+                                if view_sync.view_number != view {
+                                    debug!(
                                 "Cert view number {:?} does not match proposal view number {:?}",
-                                cert.view_number, view
+                                view_sync.view_number, view
                             );
```
ss-es, line 415: Not at all a strong opinion, but just a thought I had while reading this -- would it be better to use a custom enum? Not sure what it should be called but e.g. `pub enum ViewChangeEvidence = { Timeout(..), ViewSync(..) }`.

My worry is: if I mix up `Left` and `Right` here, I don't think I'd get a type-error because both certificates have `view_number`. And I think it would be easy to miss in code review, because the names `Left` and `Right` aren't descriptive.

jparr721:
> I see the CI failing for the pCDN, can you try pulling in main again? I made a change late last night

I am actually on latest main. Perhaps it's just flaking?

rob-maron:
> > I see the CI failing for the pCDN, can you try pulling in main again? I made a change late last night
> 
> I am actually on latest main. Perhaps it's just flaking?

Cool. I'll look into it, thanks !

crates/task-impls/src/consensus.rs
jparr721: Following up: After discussion with the team we have all decided that we prefer Salman's idea. Rolling this into my next commit.

bfish713:
This looks good.  My browser won't render the tests that are failing, which ones are they?

crates/task-impls/src/consensus.rs
```diff
@@ -392,31 +384,65 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
+
+                                // View sync certs must also be valid.
+                                if !view_sync_cert.is_valid_cert(self.quorum_membership.as_ref()) {
+                                    debug!("Invalid ViewSyncFinalize cert provided");
+                                    return;
+                                }
+                            }
                         }
+                    }
+                    if let Some(received_proposal_cert) = proposal.data.proposal_certificate.clone()
```
jparr721, line 421: @bfish713 This was the issue. A received timeout cert was not being handled and the code assumed it was somehow already stored. So I added this additional check back. Please advise if assigning at the end of this block is correct. I believe that it should be done since otherwise we lose the timeout cert in the proposal function.

jparr721: Hmm.. well that fixed 4 of the failures. Something else is going on.

jparr721: Got it. The logic was checking a stale timeout cert.

bfish713: I'm confused by the logic above where we are matching against our cert?  That seems wrong.  We just received a proposal from someone else, we should verify that it has the appropriate evidence which should only involve checking whatever is on the proposal we just got.  `self.propsal` cert should only be for creating our own proposals.  It should only be set when we formed a timeout certificate, or a got a view sync finalized certificate.

bfish713: in other words I think in the `QuorumProposalRevc` event handler we should only do
```rust
     if let Some(received_proposal_cert) = proposal.data.proposal_certificate.clone() {
        match received_proposal_cert {
            ViewChangeEvidence::Timeout(timeout_cert) => { // is it a valid TC for proposal.view - 1?}
            ViewChangeEvidence::ViewSync(view_sync_cert) => {  // is it a valid ViewSyncCert for proposal.view - 1 }
```

bfish713 (approved):

shenkeyao (approved):
The changes look good!

PR #2812, opened 2024-03-19
Bring `hotshot-types` and `hs-builder-api` into `Hotshot`

bfish713:
Merge the 3 repos

Reviewers please check if I forgot to add something(probably this would be outside of the `/src` directories) or messed up something in a Cargo.toml file


move47 (approved):
Thanks @bfish713 for this.

crates/builder-api/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "hs-builder-api"
```
ss-es, line 2: For consistency I feel like we should probably call this `hotshot-builder-api` (I think everything else related to HotShot is called `hotshot-`). But I think we should do that separately, in another PR.

@move47 Thoughts?

move47: Yeah, I had this also. We have to change in several other crates too, so thought of doing altogether.

ss-es (approved):

move47 (dismissed):
LGTM! Thanks.

crates/builder-api/Cargo.toml
bfish713: Oh I added it to this pr, hopefully that's ok

jparr721 (approved):

PR #2815, opened 2024-03-20
[Push CDN] Add whitelist update binary

rob-maron:
## This PR

(I should have split it out into slightly smaller PRs, sorry)

- Updates the Push CDN. All of the changes around `RunDef` remove the trait aliases we had to use. Instead, we implement a `RunDef`-inition and pass that in instead. You can ignore these !

- Adds the required, missing fields to the `run-config`

#### Whitelist binary
Adds a binary that will whitelist public keys and quit after it's done. It calls to the orchestrator and updates the discovery (sqlite OR redis) endpoint with the new whitelist. If it's not run, everybody is allowed to connect until the table is initialized.

The related HotShot changes:
- `get_config_without_peer` is updated to allow for pulling the list of keys without posting our own. This is so passive observers (like our whitelist binary) can pull the config without needing to participate.

## Key places to review

- `whitelist-adapter.rs`
- `crates/orchestrator/src/lib.rs:269`: Post https://github.com/EspressoSystems/HotShot/pull/2666, this has to change from 8 to 12 or the examples will be broken. I will look into it.
- `orchestrator/client.rs`

## This PR does not
Change push CDN or consensus functionality.


crates/orchestrator/src/client.rs
```diff
@@ -185,30 +185,35 @@ impl OrchestratorClient {
...
+    /// Get the config from the orchestrator.
+    /// If the identity is provided, register the identity with the orchestrator.
+    /// If not, just retrieving the config (for passive observers)
+    ///
     /// # Panics
     /// if unable to convert the node index from usize into u64
     /// (only applicable on 32 bit systems)
     #[allow(clippy::type_complexity)]
-    pub async fn get_config_without_peer<K: SignatureKey, E: ElectionConfig>(
+    pub async fn get_config<K: SignatureKey, E: ElectionConfig>(
```
dailinsubjam, line 196: All the others LGTM. The only thing is I'm more inclined to keep these comments (maybe together with the name `get_config_without_peer`) to emphasize this step haven't collected peers' info yet, so that it will not cause confusion with another two functions `get_config_after_peer_collected` and `get_config` in `trait RunDA`.

crates/orchestrator/src/lib.rs
```diff
@@ -267,7 +267,7 @@ where
         self.pub_posted.insert(node_index);
 
         // The guess is the first extra 8 bytes are from orchestrator serialization
-        pubkey.drain(..8);
+        pubkey.drain(..12);
```
dailinsubjam, line 270: I also changed this in my PR haha, would be better to also change the comment from 8 to 12.

crates/orchestrator/src/client.rs
rob-maron: Ahh okay, that makes sense. I will change it. Question: are the public keys of the peers available at this step? It definitely works for the whitelist during testing, but just curious.

crates/orchestrator/src/lib.rs
rob-maron: Haha yeah, I will look into why we need this padding soon. I'll change this comment in the meantime.

crates/orchestrator/src/client.rs
dailinsubjam: The node's own public keys are available, but its peers' public keys are not collected by it at this step yet.

rob-maron: Do you think we can unify the two functions? AKA change it so we don't return the initial config until everyone has posted their public keys? Without the need for the second function?

It looks like we already have similar logic: https://github.com/EspressoSystems/HotShot/blob/22631060ef5e4fd9ce97e3ee843cb0fa8856e09e/crates/orchestrator/src/lib.rs#L230-L235


WDYT?

dailinsubjam: Sorry I just realized `get_config_without_peer` is updated to allow for pulling the list of keys (although not sure how, looks like the function just add one more rounds of check `if let Some(identity) = identity`, would appreciate any more info on this). The function `get_complete_config` will only return the config until everyone has posted their public keys, where its first step uses `get_config_without_peer` to load configs from file, and the second step uses `post_and_wait_all_public_keys` to gather peers' public keys.

If it's true `get_config_without_peer` already pulled the list of keys and I could understand how, then the second step is not needed anymore.

rob-maron: Okay this makes sense, so instead I should get the public keys with `get_config_after_peer_collected`. It was probably just a timing thing that allowed it to work for `before_collected`

rob-maron: Updated !

crates/orchestrator/src/client.rs
```diff
@@ -244,6 +245,29 @@ impl OrchestratorClient {
         self.wait_for_fn_from_orchestrator(cur_node_index).await
     }
 
+    /// Requests the configuration from the orchestrator with the stipulation that
+    /// a successful call requires all nodes to be registered.
+    ///
+    /// Does not fail, retries internally until success.
```
dailinsubjam, line 251: Thanks for the comments.

dailinsubjam (approved):
Thanks for the update!

PR #2816, opened 2024-03-20
[Cleanup] Remove some unused dependencies

rob-maron:
Quick PR to remove some unused dependencies. Tested and works locally.


ss-es (approved):
CI passes, so looks good to me!

PR #2817, opened 2024-03-21
[BUILDER] Simple builder

### LINKED ISSUE #2559: [BUILDER] - Simple builder

For builder integration we need a simple builder internal to HotShot that fetches transactions from CDN and uses those to build blocks.

_No response_

_No response_

Simple builder implemented & tested against builder client
Includes tests to confirm builder correctly fetches and includes transactions

_No response_

QuentinI:
Closes #2559 

### This PR: 
Adds a simple builder for tests that aims to build blocks the same way HotShot does it now.

### This PR does not: 
Integrate said builder.

### DIFF
--- a/crates/testing/src/block_builder.rs
+++ b/crates/testing/src/block_builder.rs
@@ -1,4 +1,5 @@
 use std::{
+    collections::HashMap,
     num::NonZeroUsize,
     ops::{Deref, Range},
     sync::Arc,
@@ -8,6 +9,7 @@ use std::{
 use async_compatibility_layer::art::{async_sleep, async_spawn};
 use async_lock::RwLock;
 use async_trait::async_trait;
+use commit::{Commitment, Committable};
 use futures::future::BoxFuture;
 use hotshot::{traits::BlockPayload, types::SignatureKey};
 use hotshot_builder_api::{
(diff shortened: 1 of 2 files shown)


crates/testing/src/block_builder.rs
```diff
@@ -1,36 +1,151 @@
...
+    }
+
+    /// Spawn a task building blocks, configured with given options
+    #[allow(clippy::missing_panics_doc)] // ony panics on 16-bit platforms
+    pub fn run(&self, options: RandomBuilderOptions) {
+        let blocks = self.blocks.clone();
+        let (priv_key, pub_key) = (self.priv_key.clone(), self.pub_key);
+        async_spawn(async move {
+            let mut rng = SmallRng::from_entropy();
+            let time_per_block = Duration::from_millis(1000 / options.blocks_per_second);
```
jparr721, line 108: The random number here is a bit hard to parse. Could it perhaps be a named constant? For example, is 1000 the block epoch time? It's hard to know in this case.

jparr721 (commented):
A few cosmetics

crates/testing/src/block_builder.rs
QuentinI: Right, I already have no idea what this is about either :)

QuentinI:
Sorry for force-pushing, couldn't figure out how to update this branch otherwise, since it included a commit already merged to main now.

crates/testing/src/block_builder.rs
QuentinI: I re-wrote it as `Duration::from_secs(1) / options.blocks_per_second`, that should make it clear hopefully.

crates/testing/src/block_builder.rs
```diff
@@ -279,3 +225,244 @@ pub fn run_random_builder(url: Url) {
...
+            }
+            HotShotEvent::LeafDecided(leaf_chain) => {
+                let mut queue = this.transactions.write().await;
+                for leaf in leaf_chain.iter() {
+                    if let Some(ref payload) = leaf.block_payload {
+                        for txn in payload.transaction_commitments(&()) {
+                            queue.remove(&txn);
+                        }
+                    }
+                }
```
bfish713, line 356: This logic won't remove transactions if the `TransactionRecv` event happens after a leaf containing that transaction is decided.  It's pretty common with the webserver we download transactions in batches right before we become a leader.  So we'll get lots of transactions that were already decided.  I'm not sure how relevant that is for this builder if we only use as a test.  If we do bench-marking with it then we probably need to handle this case.

bfish713 (approved):
Looks good, left one comment we might need to fix.  Can be a separate PR

PR #2830, opened 2024-03-22
don't loop forever in DHT poll

### LINKED ISSUE #2820: [LIBP2P] - Fix `DHTBehaviour::poll` spinning while loops

Analysing flame graphs indicated that there is an issue in `DHTBehaviour::poll`. There are two while loops that potentially block the execution.

A similar issue was found in `DMBehaviour::poll` and was solved by a quick fix in [this PR](https://github.com/EspressoSystems/HotShot/pull/2769). Similar quick fix can be applied here.

_No response_

`DHTBehaviour::poll` should not spin in the two while loops.

_No response_

bfish713:
Closes #2820 

### This PR: 

Just loop over each request once in poll


crates/libp2p-networking/src/network/def.rs
```diff
@@ -144,25 +143,13 @@ impl NetworkDef {
...
     /// Add a direct response for a channel
     pub fn add_direct_response(&mut self, chan: ResponseChannel<Vec<u8>>, msg: Vec<u8>) {
-        self.direct_message.add_direct_response(chan, msg);
-    }
-}
-
-impl From<DMEvent> for NetworkEventInternal {
-    fn from(event: DMEvent) -> Self {
-        Self::DMEvent(event)
+        let _ = self.direct_message.send_response(chan, msg);
```
jparr721, line 152: What is the `let _` being used for here?

jparr721 (commented):
LGTM, but I'm less familiar with the previous behavior. One question

crates/libp2p-networking/src/network/def.rs
bfish713: It's suppressing an error (and clipply).  I can log the error, but this is from the other PR

bfish713: tbh I'm not sure why this shows up even though this PR is to merge into the other branch

rob-maron (dismissed):

PR #2831, opened 2024-03-22
[CX_CLEANUP] - Implement `QuorumProposalTask`

### LINKED ISSUE #2713: [CX_CLEANUP] - Implement `QuorumProposalTask`

After https://github.com/EspressoSystems/HotShot/issues/2708, we can start working on replacing `publish_proposal_if_able` with a new proposal task, similar to what we do to `vote_if_able`.

We may add more subtasks for incremental PRs and testing. 

_No response_

Design doc for the voting task: https://www.notion.so/espressosys/Refactor-Voting-Logic-9afa14dd027d453aa9b4a80a52de5e44.

New tests pass.

_No response_

jparr721:
Closes #2713

### This PR: 
Implements the `QuorumProposalTask` which is able to build a dependency chain and propose when the requirements are met. This task implements the "part 1" version of this issue as the scope is quite large. In particular, it does the following:
- Scaffolds and initializes the requisite data types
- Sets up relevant modules and exports them
- 

### This PR does not: 
- Implement exhaustive testing, that will come in a follow-on PR
- Integrate this functionality into any existing logic. This PR is exclusively the module in isolation and it does not get spawned or handle any requests.

### Key places to review: 
The handling of the dependencies in the quorum_proposal.rs file. This is a tricky bit of logic, and I want to make sure that, in particular, I set up the dependencies properly for combining an and and or dependency together. Further, while this code does not do any proposing (yet), I intend to broadcast from the `handle_dep_result` function and, as a result, would appreciate feedback at how I've constructed the types that I'll use to form the requisite payloads.


crates/task-impls/src/events.rs
```diff
@@ -44,6 +44,10 @@ pub enum HotShotEvent<TYPES: NodeType> {
     DACValidated(DACertificate<TYPES>),
     /// Send a quorum proposal to the network; emitted by the leader in the consensus task
     QuorumProposalSend(Proposal<TYPES, QuorumProposal<TYPES>>, TYPES::SignatureKey),
+    /// Dummy quorum proposal to test if the quorum proposal dependency works.
+    DummyQuorumProposalSend(Proposal<TYPES, QuorumProposal<TYPES>>, TYPES::SignatureKey),
```
jparr721, line 48: I think I might delete this before this merges and instead just send the real thing since I'll be adding testing next.

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -0,0 +1,501 @@
...
+fn validate_payload_and_metadata<TYPES: NodeType>(
+    _payload_commitment_and_metadata: CommitmentAndMetadata<<TYPES as NodeType>::BlockPayload>,
+    _event_sender: Sender<Arc<HotShotEvent<TYPES>>>,
+) -> bool {
+    true
+}
+
+/// Proposal dependency types. These types represent events that precipitate a proposal.
+#[derive(PartialEq)]
+enum ProposalDependency {
```
jparr721, line 73: Not sure how descriptive these names needed to be.

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -0,0 +1,501 @@
...
+
+    #[allow(clippy::no_effect_underscore_binding)]
+    async fn handle_dep_result(self, res: Self::Output) {
+        let mut payload_commitment = None;
+        let mut commit_and_metadata: Option<CommitmentAndMetadata<TYPES::BlockPayload>> = None;
+        let mut _quorum_certificate = None;
+        let mut _timeout_certificate = None;
+        let mut _view_sync_finalize_cert = None;
+        for event in res {
+            match event.as_ref() {
```
jparr721, line 110: Keyao had defined some custom `*Validated` responses here but that won't work for me unfortunately so I instead propagate the event (see below comment) and then that ensures that this task is alive when these events arrive, so we should have no problem receiving them and proposing if all of the requisite information is available.

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -0,0 +1,501 @@
...
+    sender: Sender<Arc<HotShotEvent<TYPES>>>,
+
+    /// Reference to consensus. The replica will require a write lock on this.
+    #[allow(dead_code)]
+    consensus: Arc<RwLock<Consensus<TYPES>>>,
+}
+impl<TYPES: NodeType> HandleDepOutput for ProposalDependencyHandle<TYPES> {
+    type Output = Vec<Arc<HotShotEvent<TYPES>>>;
+
+    #[allow(clippy::no_effect_underscore_binding)]
```
jparr721, line 102: This'll be gone when I add validation here and actually use these. In particular, I would like feedback on this structure. I _hate_ mutating, but this seemed like the best way to unpack multiple heterogeneous types without engineering myself into a black hole. I think I'll split this out into a validation function since not all of these will be set in a given proposal event.

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -0,0 +1,501 @@
...
+    #[instrument(skip_all, fields(id = self.id, latest_proposed_view = *self.latest_proposed_view), name = "Quorum vote create event dependency", level = "error")]
+    fn create_event_dependency(
+        &self,
+        dependency_type: ProposalDependency,
+        view_number: TYPES::Time,
+        event_receiver: Receiver<Arc<HotShotEvent<TYPES>>>,
+    ) -> EventDependency<Arc<HotShotEvent<TYPES>>> {
+        EventDependency::new(
+            event_receiver,
+            Box::new(move |event| {
```
jparr721, line 203: So, as I have come to understand this, we return the view number to make sure that all of the events are for the correct view, then they are collected into the `Vec<HotShotEvent>` and sent down the pipe. So I followed Keyao's lead here and just did the same thing she did. Do I need to do anything else?

crates/task/src/dependency.rs
```diff
@@ -131,12 +136,21 @@ impl<T: Clone + Send + Sync + 'static> EventDependency<T> {
         Self {
             event_rx: receiver,
             match_fn: Box::new(match_fn),
+            completed_dependency: None,
         }
     }
+
+    /// Mark a dependency as completed.
+    pub fn mark_as_completed(&mut self, dependency: T) {
```
jparr721, line 144: I _think_ I don't mind this change. We associate this task as `done` right away and provide it with a value to return. This should, in theory, return the proper result when it comes time to collect everyone. This was what I alluded to above, I need a way to, for example, initiate a proposal off of a `SendPayloadCommitmentAndMetadata` or `ViewSyncFinalize...` event, but I _also_ need to be able to capture that data. Sending this through a bespoke init function for each type would be cumbersome at best. So I came up with this which essentially just blasts the exact value back to the vec.

crates/task/src/dependency_task.rs
```diff
@@ -82,10 +82,7 @@ mod test {
     }
 
     fn eq_dep(rx: Receiver<usize>, val: usize) -> EventDependency<usize> {
-        EventDependency {
-            event_rx: rx,
-            match_fn: Box::new(move |v| *v == val),
-        }
+        EventDependency::new(rx, Box::new(move |v| *v == val))
```
jparr721, line 85: Had to swap this to account for the new field being private, so the `new` constructor was needed.

jparr721:
Got a test here that proves out the functionality of this implementation:
https://github.com/EspressoSystems/HotShot/pull/2832

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -0,0 +1,574 @@
...
+    }
+
+    /// Create and store an [`AndDependency`] combining [`EventDependency`]s associated with the
+    /// given view number if it doesn't exist.
+    fn create_dependency_task_if_new(
+        &mut self,
+        view_number: TYPES::Time,
+        event_receiver: Receiver<Arc<HotShotEvent<TYPES>>>,
+        event_sender: Sender<Arc<HotShotEvent<TYPES>>>,
+        event: Arc<HotShotEvent<TYPES>>,
```
shenkeyao, line 266: Nit: Can we add a comment for `event`? Without comment, it seems that this function will create a task specifically for this event, but it actually will create a task that marks this event as completed.

crates/task-impls/src/quorum_proposal.rs
shenkeyao: This isn't a public enum so I think it's fine.

crates/task-impls/src/quorum_proposal.rs
shenkeyao: I don't like mutating in general either. This is probably fine for this function though, since we only mutate each value once.

crates/task-impls/src/quorum_proposal.rs
shenkeyao: Not sure I understand the difference. This implementation seems very similar to the `handle_dep_result` function for the quorum vote task.

I thought the difference was that we would not have `*Validated` events before creating dependencies and would use `mark_as_completed` to complete the dependency associated with the received event after creating the dependency. But looks like now we have both. Is the `mark_as_completed` calls only to replace the `validate_*` functions in the quorum vote task, rather than completely replace the `Validated*` events?

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -0,0 +1,574 @@
...
+                        if let HotShotEvent::QuorumProposalTimeoutCertValidated(timeout_cert) =
+                            event
+                        {
+                            timeout_cert.view_number
+                        } else if let HotShotEvent::QuorumProposalQuorumCertValidated(qc) = event {
+                            qc.view_number
+                        } else {
+                            return false;
+                        }
+                    }
```
shenkeyao, line 233: We can probably merge these two enum variants into one `ProposalDependency::CertValidated` (or a better name) and use it to match the three `HotShotEvent::*CertValidated` events. They are "or" dependencies anyway.

shenkeyao (commented):
Looks good overall. Let's discuss more tomorrow!

crates/task-impls/src/quorum_proposal.rs
jparr721: This ended up changing a lot since this comment. In short, you approach ended up being the right one here. Sorry for the confusion.

crates/task-impls/src/quorum_proposal.rs
jparr721: Great idea! Added.

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -0,0 +1,477 @@
...
+            }
+            _ => {}
+        };
+
+        // We must always have the payload commitment and metadata, but can have any of the other
+        // events that include a proposable certificate
+        let proposal_dependency = AndDependency::from_deps(vec![
+            payload_commitment_dependency,
+            proposal_cert_validated_dependency,
+        ]);
```
bfish713, line 269: I'm not sure if the intention is to have the logic for the Certificates as well, but it looks like we don't have a dependency for them.  I think there should be an event dependency for QC, ViewSyncCert, or TimeoutCert as well.

If the intention is to just scaffold out the dependency spawning and make sure it works that's fine.

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -0,0 +1,477 @@
...
+
+        match event.as_ref() {
+            HotShotEvent::SendPayloadCommitmentAndMetadata(_, _, _) => {
+                payload_commitment_dependency.mark_as_completed(event.clone());
+            }
+            HotShotEvent::QuorumProposalRecv(_, _)
+            | HotShotEvent::QCFormed(_)
+            | HotShotEvent::ViewSyncFinalizeCertificate2Recv(_) => {
+                proposal_cert_validated_dependency.mark_as_completed(event);
+            }
```
bfish713, line 260: I don't think these 3 things should be grouped together.  The Proposal is a separate event.

For example, if we are leader for view `n`, to propose we'd need both a QC and Proposal from view `n-1` OR a ViewSyncFinalizedCertificate or TimeoutCertificate

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -0,0 +1,477 @@
...
+                    ProposalDependency::ProposalCertificate => match event {
+                        HotShotEvent::ViewSyncFinalizeCertificate2Recv(view_sync_cert) => {
+                            view_sync_cert.view_number
+                        }
+                        HotShotEvent::QCFormed(cert) => match cert {
+                            either::Right(timeout) => timeout.view_number,
+                            either::Left(qc) => qc.view_number,
+                        },
+                        HotShotEvent::QuorumProposalRecv(proposal, _) => proposal.data.view_number,
+                        _ => return false,
```
bfish713, line 202: Same comment as below, these events don't really fulfill the same dependency.

bfish713 (changes requested):
The way the tasks and dependencies are structured makes sense to me.  I think it's good except the logic of what the actual dependencies to propose are needs to be adjusted

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -0,0 +1,541 @@
...
+            event_receiver.clone(),
+        );
+
+        let mut payload_commitment_dependency = self.create_event_dependency(
+            ProposalDependency::PayloadAndMetadata,
+            view_number,
+            event_receiver,
+        );
+
+        match event.as_ref() {
```
jparr721, line 299: @bfish713 So I split these out to their own dependencies and handled Timeout certs separately as requested (looks like the PR comment was removed).

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -0,0 +1,541 @@
...
+                }
+            },
+            HotShotEvent::ViewSyncFinalizeCertificate2Recv(_) => {
+                view_sync_dependency.mark_as_completed(event);
+            }
+            _ => {}
+        };
+
+        // We have three cases to consider:
+        let combined = AndDependency::from_deps(vec![
```
jparr721, line 321: @bfish713 This *seems* to conform to the description you gave, but let me know if I've missed a relationship here. I went through the consensus task tests and implementation and this seems like it's the correct thing, but I'm happy to add more cases if there's additional coupling.

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -0,0 +1,541 @@
...
+
+impl<TYPES: NodeType, I: NodeImplementation<TYPES>> TaskState
+    for QuorumProposalTaskState<TYPES, I>
+{
+    type Event = Arc<HotShotEvent<TYPES>>;
+    type Output = ();
+    fn filter(&self, event: &Arc<HotShotEvent<TYPES>>) -> bool {
+        !matches!(
+            event.as_ref(),
+            HotShotEvent::QuorumProposalRecv(_, _)
```
jparr721, line 521: @bfish713 @elliedavidson @shenkeyao am I missing any events? This appears to be all of them, and I don't see any others in the consensus task that are being used to initiate a proposal.

bfish713 (approved):
Looks good now!

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -0,0 +1,541 @@
...
+        event_sender: Sender<Arc<HotShotEvent<TYPES>>>,
+        event: Arc<HotShotEvent<TYPES>>,
+    ) {
+        debug!("Attempting to make dependency task for event {:?}", event);
+        if self.propose_dependencies.get(&view_number).is_some() {
+            debug!("Task already exists");
+            return;
+        }
+
+        let mut proposal_cert_validated_dependency = self.create_event_dependency(
```
shenkeyao, line 269: Nit: similar to my previous comment, I think it's better to remove `cert` from the name.

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -0,0 +1,541 @@
...
+                    event_sender.clone(),
+                ) {
+                    return;
+                }
+
+                self.create_dependency_task_if_new(view, event_receiver, event_sender, event);
+            }
+            HotShotEvent::QuorumProposalRecv(proposal, _sender) => {
+                let view = proposal.data.get_view_number();
+                if view < self.latest_proposed_view {
```
shenkeyao, line 468: We should also skip the view that has been proposed, so this should be `<=`.

crates/task-impls/src/quorum_proposal.rs
shenkeyao: Yeah, I think this is all.

shenkeyao (commented):
LGTM! Just a change needed for view number comparison, and some minor comments about naming.

crates/task-impls/src/quorum_proposal.rs
jparr721: Fixed this in my branch by making view be `view + 1` as this was a mistake to not add one to it before.

crates/task-impls/src/quorum_proposal.rs
jparr721: Done.

Issue #2709, opened 2024-03-06
[CX_CLEANUP] - Test the dependency framework


shenkeyao:
### What is this task and why do we need to work on it?

* Sad path: Three unit tests, each sends two but misses one of `QuorumProposalRecv`, `DACRecv`, `VidDisperseRecv` events.
    * Verify there’s no `QuorumVoteRecv` event received.
* Happy path: One unit test that sends all three events with dummy fields.
    * Verify there is a `QuorumVoteRecv` event received.

### Are there any other details to include?

Design doc: https://www.notion.so/espressosys/Refactor-Voting-Logic-9afa14dd027d453aa9b4a80a52de5e44.

### What are the acceptance criteria to close this issue?

Tests pass.

PR #2833, opened 2024-03-23
[CX_CLEANUP] - Test the dependency framework

shenkeyao:
Closes https://github.com/EspressoSystems/HotShot/issues/2709.

### This PR: 
* Adds the quorum vote task upon initialization so that we can use it later.
* Adds unit tests for the quorum vote task.
* Miscs:
  * Renames events to be consistent with the diagram.
  * Adds loggings.
  * Updates `justfile`.

### This PR does not: 
* Complete the validation logic for dependencies.

### Key places to review: 
* `crates/task-impls/src/quorum_vote.rs`.
* `crates/testing/tests/quorum_vote_task.rs`.

### How to test this PR:  
* `just async_std test_quorum_vote_task` passes.


crates/testing/tests/quorum_vote_task.rs
```diff
@@ -0,0 +1,143 @@
...
+            Right(consecutive((
+                ViewChange(ViewNumber::new(2)),
+                QuorumProposalValidated(view.quorum_proposal.data.clone()),
+            ))),
+            Right(consecutive((
+                DACertificateValidated(view.da_certificate.clone()),
+                VIDShareValidated(view.vid_proposal.0.data.clone()),
+            ))),
+            Left(exact(QuorumVoteDependenciesValidated(ViewNumber::new(1)))),
+            Left(exact(DummyQuorumVoteSend(ViewNumber::new(1)))),
```
shenkeyao, line 49: Nondeterministic issue: we don't always get both of these two events, and sometimes get neither. I'm wondering if this is because the quorum vote task was shut down before receiving these two tasks triggered by the completed dependency task.

Trying to disable the shutdown or add a pause doesn't help. I'll keep investigating, but please let me know if there's an obvious bug or solution.

jparr721: I managed to fix this by pausing. Let's discuss tomorrow.

shenkeyao: I replaced `try_recv` with `recv` and timeout as we discussed today.

crates/testing/src/script.rs
```diff
@@ -107,11 +120,44 @@ pub async fn run_test_script<TYPES, S: TaskState<Event = Arc<HotShotEvent<TYPES>
         }
 
         for assert in &stage.outputs {
-            if let Ok(received_output) = test_receiver.try_recv() {
-                tracing::debug!("Test received: {:?}", received_output);
-                validate_output_or_panic(stage_number, &received_output, assert);
-            } else {
-                panic_missing_output(stage_number, assert);
+            match assert {
```
jparr721, line 123: Having to add Left and Right to all of the checks seems like it'd grow to become cumbersome over time, and the Left and Right don't really document what they're doing. What do you think about having a flag like `unordered` or something that just checks outputs without worrying about their order of arrival?

shenkeyao: I don’t like the use of `either` either 😅, but haven’t a found a better way yet. The problem with using a flag is that we can’t just say event A is unordered. We have to specify that event A and B can be any order, i.e., somehow group them together, so I figured we need to support an `either` type.

ss-es: I feel like a more general way might be to make the predicates return `pub enum PredicateState { Pass, Fail, Incomplete }` instead of a `bool`, and instead of a single `try_recv()` we loop until it passes or fails. We could then make an `all` predicate that just stores events in some internal `HashMap` or something and passes or fails when it's seen them all or gets an unexpected event. 

But we probably don't need to do this now. For now we could just change the `exact` and other predicates in `predicates.rs` to return `Left` or `Right` as appropriate, and then we won't have to write `Left` or `Right` everywhere in the actual tests (and it'll be easier to change things later).

shenkeyao: @ss-es Both ways sound good! I'll choose the latter one for this PR since it needs fewer changes, but I do like the more general approach and we can probably do it later when needed.

shenkeyao: I ended up choosing an approach combining these two ideas, with an enum supporting checking two consecutive events.

crates/hotshot/src/lib.rs
```diff
@@ -545,6 +545,13 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> SystemContext<TYPES, I> {
             &handle,
         )
         .await;
+        add_quorum_vote_task(
+            registry.clone(),
+            event_tx.clone(),
+            event_rx.activate_cloned(),
+            &handle,
+        )
+        .await;
```
ss-es, line 554: Question: is this just for testing? I don't think this is needed to run `test_quorum_vote_task` (the test should work regardless of whether we spawn the task), and this would run the task in all the examples + production.

crates/hotshot/src/tasks/task_state.rs
```diff
@@ -207,3 +207,19 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> CreateTaskState<TYPES, I>
...
+#[async_trait]
+impl<TYPES: NodeType, I: NodeImplementation<TYPES>> CreateTaskState<TYPES, I>
+    for QuorumVoteTaskState<TYPES, I>
+{
+    async fn create_from(handle: &SystemContextHandle<TYPES, I>) -> QuorumVoteTaskState<TYPES, I> {
+        QuorumVoteTaskState {
+            latest_voted_view: handle.get_cur_view().await,
+            vote_dependencies: HashMap::new(),
+            quorum_network: handle.hotshot.networks.quorum_network.clone(),
+            committee_network: handle.hotshot.networks.da_network.clone(),
```
ss-es, line 220: Just wondering: `da_network` and `committee_network` mean the same thing, right?

Should we just pick one of the names and replace the other? (no preference for which one, and we can just do this later too if it's not quick/you just want to merge this in)

crates/task-impls/src/events.rs
```diff
@@ -43,12 +43,15 @@ pub enum HotShotEvent<TYPES: NodeType> {
     /// A Data Availability Certificate (DAC) has been recieved by the network; handled by the consensus task
     DACRecv(DACertificate<TYPES>),
     /// A DAC is validated.
-    DACValidated(DACertificate<TYPES>),
+    DACertificateValidated(DACertificate<TYPES>),
```
ss-es, line 46: +1 on the name change

should we also change `DACRecv` to `DACertificateRec` too for consistency? (this is also very minor so we can just do it later)

ss-es (commented):
Sorry for not looking at it sooner! Looks good overall, only real comment is about spawning the task

crates/testing/src/predicates.rs
```diff
@@ -20,17 +18,42 @@ impl<INPUT> std::fmt::Debug for Predicate<INPUT> {
     }
 }
 
-pub fn exact<TYPES>(event: HotShotEvent<TYPES>) -> Predicate<Arc<HotShotEvent<TYPES>>>
+pub type ConsecutiveEvents<TYPES> = (Arc<HotShotEvent<TYPES>>, Arc<HotShotEvent<TYPES>>);
```
jparr721, line 21: So does this only allow two events to occur out of order?  What do you think about just extending this to be a vector of events that we process?

crates/testing/src/script.rs
```diff
@@ -108,20 +112,53 @@ pub async fn run_test_script<TYPES, S: TaskState<Event = Arc<HotShotEvent<TYPES>
...
+                    {
+                        Ok(Ok(received_output)) => {
+                            tracing::debug!("Test received: {:?}", received_output);
+                            validate_output_or_panic(stage_number, &received_output, assert);
+                        }
+                        _ => panic_missing_output(stage_number, assert),
+                    }
+                }
+                EventPredicate::Consecutive(asserts) => {
+                    match async_timeout(
```
jparr721, line 131: So is this essentially waiting until the first event happens, then waiting until the second event happens? I am having trouble understanding how this differs from two events just coexisting in the `output` vector. Is this essentially *forcing* the two events to occur right after each other?

crates/hotshot/src/lib.rs
shenkeyao: I need to send an update to resolve merge conflicts anyway, so will do the renaming as well!

crates/testing/src/predicates.rs
shenkeyao: It would be nice to extend this to be a vector, but I'm leaning towards just a tuple since it's what's needed for now, and I'd like to merge this soon to avoid more conflicts. 😄

crates/testing/src/script.rs
shenkeyao: There are two "outputs", one is the expected output and the other is the received output. For the expected output, it's just the `asserts` tuple. However, we have to call `recv_direct` twice to get the received output, which we can't just get from an `output` vector but need to construct it. The two events don't really coexist, so yes, they occur right after each other, but we just don't want to force the order.

crates/hotshot/src/lib.rs
shenkeyao: Oops, the above reply was meant for your other comments. 😅

Regarding this question -- I keep it there to save some work for my future integration work.

crates/hotshot/src/tasks/task_state.rs
shenkeyao: I didn't do this renaming (but did the other one for the event name) since we have a lot of mixed-use of "da" and "committee". I'd like to do a renaming for all in a separate PR.

crates/task-impls/HotShot_event_architecture.drawio
```diff
@@ -46,7 +46,7 @@
             </Array>
           </mxGeometry>
         </mxCell>
-        <mxCell id="WsWUFxSsiAewtnnMHRLG-8" value="&lt;div style=&quot;font-size: 11px; line-height: 18px;&quot;&gt;QuorumProposalRecv&lt;/div&gt;&lt;div style=&quot;font-size: 11px; line-height: 18px;&quot;&gt;DACRecv&lt;/div&gt;&lt;div style=&quot;font-size: 11px; line-height: 18px;&quot;&gt;TimeoutVoteRecv&lt;/div&gt;" style="edgeLabel;html=1;align=center;verticalAlign=middle;resizable=0;points=[];fontSize=11;labelBackgroundColor=none;fontColor=#000000;" vertex="1" connectable="0" parent="WsWUFxSsiAewtnnMHRLG-7">
+        <mxCell id="WsWUFxSsiAewtnnMHRLG-8" value="&lt;div style=&quot;font-size: 11px; line-height: 18px;&quot;&gt;QuorumProposalRecv&lt;/div&gt;&lt;div style=&quot;font-size: 11px; line-height: 18px;&quot;&gt;DACertificateRecv&lt;/div&gt;&lt;div style=&quot;font-size: 11px; line-height: 18px;&quot;&gt;TimeoutVoteRecv&lt;/div&gt;" style="edgeLabel;html=1;align=center;verticalAlign=middle;resizable=0;points=[];fontSize=11;labelBackgroundColor=none;fontColor=#000000;" vertex="1" connectable="0" parent="WsWUFxSsiAewtnnMHRLG-7">
```
jparr721, line 49: Looks like a global rename got you here.

shenkeyao: Good catch!

jparr721 (dismissed):

crates/hotshot/src/lib.rs
shenkeyao: I've commented this out as discussed on Zulip.

jparr721 (approved):

ss-es (approved):

Issue #2836, opened 2024-03-25
[BUILDER] - Simple Builder can include already-decided transactions


QuentinI:
### What is this task and why do we need to work on it?

> This logic won't remove transactions if the `TransactionRecv` event happens after a leaf containing that transaction is decided.  It's pretty common with the webserver we download transactions in batches right before we become a leader.  So we'll get lots of transactions that were already decided.  I'm not sure how relevant that is for this builder if we only use as a test.  If we do bench-marking with it then we probably need to handle this case.

_Originally posted by @bfish713 in https://github.com/EspressoSystems/HotShot/pull/2817#discussion_r1535940597_
            

### What are the acceptance criteria to close this issue?

Simple builder doesn't include transactions that are already decided.

QuentinI:
Closed in #2837

PR #2837, opened 2024-03-25
[BUILDER] Builder integration

### LINKED ISSUE #2667: [BUILDER] - Builder integration

Integrate builder client into hotshot

_No response_

_No response_

Hotshot queries and accepts blocks by builder

_No response_

QuentinI:
Closes #2667, #2836 

### This PR: 
- Integrates test builders into HotShot testing harness
- Replaces logic in transaction task to fetch blocks from builder

### This PR does not: 
- include tests with random builder (only simple builder is used)
- adjust VID logic to make use of commitment calculated by builder


crates/task-impls/src/transactions.rs
```diff
@@ -235,68 +130,79 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
+                .await
+            {
+                Ok(blocks) => blocks,
+                Err(err) => {
+                    error!(%err, "Couldn't get available blocks");
+                    continue;
+                }
+            };
+
+            available_blocks.sort_by_key(|block_info| block_info.block_size);
```
bfish713, line 161: Should we be sorting by the fee, intead of size?

crates/task-impls/src/transactions.rs
```diff
@@ -235,68 +130,79 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
-            //     .iter()
-            //     .filter(|(txn_hash, _txn)| !previous_used_txns.contains(txn_hash))
-            //     .collect();
-            let unclaimed_txns = all_txns;
+            // Don't try to re-claim the same block if builder advertises it again
+            if latest_block.as_ref().map_or(false, |block| {
+                block.0.block_payload.builder_commitment(&block.0.metadata) == block_info.block_hash
+            }) {
+                continue;
+            }
```
bfish713, line 172: If the builder is giving us multiple options for blocks,  and the first one matches the `latest_blocks` shouldn't we just try next block instead of starting the loop over and querying the builder again?

bfish713 (commented):
Logic looks fine to me, just a few comments.

Mabye @nyospe or @move47 can double check the changes not in testing to make sure it makes sense

crates/task-impls/src/transactions.rs
```diff
@@ -75,101 +61,21 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
-                                    true
-                                }
-                            })
-                            .collect();
+                self.api
+                    .send_event(Event {
+                        view_number: self.cur_view,
+                        event: EventType::Transactions {
+                            transactions: transactions.clone(),
+                        },
```
bfish713, line 76: I'm not sure what the expected behavior is but I think we might still want to track transactions locally and propose the from them in the case where there are not blocks.  We might have decided against that because the proposers has to pay a fee and our naive block would surely not recuperate the fee in most cases.  Worth a double check

crates/task-impls/src/transactions.rs
QuentinI: I was thinking of trying to include the most transactions instead, but it doesn't necessarily follow from bigger block size. Sorting by fee makes more sense, I'll switch to that.

crates/task-impls/src/transactions.rs
QuentinI: The one we're checking is the "best" one, so if it's the same, I don't see much need to try other ones.
Maybe in the future when we have lists of transactions we want to include or more complex metrics by which we choose blocks, but for now any other block we check will be "worse" than the `latest_block`.

crates/task-impls/src/transactions.rs
QuentinI: I would also be against it from complexity standpoint. We'd be adding an edge case here for when none of the builders (when we support multiple) proposed a block and there are transactions in our mempool, which will need to be maintained and tested and be configurable for node operator needs in case it's not profitable. I think it would be better served by simply running a separate builder of our own configured to include all transactions or somesuch.

crates/task-impls/src/transactions.rs
bfish713: I see I slightly misread the code, we are looping until the we find a block that is sufficient or we run out of time.

QuentinI: Yes, that's correct

nyospe (changes requested):
Looks good, aside from the Version integration issue.

crates/task-impls/src/transactions.rs
```diff
@@ -50,18 +39,15 @@ pub struct TransactionTaskState<
...
-    pub seen_transactions: HashSet<Commitment<TYPES::Transaction>>,
-
     /// Network for all nodes
     pub network: Arc<I::QuorumNetwork>,
 
     /// Membership for teh quorum
     pub membership: Arc<TYPES::Membership>,
 
+    /// Builder API client
+    pub builder_client: BuilderClient<TYPES, Version01>,
```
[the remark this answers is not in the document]
QuentinI: I propagated concrete version up to `add_transaction_task` in `crates/hotshot/src/tasks/mod.rs` by analogue to what I see in `add_request_network_task` in the same, but I'm not entirely sure I understand what I'm doing here and how it changes things (if it does at), so input appreciated.

bfish713:
@nyospe Are your comments resolved? If not can you make the changes?

crates/task-impls/src/transactions.rs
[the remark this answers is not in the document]
nyospe: I've cleared it for now, we need to fix this through HotShot before our first real Update happens.

nyospe (approved):

ss-es (changes requested):
Haven't had a chance to look at the changes, but I'm worried that all the async-std tests are failing. Re-running them now

ss-es:
Caveat to my approval: I think this PR breaks examples (and therefore benchmarks) on main, and it wasn't clear to me how to fix that

bfish713:
> Caveat to my approval: I think this PR breaks examples (and therefore benchmarks) on main, and it wasn't clear to me how to fix that

I assume it's because there is no builder being run in the examples.  And therefore leaders can't get blocks and don't propose.  I think if we run with a builder we can run the benchmarks

PR #2846, opened 2024-03-26
[Auto Bench] Scripts of Running Benchmarks with Nginx

### LINKED ISSUE #2826: [Auto Benchmarks] - Use nginx in scripts of running benchmarks in AWS

Current benchmarking scripts doesn't use nginx, however this is needed when `total_nodes >= 500` and `transaction_size` is large.

- [x] Use nginx in scripts of running benchmarks in AWS

After this , the script should be able to run with `total_nodes \in {500, 1000}`, `transaction_size \in {1000000 20000000}`.

_No response_

dailinsubjam:
Closes #2826 .

### This PR: 
- Adds the script for benchmarking when using nginx.
- Raise up runnable tx size to 1e7 bytes. With 50 nodes and 5 da committee it can get 2.37MBps. With tx size of 2e8 and 10/5 nodes, it can get 15.56MBps. More results in `results_init_run.csv`.
- Supports loop-over-tuples-in-bash for parameter assignment.
- Changed some `debug` and `error` log to `println`, since our Datadog does not support `info`(not sure why) and `debug`, and it takes `println` as `info`. 

### This PR does not: 
- Guarantee large total nodes (> 100) and large tx size (> 1e7 bytes) could finish runs successfully, I still meet lots of time outs using nginx.

### Key places to review: 
`scripts/aws_ecs_nginx_benchmarks.sh`

### How to test this PR: 
Follow the testing steps here: https://github.com/EspressoSystems/HotShot/pull/2793 .
And change the script to `./script/aws_ecs_nginx_benchmarks.sh`.
Remember to run nginx following the step mentioned in the script.


crates/examples/infra/mod.rs
```diff
@@ -570,10 +570,10 @@ pub trait RunDA<
         let mut total_latency = 0;
         let mut num_latency = 0;
 
-        debug!("Sleeping for {start_delay_seconds} seconds before starting hotshot!");
+        println!("Sleeping for {start_delay_seconds} seconds before starting hotshot!");
```
jparr721, line 573: Looks like this PR 2858 have similar changes. Overall I think removing logging for print statements will make parsing logs for us more difficult when not in datadog (which is a big part of the workflow). Perhaps we can just set datadog to take info logs in docker and then filtein the UI, or filter the executable as `RUST_LOG=warn,hotshot_orchestrator::config=info` per Salman's suggestion?

jparr721 (changes requested):
It might be worth a team discussion about these logs.

crates/examples/infra/mod.rs
dailinsubjam: Yes, this PR is covered by that one. I set all of these to `info` in that PR, and I'm using `RUST_LOG=warn,hotshot_orchestrator::config=info` locally.

dailinsubjam:
Coverd by https://github.com/EspressoSystems/HotShot/pull/2858

PR #2849, opened 2024-03-27
Move validate and apply off the critical path

### LINKED ISSUE #2847: [CATCHUP] - Don't block Consensus waiting for `validate_and_apply`

`validate_and_apply_header` will retry forever.  We can't simply await the results in the consensus task.  We need to move this off the critical path.

_No response_

_No response_

Catchup works when the replica can't get task from the old view.

_No response_

bfish713:
Closes #2847

### This PR: 

- Moves the state validation and storage of state/leafs to an async task
- Consensus task handles the `QuorumProposalValidated` event from this new async task.
- Changes the tests to run 

Basically this is splitting the proposal handling logic into 3 parts.  The first part is unchanged, it's the initial validation and short circuiting logic. This part is unchaged from before.  The next part is the state validation and storage, which is done in the new function `validate_proposal` and run in a spawned task.  The last step is also unchanged and comes after the part 2 task completes.  It updates our consensus information (e.g. did we decide on a view, update our locked QC etc.)  and proposes/votes if we should.

### This PR does not: 

This should keep intact the general consensus logic

### Key places to review: 

Changes to the consensus task + tests


crates/task-impls/src/consensus.rs
```diff
@@ -61,6 +61,121 @@ pub struct CommitmentAndMetadata<PAYLOAD: BlockPayload> {
 /// Alias for Optional type for Vote Collectors
 type VoteCollectorOption<TYPES, VOTE, CERT> = Option<VoteCollectionTaskState<TYPES, VOTE, CERT>>;
 
+/// Validate the state and safety and liveness of a proposal then emit the
+/// a `QuorumProposalValidated` event.
```
jparr721, line 65: nit: "the a" could be "the" or "a"

jparr721 (dismissed):
Great change, excited to incorporate this in my task.

jparr721 (approved):

PR #2852, opened 2024-03-27
Add upgrade certificate to Leaf

### LINKED ISSUE #2702: [Upgradability] - Include upgrade certificates in the commitment calculation

Currently, the commitment calculation does not include upgrade certificates. It should.

_No response_

_No response_

Upgrade certificates are included in the calculation of leaf commitments.

_No response_

ss-es:
Closes #2702

### This PR: 
- Adds upgrade certificates to the `Leaf` struct.
- Adds upgrade certificates to the `Leaf::commit` calculation.

While doing this, I also did some minor cleanup around the `Leaf` struct. Notably:
- `Leaf` now has a method `from_quorum_proposal`, which constructs the leaf directly from the quorum proposal.
- All fields of `Leaf` are private, and access to the fields now happens via the (already existing) field methods. The primary benefit here is that it should be much harder to construct invalid leaves by accident.

### This PR does not: 
The issues around the leaf commit/signature calculation are not addressed. The logic around leaf calculations should not have changed.

### Key places to review: 
The most important thing to verify is that the `consensus` logic hasn't actually changed. Previously, we were constructing leaves with some fields from our own internal state and some fields from the received quorum proposal. Now, the leaf is produced directly from the received quorum proposal and the parent commitment check happens separately.

### DIFF
--- a/crates/example-types/src/block_types.rs
+++ b/crates/example-types/src/block_types.rs
@@ -197,7 +197,7 @@ impl<TYPES: NodeType<BlockHeader = Self, BlockPayload = TestBlockPayload>> Block
         payload_commitment: VidCommitment,
         _metadata: <TYPES::BlockPayload as BlockPayload>::Metadata,
     ) -> Self {
-        let parent = &parent_leaf.block_header;
+        let parent = parent_leaf.get_block_header();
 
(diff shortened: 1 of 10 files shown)


bfish713 (approved):
Really like the addition of `Leaf::from_quorum_proposal`.  LGTM

PR #2854, opened 2024-03-28
feat: add aztec-srs

### LINKED ISSUE #1686: Secure SRS for VID
Currently we use jellyfish's `gen_srs_for_testing`, even in production. Obviously, we shouldn't do that.

[EDIT: remove `test-srs` feature flag from `Cargo.toml`, too!]

In an ideal world this SRS would be loaded from a config file. Better yet, a default SRS could be hard-coded so as to save work on io/parse.

alxiong:
Closes #1686 

### This PR: 

- add aztec's SRS behind `aztec-srs` feature flag and conditionally import it

### This PR does not: 

- turn this on by default, as the benchmarking effort by @dailinsubjam should still just use test srs, which makes no difference and could be faster than loading production-ready one from a file


dailinsubjam:
Thanks for tagging! IIUC, I don't have to turn this on when doing benchmarking, but I have to turn `gpu-vid` on for leaders who need to run vid on GPU right?

alxiong:
> but I have to turn gpu-vid on for leaders who need to run vid on GPU right?

yes, also make sure you have Nvida GPU and have cuda toolkit etc installed. 

The SRS introduced here really doesn't affect your benchmarking effort, just use test-srs for your benchmark

ggutoski:
We should remove `test-srs` feature flag from Cargo.toml, too.

alxiong:
> We should remove test-srs feature flag from Cargo.toml, too.

why? I think we need it most of the time (at least in test) or are you suggesting we only add the flag in `[dev-dependencies]`?

ggutoski:
> are you suggesting we only add the flag in `[dev-dependencies]`?

Yes. It should not be used in production.

alxiong:
Let me document my thinking for future reference

> [!NOTE]
On API design on test v.s. production SRS

AFAICT, there are two approaches: 
1. diff functions optionally enabled via feature flag
    - pros:
      - explicit in names `srs_for_test()`, arguable less error-prone for dev
    - cons:
      - harder to manage combinatorial feature flag conditions when there are more options: `#[cfg(and(not(features = "aztec-srs"), not(feature = "aleo-srs")), ...)`
      - you can't run your tests with `--all-features` in CI, instead you end up specifying all features except the mutually exclusive ones, which is an anti-pattern.
2. same function switched via rustflags
   - pros:
     - ensure mutual exclusive choice via `RUSTFLAGS` (avoiding complicated `cfg` expression)
   - cons (?):
     - different choice means different binaries

An important realization is that **both approaches force downstream users/libraries to follow the same pattern**:
- if you use `test_srs()`, then all downstream libraries would create `my_action_with_test_srs()` separate from `my_action()`, and turn them on optionally via propagating feature flags
- if you use `--cfg crypto_srs = "test"`, then all downstream would need to specify that for the compiler to select the proper code to compile

Remember: 
> features should be additive. That is, enabling a feature should not disable functionality, and it should usually be safe to enable any combination of features.
> -- [Rust book](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification)

IMO, we should adopt option 2. but the unfortunate legacy choice of 1 on Jellyfish means that it may require more refactoring if we decide to switch.
A good example I always refers to is the `backend` or `word size` choice in [dalek's curve25519](https://github.com/dalek-cryptography/curve25519-dalek/tree/main/curve25519-dalek#backends) crate.

cc @mrain @ggutoski @philippecamacho @akonring @chancharles92

alxiong:
Gus, your suggestion in https://github.com/EspressoSystems/HotShot/pull/2854#issuecomment-2030064207 will be cleaner if we switch to aforementioned option 2. but it'll be slightly messier in current "feature-flag" approach. 

for now, we either:
- default to aztec SRS: remove `aztec-srs`, add `test-srs` here which enable `jf-primitives/test-srs` -- this means HotShot users is going to download 60MB by default
- default to test SRS: keep the status quo of this current commit HEAD of this PR.

which do you prefer? @ggutoski

alxiong:
@ggutoski this is ready for review again. lmk if you want further explanation of the upstream update to `ark-srs`, but if we just care about API usage, it should be straightforwardly seen here in this PR.

crates/types/src/vid.rs
```diff
@@ -124,12 +124,28 @@ lazy_static! {
...
     };
 }
 
+// By default, use SRS from Aztec's ceremony
+#[cfg(not(feature = "test-srs"))]
+lazy_static! {
+    /// SRS comment
+    static ref KZG_SRS: UnivariateUniversalParams<E> = {
+        let srs = ark_srs::kzg10::aztec20::setup(2u64.pow(20) as usize + 2)
+            .expect("Aztec SRS fail to load");
```
jparr721, line 139: nit: failed*

crates/types/src/vid.rs
```diff
@@ -124,12 +124,28 @@ lazy_static! {
...
         .unwrap()
     };
 }
 
+// By default, use SRS from Aztec's ceremony
+#[cfg(not(feature = "test-srs"))]
+lazy_static! {
+    /// SRS comment
+    static ref KZG_SRS: UnivariateUniversalParams<E> = {
+        let srs = ark_srs::kzg10::aztec20::setup(2u64.pow(20) as usize + 2)
```
jparr721, line 138: Could we perhaps remove the magic numbers here? It's unclear what they mean and where they've come from.

flake.nix
```diff
@@ -198,6 +199,11 @@
         devShell = pkgs.mkShell {
           inherit CARGO_TARGET_DIR;
           buildInputs = [ fenixStable ] ++ buildDeps;
+          shellHook = ''
+            export ROOT_DIR=$(dirname "$(realpath ./flake.nix)")
+            export AZTEC_SRS_PATH="$ROOT_DIR/data/aztec20/kzg10-aztec20-srs-1048584.bin"
```
jparr721, line 204: Is this file always guaranteed to exist?

scripts/download_srs_aztec.sh
```diff
@@ -0,0 +1,19 @@
...
+    echo "Environment variables loaded from .env"
+else
+    echo ".env file not found"
+fi
+
+if [ -f "$AZTEC_SRS_PATH" ]; then
+    echo "SRS file $AZTEC_SRS_PATH exists"
+else
+    echo "SRS file $AZTEC_SRS_PATH does not exist, downloading ..."
+    gh release download --repo alxiong/ark-srs v0.2.0 -p "$(basename $AZTEC_SRS_PATH)" -O "$AZTEC_SRS_PATH" --skip-existing
```
jparr721, line 18: Is it wise to use the `gh` tool when not everyone will have it installed? Also, should we pehaps pin this download into a company-owned github repository instead of a personal one?

jparr721 (changes requested):

crates/types/src/vid.rs
mrain: That's some upperbound that we need to provide here.
Perhaps we should define a constant for it.

flake.nix
alxiong: no, the following line `./download_srs_aztec.sh` will download it if not exists, and skip if already exist, and this check is done every time we enter the dev shell.

crates/types/src/vid.rs
alxiong: sounds good, will define a constant and add code comment to explain where this number is coming from.

alxiong: done in [47870cb](https://github.com/EspressoSystems/HotShot/pull/2854/commits/47870cb78cbbc2fb81f615f134b1140201ffbc29)

crates/types/src/vid.rs
alxiong: fixed in [47870cb](https://github.com/EspressoSystems/HotShot/pull/2854/commits/47870cb78cbbc2fb81f615f134b1140201ffbc29)

scripts/download_srs_aztec.sh
alxiong: switch to `wget` and removed `gh` in the `buildInputs` in [47870cb](https://github.com/EspressoSystems/HotShot/pull/2854/commits/47870cb78cbbc2fb81f615f134b1140201ffbc29)

ggutoski:
> @ggutoski this is ready for review again. lmk if you want further explanation of the upstream update to `ark-srs`, but if we just care about API usage, it should be straightforwardly seen here in this PR.

Thanks! Let's talk about it in today's meeting.

jparr721:
> > @ggutoski this is ready for review again. lmk if you want further explanation of the upstream update to `ark-srs`, but if we just care about API usage, it should be straightforwardly seen here in this PR.
> 
> Thanks! Let's talk about it in today's meeting.

Just want to highlight that the build is failing here looks like due to the lazy static poisoning one of the threads. If you guys think it looks good and want to approve I'll be happy to also drop a green check once all the builds are working. Happy to help if needed.

alxiong:
let me fix some of the CI errors, they are caused by the fact that I didn't add `./download_aztec_srs.sh` in their setup

.github/workflows/build-and-test.yml
```diff
@@ -52,11 +52,13 @@ jobs:
 
       - name: Unit and integration tests for all crates in workspace
         run: |
+          export AZTEC_SRS_PATH="$PWD/data/aztec20/kzg10-aztec20-srs-1048584.bin"
+          ./scripts/download_srs_aztec.sh
```
jparr721, line 56: I wonder if this can be a justfile command or something.

jparr721 (dismissed):
This looks good, but I'd like to wait for @ggutoski or others for confirmation.

.github/workflows/build-and-test.yml
alxiong: okay! i will do that, I like single responsibility principle.

jparr721: I'll re-approve when you add it then!

alxiong: done in [c3fd331](https://github.com/EspressoSystems/HotShot/pull/2854/commits/c3fd33151aecb47f6abaa1df0fa5b3c11052a6fe)

crates/types/src/vid.rs
```diff
@@ -116,22 +117,35 @@ pub struct SmallRangeProofType(
...
-            // TODO what's the maximum possible SRS size?
-            checked_fft_size(200).unwrap(),
+            SRS_DEGREE,
         )
         .unwrap()
     };
 }
 
+// By default, use SRS from Aztec's ceremony
+#[cfg(not(feature = "test-srs"))]
```
ggutoski, line 134: Hmm. I did not properly understand until now your dilemma about "same fn" vs "different fn". 😕 I'm a little uneasy with this violation of the additivity of cargo features. My previous understanding is that `test-srs` is purely additive: it would enable a special test srs used to speed up tests, but it would not remove the aztec srs.

What's the reasoning here? Let me guess: we want to save users who run in test mode the trouble of downloading the aztec srs. Is that all? If so then I don't think that's a good enough reason to violate additivity. Test users can suck up the pain and download aztec. 😛

ggutoski (commented):
Thanks for all this work! Question below.

crates/types/src/vid.rs
alxiong: the reason is HotShot seem to want to keep SRS under the same name `static ref KZG_SRS`,
maybe we can have a separate `KZG_SRS_TEST`, which means a separate `fn vid_scheme_test()` etc. and this "extra `_test()` that accepts test srs" pattern will propagate all the way out.

which poison you want to swallow? (I'm fine with both tbh)

.github/workflows/build-and-test.yml
alxiong: ah! the env var `AZTEC_SRS_PATH` is needed in the subsequent steps, and setting env var in just command is not possible (afaik), it's only good at reading it or temporarily setting it within the scope of a just command ([see here](https://just.systems/man/en/chapter_37.html))

so I will revert back. lmk if you disagree @jparr721

crates/types/src/vid.rs
ggutoski: I've consulted others. The mob has spoken. `test-srs` shall be additive, we'll have a separate `KZG_SRS_TEST`.

alxiong: how does this look [39396a1](https://github.com/EspressoSystems/HotShot/pull/2854/commits/39396a1a1778a636c169cd35143bf3c0a36bacb3)?

note that `vid_scheme_for_test()` is not used anywhere currently, but could be in the future if needed.
atm, all test codes are invoking with production SRS, such as `block_contents.rs::precompute_vid_commitment()` which is invoked by `testing/src/block_builder.rs::build_block()`.

^^ it is obviously also fine to run unit test directly on proper SRS, at least for now it's not a memory or speed concern.

.github/workflows/build-and-test.yml
jparr721: I think that that is fine. Using a `scripts/` folder is a perfectly idiomatic thing as well.

crates/types/src/vid.rs
```diff
@@ -66,6 +67,18 @@ pub fn vid_scheme(num_storage_nodes: usize) -> VidSchemeType {
...
+/// Similar to [`vid_scheme()`], but with `KZG_SRS_TEST` for testing purpose only.
+#[cfg(feature = "test-srs")]
+pub fn vid_scheme_for_test(num_storage_nodes: usize) -> VidSchemeType {
+    let recovery_threshold = 1 << num_storage_nodes.ilog2();
+    #[allow(clippy::panic)]
+    let num_storage_nodes = u32::try_from(num_storage_nodes).unwrap_or_else(|err| {
+        panic!("num_storage_nodes {num_storage_nodes} should fit into u32\n\terror: : {err}")
+    });
+    #[allow(clippy::panic)]
+    VidSchemeType(Advz::new(num_storage_nodes, recovery_threshold, &*KZG_SRS_TEST).unwrap_or_else(|err| panic!("advz construction failure:\n\t(num_storage nodes,recovery_threshold)=({num_storage_nodes},{recovery_threshold})\n\terror: : {err}")))
```
jparr721, line 79: Since these lines are touched, I want to bring this up: can we remove the newlines and tabs from these logs? It creates non-standard outputs and can cause problems in a structured context.

jparr721 (changes requested):
Quick thing here.

crates/types/src/vid.rs
ggutoski: no objection so long as Alex is the one to do it. 😆

ggutoski (dismissed):

crates/types/src/vid.rs
alxiong: what new lines and tabs? you mean line 76?

PR #2856, opened 2024-03-28
[Builder] API changes

move47:
- ~~Serde support for General Static Committee~~ [pub trait Membership<TYPES: NodeType> isn't serde]
- Return tx_hash from tx_submit_api
- Adds a new builder address api requested in : https://github.com/EspressoSystems/espresso-sequencer/issues/1289


move47:
@nyospe I think it would be good to make `General Static Committee` serde support. 
Because, here we need SeqTypes, BLSPubKey. The builder might not have SeqTypes. Let me know your thoughts. 
https://github.com/EspressoSystems/espresso-sequencer/blob/d0ccb9e1cd5207bb489112e290237ddbd505b413/sequencer/src/context.rs#L75C1-L78C11

nyospe (dismissed):
LGTM;

Remember to link the issue.

crates/hotshot/src/traits/election/static_committee.rs
```diff
@@ -16,7 +16,8 @@ use tracing::debug;
 
 /// Dummy implementation of [`Membership`]
 
-#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+#[derive(Clone, Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
```
nyospe, line 19: I still don't understand why you're doing this, and I *really* hope you're not planning to send this over instead of the two parameters needed to construct it, as we previously discussed.

move47: No, I don't plan to send this over. Is it fine to keep the serde bound still?

move47: Removed serde bound. `pub trait Membership<TYPES: NodeType>` isn't serde

nyospe (approved):
LGTM

nyospe:
Closes https://github.com/EspressoSystems/espresso-sequencer/issues/1289

PR #2858, opened 2024-03-28
[Auto Bench] HotShot changes with GPU VID

### LINKED ISSUE #2857: [Auto Benchmarks] - HotShot config and examples with GPU VID

This is to enable GPU VID option when running examples, also prepare for benchmarking with GPU VID, for which we'll need fixed set of leader to run on an NVIDIA GPU.

- [x] Configuring feature flags (or something similar) to control the code executed on different nodes
- [x] Example changes
- [x] Config changes: use fixed set of leader
- [x] Add vid parameter to benchmarking arguments, and [choose wisely ](https://github.com/EspressoSystems/HotShot/blob/b9d501375957b21759a838cebfe8315c5528e7d4/crates/types/src/vid.rs#L58)

_No response_

See changes in `crates/examples/webserver/README.md`. There are steps of running examples.

_No response_

dailinsubjam:
Closes #2857 #2826 .

### This PR: 
- Adds a feature `fixed-leader-election`. When it's enabled, validators will only choose leader in fixed set.
- Adds a parameter `fixed_leader_for_gpuvid: usize` in HotshotConfig. If `fixed-leader-election` is enabled, this number determines the nodes with index between `[0, fixed_leader_for_gpuvid) ` will be the fixed leader set.
- Covers nginx script changes, more info https://github.com/EspressoSystems/HotShot/pull/2846 .
- Changed some logs in `hotshot_orchestrator` from `error` to `info`.

### This PR does not: 
- Run with real GPU.

### Key places to review: 
All changes, but I want to highlight `crates/examples/webserver/README.md` which includes running steps:
    - When we do real benchmarking with GPU, we'll start leaders first, they should enable both `fixed_leader_for_gpuvid` and `gpu-vid` features, use their own docker images, run on a nvidia gpu clusters.
    - Then we start other validators, they only enable `fixed_leader_for_gpuvid`, they use another common docker image and run on one of existing clusters. See [config of current cluster](https://github.com/EspressoSystems/HotShot-deploy/blob/b98f451ca8e5fcc905505c412cdcae4fd975e5ec/cluster/variables.tf).
   

 ### How to test this PR: 
See changes in `crates/examples/webserver/README.md` for steps of running tests.


crates/orchestrator/src/config.rs
```diff
@@ -218,7 +218,7 @@ impl<K: SignatureKey, E: ElectionConfig> NetworkConfig<K, E> {
         file: Option<String>,
     ) -> (NetworkConfig<K, E>, NetworkConfigSource) {
         if let Some(file) = file {
-            error!("Retrieving config from the file");
+            println!("Retrieving config from the file");
```
jparr721, line 221: Do we want to make these raw print statements? Would it be better to just info log or something?

crates/orchestrator/src/config.rs
```diff
@@ -528,6 +528,8 @@ pub struct HotShotConfigFile<KEY: SignatureKey> {
     pub staked_committee_nodes: usize,
     /// Number of non-staking committee nodes
     pub non_staked_committee_nodes: usize,
+    /// Number of fixed leader for GPU VID
```
jparr721, line 531: nit: leaders*

crates/task-impls/src/consensus.rs
```diff
@@ -269,7 +269,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
             if *self.cur_view / 100 != *new_view / 100 {
                 // TODO (https://github.com/EspressoSystems/HotShot/issues/2296):
                 // switch to info! when INFO logs become less cluttered
-                error!("Progress: entered view {:>6}", *new_view);
+                println!("Progress: entered view {:>6}", *new_view);
```
jparr721, line 272: Same log comment as before.

jparr721 (changes requested):
Looks like some of the log statements were replaced with raw println, are these from debugging?

crates/orchestrator/src/config.rs
dailinsubjam: Yes! Because I'd like to have them as `info` messages instead of `error`, however datadog does not show `info` or `debug` level message :( and datadog categorize `println` as `info`, therefore I changed to this.

justfile
```diff
@@ -41,6 +41,12 @@ build_release *ARGS:
 example *ARGS:
   cargo run --profile=release-lto --package hotshot-examples --no-default-features --example {{ARGS}}
 
+example_gpuvid *ARGS:
```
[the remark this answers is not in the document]
dailinsubjam: Aha!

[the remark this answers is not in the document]
dailinsubjam: How about rename this to `example_gpuvid_validator` and the next one to `example_gpuvid_leader`, so that users don't need to care about what exact features are enabled, maybe more intuitive?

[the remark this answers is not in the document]
mrain: But the GPU is not enabled here, right? Users could get confused.

jparr721 (commented):
LGTM I’ll let Brendon and Rob take a look since I’m less familiar with this code. Thanks for doing the log changes!!

crates/hotshot/src/traits/election/static_committee.rs
```diff
@@ -69,14 +73,26 @@ where
...
         let index = usize::try_from(*view_number % self.nodes_with_stake.len() as u64).unwrap();
         let res = self.nodes_with_stake[index].clone();
         TYPES::SignatureKey::get_public_key(&res)
     }
 
+    #[cfg(feature = "fixed-leader-election")]
+    /// Only get leader in fixed set
+    /// Index the fixed vector (first fixed_leader_for_gpuvid element) of public keys with the current view number
+    fn get_leader(&self, view_number: TYPES::Time) -> PUBKEY {
+        let index = usize::try_from(*view_number % self.fixed_leader_for_gpuvid as u64).unwrap();
```
rob-maron, line 91: General question since I'm a bit unfamiliar: are we (for now) expecting the fixed leaders to be the first n nodes? Is it moderately easy to guarantee we can have the first n nodes be the GPU ones? Maybe it is - we can probably just have them connect to the orchestrator first 😃

scripts/aws_ecs_nginx_benchmarks.sh
```diff
@@ -0,0 +1,72 @@
...
+docker push ghcr.io/espressosystems/hotshot/validator-webserver:main-async-std
+
+# ecs deploy
+ecs deploy --region us-east-2 hotshot hotshot_centralized -i centralized ghcr.io/espressosystems/hotshot/validator-webserver:main-async-std
+ecs deploy --region us-east-2 hotshot hotshot_centralized -c centralized ${orchestrator_url}
+
+# start these two dockers in another two servers and keep them running
+# enter the repo hotshot-nginx and switch to sishan/autobench
+# docker build . -f Dockerfile -t [YOUR-NAME]
+# docker run --network=host [YOUR-NAME]:latest
```
rob-maron, line 29: Looks good!

crates/examples/infra/mod.rs
```diff
@@ -188,6 +188,14 @@ pub fn read_orchestrator_init_config<TYPES: NodeType>() -> (
                 .help("Sets the url of the da webserver")
                 .required(false),
         )
+        .arg(
+            Arg::new("fixed_leader_for_gpuvid")
+                .short('f')
+                .long("fixed_leader_for_gpuvid")
+                .value_name("BOOL")
+                .help("Sets the number of fixed leader for gpu vid, only be used when leaders running on gpu")
```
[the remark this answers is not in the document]
dailinsubjam: This is only for GPU VID

crates/hotshot/src/traits/election/static_committee.rs
dailinsubjam: Yes! We're expecting the fixed leaders to be the first n nodes. Yes that's exactly my plan: we connect n validators to the orchestrator first so they'll get lower index, then start other validators. The feature [fixed-leader] is just to enable this function of choosing leaders from first n index-lower node.

crates/examples/infra/mod.rs
```diff
@@ -604,6 +622,7 @@ pub trait RunDA<
                             // this might be a obob
                             if let Some(leaf_info) = leaf_chain.first() {
                                 let leaf = &leaf_info.leaf;
+                                // use println for tmp debugging on Datadog
```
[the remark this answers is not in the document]
dailinsubjam: let me remove this comment

rob-maron:
Looks like there are merge conflicts, but I think we're good to go after 😃

rob-maron (approved):

jparr721 (approved):

PR #2859, opened 2024-03-28
[CX_CLEANUP] - Integrate QuorumProposalTask with other tasks

### LINKED ISSUE #2748: [CX_CLEANUP] - Integrate `QuorumProposalTask` with other tasks

- Spawn the quorum proposal task with other tasks.
- Remove the proposal-related code from the consensus task.

_No response_

Design doc: https://www.notion.so/espressosys/Refactor-Voting-Logic-9afa14dd027d453aa9b4a80a52de5e44.

Existing tests still pass.

_No response_

jparr721:
Closes #2748 

### This PR: 
- Spawns the `QuorumProposalTask` with the other tasks
- Validates dummy quorum proposals

### This PR does not: 

- Remove existing functionality from the `ConsensusTask` (yet)
To expand upon the above point, this code sends a `DummyQuorumProposalSend` event, utilizing the duplication of events between `ConsensusTask` and `QuorumProposalTask`. This is done because decomposing `QuorumProposalRecv` is an extreme undertaking, and would likely result in a >1000 lines of code changed PR, not to mention the supreme headache of checking the behavior.
    - This change instead opts to let `ConsensusTask` maintain ownership of the state transition and update code, allowing for *all* other pieces to be changed within `QuorumProposalTask`, limiting the scope of the change.

The flow basically lets `ConsensusTask` validate the `QuorumProposalRecv` event and, as a result, the `QuorumProposalTask` handles the `QuorumProposalValidated` event to initiate a proposal. The other expected operations take place when proposing like checking parent commitments, updating the `Consensus` state object, etc.

### Key places to review: 

 
All implemented logic for handling the events within `QuorumProposalTask`.

#### Specific View Number Handling:
- QCFormed (Timeout): View + 1
- QCFormed (Quorum Certificate): View + 1
- QuorumProposalValidated: View + 1
- ViewSyncFinalizeCertificate2Recv: View
- SendPayloadCommitmentAndMetadata: View
- UpgradeCertificateFormed: View


crates/task-impls/src/quorum_proposal.rs
```diff
@@ -80,32 +66,184 @@ struct ProposalDependencyHandle<TYPES: NodeType> {
...
+        };
+
+        let mut new_leaf = Leaf::from_quorum_proposal(&proposal);
+        new_leaf.set_parent_commitment(parent_leaf.commit());
+
+        let Ok(signature) =
+            TYPES::SignatureKey::sign(&self.private_key, new_leaf.commit().as_ref())
+        else {
+            bail!("Failed to sign leaf.commit()!");
+        };
```
ss-es, line 214: I think this could be `let signature = TYPES::SignatureKey::sign(&self.private_key, new_leaf.commit().as_ref()).context("Failed to sign leaf.commit()!")?;`

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -1,5 +1,7 @@
-use std::{collections::HashMap, sync::Arc};
+use commit::Committable;
+use std::{collections::HashMap, marker::PhantomData, sync::Arc};
 
+use anyhow::{bail, Result};
```
ss-es, line 4: +1 very big fan of this.

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -80,32 +66,184 @@ struct ProposalDependencyHandle<TYPES: NodeType> {
...
+    fn check_leader(&self, view: TYPES::Time) -> Result<()> {
+        if self.quorum_membership.get_leader(view) != self.public_key {
+            // This is expected for view 1, so skipping the logging.
+            if view != TYPES::Time::new(1) {
+                bail!(
+                    "Somehow we formed a QC but are not the leader for the next view {:?}",
+                    view
+                );
+            }
+        }
```
ss-es, line 96: I think this could be `ensure!(self.quorum_membership.get_leader(view) != self.public_key, "Somehow...", view);`.

You could `&& view != TYPES::Time::new(1)`, but I'm ambivalent. At least personally, I don't care if view 1 has unnecessary logs, it's just a single view and should be fixed relatively soon: #2873).

crates/task-impls/src/quorum_proposal.rs
ss-es, line 4: +1, very big fan

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -80,32 +66,184 @@ struct ProposalDependencyHandle<TYPES: NodeType> {
...
+    /// Find the leaf and state of the parent view to the view that we're currently proposing on.
+    fn find_parent_leaf_and_state(
+        consensus: &Consensus<TYPES>,
+    ) -> Result<(Leaf<TYPES>, Arc<TYPES::ValidatedState>)> {
+        let parent_view_number = &consensus.high_qc.get_view_number();
+
+        let Some(parent_view) = consensus.validated_state_map.get(parent_view_number) else {
+            // This should have been added by the replica?
+            bail!("Couldn't find parent view in state map, waiting for replica to see proposal\n parent view number: {}", **parent_view_number);
+        };
```
ss-es, line 109: could be `let parent_view = consensus.(...).context(format!("Couldn't find..", parent_view_number))?;`

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -80,32 +66,184 @@ struct ProposalDependencyHandle<TYPES: NodeType> {
...
+
+        // Leaf hash in view inner does not match high qc hash - Why?
+        let Some((leaf_commitment, state)) = parent_view.get_leaf_and_state() else {
+            bail!(
+                "Parent of high QC points to a view without a proposal; parent_view_number: {}, parent_view: {:?}",
+                **parent_view_number,
+                parent_view,
+            );
+        };
```
ss-es, line 118: Same as previous line.

Actually you could probably even convert `get_leaf_and_state()` and the other functions to return a `anyhow::Result`, and we can just add the error message there, and just call with `?` here. But we could do the cleanup separately if it takes more than 10 seconds

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -80,32 +66,184 @@ struct ProposalDependencyHandle<TYPES: NodeType> {
...
+
+        if leaf_commitment != consensus.high_qc.get_data().leaf_commit {
+            // NOTE: This happens on the genesis block
+            debug!(
+                "Parent leaf commitment does not equal high qc leaf commitment: {:?} != {:?}",
+                leaf_commitment,
+                consensus.high_qc.get_data().leaf_commit
+            );
+        }
```
ss-es, line 127: I think this could be an `ensure`

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -80,32 +66,184 @@ struct ProposalDependencyHandle<TYPES: NodeType> {
...
+        let Some(leaf) = consensus.saved_leaves.get(&leaf_commitment) else {
+            bail!("Failed to find high QC of parent.");
+        };
```
ss-es, line 130: I think this could be a `let leaf = ...context(..)?`

ss-es (commented):
Just some random comments about anyhow while I was looking at it.

These are really minor and feel free to ignore. I'll look at the logic some more when I get a chance!

crates/task-impls/src/consensus.rs
```diff
@@ -1485,6 +1496,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
                     ""
                 );
 
+                #[cfg(not(feature = "proposal-task"))]
```
shenkeyao, line 1499: Nit: Maybe it's just a temporary thing for debugging, but I don't think this and the same thing on line 1573 are needed, since this function already has this feature flag.

jparr721: It is a temporary debugging thing. This specific call not being present causes everything to break.

PR #2864, opened 2024-03-29
[Combined Network] Fix examples

rob-maron:
### This PR: 

- Changes the combined network's `all.rs` to launch the Push CDN
- Makes some `run-config` changes to better support the Push CDN
- Fixes some non-hotshot-testing lints

### This PR does not: 

Make any changes to the network or HotShot.

### Key places to review: 

`all.rs`

### DIFF
--- a/crates/orchestrator/run-config.toml
+++ b/crates/orchestrator/run-config.toml
@@ -66,7 +66,7 @@ mesh_n_low = 4
 mesh_outbound_min = 2
 mesh_n = 4
 online_time = 10
-base_port = 9000
+base_port = 8000
 
 [web_server_config]
 url = "http://localhost:9000"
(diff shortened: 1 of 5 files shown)


crates/orchestrator/run-config.toml
```diff
@@ -66,14 +66,17 @@ mesh_n_low = 4
 mesh_outbound_min = 2
 mesh_n = 4
 online_time = 10
-base_port = 9000
+base_port = 8000
```
jparr721, line 69: Curious, why did this change?

jparr721 (approved):
I always appreciate how extensively you comment your code. Great work.

crates/orchestrator/run-config.toml
rob-maron: The marshal port in one of the examples is `9000`, so this was giving me problems. I changed the `all.rs` example to acquire it from the config file, which is better. So I don't think we really needed this change, but it doesn't break anything

PR #2866, opened 2024-03-29
Do not send VID Dispersal through the fallback network

lukaszrzasik:
### This PR: 

* Disables sending VID dispersal via fallback network in combined network implementation.
* Makes VID dispersal proposals signed by one of the DA committee nodes valid.
* Fixes minor bugs in the request/response implemantation.

### Key places to review: 

crates/hotshot/src/traits/networking/combined_network.rs
crates/task-impls/src/consensus.rs

### How to test this PR:

All the tests should pass

### DIFF
--- a/crates/task-impls/src/consensus.rs
+++ b/crates/task-impls/src/consensus.rs
@@ -18,7 +18,7 @@ use hotshot_types::{
     event::{Event, EventType},
     message::{GeneralConsensusMessage, Proposal},
     simple_certificate::{QuorumCertificate, TimeoutCertificate, UpgradeCertificate},
-    simple_vote::{QuorumData, QuorumVote, TimeoutData, TimeoutVote},
+    simple_vote::{QuorumData, QuorumVote, TimeoutData, TimeoutVote, UpgradeProposalData},
     traits::{
         block_contents::BlockHeader,
         consensus_api::ConsensusApi,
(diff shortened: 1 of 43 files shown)


crates/task-impls/src/consensus.rs
```diff
@@ -471,6 +471,34 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
+            .validate(&disperse.signature, payload_commitment.as_ref())
+        {
+            return true;
+        }
+        // or the data was signed by one of the staked DA committee members
+        for da_member in self.committee_membership.get_staked_committee(view) {
+            if da_member.validate(&disperse.signature, payload_commitment.as_ref()) {
+                return true;
+            }
+        }
```
bfish713, line 498: Do the DA members actually sign the disperse? I thought were were passing along the leaders signature, but that might have changed when you split it up?

crates/task-impls/src/request.rs
```diff
@@ -163,7 +168,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, Ver: StaticVersionType + 'st
             delay: self.delay,
             recipients,
         };
-        let Ok(data) = Serializer::<Ver>::serialize(&request) else {
+        let Ok(data) = bincode_opts().serialize(&request) else {
```
bfish713, line 171: Hmm this should work with versioned Serializer.  Maybe the deserailize side needs to change instead?

bfish713 (dismissed):

PR #2867, opened 2024-03-29
[Sequencer] Prepare for update

rob-maron:
Needed to make one really tiny change to better support the sequencer. It looks like we changed `Leaf` to have private fields with getters and setters, but the sequencer likes to mutate fields during tests and the like. 

This PR just makes the change to add the ability to get a mutable reference to the header. I considered a setter but this seems more ergonomic.

The other change is unrelated, but removes an unnecessary clone.


ss-es (approved):
Looks good to me if this is enough for the sequencer. I think even `set_block_header` would be fine if they need that -- I only made the fields private to make it harder to construct invalid leaves accidentally.

bfish713 (approved):

PR #2870, opened 2024-03-29
Fix block builder test

ss-es:
No linked issue.

### This PR: 
The block builder test is very consistently failing in CI because of a timing issue. I believe this should fix it.


jparr721 (approved):
🙏🏽

PR #2873, opened 2024-03-30
Remove special treatment of view 1

ss-es:
### This PR: 
Removes special treatment of view 1 in the consensus task.

The root cause behind this special treatment appears to be a mistake in the relationship between `Leaf::genesis` and `QuorumCertificate::genesis`. 

With our current consensus logic, `QuorumCertificate::genesis` must depend on `Leaf::genesis`, because a quorum certificate's parent commitment must be calculated from its proposal's associated leaf. However, this relationship seems to have been inverted, with the genesis leaf being constructed from the genesis quorum certificate.

More precisely:

- `QuorumCertificate::genesis` had a parent commit that was identically 0,
- which made it impossible to construct a valid `Leaf` which could've formed the basis of this commitment,
- which made it impossible to initialize HotShot with a valid parent `Leaf` in storage for view 1,
- which forced us to special-case view 1.

This ended up having some side-effects, where we were mixing data in validating proposals. I believe this was likely benign, but these changes should still make it easier to reason about correctness.

### This PR does not: 
Remove or radically alter the concept of `Leaf`. This should, however, get us closer to making that change.


ss-es:
Pushing some additional (incomplete) work to this branch, just to record progress. 

Commits after this comment (past 1512a7b) are an experiment (that doesn't seem like it will work).

jparr721 (approved):

bfish713 (approved):

PR #2875, opened 2024-04-01
[Sequencer] Libp2p config changes

### LINKED ISSUE #2877: [Sequencer] Hotshot changes for new networks

We want to implement the combined networks into the sequencer. This ticket is to make the HotShot changes necessary to enable that.

- [ ] HotShot is ready for the sequencer

_No response_

rob-maron:
Closes #2877

# This PR:

TL;DR: Gets the sequencer ready to use our Libp2p networks.

Specifically:
1. Moves the Libp2p `from_config` function out of testing to `libp2p_network.rs`  
2. Adds key derivation functions for Libp2p keypairs that we use to generate a secondary key from a given `SignatureKey` (in this case, our consensus key). Previously we were generating this from a known seed, which is insecure
3. Changes bootstrap behavior. Instead of choosing the first n nodes to be bootstrap, all nodes are "bootstrap" nodes, but we choose `mesh_n` random nodes to connect to on startup.
4. Add the ability to push both your Libp2p public key and address to the orchestrator
5. Some cleanup - types, remove duplicate code from examples

# This PR does not:
- Change anything else about the orchestrator

# Key places for review:
- `libp2p_network.rs` - The `from_config` function


rob-maron:
@mrain checking with you on using Blake3 as a KDF - is this alright?

bfish713 (approved):

crates/hotshot/src/traits/networking/libp2p_network.rs
```diff
@@ -298,7 +305,125 @@ where
...
+/// bytes.
+pub fn derive_libp2p_keypair<K: SignatureKey>(
+    private_key: &K::PrivateKey,
+) -> anyhow::Result<Keypair> {
+    // Derive a secondary key from our primary private key
+    let derived_key = blake3::derive_key("libp2p key", &(bincode::serialize(&private_key)?));
+    let derived_key = SecretKey::try_from_bytes(derived_key)?;
+
+    // Create an `ed25519` keypair from the derived key
+    Ok(ed25519::Keypair::from(derived_key).into())
```
dailinsubjam, line 321: General question: why this secondary key pair is different from the SignatureKey scheme we used before? Do we want to add this as a new scheme for `SignatureKey`?

PR #2879, opened 2024-04-02
Allow unordered test events

ss-es:
### This PR: 
Allows for running multiple events simultaneously in `run_test_script`.

### This PR does not: 
The `test_scripts!` macro is untouched. It currently does not handle subtasks well, and makes no attempt to handle the newer predicates (an `all` predicate should result in undefined behaviour).


crates/testing/src/predicates.rs
```diff
@@ -1,21 +1,63 @@
...
+#[derive(Eq, PartialEq, Copy, Clone)]
+pub enum PredicateResult {
+    Pass,
+
+    Fail,
+
+    Incomplete,
+}
+
+impl PredicateResult {
```
jparr721, line 23: Could this just be `impl From<bool> for PredicateResult`?

jparr721 (approved):

PR #2880, opened 2024-04-02
[Deps] Remove nonbreaking pins

rob-maron:
Should make upstream crates (like the sequencer) easier to update in the future. E.g. if there are incompatible patch dependencies (e.g. syn ==2.0.53 vs syn ^2.0.10 is currently blocking the sequencer update) this would solve that by allowing us to have a version that satisfies both.

This doesn't change the version of anything, everything is still the same in the `Cargo.lock`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,16 +35,16 @@ ark-ed-on-bn254 = "0.4"
 ark-ff = "0.4"
 ark-serialize = "0.4"
 ark-std = { version = "0.4", default-features = false }
-async-broadcast = "0.7.0"
+async-broadcast = "0.7"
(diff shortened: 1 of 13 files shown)


bfish713 (approved):

rob-maron:
@jbearer - since you're most familiar with the update process - does this make sense to do, or it would it be more beneficial if I pinned the minor versions (e.g. syn = "2.0") instead of the allegedly non-breaking ones?

ss-es (approved):
Not that familiar with rust versioning, but looks fine to me (at least with my understanding of crates.io and semver). And we commit `Cargo.lock` anyway, so I don't think there's too much to worry about.

PR #2885, opened 2024-04-02
[Quick] Gate round_start_delay sleep behind #[cfg(test)]

ss-es:
### This PR: 
Gates usage of `round_start_delay` behind `#[cfg(test)]`.

### This PR does not: 
There wasn't a clear path to removing it entirely without breaking tests or adding a feature flag. This should just prevent us from inadvertently having consensus sleep in production.


ss-es:
It looks like #[cfg(test)] doesn't work the way I thought it did. Closing this for now.

PR #2892, opened 2024-04-03
remove as_ref on commitment

akonring:
Referece EspressoSystems/jellyfish#471


jparr721 (approved):

ggutoski:
Did you catch all occurrences of `.as_ref().as_ref()`? If so then we should merge this PR.

PR #2893, opened 2024-04-03
[CHORE] Update tide-disco (and others) to crates.io

jbearer:

PR #2904, opened 2024-04-04
Update the CDN to `0.1.23`

rob-maron:
Adds the feature flag `global-permits`, which allows permits to be issued for all brokers at the same time. Makes deployment easier.


ss-es (approved):

ss-es:
Should probably wait until Jeb merges #2893, I feel like `Cargo.lock` changes usually cause merge conflicts.

PR #2917, opened 2024-04-07
Add Nix shell without rust-analyzer

ss-es:
### This PR: 
This is probably just a personal pain point, but I find myself often waiting a very long time on `rust-analyzer` (which I don't use) to compile after a `flake.lock` update (which happens weekly).

This PR adds a shell without `rust-analyzer`, and uncommits `.envrc`. 

For `.envrc`: given that we a) use Nix and b) don't do anything interesting in `.envrc`, I don't think committing the latter makes sense -- it effectively just prevents you from creating your own `.envrc` to choose a non-default shell. 

Note also that our `.envrc` is very straightforward to recreate -- e.g. 1 line: `use flake .#simpleShell` -- and should only have to be done once per repository clone (and only if you use `direnv`).

### Key places to review:
Would this affect your own workflow?

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,5 +1,0 @@
-# reload when these files change
-watch_file flake.nix
-watch_file flake.lock
-# load the flake devShell
-use nix
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.envrc
 **/target
 **/result
 **/out*.txt
--- a/flake.nix
+++ b/flake.nix
@@ -177,7 +177,6 @@
             python3
             zlib.dev
             zlib.out
-            fenix.packages.${system}.rust-analyzer
             just
             pkg-config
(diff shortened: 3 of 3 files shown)


QuentinI:
Not sure if it's relevant to anyone's workflow, but I think it's worth noticing that since this PR removes RA from `buildDeps`, it makes _all_ shells except the default one RA-less.

ss-es:
> Not sure if it's relevant to anyone's workflow, but I think it's worth noticing that since this PR removes RA from `buildDeps`, it makes _all_ shells except the default one RA-less.
> 
> Also, for all direnv+nix users I can't recommend [nix-direnv](https://github.com/nix-community/nix-direnv) enough. It caches nix shell, making subsequent runs of direnv with unchanged flake way faster.

Updated!

PR #2922, opened 2024-04-08
[Builder] Vid Integration

move47:
- WIP towards interring VID calculation done by builder. i.e builder fees.

### DIFF
--- a/crates/builder-api/src/block_info.rs
+++ b/crates/builder-api/src/block_info.rs
@@ -1,4 +1,4 @@
-use std::{hash::Hash, marker::PhantomData, thread::Builder};
+use std::{hash::Hash, marker::PhantomData};
 
 use hotshot_types::{
     traits::{node_implementation::NodeType, signature_key::BuilderSignatureKey, BlockPayload},
--- a/crates/builder-api/src/block_info.rs
+++ b/crates/builder-api/src/block_info.rs
@@ -3,7 +3,7 @@ use std::{hash::Hash, marker::PhantomData};
 use hotshot_types::{
     traits::{node_implementation::NodeType, signature_key::BuilderSignatureKey, BlockPayload},
     utils::BuilderCommitment,
-    vid::VidCommitment,
+    vid::{VidCommitment, VidPrecomputeData},
 };
 use serde::{Deserialize, Serialize};
 
@@ -32,7 +32,13 @@ pub struct AvailableBlockData<I: NodeType> {
 #[serde(bound = "")]
 pub struct AvailableBlockHeaderInput<I: NodeType> {
     pub vid_commitment: VidCommitment,
-    pub signature: <<I as NodeType>::BuilderSignatureKey as BuilderSignatureKey>::BuilderSignature,
+    pub vid_precompute_data: VidPrecomputeData,
+    // signature over vid_commitment, BlockPayload::Metadata, and offered_fee
+    pub fee_signature:
+        <<I as NodeType>::BuilderSignatureKey as BuilderSignatureKey>::BuilderSignature,
+    // signature over the current response
+    pub message_signature:
+        <<I as NodeType>::BuilderSignatureKey as BuilderSignatureKey>::BuilderSignature,
     pub sender: <I as NodeType>::BuilderSignatureKey,
     pub _phantom: PhantomData<I>,
 }
--- a/crates/builder-api/src/block_info.rs
+++ b/crates/builder-api/src/block_info.rs
@@ -1,7 +1,7 @@
-use std::{hash::Hash, marker::PhantomData};
+use std::{hash::Hash, marker::PhantomData, thread::Builder};
 
 use hotshot_types::{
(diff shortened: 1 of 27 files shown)


crates/types/src/traits/signature_key.rs
```diff
@@ -139,3 +139,53 @@ pub trait SignatureKey:
...
+    type SignError: std::error::Error + Send + Sync;
+
+    /// validate the message with the builder's public key
+    fn validate_builder_signature(&self, signature: &Self::BuilderSignature, data: &[u8]) -> bool;
+
+    /// sign the message with the builder's private key
+    fn sign_builder_message(
+        private_key: &Self::BuilderPrivateKey,
+        data: &[u8],
+    ) -> Result<Self::BuilderSignature, Self::SignError>;
```
ss-es, line 190: I don't know enough to review the full PR (though it seems fine to me), but since I was thinking about this in other places recently: could this take `self` rather than `private_key` as an argument? 

As it is currently, you wouldn't be able to use `sign_builder_message`  without having direct access to the private key, which forces any struct that implements `BuilderSignatureKey` to expose the private key. But for security, the private key should probably be a private field.

I think this would also mean `BuilderPrivateKey` doesn't need to be an associated type.

(Unfortunately HotShot already propagates private keys unnecessarily as it is, so if it's not an easy thing to change here this is probably fine relative to that)

crates/types/src/traits/signature_key.rs
```diff
@@ -139,3 +139,53 @@ pub trait SignatureKey:
     #[must_use]
     fn genesis_proposer_pk() -> Self;
 }
+
+pub trait BuilderSignatureKey:
```
bfish713, line 143: I think we can just use the `SignatureKey` trait for the builder key and don't need a whole new trait.  Is there any reason to create an entirely new trait?

bfish713 (changes requested):

crates/types/src/traits/signature_key.rs
move47: https://www.notion.so/espressosys/Builder-fee-payments-4c27ef6c401544359df3ad8a7c59ef08?pvs=4#8ae01e2fe48443dfa8d3d77c82047ffa

crates/types/src/traits/signature_key.rs
move47: what about having &self, and have a different `fn get_privatekey(&self)->&Self::BuilderPrivateKey`,
```rust
fn sign_builder_message(&self, data: &[u8])-> Result<....>{
private_key = self.get_private_key();
...
..
}
```

jbearer: IMO if you have an interface to produce signatures you already have full control of the private key for all practical purposes, whether it's some wrapper struct or not

crates/task-impls/src/consensus.rs
```diff
@@ -1446,12 +1446,15 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
                 };
 
+                // TODO: Put the signature received from the builder in the block header
                 let block_header = TYPES::BlockHeader::new(
                     state,
                     &consensus.instance_state,
                     &parent_leaf,
                     null_block_commitment,
                     metadata,
+                    0,
```
jparr721, line 1456: Could we use a named constant instead of a magic number?

move47: https://github.com/EspressoSystems/HotShot/pull/2922/commits/deb3112ba4d674b16032da831337fcbcf8f2aac7

bfish713 (approved):

crates/types/src/traits/signature_key.rs
ss-es: My concern is about inadvertently misusing/leaking it rather than maliciously signing with it. For example, `BuilderPrivateKey` has a `Serialize` bound, which means it's possible to write into an array and send it out.

It's not likely we'll ever make this specific mistake, but making the field private lets you rule out a lot of misuses at the source, and it's much easier to audit a limited trait interface than it is to make sure every use of an exposed private key in the entire codebase is correct.

crates/builder-api/src/block_info.rs
```diff
@@ -13,8 +13,8 @@ pub struct AvailableBlockInfo<I: NodeType> {
     pub block_hash: BuilderCommitment,
     pub block_size: u64,
     pub offered_fee: u64,
-    pub signature: <<I as NodeType>::SignatureKey as SignatureKey>::PureAssembledSignatureType,
-    pub sender: <I as NodeType>::SignatureKey,
+    pub signature: <<I as NodeType>::BuilderSignatureKey as BuilderSignatureKey>::BuilderSignature,
```
tbro, line 16: I don't think you need the fully qualified syntax for `NodeType` b/c it is a bound of `I`.

move47:
It currently overlaps with some work done in #2948. Need to update it based on the changes in it.

move47:
Closing as part of it #2956 #2948 takes some work. The vid integration should be done in separate PR.

PR #2927, opened 2024-04-09
[CX_CLEANUP] - Integrate QuorumProposalTask with other tasks

### LINKED ISSUE #2748: [CX_CLEANUP] - Integrate `QuorumProposalTask` with other tasks

jparr721:
Closes #2748 

### This PR: 
- Spawns the `QuorumProposalTask` with the other tasks.
- Feature locks the quorum proposal task utilities.
- Cleans up the older functionality.
- Fixes some testing.
- Validates ONLY for the success case tests (this must be verified locally for now).

### This PR does not: 

- Remove existing functionality from the `ConsensusTask`.
- Integrate Quorum Proposal Task into the proposing pipeline.

### Key places to review: 

 
All implemented logic for handling the events within `QuorumProposalTask`.

#### Specific View Number Handling:
- QCFormed (Timeout): View + 1
- QCFormed (Quorum Certificate): View + 1
- QuorumProposalValidated: View + 1
- ViewSyncFinalizeCertificate2Recv: View
- SendPayloadCommitmentAndMetadata: View

This PR *only* considers the success case tests and explicitly does not handle anything else. This test should pass locally. Sometimes the web test is flaky, but the others should be consistent.
`cargo test -p hotshot-testing --test tests_1 test_success --features proposal-task -- --test-threads=1`


crates/task-impls/src/consensus.rs
```diff
@@ -243,9 +242,6 @@ pub struct ConsensusTaskState<
     /// Consensus api
     pub api: A,
 
-    /// needed to typecheck
-    pub _pd: PhantomData<I>,
```
jparr721, line 247: We use the I type now, so this wasn't needed anymore.

crates/task-impls/src/consensus.rs
```diff
@@ -870,24 +856,25 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
                     ));
             }
             HotShotEvent::QuorumProposalValidated(proposal) => {
-                let consensus = self.consensus.upgradable_read().await;
+                let consensus_reader = self.consensus.upgradable_read().await;
+                let mut consensus_writer = self.consensus.write().await;
```
ss-es, line 860: We'll always have to wait for the write lock anyway, couldn't we just take the single write lock for the entire thing?

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -80,8 +71,160 @@ struct ProposalDependencyHandle<TYPES: NodeType> {
...
+                if next_parent_leaf.get_view_number() <= consensus.last_decided_view {
+                    break;
+                }
+                next_parent_hash = next_parent_leaf.get_parent_commitment();
+            }
+            debug!("updated saved leaves");
+            // TODO do some sort of sanity check on the view number that it matches decided
+        }
+
+        // Special case: if we have a decided upgrade certificate AND it does not apply a version to the current view, we MUST propose with a null block.
```
ss-es, line 181: Is this a TODO?

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -80,8 +71,160 @@ struct ProposalDependencyHandle<TYPES: NodeType> {
...
+
+        let consensus = self.consensus.read().await;
+        let parent_view_number = &consensus.high_qc.get_view_number();
+        let mut reached_decided = false;
+
+        let Some(parent_view) = consensus.validated_state_map.get(parent_view_number) else {
+            // This should have been added by the replica?
+            error!("Couldn't find parent view in state map, waiting for replica to see proposal\n parent view number: {}", **parent_view_number);
+            return false;
+        };
```
ss-es, line 133: Does this mean the quorum proposal task will not propose on `QCFormed` if it missed the parent? 

I believe this is different from what consensus currently does, where we currently can propose off a formed QC even if we never saw the parent.

(I'm not sure exactly what we're planning to have handled by which task, just pointing out that I think this is different)

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -80,8 +71,160 @@ struct ProposalDependencyHandle<TYPES: NodeType> {
...
+
+impl<TYPES: NodeType> ProposalDependencyHandle<TYPES> {
+    /// Sends a proposal if possible from the high qc we have
+    #[allow(clippy::too_many_lines)]
+    pub async fn publish_proposal_if_able(
+        &self,
+        view: TYPES::Time,
+        event_stream: &Sender<Arc<HotShotEvent<TYPES>>>,
+        commit_and_metadata: CommitmentAndMetadata<TYPES::BlockPayload>,
+    ) -> bool {
```
ss-es, line 113: If we're making any changes while copying over consensus logic, I think it would be really worthwhile to make this return `anyhow::Result<()>`.

But if we're just trying to copy the logic as-is, it's probably better to change it in a separate PR

ss-es (commented):
I don't look at it very thoroughly, but the changes outside `quorum_proposal.rs` seem fine to me

crates/task-impls/src/quorum_proposal.rs
jparr721: Copied from consensus task, so probably yes but I think it might be an artifact from before.

crates/testing/src/predicates.rs
```diff
@@ -254,14 +255,18 @@ pub fn consensus_predicate(
 
 pub fn no_decided_upgrade_cert() -> Predicate<ConsensusTaskTestState> {
     consensus_predicate(
-        Box::new(|state| state.decided_upgrade_cert.is_none()),
+        Box::new(|state| {
+            async_block_on(async { state.consensus.read().await.decided_upgrade_cert.is_none() })
+        }),
```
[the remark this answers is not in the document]
jparr721: Yeah mine are also failing, which sucks because it hard blocks the utilization of this.

crates/task-impls/src/consensus.rs
jparr721: This one's tricky because of the mutability of the borrow for the callback. The `visit_leaf_ancestors` method requires an immutable borrow. I was thinking we could maybe change the signature, but it seems like that creates the need to change a couple spots (and add some more contention for write locks). But I'd be open to ideas.

ss-es: Oh, I see! That makes sense, I think this is fine as-is then

ss-es: Actually, looking at this again -- I think this is what causes the deadlock you were getting (the `async_block_on` in the predicates is probably fine, or at the very least not the sole issue). 

The write lock can't be taken until the read lock is dropped, so this line will just await forever. I think the only way to actually do this is to create a temporary `let mut decided_upgrade_cert = None`, then conditionally update `consensus.decided_upgrade_cert` right after the read lock gets upgraded

crates/task-impls/src/quorum_proposal.rs
jparr721: Marking this for next PR. I definitely plan to re-add the anyhow logic from my last change. I was hoping to avoid making this PR more of a headache to review than it already is.

shenkeyao (approved):

crates/task-impls/src/quorum_proposal.rs
shenkeyao: We can probably get rid of this TODO since the dependency task makes sure the view number always matches.

PR #2934, opened 2024-04-10
Only Append VID and Payload once Right Before Voting

bfish713:
### This PR: 

- Moves the calls to storage of VID info until just before we send our vote
- Fixes a bug in the webserver which causes VID shares to be only shared for one key
- Checks if the VID share we are receiving is for our key, both on the receival event and before voting
- Passes the correct vid share into the task in the unit tests

The goal of this PR is to not call the sequencer storage for the same thing multiple times, and to all ask to store things we are sure relate to a valid proposal.  This means we only store the data if we we have seen and voted on a valid proposal for the view of this data.

### Key places to review: 

The places where we moved `append_vid`


crates/hotshot/src/traits/networking/web_server_network.rs
```diff
@@ -911,9 +927,11 @@ impl<TYPES: NodeType + 'static, NetworkVersion: StaticVersionType + 'static>
...
 
                 // Cancel old, stale tasks
-                task_map
-                    .prune_tasks(view_number, ConsensusIntentEvent::CancelPollForVIDDisperse)
-                    .await;
+                if view_number > 2 {
+                    task_map
+                        .prune_tasks(view_number, ConsensusIntentEvent::CancelPollForVIDDisperse)
+                        .await;
+                }
```
ss-es, line 934: Why only for view_number > 2?

(or wait -- is this a web server thing? if so nevermind)

crates/task-impls/src/quorum_vote.rs
```diff
@@ -628,7 +628,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumVoteTaskState<TYPES, I
                     .vid_shares
```
ss-es, line 628: should this `storage.write()` on 621 also be moved right before the quorum vote task's QuorumVoteSend?

ss-es (dismissed):
`append_da` seems to not have changed, and I think the `append_vid` move makes sense to me

crates/task-impls/src/quorum_vote.rs
bfish713: Yes, good catch.  I didn't pay enough attention to the new task, I'll update with this change

crates/hotshot/src/traits/networking/web_server_network.rs
bfish713: Yeah it's some webserver weirdness, we can get by without it probably but for now I'll keep it in

bfish713:
> `append_da` seems to not have changed, and I think the `append_vid` move makes sense to me

I had originally done this for append_da but we need to make sure it's stored before we even send the DA vote so it needs to stay put.  Forgot to update the PR description

crates/task-impls/src/consensus.rs
```diff
@@ -1223,30 +1235,27 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
...
                     .await
                     .vid_shares
                     .entry(view)
                     .or_default()
-                    .insert(disperse.data.recipient_key.clone(), disperse.data.clone());
-
+                    .insert(disperse.data.recipient_key.clone(), disperse.clone());
+                if disperse.data.recipient_key != self.public_key {
+                    self.quorum_network
+                        .inject_consensus_info(ConsensusIntentEvent::PollForVIDDisperse(
```
shenkeyao, line 1247: Why do we need to cancel the poll if it's not canceled before?

crates/task-impls/src/quorum_vote.rs
shenkeyao: Can we update the changes to the `Poll` and `CancelPoll` as well?

crates/task-impls/src/consensus.rs
bfish713: Ah we don't I added this when debugging because I thought polling was cancelled after each VID share was received

shenkeyao (approved):

PR #2935, opened 2024-04-10
Add (integrated) builder support to examples

### LINKED ISSUE #2919: [Builder] - Update examples with simple builder

After getting builders merged in, we'll need to update examples so that they can run with simple builder.

_No response_

Related PR: https://github.com/EspressoSystems/HotShot/pull/2837

We should be able to run benchmarks with simple builders.
A testing command: `just async_std example all-webserver -- --config_file ./crates/orchestrator/run-config.toml --orchestrator_url http://localhost:4444`

_No response_

QuentinI:
Closes #2919

### This PR: 
Adds support for integrated/external builder to orchestrator.
- Setting `builder` in `run-config.toml` to `Random` or `Simple` will start corresponding builders alongside nodes and set `builder_url` accordingly, setting it to `External` will leave `builder_url` untouched, allowing one to run examples with external builder.
- Adds `random_builder` section to orchestrator config
- Adds `--builder` flag to examples

### This PR does not: 
Add any support for launching external builders in-network (setting num_nodes_without_stake, etc.).

### Key places to review: 
`crates/orchestrator/*`

### How to test this PR:
`just async_std example all-combined -- --config_file crates/orchestrator/run-config.toml --builder <external|random|simple>`


crates/examples/infra/mod.rs
```diff
@@ -190,6 +193,15 @@ pub fn read_orchestrator_init_config<TYPES: NodeType>() -> (
...
         )
+        .arg(
+            Arg::new("builder")
+                .short('b')
+                .long("builder")
+                .value_name("BUILDER_TYPE")
+                .value_parser(value_parser!(BuilderType))
+                .help("Sets type of builder. `simple` or `random` to run corresponding integrated builder, `external` to use the one specified by `[config.builder_url]` in config")
+                .required(false),
+        )
```
rob-maron, line 204: Can we perhaps add a default value (if there isn't one)?

crates/examples/infra/mod.rs
```diff
@@ -250,6 +262,9 @@ pub fn read_orchestrator_init_config<TYPES: NodeType>() -> (
         };
         config.da_web_server_config = Some(updated_da_web_server_config);
     }
+    if let Some(builder_type) = matches.get_one::<BuilderType>("builder") {
+        config.builder = *builder_type;
+    }
```
rob-maron, line 267: Nice 🚀

crates/examples/infra/mod.rs
```diff
@@ -1027,10 +1042,42 @@ pub async fn main_entry_point<
...
+            builder_task
+        }
+    };
+
     info!("Initializing networking");
     let run = RUNDA::initialize_networking(run_config.clone(), args.advertise_address).await;
     let hotshot = run.initialize_state_and_hotshot().await;
 
+    if let Some(task) = builder_task {
+        task.start(Box::new(hotshot.get_event_stream()));
```
rob-maron, line 1078: Does this run a builder for each node? If so, I think instead we should consider a separate (simple) binary that just runs the simple builder. Then in `all.rs` we could add a call to start the simple builder, and when we have large, multi-node runs, we can run the binary

crates/examples/infra/mod.rs
QuentinI: There is, clap respects `Default` impl for `BuilderType` (which is `Simple`)

crates/examples/infra/mod.rs
QuentinI: Yeah, it runs a builder for each node.
Running a single builder per network would probably make more sense, honestly. WDYT about adding a cmdline/config option for amount of builders to run and defaulting it to 1?
We need to start the simple builder together with a hotshot node, because it (for now?) relies on hs event stream to gather transactions.

rob-maron: Ah right, yes this makes sense to do but only if it's not too much of a change. It shouldn't matter _too much_ because we are generally doing small runs with `all.rs`.

For benchmarking runs, we will need a builder binary, but assuming that's out of scope for this one

QuentinI: > Ah right, yes this makes sense to do but only if it's not too much of a change. It shouldn't matter too much because we are generally doing small runs with all.rs.

Yeah, it shouldn't be too much work, I'll add it today.

> For benchmarking runs, we will need a builder binary, but assuming that's out of scope for this one

Any explicit support for external builder binary is out of scope for this, I'm going to discuss how best to do it with builder team after this one.

QuentinI: Self-quoting over from Zulip
> Meanwhile, I realised that properly supporting running arbitrary number of integrated builders with examples is a bit tricky, since there needs to be some back-and-forth with orchestrator to properly assign builder addresses to nodes. Do you mind if we merge #2935 as-is for now, and I'll open a separate issue to allow running arbitrary number of integrated builders?

jparr721 (dismissed):
LGTM

rob-maron (dismissed):

PR #2940, opened 2024-04-10
Minor rustfmt fixes

ss-es:
No linked issue.

### This PR: 
- Correctly sets the rust edition in `rustfmt.toml`, so that a raw invocation of `rustfmt` (as opposed to `cargo fmt`) does not error.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Install Rust Stable
         uses: dtolnay/rust-toolchain@nightly
         with:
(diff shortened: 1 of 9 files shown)


bfish713:
One question about this, does this require us to use nightly to properly format and have PRs pass CI?  I think for those of us not using nix cargo fmt gives warnings because the two features we enable in rustformat.toml are nightly only

ss-es:
Sorry, took me a while to get back to this.

@bfish713 that was correct, though I just reverted that part of the PR. It should now just work as before.

jparr721 (approved):

PR #2945, opened 2024-04-11
Update Testing Framework to Support Async

jparr721:
Closes #<ISSUE_NUMBER>

### This PR: 
Adds the ability to use async within the testing environment for the purpose of unpacking async structures like RwLocks in a way that avoids deadlocking in the test threads. We are now able to define custom trait implementations with behavior that is able to be unique to the test case. This came as a response to have background async behavior without needing to argue with the borrow checker when trying to take ownership of data that we cannot (or should not) clone.

Specifically, this introduces the `Predicate` trait, which has the following structure:
```rust
#[async_trait]
pub trait Predicate<INPUT>: std::fmt::Debug {
    async fn evaluate(&self, input: &INPUT) -> PredicateResult;
    async fn info(&self) -> String;
}
```

This is very similar to the struct that it replaces, except we can now define this behavior over any callback. We may now define a predicate callback for a given input of any type, exactly as it worked before. An example implementation of this can be seen here:
```rust
type EventCallback<TYPES> = Arc<dyn Fn(Arc<HotShotEvent<TYPES>>) -> bool + Send + Sync>;

pub struct EventPredicate<TYPES>
where
    TYPES: NodeType + Send + Sync,
{
    check: EventCallback<TYPES>,
    info: String,
}

impl<TYPES: NodeType> std::fmt::Debug for EventPredicate<TYPES> {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        write!(f, "{}", self.info)
    }
}

#[async_trait]
impl<TYPES> Predicate<Arc<HotShotEvent<TYPES>>> for EventPredicate<TYPES>
where
    TYPES: NodeType + Send + Sync + 'static,
{
    async fn evaluate(&self, input: &Arc<HotShotEvent<TYPES>>) -> PredicateResult {
        PredicateResult::from((self.check)(input.clone()))
    }

    async fn info(&self) -> String {
        self.info.clone()
    }
}
```

Here, we get a pretty straightforward function, but this gives us an async context to work with, so if the `self.check` method needs to do something async, we now can accomplish this in a way that doesn't require arduous blocking or rewriting of the implementation.

Furthermore, I have upgraded all of the one-liner predicates to automatically build the values that they're looking for. Consider the `exact` predicate:
```rust
pub fn exact<TYPES>(event: HotShotEvent<TYPES>) -> Box<EventBooleanPredicate<TYPES>>
where
    TYPES: NodeType,
{
    let info = format!("{:?}", event);
    let event = Arc::new(event);

    let check: BooleanCallback<TYPES> = Arc::new(move |e: Arc<HotShotEvent<TYPES>>| {
        let event_clone = event.clone();
        *e == *event_clone
    });

    Box::new(EventBooleanPredicate { check, info })
}
```

This predicate is a good example of how you can easily build a callback function for any specific behavior that you'd like to implement. This background function, crucially, may also be an async callback (if you can get the data into it).

### DIFF
--- a/crates/macros/src/lib.rs
+++ b/crates/macros/src/lib.rs
@@ -364,7 +364,7 @@ pub fn test_scripts(input: proc_macro::TokenStream) -> TokenStream {
 
                         let assert = &mut output_asserts[#output_index_names];
 
-                        validate_output_or_panic_in_script(stage_number, #script_names.to_string(), &received_output, assert);
+                        validate_output_or_panic_in_script(stage_number, #script_names.to_string(), &received_output, &**assert).await;
 
                         #output_index_names += 1;
                     }
@@ -392,7 +392,7 @@ pub fn test_scripts(input: proc_macro::TokenStream) -> TokenStream {
 
                         let mut assert = &mut output_asserts[#output_index_names];
 
-                        validate_output_or_panic_in_script(stage_number, #script_names.to_string(), &received_output, assert);
+                        validate_output_or_panic_in_script(stage_number, #script_names.to_string(), &received_output, &**assert).await;
 
                         #output_index_names += 1;
                     }
@@ -410,7 +410,7 @@ pub fn test_scripts(input: proc_macro::TokenStream) -> TokenStream {
             let task_state_asserts = &mut #task_expectations[stage_number].task_state_asserts;
 
             for assert in task_state_asserts {
-                validate_task_state_or_panic_in_script(stage_number, #script_names.to_string(), #task_names.state(), assert);
+                validate_task_state_or_panic_in_script(stage_number, #script_names.to_string(), #task_names.state(), &**assert).await;
             }
         )*
     } }
--- a/crates/testing/tests/tests_1/consensus_task.rs
+++ b/crates/testing/tests/tests_1/consensus_task.rs
@@ -6,8 +6,7 @@ use hotshot_testing::task_helpers::key_pair_for_id;
 use hotshot_testing::test_helpers::permute_input_with_index_order;
 use hotshot_testing::{
     predicates::{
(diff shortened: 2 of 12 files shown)


crates/macros/src/lib.rs
```diff
@@ -364,7 +364,7 @@ pub fn test_scripts(input: proc_macro::TokenStream) -> TokenStream {
 
                         let assert = &mut output_asserts[#output_index_names];
 
-                        validate_output_or_panic_in_script(stage_number, #script_names.to_string(), &received_output, assert);
+                        validate_output_or_panic_in_script(stage_number, #script_names.to_string(), &received_output, &**assert).await;
```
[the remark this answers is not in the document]
shenkeyao: Nvm! Saw the comment in another file.

shenkeyao (approved):
Since this is all test code changes, I think we can merge it despite the soft code freeze.

crates/testing/tests/tests_1/consensus_task.rs
```diff
@@ -65,7 +64,7 @@ async fn test_consensus_task() {
             quorum_proposal_validated(),
             exact(QuorumVoteSend(votes[0].clone())),
         ],
-        asserts: vec![is_at_view_number(1)],
+        asserts: vec![],
```
ss-es, line 67: Were these asserts removed accidentally, or were they conceptually wrong?

jparr721: Yeah they honestly just didn't make sense to have since we're basically replicating what a view change is. I made this edit on day 2 here, so I had *fundamentally* no idea what was going on.

PR #2946, opened 2024-04-11
[TECH-DEBT] Have `BlockPayload::get_transactions` return an iterator

jbearer:
This means that the payload is no longer required to explicitly store a list of transactions as a field. We also default-impl a couple other trait functions using this iterator.


bfish713 (approved):

jbearer:
Failing tests seem to be failing on main as well

bfish713:
> Failing tests seem to be failing on main as well

Yeah examples are broken in CI due to builder changes, not related to this

bfish713 (approved):

PR #2956, opened 2024-04-12
Miscellaneous Builder changes

move47:
- Validates builder message signatures.
- Add the builder fee signatures.
- Removes the spawn blocking over vid calls. 
- Changes the first API call to be based on builder commitment instead of VID commitment.


crates/task-impls/src/helpers.rs
```diff
@@ -60,6 +60,29 @@ pub async fn calculate_vid_disperse<TYPES: NodeType>(
...
+/// # Panics
+/// Panics if the VID calculation fails, this should not happen.
+pub async fn calculate_vid_disperse_using_precompute_data<TYPES: NodeType>(
+    txns: Vec<u8>,
+    membership: Arc<TYPES::Membership>,
+    view: TYPES::Time,
+    pre_compute_data: VidPrecomputeData,
+) -> VidDisperse<TYPES> {
+    let num_nodes = membership.total_nodes();
+    let vid_disperse = spawn_blocking(move || {
```
jparr721, line 74: Can we just move the #[allow(panic)] to the top of the function?

crates/task-impls/src/helpers.rs
```diff
@@ -60,6 +60,29 @@ pub async fn calculate_vid_disperse<TYPES: NodeType>(
...
+        #[allow(clippy::panic)]
+        vid_scheme(num_nodes).disperse_precompute(&txns, &pre_compute_data).unwrap_or_else(|err|panic!("VID disperse failure:\n\t(num_storage nodes,payload_byte_len)=({num_nodes},{})\n\terror: : {err}", txns.len()))
```
jparr721, line 76: Two things:
1. It looks like the formatter missed this, it seems like `|err|panic` should have a space.
2. I don't think it's a good idea to add new lines and tabs to error logs as it'll make it more difficult to parse when reading and applying structured logging.
3. You shouldn't need to panic here. Instead, you can remove `error: {err}` entirely and simply use `expect`.

jparr721: As an addendum, I'm not crystal clear why we spawn this as a blocking task? Can't you just call this would needing a background thread at all?

crates/task-impls/src/transactions.rs
```diff
@@ -144,16 +147,21 @@ impl<
     #[instrument(skip_all, fields(id = self.id, view = *self.cur_view), name = "Transaction Handling Task", level = "error")]
     async fn wait_for_block(
         &self,
-    ) -> Option<(AvailableBlockData<TYPES>, AvailableBlockHeaderInput<TYPES>)> {
+    ) -> Option<(
```
jparr721, line 150: This is becoming a bit hard to parse. We should maybe make this into a struct.

jparr721 (changes requested):

crates/task-impls/src/helpers.rs
nyospe: I concur on the first to points. For the third... isn't `disperse_precompute` a bit long running, even with the precompute? I agree that `spawn_blocking` here is not helpful, but not on the reason. The advantage of `spawn` is that you can launch it allow it to run in a different thread, then come back and await (on the `JoinHandle`) after doing other stuff in parallel. It's a way to leverage the advantages of multithreaded concurrency in an async context. By contrast, `spawn_blocking` will yield on the calling async context while the spawned task gets to run (effectively) as if it were synchronous. This will allow other async tasks to progress, but not this current async task.

Crucially, either of these would only execute when you call `.await` on the future returned by `calculate_vid_disperse_using_precompute_data`, which isn't where the advantage of using `spawn` would be. To avoid serial execution of independent operations, you would want to `spawn` around `calculate_vid_disperse_using_precompute_data` and then await the result after completion of whatever other work needed doing in the calling context.

I suspect this is using `spawn_blocking` because of an (incorrect) rule of thumb that was widely promoted a few years ago, that recommended using it for longer (tens to hundreds of milliseconds) CPU-bound operations. This is still a one-off operation that eventually completes, and would be better handled with something like `rayon` or one of the more sophisticated threadpool crates.

crates/task-impls/src/helpers.rs
move47: https://github.com/EspressoSystems/HotShot/pull/2956/commits/8d15da890e00f3ab93edcd880b5fb1160b95768d

crates/task-impls/src/helpers.rs
move47: I guess it's fine since the existing api was following the same pattern. https://github.com/EspressoSystems/HotShot/blob/15c776d32001ce88edf2a87f3c520301da22a582/crates/task-impls/src/helpers.rs#L45

crates/task-impls/src/transactions.rs
move47: https://github.com/EspressoSystems/HotShot/pull/2956/commits/8d15da890e00f3ab93edcd880b5fb1160b95768d

crates/types/src/traits/block_contents.rs
```diff
@@ -124,6 +124,21 @@ pub fn vid_commitment(
...
+#[must_use]
+pub fn precompute_vid_commitment(
+    encoded_transactions: &Vec<u8>,
+    num_storage_nodes: usize,
+) -> (
+    <VidSchemeType as VidScheme>::Commit,
+    <VidSchemeType as Precomputable>::PrecomputeData,
+) {
+    #[allow(clippy::panic)]
+     vid_scheme(num_storage_nodes).commit_only_precompute(encoded_transactions).unwrap_or_else(|err| panic!("VidScheme::commit_only failure:\n\t(num_storage_nodes,payload_byte_len)=({num_storage_nodes},{}\n\t{err}", encoded_transactions.len()))
```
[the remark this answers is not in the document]
move47: https://github.com/EspressoSystems/HotShot/pull/2956/commits/9eea1144928da906c5d7ba3dab4e4ac0be962111

move47:
@jparr721 Can I get review again?

jparr721:
> @jparr721 Can I get review again?

Looks like the error messages still have the newlines and tab characters, and it looks like we're still explicitly panicking. Can we remove both of those and I can re-remove?

jparr721:
@nyospe Looks fine to me, but I am not deep on these changes. Mind taking a look as well?

crates/task-impls/src/transactions.rs
```diff
@@ -29,6 +31,18 @@ use crate::{
     helpers::broadcast_event,
 };
 
+/// Builder Provided Responses
+pub struct BuilderResponses<TYPES: NodeType> {
```
nyospe, line 35: I'm not clear on why we're bundling this all in one struct. Are we collecting what we get from each phase with the builder?

crates/task-impls/src/transactions.rs
```diff
@@ -186,9 +199,32 @@ impl<
                 continue;
             };
 
+            // Verify signature over chosen block instead of
```
nyospe, line 202: 👍

crates/task-impls/src/transactions.rs
move47: Yes, we are bundling all the responses from the builder. Later, we might filter and store what we need.

crates/builder-api/src/block_info.rs
```diff
@@ -33,7 +33,12 @@ pub struct AvailableBlockData<TYPES: NodeType> {
 #[serde(bound = "")]
 pub struct AvailableBlockHeaderInput<TYPES: NodeType> {
     pub vid_commitment: VidCommitment,
-    pub signature:
+    pub vid_precompute_data: VidPrecomputeData,
```
nyospe, line 36: This part looks good.

crates/task-impls/src/helpers.rs
```diff
@@ -42,15 +42,38 @@ pub async fn broadcast_event<E: Clone + std::fmt::Debug>(event: E, sender: &Send
...
     view: TYPES::Time,
 ) -> VidDisperse<TYPES> {
     let num_nodes = membership.total_nodes();
     let vid_disperse = spawn_blocking(move || {
-        #[allow(clippy::panic)]
-        vid_scheme(num_nodes).disperse(&txns).unwrap_or_else(|err|panic!("VID disperse failure:\n\t(num_storage nodes,payload_byte_len)=({num_nodes},{})\n\terror: : {err}", txns.len()))
+        vid_scheme(num_nodes).disperse(&txns).unwrap_or_else(|err| panic!("VID disperse failure:(num_storage nodes,payload_byte_len)=({num_nodes},{}) error: {err}", txns.len()))
+    })
+    .await;
+    #[cfg(async_executor_impl = "tokio")]
```
nyospe, line 56: I have questions. I don't think there's any other point in the HotShot repo where we hardcode only tokio. Maybe it would be worth unravelling this spawn_blocking just to avoid that.

crates/task-impls/src/helpers.rs
```diff
@@ -42,15 +42,38 @@ pub async fn broadcast_event<E: Clone + std::fmt::Debug>(event: E, sender: &Send
...
+/// Panics if the VID calculation fails, this should not happen.
+#[allow(clippy::panic)]
+pub async fn calculate_vid_disperse_using_precompute_data<TYPES: NodeType>(
+    txns: Vec<u8>,
+    membership: Arc<TYPES::Membership>,
+    view: TYPES::Time,
+    pre_compute_data: VidPrecomputeData,
+) -> VidDisperse<TYPES> {
+    let num_nodes = membership.total_nodes();
+    let vid_disperse = spawn_blocking(move || {
```
nyospe, line 75: This one too...

crates/task-impls/src/transactions.rs
nyospe: Doesn't that defeat the point of, well, having an external builder?

move47: Sorry, I didn't get wdym?

crates/task-impls/src/helpers.rs
move47: https://github.com/EspressoSystems/HotShot/pull/2956/commits/5408f863bc99171c7e299ac0460875145d98252d

crates/task-impls/src/helpers.rs
move47: https://github.com/EspressoSystems/HotShot/pull/2956/commits/5408f863bc99171c7e299ac0460875145d98252d

crates/builder-api/src/data_source.rs
```diff
@@ -15,7 +14,7 @@ pub trait BuilderDataSource<TYPES: NodeType> {
     /// To get the list of available blocks
     async fn get_available_blocks(
         &self,
-        for_parent: &VidCommitment,
+        for_parent: &BuilderCommitment,
```
nyospe, line 17: 👍

crates/task-impls/src/transactions.rs
nyospe: Never mind, we can fix this during the transition to multi-builder auctions

nyospe (approved):
Okay, I think this is ready now.

jparr721 (approved):

Issue #2951, opened 2024-04-11
[CX_CLEANUP] - Create a new task for `QuorumProposalRecv` events 


jparr721:
### What is this task and why do we need to work on it?

`QuorumProposalRecv` is an event that is handled by the consensus task when a new quorum proposal is received. This task is also responsible for validating multiple pieces of the proposal before submitting a `QuorumProposalValidated` event. Since this sequence of steps is now being split across the `QuorumProposalTask` and `QuorumVoteTask`, we need to avoid duplicating this behavior by instead creating a new task whose sole objective is to handle receiving and validating the quorum proposal and sending the validated event.

### What are the acceptance criteria to close this issue?

A new task is created for handling quorum proposals, it has unit tests created and the logic from the consensus task is able to be successfully deleted.

PR #2963, opened 2024-04-15
Add The `ProposeNow` and `VoteNow` events.

jparr721:
Closes #<ISSUE_NUMBER>

### This PR: 
Adds the ability to initiate a `ProposeNow` and `VoteNow` event which enables the initiator to be able to submit a proposal with their internal state immediately, without waiting for the appropriate sequence to have taken place. The relevant code for when such an event can happen is within the liveness check in consensus:
```rust
if liveness_check {
    self.current_proposal = Some(proposal.data.clone());
    let new_view = proposal.data.view_number + 1;

    // This is for the case where we form a QC but have not yet seen the previous proposal ourselves
    let should_propose = self.quorum_membership.get_leader(new_view)
        == self.public_key
        && high_qc.view_number
            == self.current_proposal.clone().unwrap().view_number;
    let qc = high_qc.clone();
    if should_propose {
        debug!(
            "Attempting to publish proposal after voting; now in view: {}",
            *new_view
        );
        self.publish_proposal_if_able(qc.view_number + 1, &event_stream)
            .await;
    }
    if self.vote_if_able(&event_stream).await {
        self.current_proposal = None;
    }
}
```
For `QuorumProposal` task, we need the ability to initiate a proposal in this case, but it does not occur within a `QuorumProposalValidated` event, which breaks the dependency tasks for voting and proposing. Instead, we must rely on an additional method for proposing and voting respectively.

The presumption with the liveness check call is that we at least *sometimes* have the ability to propose/vote in this block and would have the data available which would be required to do so. So, by making the new event type, we allow the sender to simply send out whatever data that they have which would make the proposal or vote valid.

So, within the new `QuorumProposalRecv` task (tracking: #2951), we'll be able to do something like this:
```rust
if liveness_check {
    self.current_proposal = Some(proposal.data.clone());
    let new_view = proposal.data.view_number + 1;

    // This is for the case where we form a QC but have not yet seen the previous proposal ourselves
    let should_propose = self.quorum_membership.get_leader(new_view)
        == self.public_key
        && high_qc.view_number
            == self.current_proposal.clone().unwrap().view_number;
    let qc = high_qc.clone();
    if should_propose {
        debug!(
            "Attempting to publish proposal after voting; now in view: {}",
            *new_view
        );
        broadcast_event(
            Arc::new(HotShotEvent::ProposeNow(...)),
            event_stream,
        )
        .await;
    }
    broadcast_event(
            Arc::new(HotShotEvent::VoteNow(...)),
            event_stream,
        )
        .await;
}
```
Which will then get to where it needs to be.

### DIFF
--- a/crates/task-impls/src/quorum_proposal.rs
+++ b/crates/task-impls/src/quorum_proposal.rs
@@ -57,6 +57,9 @@ enum ProposalDependency {
 
     /// For the `QuroumProposalRecv` event.
     Proposal,
+
+    /// For the `ProposeNow` event.
+    ProposeNow,
 }
 
 /// Handler for the proposal dependency
@@ -228,7 +231,7 @@ impl<TYPES: NodeType> ProposalDependencyHandle<TYPES> {
 }
 
 impl<TYPES: NodeType> HandleDepOutput for ProposalDependencyHandle<TYPES> {
-    type Output = Vec<Vec<Arc<HotShotEvent<TYPES>>>>;
+    type Output = Vec<Vec<Vec<Arc<HotShotEvent<TYPES>>>>>;
 
     #[allow(clippy::no_effect_underscore_binding)]
     async fn handle_dep_result(self, res: Self::Output) {
@@ -237,7 +240,7 @@ impl<TYPES: NodeType> HandleDepOutput for ProposalDependencyHandle<TYPES> {
         let mut _quorum_certificate = None;
         let mut _timeout_certificate = None;
         let mut _view_sync_finalize_cert = None;
-        for event in res.iter().flatten() {
+        for event in res.iter().flatten().flatten() {
             match event.as_ref() {
                 HotShotEvent::QuorumProposalValidated(proposal, _) => {
                     let proposal_payload_comm = proposal.block_header.payload_commitment();
@@ -401,6 +404,13 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalTaskState<TYPE
                             return false;
                         }
                     }
+                    ProposalDependency::ProposeNow => {
+                        if let HotShotEvent::ProposeIfAble(view, _) = event {
+                            *view
+                        } else {
+                            return false;
+                        }
+                    }
                 };
                 let valid = event_view == view_number;
                 if valid {
@@ -417,7 +427,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalTaskState<TYPE
         view_number: TYPES::Time,
         event_receiver: Receiver<Arc<HotShotEvent<TYPES>>>,
         event: Arc<HotShotEvent<TYPES>>,
-    ) -> AndDependency<Vec<Arc<HotShotEvent<TYPES>>>> {
+    ) -> AndDependency<Vec<Vec<Arc<HotShotEvent<TYPES>>>>> {
         let mut proposal_dependency = self.create_event_dependency(
             ProposalDependency::Proposal,
             view_number,
@@ -445,6 +455,12 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalTaskState<TYPE
         let mut payload_commitment_dependency = self.create_event_dependency(
             ProposalDependency::PayloadAndMetadata,
             view_number,
+            event_receiver.clone(),
+        );
+
+        let mut propose_now_dependency = self.create_event_dependency(
+            ProposalDependency::ProposeNow,
+            view_number,
             event_receiver,
         );
 
@@ -454,31 +470,8 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalTaskState<TYPE
         );
 
         match event.as_ref() {
-            HotShotEvent::ProposeNow(view, proposal_dependency_data) => {
-                payload_commitment_dependency.mark_as_completed(
-                    HotShotEvent::SendPayloadCommitmentAndMetadata(
-                        proposal_dependency_data.commitment_and_metadata.commitment,
-                        proposal_dependency_data
-                            .commitment_and_metadata
-                            .metadata
-                            .clone(),
-                        *view,
-                    )
-                    .into(),
-                );
-
-                match &proposal_dependency_data.secondary_proposal_information {
-                    hotshot_types::consensus::SecondaryProposalInformation::QuorumProposalAndCertificate(quorum_proposal, leaf, qc) => {
-                        proposal_dependency.mark_as_completed(HotShotEvent::QuorumProposalValidated(quorum_proposal.clone(), leaf.clone()).into());
-                        qc_dependency.mark_as_completed(HotShotEvent::QCFormed(either::Left(qc.clone())).into());
-                    } ,
-                    hotshot_types::consensus::SecondaryProposalInformation::Timeout(tc) => {
-                        timeout_dependency.mark_as_completed(HotShotEvent::QCFormed(either::Right(tc.clone())).into());
-                    },
-                    hotshot_types::consensus::SecondaryProposalInformation::ViewSync(vs) => {
-                        view_sync_dependency.mark_as_completed(HotShotEvent::ViewSyncFinalizeCertificate2Recv(vs.clone()).into());
-                    },
-                };
+            HotShotEvent::ProposeIfAble(..) => {
+                propose_now_dependency.mark_as_completed(event.clone())
             }
             HotShotEvent::SendPayloadCommitmentAndMetadata(..) => {
                 payload_commitment_dependency.mark_as_completed(event.clone());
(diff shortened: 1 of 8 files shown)


crates/task-impls/src/quorum_proposal.rs
```diff
@@ -461,83 +448,111 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalTaskState<TYPE
...
+                        qc_dependency.mark_as_completed(HotShotEvent::QCFormed(either::Left(qc.clone())).into());
+                    } ,
+                    hotshot_types::consensus::SecondaryProposalInformation::Timeout(tc) => {
+                        timeout_dependency.mark_as_completed(HotShotEvent::QCFormed(either::Right(tc.clone())).into());
+                    },
+                    hotshot_types::consensus::SecondaryProposalInformation::ViewSync(vs) => {
+                        view_sync_dependency.mark_as_completed(HotShotEvent::ViewSyncFinalizeCertificate2Recv(vs.clone()).into());
+                    },
+                };
+            }
```
shenkeyao, line 482: My understanding of this part is that it marks applicable dependencies as completed, which differ due to the SecondaryProposalInformation, but no matter which dependencies they are, we are basically completing sufficient dependencies so that we can send the proposal.

If my above understanding is correct, is this equivalent to defining the task completion as something like OrDependency(ProposalNow, CurrentDependencyLogic)? In this way, we don't need to do these mark_as_completed operations anymore.

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -461,83 +448,111 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalTaskState<TYPE
...
-                    AndDependency::from_deps(vec![view_sync_dependency]),
-                ]),
-            ])
-        };
+            secondary_deps.push(AndDependency::from_deps(vec![qc_dependency]));
+        }
+
+        AndDependency::from_deps(vec![
+            OrDependency::from_deps(vec![AndDependency::from_deps(vec![
+                payload_commitment_dependency,
```
shenkeyao, line 523: Nit: Could we simplify this? I don't think we need both `Or` and `And`.

crates/task-impls/src/quorum_vote.rs
```diff
@@ -250,19 +250,47 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumVoteTaskState<TYPES, I
...
+                vid_dependency.mark_as_completed(
+                    HotShotEvent::VIDShareValidated(vote_dependency_data.disperse_proposal.clone())
+                        .into(),
+                );
+                dac_dependency.mark_as_completed(
+                    HotShotEvent::DACertificateValidated(vote_dependency_data.da_cert.clone())
+                        .into(),
+                );
+            };
+        }
```
shenkeyao, line 291: Similar question as in `quorum_proposal.rs` about whether we can change the dependency setting to avoid marking all dependencies.

crates/types/src/vote.rs
```diff
@@ -180,3 +182,20 @@ impl<TYPES: NodeType, VOTE: Vote<TYPES>, CERT: Certificate<TYPES, Voteable = VOT
 
 /// Mapping of commitments to vote tokens by key.
 type VoteMap2<COMMITMENT, PK, SIG> = HashMap<COMMITMENT, (U256, BTreeMap<PK, (SIG, COMMITMENT)>)>;
+
+/// Payload for the `HotShotEvents::VoteNow` event type. The proposal and leaf are
+/// obtained via a `QuorumProposalValidated` event being processed.
+#[derive(Eq, Hash, PartialEq, Debug, Clone)]
+pub struct VoteDependencyData<TYPES: NodeType> {
```
shenkeyao, line 189: Is this used anywhere or just a placeholder?

shenkeyao (commented):
Besides my other comments, I have a more high-level question:

From the PR description, it looks like you plan to modify
```rust
if should_propose {
    ...
    self.publish_proposal_if_able...
}
```
to 
```rust
if should_propose {
    ...
    broadcast_event(
        Arc::new(HotShotEvent::ProposeNow(...)),
        ...
    )
}
```

I.e., instead of checking if we have all the info to propose, we just emit `ProposeNow`. I'd assume that we will collect data to be passed to `ProposaNow`, otherwise we can't guarantee the ability to propose. However, if so, how is this different from just having the new `QuorumProposalRecv` task send an event signaling the proposal is valid, and having the `QuorumProposal` task get all the info from other events and send the proposal? I feel it's not the `QuorumProposalRecv` task's job to collect all the data.

I agree that we need an additional event to indicate that the `QuorumProposalRecv` task is completed for a view, but I just don't think it's correct to rely on it to start the proposing or voting. Maybe I misunderstood your design or the necessity of doing so. Would like to hear your thoughts!

crates/task-impls/src/quorum_proposal.rs
jparr721: Oh that's a good point!

crates/task-impls/src/quorum_proposal.rs
jparr721: Unfortunately we do due to how the type system works. If the heterogenous vector doesn't have the same level of indirection the compiler won't let it through.

crates/types/src/vote.rs
jparr721: It's used in the event type and I also use it in the unit test (pushing in a sec)

crates/task-impls/src/quorum_proposal.rs
jparr721: Also, at the highest level, it requires that an `AndDependency` is the top-level value if any layer uses an And dependency due to how the task is implemented.

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -586,6 +610,14 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalTaskState<TYPE
         event_sender: Sender<Arc<HotShotEvent<TYPES>>>,
     ) {
         match event.as_ref() {
+            HotShotEvent::ProposeNow(view, _) => {
+                self.create_dependency_task_if_new(
```
jparr721, line 614: Explicit validation is coming to this before the final integration with consensus.

crates/task-impls/src/quorum_vote.rs
```diff
@@ -109,6 +111,16 @@ impl<TYPES: NodeType, S: Storage<TYPES> + 'static> HandleDepOutput
...
                     }
                 }
+                HotShotEvent::VoteNow(_, vote_dependency_data) => {
+                    payload_commitment = Some(
+                        vote_dependency_data
+                            .quorum_proposal
+                            .block_header
+                            .payload_commitment(),
+                    );
+                    leaf = Some(vote_dependency_data.leaf.clone());
```
jparr721, line 121: @shenkeyao looks like this code gets the payload_commitment from one of 2 possible sources, not sure if this will change in your upcoming work, but I can shrink the payload size if we only care about one possible option.

shenkeyao: I don't think we need to set `payload_commitment` if the event is `VoteNow`. We want to check if the commitments from the other two events match, but if we get `VoteNow` we know we can vote immediately.

shenkeyao (commented):
Looks good except for a few minor comments. There are merge conflicts but I can send a quick update to fix the conflicts.

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -271,6 +275,21 @@ impl<TYPES: NodeType> HandleDepOutput for ProposalDependencyHandle<TYPES> {
...
+                    match &pdd.secondary_proposal_information {
+                        hotshot_types::consensus::SecondaryProposalInformation::QuorumProposalAndCertificate(quorum_proposal, quorum_certificate) => {
+                            _quorum_certificate = Some(quorum_certificate.clone());
+                            payload_commitment = Some(quorum_proposal.block_header.payload_commitment());
+                        },
+                        hotshot_types::consensus::SecondaryProposalInformation::Timeout(tc) => {
+                            _timeout_certificate = Some(tc.clone());
+                        }
+                        hotshot_types::consensus::SecondaryProposalInformation::ViewSync(vsc) => {
+                            _view_sync_finalize_cert = Some(vsc.clone());
```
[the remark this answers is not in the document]
shenkeyao: Oh, are we going to use them later when completing the integration?

crates/task-impls/src/quorum_vote.rs
shenkeyao: I sent an update to remove the `payload_commitment` setting from here.

shenkeyao (approved):
I sent two updates to fix the merge conflicts and addressed my own comment. I'm going to approve this PR so you can merge it if you think my comments and changes look good. Otherwise, feel free to revert my changes or add more fixes.

crates/task-impls/src/quorum_proposal.rs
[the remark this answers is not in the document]
jparr721: Yep!

PR #2968, opened 2024-04-16
Remove special casing for view 1 (again)

ss-es:
No linked issue.

### This PR: 
Adds back #2873 (reverted by #2952).

The only change over #2873 is that we now broadcast a leaf decide event on startup, which seems to be required by `hotshot-query-service`.


jparr721:
Do we have any way to run automated tests against query service to make sure this works? If you've gone and validated against query service, I'd love to have those steps in "How to test" for my own learning.

rob-maron (dismissed):

rob-maron (dismissed):

rob-maron (approved):

PR #2970, opened 2024-04-16
partial commit

jparr721:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/crates/task-impls/src/quorum_proposal.rs
+++ b/crates/task-impls/src/quorum_proposal.rs
@@ -1,6 +1,10 @@
-use crate::{consensus::validate_proposal, helpers::AnyhowTracing};
+use crate::{
+    consensus::{update_view, validate_proposal},
+    helpers::AnyhowTracing,
+};
 use std::{collections::HashMap, marker::PhantomData, sync::Arc, time::Duration};
 
+use anyhow::{ensure, Result};
 use async_broadcast::{Receiver, Sender};
 use async_compatibility_layer::art::{async_sleep, async_spawn};
 use async_lock::{RwLock, RwLockUpgradableReadGuard};
@@ -724,7 +728,22 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalTaskState<TYPE
                 }
 
                 // NOTE: We could update our view with a valid TC but invalid QC, but that is not what we do here
-                self.update_view(view, &event_sender).await;
+                if let Err(e) = update_view::<TYPES, I>(
+                    self.public_key.clone(),
+                    view,
+                    &event_sender,
+                    self.quorum_membership.clone(),
+                    self.quorum_network.clone(),
+                    self.timeout,
+                    self.consensus.clone(),
+                    &mut self.cur_view,
+                    &mut self.timeout_task,
+                )
+                .await
+                {
+                    // This isn't typically a serious end-it-all failure.
+                    warn!("Failed to update view; error = {:?}", e);
+                }
 
                 let consensus = self.consensus.upgradable_read().await;
 
@@ -817,103 +836,6 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalTaskState<TYPE
             }
             _ => {}
         }
-    }
-
-    /// Must only update the view and GC if the view actually changes
-    #[instrument(skip_all, fields(
-        id = self.id,
-        view = *self.cur_view
-    ), name = "Consensus update view", level = "error")]
-    async fn update_view(
-        &mut self,
-        new_view: TYPES::Time,
-        event_stream: &Sender<Arc<HotShotEvent<TYPES>>>,
-    ) -> bool {
-        if *self.cur_view < *new_view {
-            debug!(
-                "Updating view from {} to {} in consensus task",
-                *self.cur_view, *new_view
-            );
-
-            if *self.cur_view / 100 != *new_view / 100 {
-                // TODO (https://github.com/EspressoSystems/HotShot/issues/2296):
-                // switch to info! when INFO logs become less cluttered
-                error!("Progress: entered view {:>6}", *new_view);
-            }
-
-            // cancel the old timeout task
-            if let Some(timeout_task) = self.timeout_task.take() {
-                cancel_task(timeout_task).await;
-            }
-            self.cur_view = new_view;
-
-            // Poll the future leader for lookahead
-            let lookahead_view = new_view + LOOK_AHEAD;
-            if self.quorum_membership.get_leader(lookahead_view) != self.public_key {
-                self.quorum_network
-                    .inject_consensus_info(ConsensusIntentEvent::PollFutureLeader(
-                        *lookahead_view,
-                        self.quorum_membership.get_leader(lookahead_view),
-                    ))
-                    .await;
-            }
-
-            // Start polling for proposals for the new view
-            self.quorum_network
-                .inject_consensus_info(ConsensusIntentEvent::PollForProposal(*self.cur_view + 1))
-                .await;
-
-            self.quorum_network
-                .inject_consensus_info(ConsensusIntentEvent::PollForDAC(*self.cur_view + 1))
-                .await;
-
-            if self.quorum_membership.get_leader(self.cur_view + 1) == self.public_key {
-                debug!("Polling for quorum votes for view {}", *self.cur_view);
-                self.quorum_network
-                    .inject_consensus_info(ConsensusIntentEvent::PollForVotes(*self.cur_view))
-                    .await;
-            }
-
-            broadcast_event(Arc::new(HotShotEvent::ViewChange(new_view)), event_stream).await;
-
-            // Spawn a timeout task if we did actually update view
-            let timeout = self.timeout;
-            self.timeout_task = Some(async_spawn({
-                let stream = event_stream.clone();
-                // Nuance: We timeout on the view + 1 here because that means that we have
-                // not seen evidence to transition to this new view
-                let view_number = self.cur_view + 1;
-                async move {
-                    async_sleep(Duration::from_millis(timeout)).await;
-                    broadcast_event(
-                        Arc::new(HotShotEvent::Timeout(TYPES::Time::new(*view_number))),
-                        &stream,
-                    )
-                    .await;
-                }
-            }));
-            let consensus = self.consensus.upgradable_read().await;
-            consensus
-                .metrics
-                .current_view
-                .set(usize::try_from(self.cur_view.get_u64()).unwrap());
-            // Do the comparison before the subtraction to avoid potential overflow, since
-            // `last_decided_view` may be greater than `cur_view` if the node is catching up.
-            if usize::try_from(self.cur_view.get_u64()).unwrap()
-                > usize::try_from(consensus.last_decided_view.get_u64()).unwrap()
-            {
-                consensus.metrics.number_of_views_since_last_decide.set(
-                    usize::try_from(self.cur_view.get_u64()).unwrap()
-                        - usize::try_from(consensus.last_decided_view.get_u64()).unwrap(),
-                );
-            }
-            let mut consensus = RwLockUpgradableReadGuard::upgrade(consensus).await;
-            consensus.update_view(new_view);
-            drop(consensus);
-
-            return true;
-        }
-        false
     }
 }
 
--- a/crates/task-impls/src/quorum_proposal_recv.rs
+++ b/crates/task-impls/src/quorum_proposal_recv.rs
@@ -1,0 +1,400 @@
+use std::{collections::BTreeMap, sync::Arc};
+
+use crate::{
+    consensus::{update_view, validate_proposal},
+    events::HotShotEvent,
+    helpers::{broadcast_event, cancel_task, AnyhowTracing},
+};
+use async_broadcast::Sender;
+use async_compatibility_layer::art::{async_sleep, async_spawn};
+use async_lock::{RwLock, RwLockUpgradableReadGuard};
+#[cfg(async_executor_impl = "async-std")]
+use async_std::task::JoinHandle;
+use committable::Committable;
+use futures::future::{join_all, FutureExt};
+use hotshot_task::task::{Task, TaskState};
+use hotshot_types::{
+    consensus::Consensus,
+    data::{Leaf, QuorumProposal, ViewChangeEvidence},
+    event::{Event, EventType, LeafInfo},
+    simple_certificate::UpgradeCertificate,
+    traits::{
+        election::Membership,
+        network::{ConnectedNetwork, ConsensusIntentEvent},
+        node_implementation::{ConsensusTime, NodeImplementation, NodeType},
+        signature_key::SignatureKey,
+        storage::Storage,
+        ValidatedState,
+    },
+    utils::{View, ViewInner},
+    vote::{Certificate, HasViewNumber},
+};
+#[cfg(async_executor_impl = "tokio")]
+use tokio::task::JoinHandle;
+use tracing::{debug, error, instrument, warn};
+
+/// The state for the quorum proposal task. Contains all of the information for
+/// handling [`HotShotEvent::QuorumProposalRecv`] events.
+pub struct QuorumProposalRecvTaskState<TYPES: NodeType, I: NodeImplementation<TYPES>> {
+    /// Our public key
+    pub public_key: TYPES::SignatureKey,
+
+    /// Our Private Key
+    pub private_key: <TYPES::SignatureKey as SignatureKey>::PrivateKey,
+
+    /// Reference to consensus. The replica will require a write lock on this.
+    pub consensus: Arc<RwLock<Consensus<TYPES>>>,
+
+    /// View number this view is executing in.
+    pub cur_view: TYPES::Time,
+
+    /// Network for all nodes
+    pub quorum_network: Arc<I::QuorumNetwork>,
+
+    /// Membership for Quorum Certs/votes
+    pub quorum_membership: Arc<TYPES::Membership>,
+
+    /// Membership for Timeout votes/certs
+    pub timeout_membership: Arc<TYPES::Membership>,
+
+    /// timeout task handle
+    pub timeout_task: Option<JoinHandle<()>>,
+
+    /// View timeout from config.
+    pub timeout: u64,
+
+    /// Output events to application
+    pub output_event_stream: async_broadcast::Sender<Event<TYPES>>,
+
+    /// This node's storage ref
+    pub storage: Arc<RwLock<I::Storage>>,
+
+    /// The most recent proposal we have, will correspond to the current view if Some()
+    /// Will be none if the view advanced through timeout/view_sync
+    pub current_proposal: Option<QuorumProposal<TYPES>>,
+
+    /// Spawned tasks related to a specific view, so we can cancel them when
+    /// they are stale
+    pub spawned_tasks: BTreeMap<TYPES::Time, Vec<JoinHandle<()>>>,
+
+    /// The node's id
+    pub id: u64,
+}
+
+impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalRecvTaskState<TYPES, I> {
+    /// Cancel all tasks that have been spawned before the provided view.
+    async fn cancel_tasks(&mut self, view: TYPES::Time) {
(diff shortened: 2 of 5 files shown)


crates/task-impls/src/quorum_proposal_recv.rs
```diff
@@ -0,0 +1,409 @@
...
+                            broadcast_event(
+                                Arc::new(HotShotEvent::ProposeNow(
+                                    qc.view_number + 1,
+
+                                ))
+                            )
+                            // self.publish_proposal_if_able(qc.view_number + 1, &event_stream)
+                            //     .await;
+                        }
+                        if self.vote_if_able(&event_stream).await {
```
shenkeyao, line 351: Similar to the above, we need to replace this function with vote validity checks here. I can work on this.

crates/task-impls/src/quorum_proposal_recv.rs
```diff
@@ -0,0 +1,409 @@
...
+                            let pdd = {
+
+                            }
+                            broadcast_event(
+                                Arc::new(HotShotEvent::ProposeNow(
+                                    qc.view_number + 1,
+
+                                ))
+                            )
+                            // self.publish_proposal_if_able(qc.view_number + 1, &event_stream)
```
shenkeyao, line 348: Looks like the code inside `publish_proposal_if_able` isn't pasted to here yet. We can do that later, but I just wanted to leave this comment in case we forget.

crates/task-impls/src/quorum_proposal_recv.rs
```diff
@@ -0,0 +1,409 @@
...
+
+impl<TYPES: NodeType, I: NodeImplementation<TYPES>> TaskState
+    for QuorumProposalRecvTaskState<TYPES, I>
+{
+    type Event = Arc<HotShotEvent<TYPES>>;
+    type Output = ();
+    fn filter(&self, event: &Arc<HotShotEvent<TYPES>>) -> bool {
+        !matches!(
+            event.as_ref(),
+            HotShotEvent::QuorumProposalRecv(..) | HotShotEvent::QuorumProposalValidated(..),
```
shenkeyao, line 393: Why is the `QuorumProposalValidated` event needed here?

shenkeyao:
Another thing we need to fix: In the new task we call `validate_proposal`, which is a function in the consensus task. I think we need to either modify that function to emit `ProposeNow` and `VoteNow` when able, or remove that call and directly do validations in the new task and send `*Now` events. I think both can work but the latter is probably better since we don't have to make `validate_proposal` work for both tasks.

crates/task-impls/src/quorum_proposal_recv.rs
jparr721: Mistake on my part, removing

crates/task-impls/src/consensus/proposal_helpers.rs
```diff
@@ -566,6 +567,12 @@ pub async fn publish_proposal_if_able<TYPES: NodeType>(
     }
 }
 
+#[cfg(feature = "dependency-tasks")]
+type TempraryProposalCombinedType<TYPES, I, A> = QuorumProposalRecvTaskState<TYPES, I>;
+
+#[cfg(not(feature = "dependency-tasks"))]
+type TempraryProposalCombinedType<TYPES, I, A> = ConsensusTaskState<TYPES, I, A>;
```
shenkeyao, line 574: Curious why we need these types. Looks like they are the same as the existing `*State` structs. Will this change later?

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -851,103 +867,6 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalTaskState<TYPE
             _ => {}
         }
     }
-
-    /// Must only update the view and GC if the view actually changes
-    #[instrument(skip_all, fields(
-        id = self.id,
-        view = *self.cur_view
-    ), name = "Consensus update view", level = "error")]
-    async fn update_view(
```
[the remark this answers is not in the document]
shenkeyao: Even nicer that it's moved to a separate `view_change` module. 😃

crates/task-impls/src/quorum_proposal_recv.rs
```diff
@@ -0,0 +1,164 @@
...
+                                VoteDependencyData {
+                                    quorum_proposal: current_proposal,
+                                },
+                            )),
+                            &event_stream,
+                        )
+                        .await;
+                        // if self.vote_if_able(&event_stream).await {
+                        //     self.current_proposal = None;
+                        // }
```
shenkeyao, line 131: I think we should still reset `current_proposal` though we don't have to call `vote_if_able` here.

shenkeyao: Actually, do we need `self.curren_proposal`? Looks like we don't use it anywhere in this task.

crates/task-impls/src/quorum_proposal_recv.rs
```diff
@@ -0,0 +1,164 @@
...
+                    .await
+                {
+                    Ok(Some(current_proposal)) => {
+                        self.current_proposal = Some(current_proposal);
+
+                        // Build the parent leaf since we didn't find it during the proposal check.
+                        let consensus = self.consensus.read().await;
+
+                        broadcast_event(
+                            Arc::new(HotShotEvent::VoteNow(
```
shenkeyao, line 120: Thinking about this more, I don't think we can send `VoteNow` here since `handle_quorum_proposal_recv` doesn't call `vote_if_able` to check the eligibility to vote.

shenkeyao:
I fixed the lint errors, except for the documentation for the `Temporary*` types because I'm not sure why we need them. I added a `TODO` there.

shenkeyao:
Fixed more lints for the build with the `dependency-task` feature. There are a few errors left related to `A: ConsensusApi` which I think we can remove from the consensus task state. I'm going to do this in a separate PR since it's unrelated to the proposal refactor.

shenkeyao:
> Fixed more lints for the build with the `dependency-task` feature. There are a few errors left related to `A: ConsensusApi` which I think we can remove from the consensus task state. I'm going to do this in a separate PR since it's unrelated to the proposal refactor.

I've removed `ConsensusApi` in https://github.com/EspressoSystems/HotShot/pull/3025. I think we can merge that one first, then the build fix (with the `dependency-task` feature) for this PR should be easy.

crates/task-impls/src/quorum_proposal_recv.rs
jparr721: Yeah we can blow it away, it was legacy for when I was trying to make this work within the Consensus task.

shenkeyao:
I fixed the build for compiling with and without `dependency_tasks` feature in this PR: https://github.com/EspressoSystems/HotShot/pull/3036. Didn't push directly to here since it has some changes besides the build fix.

crates/task-impls/src/quorum_proposal_recv.rs
jparr721: Isn't eligibility determined somewhat arbitrarily? It calls it only if we have a `current_proposal`, meaning that in this case where we receive an output from this function we should still be able to *try* to vote.

crates/hotshot/src/lib.rs
```diff
@@ -292,7 +292,10 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> SystemContext<TYPES, I> {
             locked_view: anchored_leaf.get_view_number(),
             high_qc: initializer.high_qc,
             metrics: Arc::clone(&consensus_metrics),
+            decided_upgrade_cert: None,
```
jparr721, line 295: Gating these proved to be impossible. I'll keep a watch for any usage of these vars in coming PRs.

crates/task-impls/src/consensus/proposal_helpers.rs
```diff
@@ -975,37 +1009,41 @@ pub async fn handle_quorum_proposal_validated<TYPES: NodeType, I: NodeImplementa
...
 
-    let new_view = task_state.current_proposal.clone().unwrap().view_number + 1;
-    // In future we can use the mempool model where we fetch the proposal if we don't have it, instead of having to wait for it here
-    // This is for the case where we form a QC but have not yet seen the previous proposal ourselves
-    let should_propose = task_state.quorum_membership.get_leader(new_view) == task_state.public_key
-        && consensus.high_qc.view_number
-            == task_state.current_proposal.clone().unwrap().view_number;
-    // todo get rid of this clone
-    let qc = consensus.high_qc.clone();
+    #[cfg(not(feature = "dependency-tasks"))]
```
jparr721, line 1012: This is a weird looking diff, but everything including line 1045 is feature gated.

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -988,8 +780,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> TaskState
     fn filter(&self, event: &Arc<HotShotEvent<TYPES>>) -> bool {
         !matches!(
             event.as_ref(),
-            HotShotEvent::QuorumProposalRecv(_, _)
```
jparr721, line 991: We no longer need to handle this since we have a dedicated task.

justfile
```diff
@@ -176,11 +176,16 @@ fmt:
   echo Running cargo fmt
   cargo fmt
 
-fmt_lint: 
+fmt_lint:
   echo Formatting and linting
   cargo fmt
   cargo clippy --workspace --examples --bins --tests -- -D warnings
 
+fmt_lint_dependency_tasks:
```
jparr721, line 184: Thanks keyao for adding this!

crates/hotshot/src/lib.rs
shenkeyao: Ahh, I see. I had the same question in https://github.com/EspressoSystems/HotShot/pull/3036 and didn't realize it's actually impossible.

Can we instead have two `Consensus` structs with different fields? The code may look ugly but then we don't have to keep a watch to avoid misusage.

crates/task-impls/src/quorum_proposal_recv.rs
```diff
@@ -0,0 +1,206 @@
...
+    #[allow(unused_variables)]
+    pub async fn handle(
+        &mut self,
+        event: Arc<HotShotEvent<TYPES>>,
+        event_stream: Sender<Arc<HotShotEvent<TYPES>>>,
+    ) {
+        #[cfg(feature = "dependency-tasks")]
+        if let HotShotEvent::QuorumProposalRecv(proposal, sender) = event.as_ref() {
+            match handle_quorum_proposal_recv(proposal, sender, event_stream.clone(), self).await {
+                Ok(Some(current_proposal)) => {
```
shenkeyao, line 122: As mentioned offline, I think we should call `cancel_tasks` for this view if the result is `Ok`.

Otherwise, looks like we never cancel the spawned tasks and can just remove the `cancel_tasks` function.

crates/types/src/consensus.rs
```diff
@@ -75,6 +78,17 @@ pub struct Consensus<TYPES: NodeType> {
...
+    /// The most recent upgrade certificate this node formed.
+    /// Note: this is ONLY for certificates that have been formed internally,
+    /// so that we can propose with them.
+    ///
+    /// Certificates received from other nodes will get reattached regardless of this fields,
+    /// since they will be present in the leaf we propose off of.
+    pub formed_upgrade_certificate: Option<UpgradeCertificate<TYPES>>,
+
+    /// most recent decided upgrade certificate
+    pub decided_upgrade_cert: Option<UpgradeCertificate<TYPES>>,
```
shenkeyao, line 91: See my comment in lib.rs. It's probably better to have two `Consensus` structs to make sure we always pass in the correct fields.

If this is too complicated (shouldn't be?), I think we should at least add a comment here.

crates/types/src/vote.rs
```diff
@@ -190,7 +190,8 @@ pub struct VoteDependencyData<TYPES: NodeType> {
     /// The quorum proposal (not necessarily valid).
     pub quorum_proposal: QuorumProposal<TYPES>,
 
-    /// The leaf we've obtained from the `QuorumProposalValidated` event.
+    /// The leaf we've obtained from the `QuorumProposalValidated` event. This is the
```
shenkeyao, line 193: Nit: Can we rename this to `parent_leaf`?

shenkeyao (commented):
I have some comments/questions about task cancellation and feature gates. Looks good overall!

crates/task-impls/src/quorum_proposal_recv.rs
```diff
@@ -0,0 +1,206 @@
...
+                    .await
+                    {
+                        Ok((parent_leaf, _ /* state */)) => parent_leaf,
+                        Err(e) => {
+                            warn!(?e, "Failed to get parent leaf and state");
+                            return;
+                        }
+                    };
+
+                    // TODO: Can we send `VoteNow` without calling `vote_if_able` before?
```
shenkeyao, line 140: Nit: This TODO is no longer applicable. We can remove it.

crates/hotshot/src/lib.rs
ss-es: Is it the upgrade certificates in particular? What were you trying to gate them on?

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -739,126 +738,16 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalTaskState<TYPE
...
-                        .storage
-                        .write()
-                        .await
-                        .update_high_qc(justify_qc.clone())
-                        .await
-                    {
-                        warn!("Failed to store High QC not voting. Error: {:?}", e);
-                        return;
-                    }
+                    warn!(?e, "Failed to handle QuorumProposalValidated event");
```
ss-es, line 748: This should probably be `debug`, and formatted with `{e}` rather than `?e`

crates/task-impls/src/consensus/proposal_helpers.rs
```diff
@@ -975,37 +1008,41 @@ pub async fn handle_quorum_proposal_validated<TYPES: NodeType, I: NodeImplementa
...
+        if should_propose {
+            debug!(
+                "Attempting to publish proposal after voting; now in view: {}",
+                *new_view
+            );
+            if let Err(e) = task_state
+                .publish_proposal(qc.view_number + 1, event_stream.clone())
+                .await
+            {
+                warn!("Failed to propose; error = {e:?}");
```
ss-es, line 1036: Should this be `warn`?

crates/task-impls/src/consensus/proposal_helpers.rs
```diff
@@ -873,7 +894,15 @@ pub async fn handle_quorum_proposal_validated<TYPES: NodeType, I: NodeImplementa
...
-                            task_state.decided_upgrade_cert = Some(cert.clone());
+                            #[cfg(not(feature = "dependency-tasks"))]
+                            {
+                                task_state.decided_upgrade_cert = Some(cert.clone());
+                            }
+
+                            #[cfg(feature = "dependency-tasks")]
+                            {
+                                decided_upgrade_cert = Some(cert.clone());
+                            }
```
ss-es, line 905: If I understand correctly what's happening here, I think you can move `decided_upgrade_cert` out of `ConsensusTaskState` and just have it in `Consensus`. We take a lock on consensus in this function anyway, I don't see why we can't just store this there (and avoid gating on the feature).

shenkeyao:
Have you tried the proposal and vote task tests? They both hang for me locally, which I think is probably caused by a previous PR but we should fix it (not necessarily in this PR).

crates/task-impls/src/quorum_proposal.rs
jparr721: Yes! I had made this change before our discussion yesterday. Thanks for pointing this out.

crates/hotshot/src/lib.rs
jparr721: No, it was moving them *into* the Consensus struct. Rust doesn't allow you to gate expressions like this. I actually figured it out after this rolled. Let me tinker around today.

crates/task-impls/src/consensus/proposal_helpers.rs
jparr721: So moving `decided_upgrade_cert` out of `Consensus` already exists, this is required due to the `consensus` variable being a read lock, not a write lock, when this callback is invoked. As a result, I need to be able to capture the value. I tried doing it with a write lock, but the `consensus.visit_leaf_anscestors` call is an immutable reference, so the compiler gets angry. The next-best option in this case was to simply have a variable that I allocate and assign.

crates/task-impls/src/consensus/proposal_helpers.rs
jparr721: Fixed.

crates/task-impls/src/quorum_proposal_recv.rs
jparr721: Done

crates/types/src/consensus.rs
jparr721: This is just headaches on headaches. I added the separate struct, but initializing was not quite doable without defining a super `new` method or trying to implement Default so it would smooth over the fields that are default None. I'll add a comment and, for now, rename the variables to include something like `dontuse`

jparr721:
> Have you tried the proposal and vote task tests? They both hang for me locally, which I think is probably caused by a previous PR. Let's fix it! Not necessarily in this PR though.

Cutting a branch with CI changes and testing now.

crates/task-impls/src/quorum_vote.rs
```diff
@@ -361,230 +364,17 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumVoteTaskState<TYPES, I
...
-                    &self.output_event_stream,
-                )
-                .await;
-                broadcast_event(
-                    Arc::new(HotShotEvent::QuorumProposalValidated(
-                        proposal.data.clone(),
-                        parent_leaf,
-                    )),
+            HotShotEvent::QuorumProposalValidated(proposal, _leaf) => {
+                // This task simultaneously does not rely on the state updates of the `handle_quorum_proposal_validated`
```
jparr721, line 368: We no longer need to process `QuorumProposalRecv` events here.

crates/testing/tests/tests_1/quorum_vote_task.rs
```diff
@@ -23,26 +23,31 @@ async fn test_quorum_vote_task_success() {
...
         inputs: vec![
-            QuorumProposalRecv(view.quorum_proposal.clone(), view.leader_public_key),
-            DACertificateRecv(view.da_certificate.clone()),
-            VIDShareRecv(view.vid_proposal.0[0].clone()),
+            QuorumProposalValidated(proposals[1].data.clone(), leaves[0].clone()),
+            DACertificateRecv(dacs[1].clone()),
+            VIDShareRecv(vids[1].0[0].clone()),
         ],
         outputs: vec![
-            exact(ViewChange(ViewNumber::new(2))),
```
shenkeyao, line 41: Probably unrelated to this PR, but I'm wondering why we no longer have `ViewChange` here and the other tests in this file. I think the vote task still needs this event to call `PollForVIDDisperse`.

crates/task-impls/src/consensus/mod.rs
```diff
@@ -315,17 +308,19 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> ConsensusTaskState<TYPES, I>
         false
     }
 
+    /// Dummy proposal
     #[cfg(feature = "dependency-tasks")]
+    #[allow(clippy::unused_async, clippy::no_effect_underscore_binding)]
```
shenkeyao, line 313: Nit: can we remove this `publish_proposal` so that we know we'll propose whenever we call `publish_proposal`? Similarly, we have only one `vote_if_able`.

shenkeyao (approved):
Looks good. Thanks for addressing all the comments. I can send a quick fix for the last nit comment I left.

crates/task-impls/src/consensus/mod.rs
shenkeyao: I've sent an update to remove this.

shenkeyao (approved):
Approving since I've fixed the comments from myself. Feel free to modify if needed!

shenkeyao (approved):

PR #2979, opened 2024-04-17
Fix builder client URL

sveitser:


crates/task-impls/src/transactions.rs
```diff
@@ -192,9 +192,14 @@ impl<
...
             {
-                Ok(blocks) => blocks,
+                Ok(blocks) => {
+                    tracing::debug!("Got available blocks: {:?}", blocks);
+                    blocks
+                },
                 Err(err) => {
                     error!(%err, "Couldn't get available blocks");
+                    // pause a bit
+                    async_sleep(Duration::from_millis(100)).await;
```
QuentinI, line 202: Oh, good one! Guess I've been DoSing the builder all this time 😅
We should probably do the same on claim calls down the line.

PR #2980, opened 2024-04-17
[CX_CLEANUP] - Decompose Consensus Task into a Smaller Module

### LINKED ISSUE #2981: [CX_CLEANUP] - Decompose Consensus Task into a Smaller Module

The consensus task is extremely large and requires significant effort to be able to decompose it fully into the proposal and vote tasks. Part of this work means the gradual movement of this logic into separable functions such that the same behavior can exist without the need to arduously manage the implementations in two or more places. For instance, consider validating a proposal:
```rust
/// Validate the state and safety and liveness of a proposal then emit
/// a `QuorumProposalValidated` event.
#[allow(clippy::too_many_arguments)]
pub async fn validate_proposal<TYPES: NodeType>(
    proposal: Proposal<TYPES, QuorumProposal<TYPES>>,
    parent_leaf: Leaf<TYPES>,
    decided_upgrade_certificate: Option<UpgradeCertificate<TYPES>>,
    quorum_membership: Arc<TYPES::Membership>,
    parent_state: Arc<TYPES::ValidatedState>,
    view_leader_key: TYPES::SignatureKey,
    event_stream: Sender<Arc<HotShotEvent<TYPES>>>,
    sender: TYPES::SignatureKey,
    event_sender: Sender<Event<TYPES>>,
    storage: Arc<RwLock<impl Storage<TYPES>>>,
) -> Result<()> {
```

This function was split out to be able to more-easily manage the complexity of the `QuorumProposalRecv` handler. Without this, we were forced to have a massive event handler, which resulted in extreme pain when trying to refactor the task. This work formally scopes this exact refactoring work.

We need to decompose methods whose behavior may need to be replicated during the process of decomposing the monolithic consensus task. Specifically, we will do the following:
1. Move `consensus.rs` into a `consensus` module.
2. Create separate files for the different pieces of functionality that need to be created (ex. `proposal.rs` for proposing)
3. Move any other relevant functions in as low-touch of a way as possible.

_No response_

The unit tests must be able to pass as-is. This is intended to be primarily a cosmetic and complexity reducing change. If a change results in a test failure (that is not the result of a function signature change, for example), then we should not make the change. The intention is to initially provide a *surface* for larger changes to result in smaller PRs.

_No response_

jparr721:
Closes #2981 

### This PR: 
The consensus task is extremely large and requires significant effort to be able to decompose it fully into the proposal and vote tasks. Part of this work means the gradual movement of this logic into separable functions such that the same behavior can exist without the need to arduously manage the implementations in two or more places. For instance, consider validating a proposal:
```rust
/// Validate the state and safety and liveness of a proposal then emit
/// a `QuorumProposalValidated` event.
#[allow(clippy::too_many_arguments)]
pub async fn validate_proposal<TYPES: NodeType>(
    proposal: Proposal<TYPES, QuorumProposal<TYPES>>,
    parent_leaf: Leaf<TYPES>,
    decided_upgrade_certificate: Option<UpgradeCertificate<TYPES>>,
    quorum_membership: Arc<TYPES::Membership>,
    parent_state: Arc<TYPES::ValidatedState>,
    view_leader_key: TYPES::SignatureKey,
    event_stream: Sender<Arc<HotShotEvent<TYPES>>>,
    sender: TYPES::SignatureKey,
    event_sender: Sender<Event<TYPES>>,
    storage: Arc<RwLock<impl Storage<TYPES>>>,
) -> Result<()> {
```

This function was split out to be able to more-easily manage the complexity of the `QuorumProposalRecv` handler. Without this, we were forced to have a massive event handler, which resulted in extreme pain when trying to refactor the task. This work pulls the relevant duplicated functions into their own places so that way other parts of the code can use them.

This PR does the following work:
1. Creates the new `consensus` module.
2. Moves relevant functions into their logical location.
3. Updates relevant import statements.

### This PR does not: 

This PR, by design, does *not* change any logic or tests. The idea is to first *move* things to enable breaking changes to operate with fewer lines touched.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1292,7 +1292,7 @@ checksum = "8cd6604a82acf3039f1144f54b8eb34e91ffba622051189e71b781822d5ee1f5"
 [[package]]
 name = "cdn-broker"
 version = "0.1.0"
-source = "git+https://github.com/EspressoSystems/Push-CDN?tag=0.1.24#5e0689dd07e3a5692b272afb176be43d26002f3d"
+source = "git+https://github.com/EspressoSystems/Push-CDN?tag=0.1.25#a506712f553b4800ce3a36e62341c839153aeab3"
 dependencies = [
  "async-std",
  "cdn-proto",
@@ -1315,7 +1315,7 @@ dependencies = [
 [[package]]
 name = "cdn-client"
 version = "0.1.0"
-source = "git+https://github.com/EspressoSystems/Push-CDN?tag=0.1.24#5e0689dd07e3a5692b272afb176be43d26002f3d"
+source = "git+https://github.com/EspressoSystems/Push-CDN?tag=0.1.25#a506712f553b4800ce3a36e62341c839153aeab3"
 dependencies = [
  "async-std",
  "cdn-proto",
@@ -1331,7 +1331,7 @@ dependencies = [
 [[package]]
 name = "cdn-marshal"
 version = "0.1.0"
-source = "git+https://github.com/EspressoSystems/Push-CDN?tag=0.1.24#5e0689dd07e3a5692b272afb176be43d26002f3d"
+source = "git+https://github.com/EspressoSystems/Push-CDN?tag=0.1.25#a506712f553b4800ce3a36e62341c839153aeab3"
 dependencies = [
  "async-std",
  "cdn-proto",
@@ -1346,7 +1346,7 @@ dependencies = [
 [[package]]
 name = "cdn-proto"
 version = "0.1.0"
-source = "git+https://github.com/EspressoSystems/Push-CDN?tag=0.1.24#5e0689dd07e3a5692b272afb176be43d26002f3d"
+source = "git+https://github.com/EspressoSystems/Push-CDN?tag=0.1.25#a506712f553b4800ce3a36e62341c839153aeab3"
 dependencies = [
  "anyhow",
  "ark-serialize",
--- a/crates/task-impls/src/consensus/view_change.rs
+++ b/crates/task-impls/src/consensus/view_change.rs
@@ -25,7 +25,11 @@ use async_std::task::JoinHandle;
 #[cfg(async_executor_impl = "tokio")]
 use tokio::task::JoinHandle;
 
-/// Update the view if it actually changed. Returns the spawned timeout task join handle.
+/// Update the view if it actually changed, takes a mutable reference to the `cur_view` and the
+/// `timeout_task` which are updated during the operation of the function.
+///
+/// # Errors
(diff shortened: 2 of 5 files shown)


crates/task-impls/src/consensus/view_change.rs
```diff
@@ -0,0 +1,132 @@
...
+pub(crate) async fn update_view<TYPES: NodeType, I: NodeImplementation<TYPES>>(
+    public_key: TYPES::SignatureKey,
+    new_view: TYPES::Time,
+    event_stream: &Sender<Arc<HotShotEvent<TYPES>>>,
+    quorum_membership: Arc<TYPES::Membership>,
+    quorum_network: Arc<I::QuorumNetwork>,
+    timeout: u64,
+    consensus: Arc<RwLock<Consensus<TYPES>>>,
+    cur_view: &mut TYPES::Time,
+    timeout_task: &mut Option<JoinHandle<()>>,
```
rob-maron, line 43: Question: what was the reason for changing this function from being a member of the consensus task state struct?

rob-maron (approved):
Looks good, just one question

crates/task-impls/src/consensus/view_change.rs
jparr721: We need to be able to use it within the `QuorumProposalRecv` task. In order to process that specific event, we need to call `update_view` at some point, but we cannot without inheriting a ton of the baggage of the consensus task.

ss-es (commented):
Looks fine to me, but 3 general questions/comments:

1) is the plan to merge this soon (i.e. before #2968)? I can start rebasing if so

2) For some reason, I thought the standard practice was to have a `consensus.rs` at the same level as `consensus/` that re-exports things in `consensus/`. But I don't personally have a strong opinion, and I couldn't find a strong reason either way in a quick search.

3) A thought I had while looking at some of these functions earlier was that they should really be methods in an `impl ConsensusTaskState {}` and just take `&Self`. I think putting them in separate files precludes that?

Cargo.lock
```diff
@@ -1331,7 +1331,7 @@ dependencies = [
 [[package]]
 name = "cdn-marshal"
 version = "0.1.0"
-source = "git+https://github.com/EspressoSystems/Push-CDN?tag=0.1.24#5e0689dd07e3a5692b272afb176be43d26002f3d"
+source = "git+https://github.com/EspressoSystems/Push-CDN?tag=0.1.25#a506712f553b4800ce3a36e62341c839153aeab3"
 dependencies = [
  "async-std",
  "cdn-proto",
```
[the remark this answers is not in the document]
jparr721: Fixed!

ss-es (approved):

shenkeyao (approved):
LGTM

PR #2987, opened 2024-04-18
[Builder] Use correct commitment when querying builder

QuentinI:
Closes #0000

### This PR: 
Uses last known `BuilderCommitment` when querying builder for available blocks instead of last *decided* `BuilderCommitment`

### Key places to review: 
`crates/task-impls/src/transactions.rs`

### DIFF
--- a/crates/task-impls/src/transactions.rs
+++ b/crates/task-impls/src/transactions.rs
@@ -12,15 +12,17 @@ use hotshot_builder_api::block_info::{
 use hotshot_task::task::{Task, TaskState};
 use hotshot_types::{
     consensus::Consensus,
+    data::Leaf,
     event::{Event, EventType},
     traits::{
         block_contents::{BlockHeader, BuilderFee},
         consensus_api::ConsensusApi,
         election::Membership,
-        node_implementation::{NodeImplementation, NodeType},
+        node_implementation::{ConsensusTime, NodeImplementation, NodeType},
         signature_key::{BuilderSignatureKey, SignatureKey},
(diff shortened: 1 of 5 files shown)


ss-es (approved):
Makes sense to me, but the builder team should probably also have a look.

crates/task-impls/src/transactions.rs
```diff
@@ -164,11 +166,47 @@ impl<
         None
     }
 
+    /// Get last known builder commitment from consensus.
+    async fn latest_known_builder_commitment(&self) -> BuilderCommitment {
+        let consensus = self.consensus.read().await;
+
+        let mut prev_view = TYPES::Time::new(self.cur_view.saturating_sub(1));
+
+        // Search through all previous views...
```
move47, line 175: Does this mean searching through all previous decided views? If yes, isn't that inefficient?

QuentinI: We'll stop almost immediately in practice, I don't think it's a huge concern.
Ideally we shoild explicitly store last known builder commitment, you are right

ss-es: I believe it searches until it hits a decided view, and breaks out of the loop (and the rest of the function) at that point.

We could do a binary search or something else more asymptotically worst-case efficient. But I don't think we'll ever be that far from a decided leaf, so I feel like a simple linear backwards search makes more sense.

move47: Thanks! let's merge this.

PR #3000, opened 2024-04-19
Fix builder URL, part 2

QuentinI:
Closes #0000

### DIFF
--- a/crates/testing/src/block_builder.rs
+++ b/crates/testing/src/block_builder.rs
@@ -8,6 +8,8 @@ use std::{
 
 use async_compatibility_layer::art::{async_sleep, async_spawn};
 use async_lock::RwLock;
+#[cfg(async_executor_impl = "async-std")]
+use async_std::task::spawn_blocking;
 use async_trait::async_trait;
 use committable::{Commitment, Committable};
 use futures::{future::BoxFuture, Stream, StreamExt};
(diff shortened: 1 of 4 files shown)


crates/testing/src/block_builder.rs
```diff
@@ -527,10 +533,14 @@ fn build_block<TYPES: NodeType>(
...
     let commitment = block_payload.builder_commitment(&metadata);
 
-    let (vid_commitment, precompute_data) = precompute_vid_commitment(
-        &block_payload.encode().unwrap().collect(),
-        num_storage_nodes,
-    );
+    let encoded_payload = block_payload.encode().unwrap().collect();
+    let vid_result =
+        spawn_blocking(move || precompute_vid_commitment(&encoded_payload, num_storage_nodes))
+            .await;
```
ss-es, line 539: Not a big deal since it's just in testing, though I'm slightly worried about this `spawn_blocking` and would need to think about it before I could definitely say "it's fine" (maybe someone else who's more familiar with async than me knows more).

I don't think there's any reason we wouldn't want to just block on this calculation in the test though, right?

ss-es (dismissed):

crates/testing/src/block_builder.rs
QuentinI: It's by analogy with `crates/task-impls/src/helpers.rs`
I think we've decided in Zulip that that's what we want for cpu-bound operations in async context

ss-es: My understanding was: the reason we use `spawn_blocking` is to allow the executor to switch to another async thread while performing a sync calculation. With HotShot in its entirety, I think this makes sense -- there's lots of async work that could be done (by other tasks) while waiting for the vid calculation to return. In the context of the test, I don't think there's any work that could be done while waiting for `build_block` to return.

And I was under the impression that `spawn_blocking` should be used sparingly, because you could deadlock the executor if it gets nested.

...but after writing this, I think I also convinced myself this is a good thing -- this would only ever make deadlocks more likely in testing, which means we would catch it, and it might even be a good thing if the tests are more susceptible to deadlocks.

ss-es (dismissed):

ss-es (approved):

PR #3006, opened 2024-04-19
Propose empty block

rob-maron:
No linked issue

### This PR: 

- Sets the builder to propose an empty block if it does not have one in time to propose
- Adds a timeout on the available blocks request for the propose time

### DIFF
--- a/crates/task-impls/src/transactions.rs
+++ b/crates/task-impls/src/transactions.rs
@@ -161,8 +161,7 @@ impl<
                     );
 
                     // Calculate the builder fee for the empty block
-                    let Some(builder_fee) =
-                        null_block::builder_fee(self.membership.total_nodes() as usize)
+                    let Some(builder_fee) = null_block::builder_fee(self.membership.total_nodes())
                     else {
                         error!("Failed to get builder fee");
                         return None;
--- a/crates/task-impls/src/transactions.rs
+++ b/crates/task-impls/src/transactions.rs
@@ -290,7 +290,9 @@ impl<
 
                 // We timed out while getting available blocks
                 Err(err) => {
-                    error!(%err, "Timeout while getting available blocks");
+                    if latest_block.is_none() {
+                        error!(%err, "Timeout while getting available blocks");
+                    }
                     break;
                 }
             };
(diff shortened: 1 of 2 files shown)


jbearer (approved):

crates/task-impls/src/transactions.rs
```diff
@@ -231,25 +262,38 @@ impl<
...
                     // pause a bit
                     async_sleep(Duration::from_millis(100)).await;
                     continue;
                 }
+
+                // We timed out while getting available blocks
+                Err(err) => {
+                    error!(%err, "Timeout while getting available blocks");
+                    // pause a bit
+                    async_sleep(Duration::from_millis(100)).await;
```
QuentinI, line 293: Not sure I understand this bit. If timeout is `propose_max_round_time()`, aren't we are out of time and need to propose whatever we have ASAP, not try again?

crates/task-impls/src/transactions.rs
```diff
@@ -231,25 +262,38 @@ impl<
                 continue;
             };
 
-            let mut available_blocks = match self
-                .builder_client
-                .get_available_blocks(
+            let mut available_blocks = match async_compatibility_layer::art::async_timeout(
+                self.api.propose_max_round_time(),
```
QuentinI, line 266: Maybe tighten it to `self.api.propose_max_round_time() - task_start_time.elapsed()`?

crates/task-impls/src/transactions.rs
bfish713: Agree, I think just `return None;` here

rob-maron: That makes sense, this was to just a stopgap to prevent us fromn being dependent on the builder, which can hang sometimes

QuentinI: > Agree, I think just `return None;` here

We may already have something in `latest_block`, should be `break;`

crates/task-impls/src/transactions.rs
```diff
@@ -231,25 +262,37 @@ impl<
...
                     error!(%err, "Couldn't get available blocks");
                     // pause a bit
                     async_sleep(Duration::from_millis(100)).await;
                     continue;
                 }
+
+                // We timed out while getting available blocks
+                Err(err) => {
+                    error!(%err, "Timeout while getting available blocks");
+                    return None;
```
QuentinI:
```suggestion
                    break;
```
We may have something in `latest_block` from previous iteration

QuentinI (dismissed):
Looks good!

QuentinI (approved):

PR #3013, opened 2024-04-22
Nodes subscribe to `transactions` topic only when they need transactions

### LINKED ISSUE #1992: Implement Libp2p optimistic transaction subscriptions
Related to #1991. 

For censorship resistance purposes, right now all Libp2p nodes gossip about all transactions. Design a way to subscribe to relevant transactions, most likely subscribing only when you are soon-to-be leader. This helps prevent duplicate transactions.

lukaszrzasik:
Closes #1992

### This PR: 

* Creates three new methods in the `ConnectedNetwork` trait:
* `subscribe_transactions` which subscribes to the new `Transactions` topic
* `unsubscribe_transactions` which unsubscribes from the `Transactions` topic
* `publish_transaction` which publishes a message to the `Transactions` topic
* implements these three methods for networks: `Libp2pNetwork`, `PushCdnNetwork` and `CombinedNetworks`
* subscribes the node to the topic when it is to become the leader in the next view
* unsubscribes the node from the topic when the view has been updated and the node is not the leader in the next view

### DIFF
--- a/crates/hotshot/src/traits/networking/combined_network.rs
+++ b/crates/hotshot/src/traits/networking/combined_network.rs
@@ -415,12 +415,12 @@ impl<TYPES: NodeType> ConnectedNetwork<Message<TYPES>, TYPES::SignatureKey>
         &self,
         message: Message<TYPES>,
         bind_version: VER,
-    ) -> anyhow::Result<()> {
+    ) -> Result<(), NetworkError> {
         let primary = self.primary().clone();
         let secondary = self.secondary().clone();
         let primary_message = message.clone();
         let secondary_message = message.clone();
-        Ok(self.send_both_networks(
+        self.send_both_networks(
             message,
             async move {
                 primary
@@ -433,7 +433,7 @@ impl<TYPES: NodeType> ConnectedNetwork<Message<TYPES>, TYPES::SignatureKey>
                     .await
             },
         )
-        .await?)
+        .await
     }
 
     async fn vid_broadcast_message<VER: StaticVersionType + 'static>(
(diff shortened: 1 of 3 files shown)


bfish713 (commented):
From my understanding this PR allows us to subscribe to transactions for views which we are going to be leader, just before being leader.  I think this made sense for previous versions of the code where we were building blocks ourselves.  Now hotshots job as it relates to transactions is to gossip/share them with any node who cares to create a public mempool.  The builder also listens to this topic (right now via a hotshot node) to get transactions to build blocks.  In other words any node that can receive transactions from a user needs to be able to send it to all HotShot nodes that care (query server nodes).  For libp2p if a node is not in the topic, it can't necessarily gossip to that topic and joing/leaving a topic sporadically is not something you want to do either.

The code changes for subscribing/unsubscribing looks good, we probably want a different way to decide if a node cares about the transaction topic.

crates/hotshot/src/traits/networking/combined_network.rs
```diff
@@ -411,6 +411,31 @@ impl<TYPES: NodeType> ConnectedNetwork<Message<TYPES>, TYPES::SignatureKey>
         .await
     }
 
+    async fn publish_transaction<VER: StaticVersionType + 'static>(
```
jparr721, line 414: If we want, this could be an anyhow and we just map the error from whatever legacy error type.

crates/task-impls/src/network.rs
```diff
@@ -408,6 +410,18 @@ impl<
                     self.version = version;
                     return None;
                 }
+                HotShotEvent::SubscribeTransactions => {
+                    if let Err(e) = self.channel.subscribe_transactions().await {
+                        warn!("Could not subscribe to transactions: {:?}", e);
```
jparr721, line 415: These could be `warn!(?e, "...")` if you want.

jparr721 (commented):
Looks good.

crates/task-impls/src/network.rs
jparr721: Yeah by all means feel free to leave it as is! Just tossing it out there.

crates/hotshot/src/traits/networking/combined_network.rs
jparr721: Yeah I think based on our discussion earlier it might be worth evaluating if this is internal or external, and we can do errors depending on what it ends up being.

bfish713:
> 2\. publish_transaction_async

`publish_transaction_async` is the function that is doing the gossiping (it's handled at the networking layer).  So basically a node gets a transaction, then it broadcast that transaction to all the other nodes, and finally each node upon receiving the broadcast transaction sends the transaction to the builder.

rob-maron:
The CDN changes look good 👍

bfish713:
Should we close this (and revive the branch if we really need this)?  I think we aren't going to work on this anytime soon

PR #3014, opened 2024-04-22
Demote update_view log

ss-es:
### This PR: 
Demotes the log produced by the call to `update_view` in consensus to `trace`, down from `warn`.

### Key places to review: 
Should it be any higher?

Note that this is a relatively large log that happens at least once every view, because we call `update_view` twice each view (and the second call will always fail). 

But I don't think it produces any useful information. View changes are logged on seeing a `ViewChange` event, and almost all (if not all) errors from `update_view` are expected anyway.

### DIFF
--- a/crates/task-impls/src/consensus/mod.rs
+++ b/crates/task-impls/src/consensus/mod.rs
@@ -438,7 +438,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> + 
                 )
                 .await
                 {
-                    warn!("Failed to update view; error = {e:?}");
+                    tracing::trace!("Failed to update view; error = {e:?}");
                 }
 
                 let consensus = self.consensus.upgradable_read().await;
@@ -1066,7 +1066,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> + 
                 )
                 .await
                 {
-                    warn!("Failed to update view; error = {e:?}");
+                    tracing::trace!("Failed to update view; error = {e:?}");
                     return;
                 }
 


jparr721 (dismissed):

crates/task-impls/src/consensus/mod.rs
```diff
@@ -1050,6 +1051,9 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
                     }
                 }
 
+                // Clear the payload commitment and metadata
+                self.payload_commitment_and_metadata = None;
+
```
bfish713, line 1056: I don't think we want to clear this here because we could have set the option before a view change event and then clear it before we can propose with it

ss-es: Yeah, I suspected this would cause problems -- changed to a clear based on view number!

crates/task-impls/src/consensus/proposal_helpers.rs
```diff
@@ -276,7 +276,7 @@ pub fn validate_proposal_view_and_certs<TYPES: NodeType>(
     if proposal.data.justify_qc.get_view_number() != view - 1 {
         let received_proposal_cert =
             proposal.data.proposal_certificate.clone().context(format!(
-"Quorum proposal for view {} needed a timeout or view sync certificate, but did not have one",
+                "Quorum proposal for view {} needed a timeout or view sync certificate, but did not have one",
```
jparr721, line 279: Oh good catch, rustfmt is so weird sometimes.

crates/task-impls/src/consensus/proposal_helpers.rs
```diff
@@ -487,9 +488,14 @@ async fn publish_proposal_from_commitment_and_metadata<TYPES: NodeType>(
     // FIXME - This is not great, and will be fixed later.
     // If it's > July, 2024 and this is still here, something has gone horribly wrong.
     let cnm = commitment_and_metadata
-        .clone()
+        .take()
```
jparr721, line 491: So then should we remove the below setting commit_and_metadata to None explicitly?

jparr721 (approved):
Great work!

crates/types/src/consensus.rs
```diff
@@ -406,6 +406,8 @@ pub struct CommitmentAndMetadata<TYPES: NodeType> {
     pub metadata: <TYPES::BlockPayload as BlockPayload>::Metadata,
     /// Builder fee data
     pub fee: BuilderFee<TYPES>,
+    /// View number this block is for
+    pub block_view: TYPES::Time,
```
jparr721, line 410: Quick question about this: how does this differ from `cur_view` and does this clobber that value? Like, will these values ever differ?

ss-es: My understanding is that we request 1 view in advance, so `cur_view` isn't really descriptive. I think `block_view` is more accurate; we're requesting a block to propose with in a specific (future) view. But I'm not that familiar with this code so there might be some edge cases here that I'm missing.

(Also note that the struct does not have a `cur_view`.)

shenkeyao:
> Question for reviewers: is there a reason we don't exit early if `update_view` fails on `QuorumProposalRecv`? I haven't fully traced through the logic, but I can investigate unless there's a reason for this.

Looks like `update_view` can only fail if one of the `get_u64()` calls in that function returns an error. Because it's not a consensus or safety error, I think it makes sense to still proceed to the code after it.

PR #3021, opened 2024-04-23
Conditional DA

rob-maron:
No linked issue.

### This PR: 

- Changes the orchestrator to supply whether or not nodes are DA
- Changes certificate signatures to require a threshold based on the  number of DA nodes connected instead of the expected value
- Changes [PushCDN/Libp2p] networks to not subscribe to DA if they are not
- Changes Libp2p to use the `da_keys` specified therein instead of by node index
- Edits some configs to support this

### This PR does not: 

- Automatically make the sequencer do this upstream

### Key places to review: 

The certificate signature threshold change for correctness

### DIFF
--- a/crates/examples/infra/mod.rs
+++ b/crates/examples/infra/mod.rs
@@ -296,7 +296,7 @@ pub fn load_config_from_file<TYPES: NodeType>(
     // but its type is too complex to load so we'll generate it from seed now.
     // Also this function is only used for orchestrator initialization now, so this value doesn't matter
     config.config.my_own_validator_config =
-        ValidatorConfig::generated_from_seed_indexed(config.seed, config.node_index, 1);
+        ValidatorConfig::generated_from_seed_indexed(config.seed, config.node_index, 1, true);
     // initialize it with size for better assignment of peers' config
     config.config.known_nodes_with_stake =
         vec![PeerConfig::default(); config.config.num_nodes_with_stake.get() as usize];
@@ -403,7 +403,8 @@ pub trait RunDA<
         });
 
         let committee_election_config = TYPES::Membership::default_election_config(
-            config.config.da_staked_committee_size.try_into().unwrap(),
+            // Use the number of _actual_ DA nodes connected for the committee size
+            config.config.known_da_nodes.len() as u64,
             config.config.num_nodes_without_stake as u64,
         );
         let networks_bundle = Networks {
@@ -751,9 +752,9 @@ where
             private_key: key.private_key,
         };
 
-        // See if we should be DA
+        // See if we should be DA, subscribe to the DA topic if so
         let mut topics = vec![Topic::Global];
-        if config.node_index < config.config.da_staked_committee_size as u64 {
+        if config.config.my_own_validator_config.is_da {
             topics.push(Topic::DA);
         }
 
@@ -767,6 +768,9 @@ where
             keypair,
         )
         .expect("failed to create network");
+
+        // Wait for the network to be ready
+        network.wait_for_ready().await;
 
         PushCdnDaRun {
             config,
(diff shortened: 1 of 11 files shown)


crates/examples/infra/mod.rs
```diff
@@ -1038,6 +1044,7 @@ pub async fn main_entry_point<
             my_own_validator_config,
             args.advertise_address,
             Some(libp2p_public_key),
+            true,
```
jparr721, line 1047: Is it possible to get a comment on the true and the false above so a reader knows what it's doing? Perhaps something like `true /* is_da */` or a constant `IS_DA`

crates/orchestrator/src/config.rs
```diff
@@ -710,20 +730,34 @@ impl<KEY: SignatureKey, E: ElectionConfig> From<ValidatorConfigFile> for HotShot
 
 impl<KEY: SignatureKey> Default for HotShotConfigFile<KEY> {
     fn default() -> Self {
+        let staked_committee_nodes: usize = 5;
```
jparr721, line 733: Is 5 significant here? Or is it just a decided on number? If it's a constant, should it perhaps be notated as such?

jparr721 (commented):
LGTM, couple of questions. Will defer to someone more familiar for an approval.

crates/examples/infra/mod.rs
rob-maron: Yes, good point. This is actually `indexed_da` which controls whether or not being DA is governed by the node index. It's because we both get the node index and send our DA status within this function, so it needed to be done in here. It probably makes more sense to break the function out, but I wanted to make it work with minimal changes

crates/orchestrator/src/config.rs
```diff
@@ -710,20 +730,35 @@ impl<KEY: SignatureKey, E: ElectionConfig> From<ValidatorConfigFile> for HotShot
...
+
+        // Aggregate the DA nodes
+        let mut known_da_nodes = HashSet::new();
+
         let gen_known_nodes_with_stake = (0..10)
             .map(|node_id| {
-                let cur_validator_config: ValidatorConfig<KEY> =
-                    ValidatorConfig::generated_from_seed_indexed([0u8; 32], node_id, 1);
+                let mut cur_validator_config: ValidatorConfig<KEY> =
+                    ValidatorConfig::generated_from_seed_indexed([0u8; 32], node_id, 1, false);
```
ss-es, line 742: Is `is_da` meant to be `false` here?

ss-es (approved):
I think this looks fine to me, though I don't know if I'd be able to tell if there was something wrong

crates/examples/infra/mod.rs
dailinsubjam: Just want to check my understanding, is the flow in examples be like: a node post and register its key pairs to the orchestrator -> the orchestrator give it a `node_index` and use this `node_index` to determine whether it is a da node. 
While in other scenarios it's possible that they do not use `node_index` to determine whether a node is part of DA committee, then they need to call `get_complete_config(..., false)` .

crates/orchestrator/api.toml
```diff
@@ -32,11 +32,13 @@ Get the latest temporary node index only for generating validator's key pair for
 
 # POST the node's node index to generate public key for pubkey collection
```
dailinsubjam, line 33: Let's remove `to generate public key` if it's easy to update, but feel free to leave it to later PR.

dailinsubjam (approved):
Thanks for the detailed comment! LGTM. BTW which file includes the certificate signature threshold change?

crates/examples/infra/mod.rs
rob-maron: That's correct, the only difference being that the orchestrator does not determine who is DA, that happens locally after the node gets its `node_index`. Then in the second call to the orchestrator it tells the orchestrator it is DA

rob-maron:
> Thanks for the detailed comment! LGTM. BTW which file includes the certificate signature threshold change?

This should be included in the changes in `crates/testing/src/task_helpers.rs`

When we do the sequencer, I will have to update that one in a similar fashion

PR #3022, opened 2024-04-23
Move instance_state out of lock

bfish713:
Moves `instance_state` into it's own arc that gets passed around.  This fixes the issue where we hold a consensus lock while we do catchup in the sequencher.

### DIFF
--- a/crates/hotshot/src/lib.rs
+++ b/crates/hotshot/src/lib.rs
@@ -113,7 +113,6 @@ pub struct Memberships<TYPES: NodeType> {
 }
 
 /// Holds the state needed to participate in `HotShot` consensus
-#[derive(Clone)]
 pub struct SystemContext<TYPES: NodeType, I: NodeImplementation<TYPES>> {
     /// The public key of this node
     public_key: TYPES::SignatureKey,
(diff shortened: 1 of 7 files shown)


crates/hotshot/src/lib.rs
```diff
@@ -180,7 +202,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> SystemContext<TYPES, I> {
     ) -> Result<Arc<Self>, HotShotError<TYPES>> {
         debug!("Creating a new hotshot");
 
-        let consensus_metrics = Arc::new(metrics);
+        let consensusmetrics = Arc::new(metrics);
```
ss-es, line 205: should this be `consensus_metrics`?

ss-es (approved):
Looks good to me

crates/hotshot/src/lib.rs
bfish713: yeah find and replace of _metrics got me

crates/hotshot/src/lib.rs
```diff
@@ -159,6 +161,26 @@ pub struct SystemContext<TYPES: NodeType, I: NodeImplementation<TYPES>> {
     /// Reference to the internal storage for consensus datum.
     pub storage: Arc<RwLock<I::Storage>>,
 }
+impl<TYPES: NodeType, I: NodeImplementation<TYPES>> Clone for SystemContext<TYPES, I> {
```
nyospe, line 164: Ouch, caught by the classic `Arc<T::X>` where `T::X` not locally constrained to `Clone` issue.

ss-es (approved):

PR #3040, opened 2024-04-25
Move `QuorumProposalValidated` to its own function

jparr721:
Closes #<ISSUE_NUMBER>

### This PR: 
While making the `QuorumProposalRecv` task, I noticed that I need to be able to gate the proposal behavior within the `QuorumProposalValidated` event. To do this without duplicating the functionality or making the PR extremely massive I have opted to push the change here so that way it's easier to review.

### This PR does not: 
No functionality is changed, it is copied verbatim.

### DIFF
--- a/crates/task-impls/src/consensus/mod.rs
+++ b/crates/task-impls/src/consensus/mod.rs
@@ -1,21 +1,17 @@
-use std::{
-    collections::{BTreeMap, HashMap, HashSet},
-    sync::Arc,
-};
+use std::{collections::BTreeMap, sync::Arc};
 
 use anyhow::Result;
 use async_broadcast::Sender;
-use async_lock::{RwLock, RwLockUpgradableReadGuard};
+use async_lock::RwLock;
 #[cfg(async_executor_impl = "async-std")]
 use async_std::task::JoinHandle;
-use chrono::Utc;
 use committable::Committable;
 use futures::future::join_all;
 use hotshot_task::task::{Task, TaskState};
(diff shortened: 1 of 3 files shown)


crates/task-impls/src/consensus/mod.rs
```diff
@@ -385,186 +381,11 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> ConsensusTaskState<TYPES, I>
...
-                        warn!("Failed to propose; error = {e:?}");
-                    };
-                }
-
-                if !self.vote_if_able(&event_stream).await {
-                    return;
+                if let Err(e) =
+                    handle_quorum_proposal_validated(proposal, event_stream.clone(), self).await
+                {
+                    warn!(?e, "Failed to handle QuorumProposalValidated event");
```
ss-es, line 387: I think this should be `debug` or `info` at best, not `warn` since the function calls `vote_if_able`, which can fail for multiple reasons (some benign).

ss-es: Also, I think `?e` uses the `Debug` formatter which prints a backtrace (which I don't think we want here, that's what made the `update_view` log really verbose last time)

I think `warn!("Failed to handle ...: {e}");` might avoid that, but I'd have to test (can check later tonight)

jparr721: No problem, let me check this one

ss-es (approved):
Looks good to me

PR #3041, opened 2024-04-25
Simplify interaction with builder

jbearer:
* Remove block building parameters (`min_transactions`, `max_transactions`, `propose_min_round_time`, `propose_max_round_time`) from HotShot, let the builder deal with that
* Loop until we successfully get blocks (or time out), but no more than that; take a block from the first set that we receive
* Choose block by max fee per byte, not max total fee
* Apply timeout to entire builder interaction, not just `get_available_blocks`


config/ValidatorConfigOutput
```diff
@@ -1 +1 @@
...
...123227923636811928330606075281145117212394 + 15807017392833049888165434456991157794698032464874424842715555348468160607934 * u), QuadExtField(7996517616082121122160563552650547601395271017260499735456299700133762512689 + 7504045709281061282278228438613345070383424761478787301859187055302953740948 * u), QuadExtField(1515973040548822760825076242090160370742046237881440422068330135941139244581 + 20251846261653098602911417004145145971080304248810966341160788194007704966108 * u))), private_key: SignKey(BigInt([3505488234151006356, 6655477166151225138, 3291219027844407676, 2153641080015542578])), stake_value: 1, state_key_pair: StateKeyPair(KeyPair { sk: SignKey(BigInt([2822822805887490846, 6664316196088353173, 4926510007447087464, 116097479308258694])), vk: VerKey(Projective { x: BigInt([11315198235793138814, 4744451806709910489, 6921831025042192557, 1125393823825936625]), y: BigInt([13035879815613524256, 18225673961538637854, 12006860967936477969, 1516668567229692859]), t: BigInt([13450777528397789701, 12242009376162249168, 12596256366242272750, 3368076418495976469]), z: BigInt([10465708325245823445, 13967918689717629445, 14943426723808572731, 621075342718756551]) }) }), is_da: false }
```
ss-es, line 1: I'm not sure what this file is or what this change is

cc: @mrain I think you were the last person to edit this file

ss-es (approved):
Makes sense to me and it appears the examples are still committing a normal amount of transactions in CI, so I think this should be fine

Approving, but I'm not that familiar with the builder (if you wanted to wait for a more careful review from someone else)

config/ValidatorConfigOutput
jbearer: Huh I also don't know what this is or where this came from. Definitely didn't mean to change it

jbearer: It's in the .gitignore so maybe it got accidentally committed at some point?

crates/hotshot/src/lib.rs
```diff
@@ -690,20 +690,8 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> ConsensusApi<TYPES, I>
         self.hotshot.config.num_nodes_with_stake
     }
 
-    fn propose_min_round_time(&self) -> Duration {
```
nyospe, line 693: H-uh. I guess I should've checked whether someone else was doing this.

crates/task-impls/src/transactions.rs
```diff
@@ -243,175 +244,140 @@ impl<
...
                 self.api
-                    .propose_max_round_time()
+                    .builder_timeout()
                     .saturating_sub(task_start_time.elapsed()),
-                self.builder_client.get_available_blocks(
-                    parent_commitment,
-                    self.public_key.clone(),
-                    &request_signature,
-                ),
+                self.get_block_from_builder(parent_comm, &parent_comm_sig),
```
nyospe, line 264: Ohhhh... so, okay, this doesn't outright make things worse, but... it doesn't make them better either.

So, the implementation here basically took what the builder API was doing, said, "eh, let's ignore all of this and why it was designed this way, and take the most reductive and least efficient approach we can manage". And now that's been tied up with a bow into a nice opaque function.

We have three API calls.

The first should allow HotShot to query any number of builders for their available blocks, and (assuming we ever factor size) wait some amount of time to collect all results that arrive, and choose between the offerings based on fee and size.

The second should fetch one of those blocks - with everything needed to start the next step, e.g. sending it to the DA committee. It should return as close to immediately as it takes to send the response.

The third won't return immediately, not if it's called too fast, because it only starts computing the VID precompile after it builds the block. (note: we may even change it to start computing after the second API request is received)

What this did after the loop wasn't great, but at least it didn't bury the discreet parts even one level of coupling deeper...

nyospe (commented):
Thank you for getting rid of the loop.

I'm glad that part got done, but I'm not thrilled with where this went...

nyospe:
So the issue here is, you set out to "simplify interaction with the builder", but the interaction had already been oversimplified.

crates/hotshot/src/lib.rs
jbearer: Oop, me too 🙊

crates/task-impls/src/transactions.rs
jbearer: I'm not sure the "burying one level deeper" is really a problem. The only reason for having a separate function there is to make the retry loop a little simpler. We would want something along these lines regardless, even if it eventually only does the first API call in the inner loop. Other than that, the logic here is exactly the same as before except we won't call the same builder multiple times until we get a certain number of transactions or whatever.

I totally understand what you're saying about the correct way to use the API, and that would absolutely be an improvement. But, that's a bigger change to consensus than I want to make right now (especially the starting DA before VID part) and I'm inclined to go ahead with something along these lines anyways, because
* it will eliminate a variable when debugging the builder (problems arising due to repeat requests, plus this has slightly better timeout handling)
* it will simplify the config down in the sequencer by removing some parameters

nyospe (approved):
Okay. It'll need refactoring later, but this does solve the immediate issue.

crates/task-impls/src/builder.rs
```diff
@@ -73,7 +73,9 @@ impl<TYPES: NodeType, Ver: StaticVersionType> BuilderClient<TYPES, Ver> {
     /// If the URL is malformed.
     pub fn new(base_url: impl Into<Url>) -> Self {
         Self {
-            inner: Client::new(base_url.into().join("block_info").unwrap()),
+            inner: Client::builder(base_url.into().join("block_info").unwrap())
+                .set_timeout(Some(Duration::from_secs(2)))
```
QuentinI, line 77: Don't we want it to be equal to `builder_timeout`?

jbearer: I was thinking the client shouldn't know about consensus parameters, and this would just be a general "a simple HTTP request should never take longer than 2s or something is wrong"

jbearer: I think it's fine because we have the `builder_timeout` logic in consensus anyways

PR #3045, opened 2024-04-26
Fix conditional DA: remove node indexing

rob-maron:
No issue linked.

### This PR: 

Fixes a bug wherein node indexing breaks the conditional DA changes.
- Removes `ElectionConfig` (a new type will be added for benchmarking purposes)
- Changes the membership constructor to reflect the changes

I also had to change the testbed to support generating a DAC with the DA membership instead of the quorum membership. As it turns out, the tests were actually pinning DA nodes == quorum nodes.

See: https://github.com/EspressoSystems/HotShot/blob/1bd6c0b12dd30547dd7507c57f32acaf80bce828/crates/testing/src/task_helpers.rs#L64-L76
 

### Key places to review: 

- `static_committee.rs`
- All places where the test helpers were changed

### How to test this PR:

- Has been tested with HotShot
- Am testing with the sequencer extensively tomorrow


rob-maron:
Drafting to fix test

crates/examples/infra/mod.rs
```diff
@@ -314,11 +311,7 @@ pub async fn run_orchestrator<
     OrchestratorArgs { url, config }: OrchestratorArgs<TYPES>,
 ) {
     println!("Starting orchestrator",);
-    let _result = hotshot_orchestrator::run_orchestrator::<
-        TYPES::SignatureKey,
-        TYPES::ElectionConfigType,
-    >(config, url)
-    .await;
+    let _result = hotshot_orchestrator::run_orchestrator::<TYPES::SignatureKey>(config, url).await;
```
jparr721, line 314: Curious, why do we bother storing this value?

crates/examples/infra/mod.rs
```diff
@@ -416,22 +396,22 @@ pub trait RunDA<
...
             vid_membership: <TYPES as NodeType>::Membership::create_election(
                 known_nodes_with_stake.clone(),
-                quorum_election_config.clone(),
+                known_nodes_with_stake.clone(),
                 config.config.fixed_leader_for_gpuvid,
             ),
             view_sync_membership: <TYPES as NodeType>::Membership::create_election(
                 known_nodes_with_stake.clone(),
-                quorum_election_config,
+                known_nodes_with_stake.clone(),
```
jparr721, line 414: does this last one need to clone? Not sure what the type is, but we can probably `move` the final invocation.

crates/hotshot/src/traits/election/static_committee.rs
```diff
@@ -123,44 +106,41 @@ where
...
+        all_nodes.retain(|entry| entry.stake_table_entry.get_stake() > U256::from(0));
+
+        debug!(
+            "Election Membership Size: {}",
+            committee_nodes_with_stake.len()
+        );
+
         Self {
-            nodes_with_stake,
+            all_nodes_with_stake: all_nodes
```
jparr721, line 140: This might be able to be `into_iter`? Since we pass ownership to the function above, we can just consume it here if it's not being used.

crates/orchestrator/src/config.rs
```diff
@@ -585,7 +579,7 @@ pub struct HotShotConfigFile<KEY: SignatureKey> {
     pub known_nodes_with_stake: Vec<PeerConfig<KEY>>,
     #[serde(skip)]
     /// The known DA nodes' public key and stake values
-    pub known_da_nodes: HashSet<PeerConfig<KEY>>,
+    pub known_da_nodes: Vec<PeerConfig<KEY>>,
```
jparr721, line 582: Curious: why this change?

crates/testing/src/task_helpers.rs
```diff
@@ -84,22 +70,22 @@ pub async fn build_system_handle(
...
         vid_membership: <TestTypes as NodeType>::Membership::create_election(
             known_nodes_with_stake.clone(),
-            quorum_election_config.clone(),
+            known_nodes_with_stake.clone(),
             config.fixed_leader_for_gpuvid,
         ),
         view_sync_membership: <TestTypes as NodeType>::Membership::create_election(
             known_nodes_with_stake.clone(),
-            quorum_election_config,
+            known_nodes_with_stake.clone(),
```
jparr721, line 88: same remark as above about moving the last call. Looks like there's a couple more spots below this. I won't mark all of them to avoid spamming.

crates/types/src/lib.rs
```diff
@@ -172,7 +170,7 @@ pub struct HotShotConfig<KEY: SignatureKey, ELECTIONCONFIG: ElectionConfig> {
     /// List of known node's public keys and stake value for certificate aggregation, serving as public parameter
     pub known_nodes_with_stake: Vec<PeerConfig<KEY>>,
     /// All public keys known to be DA nodes
-    pub known_da_nodes: HashSet<PeerConfig<KEY>>,
+    pub known_da_nodes: Vec<PeerConfig<KEY>>,
```
jparr721, line 173: Why this change?

jparr721 (commented):
LGTM, general comments.

crates/examples/infra/mod.rs
rob-maron: No idea lol, the change I made was to remove the generics. I can remove this though since it seems we're not using it

crates/examples/infra/mod.rs
rob-maron: No, good point

crates/hotshot/src/traits/election/static_committee.rs
rob-maron: Sure, I can make this change

crates/orchestrator/src/config.rs
rob-maron: Yeah so,

I spent a bit debugging why at one point these changes broke the DA validation. It turns out that the way we've implemented the QC, everyone has to have this in the same order. The HashSet was not preserving order when sent over the wire, so it was breaking validation

crates/types/src/lib.rs
rob-maron: Same as above, the ordering of the peers broke the validation because they're pulled in as the stake table

crates/orchestrator/src/config.rs
jparr721: Ahh, makes sense!

jparr721 (dismissed):

jparr721 (approved):

crates/examples/infra/mod.rs
```diff
@@ -393,27 +393,26 @@ pub trait RunDA<
...
+            config.config.fixed_leader_for_gpuvid,
+        );
+
+        // Create the quorum membership from all nodes, specifying the committee
+        // as the known da nodes
+        let da_membership = <TYPES as NodeType>::Membership::create_election(
+            known_nodes_with_stake.clone(),
+            known_nodes_with_stake,
+            config.config.fixed_leader_for_gpuvid,
+        );
```
rob-maron, line 409: Here

Issue #3037, opened 2024-04-25
[CX_CLEANUP] - Replace proposal handling in proposal and vote tasks with the new task


shenkeyao:
### What is this task and why do we need to work on it?

Quorum proposal handling in the proposal and vote tasks is duplicated, and can be replaced by the new `QuorumProposalRecv` task.

### What are the acceptance criteria to close this issue?

Quorum proposal is now only handled by the `QuorumProposalRecv` task.

shenkeyao:
Closed by https://github.com/EspressoSystems/HotShot/pull/3052.

PR #3052, opened 2024-04-26
[CX_CLEANUP] - Add Testing for QuorumProposalRecv Task

### LINKED ISSUE #3050: [CX_CLEANUP] - Add Testing for QuorumProposalRecv Task

The `QuorumProposalRecv` task needs testing before it can be integrated. We just need tests for the single event. We will also add `dependency-tasks` to CI so that way we can make sure that breaking changes are not being introduced.

_No response_

_No response_

Tests passing.

_No response_

jparr721:
Closes #3050 and #3037

### This PR: 
Tests the `QuorumProposalRecv` task and adds the `dependency-tasks` feature to CI. This PR explicitly seeks to fix only the `test-ci-1` suite as fixing all tests would make this PR a lot larger than it already is.

Highlights:
- Adds the quorum proposal recv task to `hotshot/src/lib.rs`
- Adds the quorum vote task polls and quorum proposal rec polls to `hotshot/src/tasks/mod.rs`
- Adds task state for quorum proposal recv
- Gates the liveness check but does *not* yet replace that code (follow-on PR will address)
- Begins the process of removing `cur_view` by relying instead on `latest_proposed_view` within quorum proposal task
- Fully removes bespoke code from quorum proposal task

### This PR does not: 
This PR does not attempt to fix the dependency-tasks failures. That will be in a follow-on PR, the expectation is that most of them will fail for now until we have integrated everything together.

### Key places to review: 
Please look specifically at `quorum_proposal` with special emphasis on the view number in the `QuorumProposalRecv` event.

This suite should pass
`just async_std test-ci-1 --features "dependency-tasks"`


crates/task-impls/src/consensus/view_change.rs
```diff
@@ -41,8 +41,8 @@ pub(crate) async fn update_view<TYPES: NodeType, I: NodeImplementation<TYPES>>(
     consensus: Arc<RwLock<Consensus<TYPES>>>,
     cur_view: &mut TYPES::Time,
     timeout_task: &mut Option<JoinHandle<()>>,
+    double_send: bool,
```
jparr721, line 44: This was why we got duplicate view changes, it recursively broadcasted a view change in the `ViewChange` event handler in the consensus task. I'll fix the naming later.

crates/task-impls/src/consensus/mod.rs
```diff
@@ -747,6 +752,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> ConsensusTaskState<TYPES, I>
                 let consensus = self.consensus.read().await;
                 consensus.metrics.number_of_timeouts.add(1);
             }
+            #[cfg(not(feature = "dependency-tasks"))]
```
shenkeyao, line 755: As mentioned in [Zulip](https://espresso.zulipchat.com/#narrow/stream/397846-Consensus-Team/topic/Proposal.20Task/near/436226938), looks like all the logic inside this `SendPayloadCommitmentAndMetadata` handling has been moved to the quorum proposal recv task, except for the `self.payload_commitment_and_metadata` setting. I think we should set it in the quorum proposal recv task as well, unless we find it's unnecessary, in which case we should remove it from here as well.

crates/hotshot/src/tasks/mod.rs
shenkeyao, line 1: Ahh looks like we missed an `inject_quorum_vote_polls` function. It's probably fine for now since the consensus task will poll for those vote-related events as well, but it's better to have a separate function for the vote task.

crates/task-impls/src/consensus/view_change.rs
shenkeyao: Hmm why do we need the double send?

crates/task-impls/src/quorum_proposal_recv.rs
```diff
@@ -120,7 +120,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalRecvTaskState<
         if let HotShotEvent::QuorumProposalRecv(proposal, sender) = event.as_ref() {
             match handle_quorum_proposal_recv(proposal, sender, event_stream.clone(), self).await {
                 Ok(Some(current_proposal)) => {
-                    self.cancel_tasks(proposal.data.get_view_number() + 1).await;
+                    tracing::error!("Cancelling tasks and voting");
```
shenkeyao, line 123: I think we can still call `cancel_tasks` but for the current view, not `+ 1`.

crates/task-impls/src/quorum_vote.rs
```diff
@@ -334,7 +336,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumVoteTaskState<TYPES, I
             );
 
             // Cancel the old dependency tasks.
-            for view in (*self.latest_voted_view + 1)..=(*new_view) {
+            for view in (*self.latest_voted_view)..=(*new_view) {
```
shenkeyao, line 339: Do we need to include `self.latest_voted_view` here? E.g., if `self.latest_voted_view` is 3, then previously, the `new_view` is 3, so we've already canceled the task for view 3. Now we only have to cancel the tasks from view 4.

crates/testing/tests/tests_1/libp2p.rs
```diff
@@ -44,6 +44,7 @@ async fn libp2p_network() {
 /// libp2p network test with failures
 #[cfg_attr(async_executor_impl = "tokio", tokio::test(flavor = "multi_thread"))]
 #[cfg_attr(async_executor_impl = "async-std", async_std::test)]
+#[ignore]
```
shenkeyao, line 47: Is this test not working?

crates/testing/tests/tests_1/quorum_proposal_task.rs
```diff
@@ -50,11 +47,11 @@ async fn test_quorum_proposal_task_quorum_proposal() {
 
     // We need to propose as the leader for view 2, otherwise we get caught up with the special
```
shenkeyao, line 48: Nit: update the comment to "view 1".

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -601,16 +491,16 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalTaskState<TYPE
...
-            info!(
-                "Updating next proposal view from {} to {} in the quorum proposal task",
+            debug!(
+                "Updating latest proposed view from {} to {} in the Quorum Proposal task",
                 *self.latest_proposed_view, *new_view
             );
 
             // Cancel the old dependency tasks.
-            for view in (*self.latest_proposed_view + 1)..=(*new_view) {
+            for view in (*self.latest_proposed_view)..=(*new_view) {
```
shenkeyao, line 503: Similar to my comment in `quorum_vote.rs`, I think we can keep the `+1`.

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -700,10 +588,13 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalTaskState<TYPE
                 _fee,
             ) => {
                 let view = *view;
-                debug!(
-                    "Got payload commitment {:?} for view {view:?}",
-                    payload_commitment
-                );
+
+                if self.quorum_membership.get_leader(view) == self.public_key {
```
shenkeyao, line 592: Shouldn't we skip the dependency creation if we are not the leader?

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -730,29 +621,19 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalTaskState<TYPE
...
                 let view = certificate.view_number;
 
-                if self.quorum_membership.get_leader(view) == self.public_key {
-                    debug!(
-                        "Attempting to publish proposal after forming a View Sync Finalized Cert for view {}",
-                        *certificate.view_number
-                    );
-                    self.create_dependency_task_if_new(view, event_receiver, event_sender, event);
-                }
+                self.create_dependency_task_if_new(view, event_receiver, event_sender, event);
```
shenkeyao, line 624: Same here, shouldn't we create the dependency only if we are the leader?

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -763,7 +644,14 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalTaskState<TYPE
...
                 let view = proposal.data.view_number;
                 if !self.update_latest_proposed_view(view).await {
-                    warn!("Failed to update latest proposed view");
+                    tracing::trace!("Failed to update latest proposed view");
+                    return;
+                }
+            }
+            HotShotEvent::QuorumProposalRecv(proposal, _) => {
+                let view = proposal.data.view_number;
+                if !self.update_latest_proposed_view(view).await {
```
shenkeyao, line 653: I'm not sure if we should call `update_latest_proposed_view` when receiving a proposal. If the proposal is invalid (since it may be from another node), wouldn't this cause an incorrect view update? I feel it's safer to only call `update_latest_proposed_view` if we are the proposal sender or after validating the proposal.

shenkeyao (commented):
I had some comments in places that I think need some modifications, but I don't think any of them should cause the tests to fail.

crates/testing/tests/tests_1/libp2p.rs
jparr721: Just noisy, so I ignore it until I get the non failure one going.

crates/task-impls/src/quorum_proposal.rs
jparr721: Wouldn't this statement imply that we are?

crates/task-impls/src/quorum_proposal.rs
jparr721: Ah yeah that makes sense.

crates/task-impls/src/quorum_proposal.rs
jparr721: We do, but editing *each* call site was kind of a code smell, so I changed it so the function itself will just ignore non leader initiations.

crates/task-impls/src/quorum_proposal_recv.rs
jparr721: This was just me tinkering around

crates/task-impls/src/quorum_vote.rs
jparr721: Yeah I was trying to debug if I had any long-running tasks laying around. I've reverted this

crates/task-impls/src/consensus/mod.rs
jparr721: It was unnecessary and it was removed. Thanks for highlighting!

shenkeyao (approved):
The update looks good!

QuentinI (dismissed):
Don't fully understand the consensus logic changes, so can't vouch for that
Otherwise LGTM

.github/workflows/build-and-test.yml
```diff
@@ -56,6 +56,47 @@ jobs:
...
           RUST_BACKTRACE: full
+  test-dependency-tasks:
+    strategy:
+      matrix:
+        just_variants:
+          - async-std
+          - tokio
+        test_suites:
+          - test-ci-1
+          # - test-ci-2
```
jparr721, line 67: These are intentionally commented out.

crates/task-impls/src/quorum_vote.rs
```diff
@@ -487,22 +487,6 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumVoteTaskState<TYPES, I
                     return;
                 }
             }
-            HotShotEvent::ViewChange(new_view) => {
```
jparr721, line 490: @shenkeyao This was erroneously killing polling for views since the latest voted view operates distinctly from `cur_view`. It appears as well that this doesn't actually update any state? Since we left ViewChange alive in consensus, I removed this to fix the web tests failing. Now everything seems to be working.

config/ValidatorConfigOutput
dailinsubjam, line 1: We could also add this to `.gitignore` as it might be generated locally if the test `gen_key_pair` is included in any general tests.

dailinsubjam (commented):
CI part and testing part LGTM. But for changes in Consensus might need someone more familiar to approve.

config/ValidatorConfigOutput
jparr721: It's in there, but gitignore will not retroactively remove it.

PR #3054, opened 2024-04-26
Add configurable start threshold for orchestrator

ss-es:
No linked issue.

### This PR: 
Adds a configurable start threshold for the orchestrator

### DIFF
--- a/crates/orchestrator/src/lib.rs
+++ b/crates/orchestrator/src/lib.rs
@@ -334,7 +334,11 @@ where
     fn post_ready(&mut self) -> Result<(), ServerError> {
         self.nodes_connected += 1;
         println!("Nodes connected: {}", self.nodes_connected);
-        if self.nodes_connected >= (self.config.config.num_nodes_with_stake.get() as u64) {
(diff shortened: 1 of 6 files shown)


crates/orchestrator/src/lib.rs
```diff
@@ -334,7 +334,11 @@ where
     fn post_ready(&mut self) -> Result<(), ServerError> {
         self.nodes_connected += 1;
         println!("Nodes connected: {}", self.nodes_connected);
-        if self.nodes_connected >= (self.config.config.num_nodes_with_stake.get() as u64) {
+        // i.e. nodes_connected >= num_nodes_with_stake * (start_threshold.0 / start_threshold.1)
+        if self.nodes_connected * self.config.config.start_threshold.1
```
jparr721, line 338: How come we aren't just doing the formula as described in the comment?

ss-es: You can't divide a `u64` and compare to a `u64`, and I think casting to `f64` is best avoided. You also don't have to worry about 0 in the denominator this way.

jbearer (approved):
LGTM

PR #3055, opened 2024-04-26
Add endpoint to force start the orchestrator

ss-es:
No linked issue.

### This PR: 
Adds an endpoint to the orchestrator's API that we can use to force-start it, even if we haven't yet met the threshold for the number of nodes connected.

Additionally:
- nodes no longer try to register their public key along with their `node_index`; instead, the node now sends the orchestrator its public key, and the orchestrator tells it what its `node_index` is.
- `get_complete_config` no longer attempts to load a config from a file. It gets the config straight from the orchestrator and only sets the node's `node_index`.

### Key places to review: 
You can ignore the `task-impls/` folder (just `rustfmt` changes)

The changes in `get_complete_config` should probably be looked at closely


crates/orchestrator/api.toml
```diff
@@ -22,7 +22,7 @@ This must be a POST request so we can update the OrchestratorState in the server
 received from the 'identity' endpoint
 """
 
-# GET the latest temporary node index only for generating validator's key pair
+# POST the latest temporary node index only for generating validator's key pair
```
elliedavidson, line 25: Why did we change this to POST?

crates/orchestrator/src/lib.rs
```diff
@@ -340,15 +355,63 @@ where
     // Assumes nodes do not post 'ready' twice
     // TODO ED Add a map to verify which nodes have posted they're ready
     fn post_ready(&mut self) -> Result<(), ServerError> {
+        if !self.accepting_connections {
```
elliedavidson, line 358: I don't think we want this here.  We should stop allowing people to register new keys once we've restarted.  We still want people to post whether or not they're ready, though.  I think we want to put this in the register keys function.

crates/orchestrator/src/lib.rs
```diff
@@ -248,6 +259,8 @@ where
     // Assumes nodes will set their own index that they received from the
     // 'identity' endpoint
     fn post_getconfig(&mut self, _node_index: u16) -> Result<NetworkConfig<KEY>, ServerError> {
+        self.manual_start_allowed = false;
```
elliedavidson, line 262: I don' t think we want this here.  Iirc, this function gets the initial config before keys have been registered.  (We actually don't really need this function at all and imo it just adds extra confusion, but that can be for another PR :) ).

crates/orchestrator/src/lib.rs
```diff
@@ -81,6 +82,10 @@ struct OrchestratorState<KEY: SignatureKey> {
     bench_results: BenchResults,
     /// The number of nodes that have posted their results
     nodes_post_results: u64,
+    /// Whether the orchestrator can be started manually
+    manual_start_allowed: bool,
+    /// Whether the orchestrator can be started manually
```
elliedavidson, line 87: Let's change the comment on accepting connections to be more descriptive :)

elliedavidson (changes requested):

crates/orchestrator/api.toml
dailinsubjam: I guess because now `&self` is mutable?

dailinsubjam (dismissed):

dailinsubjam (commented):
Oops, looks like examples are stuck...

dailinsubjam (approved):

crates/orchestrator/src/lib.rs
```diff
@@ -341,14 +363,52 @@ where
...
+                status: tide_disco::StatusCode::Forbidden,
+                message: format!("We cannot manually start the network, because we have {} public keys registered.", self.pub_posted.len())
+            });
+        }
+
+        if self.config.manual_start_password == Some(password) {
+            self.accepting_connections = false;
+            self.manual_start_allowed = false;
+            self.peer_pub_ready = true;
             self.start = true;
```
elliedavidson, line 409: I don't think we want this here.  In my view, a manual restart should stop the orchestrator from registering keys and make the orchestrator start sending out the config.  We still want all the nodes that have registered keys to post that they are "ready" to start before setting self.start to true.

elliedavidson (changes requested):

crates/orchestrator/src/lib.rs
ss-es: Removed!

Also just reorganized `post_manual_start` a bit. We now return immediately if the password is wrong or it's too late to manually start. If everything is fine, we set `num_nodes_with_stake` to `self.pub_posted.len()` and signal `peer_pub_ready`.

dailinsubjam (commented):
I didn't see anything wrong from looking at the code. But to guarantee its correctness, is it possible to do some simple tests? Say we only start 7 nodes when the requirement is (8, 10) of total nodes, then force manual start at node 5 by calling `client.post("api/manual_start")`. 
I did a little try but keep getting `Error 400: Peer's public configs are not ready`, however the reason might be that I placed client call at unreacheble place.

crates/orchestrator/src/lib.rs
```diff
@@ -269,41 +278,64 @@ where
...
+            .known_nodes_with_stake
+            .push(staked_pubkey.clone());
 
         // If the node wants to be DA, add it to the list of known DAs
         if is_da {
-            self.config
-                .config
-                .known_da_nodes
-                .push(register_pub_key_with_stake);
+            self.config.config.known_da_nodes.push(staked_pubkey);
```
dailinsubjam, line 311: As mentioned above, here I guess `is_da` is always `false`, because we set it at a later step.

crates/orchestrator/src/config.rs
```diff
@@ -291,50 +294,35 @@ impl<K: SignatureKey> NetworkConfig<K> {
...
-        run_config.config.known_nodes_with_stake = updated_config.config.known_nodes_with_stake;
-        run_config.config.known_da_nodes = updated_config.config.known_da_nodes;
-
-        info!("Retrieved config; our node index is {node_index}.");
-        Ok((run_config, source))
+        info!(
+            "Retrieved config; our node index is {}.",
+            run_config.node_index
+        );
+        Ok((run_config, NetworkConfigSource::Orchestrator))
```
dailinsubjam, line 325: unfortunatelly we might still need one more round of communication after setting `is_da`

dailinsubjam (commented):
Left some DA - related comments. Although I'm not sure whether I miss anything, and why this is not tested out. cc @rob-maron

dailinsubjam (commented):
I've tested dup key registration and `known_da_nodes`, I think they behave as expected.

dailinsubjam (approved):
Thanks for the great work! Let's also wait for Ellie's review.

crates/orchestrator/src/lib.rs
```diff
@@ -269,41 +278,78 @@ where
...
+                    .unwrap()
+                    .bootstrap_nodes
+                    .push((libp2p_public_key, libp2p_address));
+            }
+        }
+
+        println!("Posted public key for node_index {node_index}");
+
+        // node_index starts at 0, so once it matches `num_nodes_with_stake`
+        // we will have registered one node too many. hence, we want `node_index + 1`.
```
elliedavidson, line 347: I appreciate this comment explaining the logic!

elliedavidson (approved):

PR #3066, opened 2024-04-29
Add VidCommon to `validate_and_apply_header` and `Header::new`

### LINKED ISSUE #2744: [CX_CLEANUP] - Obtain VID Payload Size During Validation

We plan to enforce a base fee per byte. To do this, we need to know the size of the proposed payload during validation (also in `Header::new` where the fee info gets proposed). So we need to pass `VidCommon` into the block trait functions. This means that in the consensus task, we can't call `validate_and_apply_header` until we have received the VID dispersal, so that call would have to get moved into `vote_if_able`, which is a pretty significant logic change.

After the voting dependancies are collected we'll know that we have the VID information and can do all our validation. Also it will mean we won't be running the `validated_and_apply_header` logic which is blocking on the main consensus task.

As part of the refactoring efforts to the `vote_if_able` function, we need to support the ability to make this data available for calculation purposes.

_No response_

_No response_

Tests to ensure that the requisite information is indeed available.

_No response_

bfish713:
Closes #2744

### This PR: 

- Adds `VidCommon` as a parameter to `validate_and_apply_header` and `Header::new`
- Moves validation of state (`validate_and_apply_header`) until just before voting (i.e. moves the call to that function into `vote_if_able`
- Moves `vote_if_able` to a standalone function and spawns it as a task so it doesn't block forever on `validate_and_apply_header`
 - resets `current_proposal` after sending a vote (on the event) because we can't modify it from the task which runs `vote_if_able`
 - Gets our VID share (which we created) before proposing, and use it to create the new header (pass common to `Header::new`
 - Store all the VID shares right after we create them so we know they are available before proposing.
 - fixes tests by adding the share received event to unit tests  in views we want to test proposing because we now need that information to propose (it's stored automatically in the real workflow)

### Key places to review: 

The changes to the consensus logic


crates/task-impls/src/quorum_vote.rs
```diff
@@ -350,6 +350,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumVoteTaskState<TYPES, I
 
     /// Handle a vote dependent event received on the event stream
     #[instrument(skip_all, fields(id = self.id, latest_voted_view = *self.latest_voted_view), name = "Quorum vote handle", level = "error")]
+    #[allow(unused_variables)]
```
shenkeyao, line 353: Nit: Why do we need this? Looks like all variables are used.

crates/testing/tests/tests_1/upgrade_task.rs
```diff
@@ -255,7 +255,7 @@ async fn test_upgrade_and_consensus_task() {
         upgrade_vote_recvs,
         vec![QuorumProposalRecv(proposals[1].clone(), leaders[1])],
         vec![
-            DACertificateRecv(dacs[1].clone()),
+            VIDShareRecv(get_vid_share(&vids[1].0, handle.get_public_key())),
```
shenkeyao, line 258: Not sure if this is intentional. Why do we replace `DACertificateRecv` with `VIDShareRecv` rather than just add `VIDShareRecv`?

crates/task-impls/src/consensus/mod.rs
```diff
@@ -143,6 +145,182 @@ pub struct ConsensusTaskState<TYPES: NodeType, I: NodeImplementation<TYPES>> {
...
+        .write()
+        .await
+        .update_undecided_state(
+            consensus.saved_leaves.clone(),
+            consensus.validated_state_map.clone(),
+        )
+        .await
+    {
+        warn!("Couldn't store undecided state.  Error: {:?}", e);
+    }
```
shenkeyao, line 260: Looks like we need to add this to the quorum vote task as well.

crates/task-impls/src/consensus/mod.rs
```diff
@@ -156,132 +334,13 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> ConsensusTaskState<TYPES, I>
...
-            *self.cur_view
-        );
+    /// Ignores old vote behavior and lets `QuorumVoteTask` take over.
+    #[cfg(feature = "dependency-tasks")]
+    async fn vote_if_able(&mut self, _event_stream: &Sender<Arc<HotShotEvent<TYPES>>>) -> bool {
         false
     }
 
-    /// Validate the VID disperse is correctly signed and has the correct share.
+    /// Validates whether the VID Dispersal Proposal is correctly signed
```
shenkeyao, line 343: Nit: This seems to be the old documentation before I added the VID share validation. It can be updated after syncing with main.

crates/task-impls/src/consensus/proposal_helpers.rs
```diff
@@ -1048,11 +1022,36 @@ pub async fn handle_quorum_proposal_validated<TYPES: NodeType, I: NodeImplementa
             };
         }
 
-        ensure!(
-            task_state.vote_if_able(&event_stream).await,
-            "Failed to vote"
-        );
-        task_state.current_proposal = None;
```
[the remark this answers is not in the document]
shenkeyao: After reviewing changes in `consensus/mod.rs`, I think it makes sense to remove this line, since we'll do this when handling the `QuorumVoteSend` event. But I also think we should do the same for the dependency tasks.

crates/testing/tests/tests_1/quorum_proposal_task.rs
```diff
@@ -37,7 +37,7 @@ fn make_payload_commitment(
     let encoded_transactions = Vec::new();
     vid.commit_only(&encoded_transactions).unwrap()
 }
-
+#[ignore]
```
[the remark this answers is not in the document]
shenkeyao: Yeah, they fail due to the dependency task refactor. We haven't fixed them yet but they are not part of the current consensus logic.

crates/task-impls/src/consensus/mod.rs
jparr721: We should probably just call this function and then use gates to call your task's specific code. I think that'll enable us to prevent sources of error from cropping up.

crates/testing/tests/tests_1/quorum_proposal_task.rs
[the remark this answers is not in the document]
shenkeyao: Oh isn't this part of the test failure that we were trying to fix?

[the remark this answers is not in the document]
bfish713: I added `todo()` to the code so they will all fail

crates/task-impls/src/consensus/mod.rs
shenkeyao: Yeah, I think that's better. Then we don't have to make sure of consistent logic in multiple places.

crates/testing/tests/tests_1/quorum_proposal_task.rs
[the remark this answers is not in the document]
shenkeyao: I think I confused @ss-es and @jparr721 by saying the tests failed due to the dependency task refactor -- They only failed for me due to cache and VPN issues which have been fixed. So yeah, the failure here is unrelated to the refactor.

crates/task-impls/src/consensus/mod.rs
```diff
@@ -958,15 +958,17 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> ConsensusTaskState<TYPES, I>
...
                     };
                 }
             }
             HotShotEvent::QuorumVoteSend(vote) => {
                 let Some(proposal) = &self.current_proposal else {
                     return;
                 };
-                if proposal.get_view_number() == vote.get_view_number() {
+                if proposal.get_view_number() == vote.get_view_number()
+                    && self.public_key != self.quorum_membership.get_leader(vote.get_view_number())
```
shenkeyao, line 970: Why can't we reset `current_proposal` if we are the leader?

bfish713: We can reset, I thought we needed it to propose and that we might be setting it to none too soon, but we don't need it to propose.

crates/task-impls/src/quorum_vote.rs
bfish713: I think it has to do with the function not being called? maybe it can be removed now but clippy was complaining

crates/testing/tests/tests_1/upgrade_task.rs
bfish713: Oh this is because we don't need the DAC to propose, but we now do need the VID share to propose.

crates/task-impls/src/consensus/helpers.rs
```diff
@@ -511,7 +491,7 @@ async fn publish_proposal_from_commitment_and_metadata<TYPES: NodeType>(
     // FIXME - This is not great, and will be fixed later.
     // If it's > July, 2024 and this is still here, something has gone horribly wrong.
     let cnm = commitment_and_metadata
-        .take()
+        .clone()
```
jparr721, line 494: If we're cloning, and we don't have this as mutable anymore, we might be able to blow away the FIXME here. The issue that the FIXME was added to alert to was that this whole process was super obtuse due to needing to maintain backwards compatibility with the state handling in the consensus task.

crates/task-impls/src/consensus/mod.rs
```diff
@@ -374,10 +334,32 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> ConsensusTaskState<TYPES, I>
                     .await
                 {
                     Ok(Some(current_proposal)) => {
+                        error!("voting for liveness");
```
jparr721, line 337: Should this be here?

crates/task-impls/src/consensus/mod.rs
```diff
@@ -388,7 +370,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> ConsensusTaskState<TYPES, I>
                 if let Err(e) =
                     handle_quorum_proposal_validated(proposal, event_stream.clone(), self).await
                 {
-                    debug!("Failed to handle QuorumProposalValidated event {e:#}");
+                    warn!("Failed to handle QuorumProposalValidated event {e:#}");
```
jparr721, line 373: What made this upgrade in log level?

crates/task-impls/src/consensus/mod.rs
```diff
@@ -589,7 +593,9 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> ConsensusTaskState<TYPES, I>
                 // Adding `+ 1` on the LHS rather than `- 1` on the RHS, to avoid the overflow
                 // error due to subtracting the genesis view number.
                 if view + 1 < self.cur_view {
-                    info!("Throwing away VID disperse data that is more than one view older");
+                    tracing::info!(
```
jparr721, line 596: Does this need the `tracing::` namespace?

crates/task-impls/src/consensus/mod.rs
```diff
@@ -616,9 +622,31 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> ConsensusTaskState<TYPES, I>
...
+                };
+                let upgrade = self.decided_upgrade_cert.clone();
+                let pub_key = self.public_key.clone();
+                let priv_key = self.private_key.clone();
+                let consensus = Arc::clone(&self.consensus);
+                let storage = Arc::clone(&self.storage);
+                let quorum_mem = Arc::clone(&self.quorum_membership);
+                let committee_mem = Arc::clone(&self.committee_membership);
+                let instance_state = Arc::clone(&self.instance_state);
+                let handle = async_spawn(async move {
```
jparr721, line 636: Looks like this exact block is copied a couple of times, is it worth perhaps defining a helper function for it?

crates/task-impls/src/consensus/mod.rs
```diff
@@ -829,6 +860,46 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> ConsensusTaskState<TYPES, I>
...
+                }
+            }
+            HotShotEvent::QuorumProposalSend(proposal, _) => {
+                if self
+                    .payload_commitment_and_metadata
+                    .as_ref()
+                    .is_some_and(|p| p.block_view <= proposal.data.get_view_number())
+                {
+                    self.payload_commitment_and_metadata = None;
+                }
```
jparr721, line 892: Is it correct in my understanding that we only want to wipe the payload commitment and metadata when we successfully propose? Would this not basically just become stale if we didn't propose again after that?

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -102,6 +108,148 @@ struct ProposalDependencyHandle<TYPES: NodeType> {
     id: u64,
 }
 
+impl<TYPES: NodeType> ProposalDependencyHandle<TYPES> {
+    /// Sends a proposal if possible from the high qc we have
+    #[allow(clippy::too_many_lines)]
+    #[allow(dead_code)]
+    pub async fn publish_proposal_if_able(
```
jparr721, line 115: How come we added this here instead of just calling the function in helpers?

crates/task-impls/src/transactions.rs
```diff
@@ -280,7 +280,7 @@ impl<
 
                 // We failed to get a block
                 Ok(Err(err)) => {
-                    error!(%err, "Couldn't get a block");
+                    tracing::warn!(%err, "Couldn't get a block");
```
jparr721, line 283: I think we can remove `tracing::` here.

crates/task-impls/src/vid.rs
```diff
@@ -71,10 +72,25 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
                     *view_number,
                 )
                 .await;
+                let payload_commitment = vid_disperse.payload_commitment;
+                let shares = VidDisperseShare::from_vid_disperse(vid_disperse.clone());
+                let mut consensus = self.consensus.write().await;
+                for share in shares {
+                    let s = share.clone();
```
jparr721, line 79: Instead of needing to copy each share, would we just use `into_iter` over `shares` so we can consume it? If I'm understanding correctly, it looks like iterating over shares returns `&share`, but we need to own it to call `to_proposal`, but I *think* if we use `shares.into_iter()` we can obtain a `share`. This might be an oversimplification since I haven't run it.

crates/testing/src/test_runner.rs
```diff
@@ -311,7 +311,7 @@ where
                         }
                     }
                     Err(e) => {
-                        panic!("Error Joining the test task {:?}", e);
+                        tracing::error!("Error Joining the test task {:?}", e);
```
jparr721, line 314: Should this just be `error`

crates/testing/tests/tests_1/quorum_proposal_task.rs
```diff
@@ -184,6 +184,7 @@ async fn test_quorum_proposal_task_quorum_proposal_view_gt_1() {
     run_test_script(script, quorum_proposal_task_state).await;
 }
 
+#[ignore]
```
jparr721, line 187: Seems like breaking these will make the dependency task refactor more difficult. Perhaps I can take a look at these before this merges and try to get them working (minimally)?

crates/testing/tests/tests_5/combined_network.rs
```diff
@@ -72,7 +70,7 @@ async fn test_combined_network_cdn_crash() {
         // allow more time to pass in CI
         completion_task_description: CompletionTaskDescription::TimeBasedCompletionTaskBuilder(
             TimeBasedCompletionTaskDescription {
-                duration: Duration::from_secs(120),
+                duration: Duration::from_secs(220),
```
jparr721, line 73: How come we upped the time here?

crates/types/src/consensus.rs
```diff
@@ -55,6 +55,9 @@ pub struct Consensus<TYPES: NodeType> {
     /// View number that is currently on.
     pub cur_view: TYPES::Time,
 
+    /// View we proposed in last.  To prevent duplicate proposals
+    pub last_proposed_view: TYPES::Time,
```
jparr721, line 59: How does this differ from `cur_view`?

crates/task-impls/src/consensus/helpers.rs
```diff
@@ -393,7 +373,7 @@ pub async fn get_parent_leaf_and_state<TYPES: NodeType>(
 /// case proposal scenario.
 #[allow(clippy::too_many_lines)]
 #[allow(clippy::too_many_arguments)]
-async fn publish_proposal_from_upgrade_cert<TYPES: NodeType>(
+pub async fn publish_proposal_from_upgrade_cert<TYPES: NodeType>(
```
shenkeyao, line 376: Why is the `pub` needed? Looks like it's not called outside of this crate.

crates/task-impls/src/consensus/helpers.rs
shenkeyao: I also think the original backward state handling needs some rework, but I'd like to be more certain if we can make `commitment_and_metadata` immutable and replace `take` with `clone` here.

We probably can, since it looks like we don't have a logic where we do an operation depending on whether `commitment_and_metadata` is `Some`, except for setting it to `None` which is fine. If so, we can simplify the mutable handling in some other places as well (not necessarily in this PR), but I just want to make sure this is the case and that we are not skipping the state update unintentionally.

crates/task-impls/src/quorum_proposal.rs
shenkeyao: Looks like `publish_proposal_if_able` in `helpers` is imported in line 36 and we are actually using that instead of the function here. I guess this is a merge issue and we can just remove it.

crates/testing/tests/tests_1/proposal_ordering.rs
```diff
@@ -100,14 +100,17 @@ async fn test_ordering_with_specific_order(input_permutation: Vec<usize>) {
         ),
     ];
 
-    let view_2_inputs = permute_input_with_index_order(inputs, input_permutation);
+    let mut view_2_inputs = permute_input_with_index_order(inputs, input_permutation);
+    view_2_inputs.insert(0, DACertificateRecv(dacs[1].clone()));
+    view_2_inputs.insert(0, VIDShareRecv(get_vid_share(&vids[2].0, handle.get_public_key())));
+    view_2_inputs.insert(0, VIDShareRecv(get_vid_share(&vids[1].0, handle.get_public_key())));
```
shenkeyao, line 106: Why do we need the VID shares for two views?

crates/types/src/consensus.rs
shenkeyao: I guess the reason to separate `cur_view` and `last_propose_view` is that the former is for _any node_ to keep track of the view number it's on, whereas the latter is for a _leader_ to avoid duplicate proposals.

However, I don't think we are using `cur_view` except upon initialization (we are using the `cur_view` in each task instead), so we can probably just combine these two parameters.

crates/testing/tests/tests_1/quorum_proposal_task.rs
shenkeyao: +1 I can also help with this.

crates/task-impls/src/consensus/mod.rs
```diff
@@ -829,6 +860,46 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> ConsensusTaskState<TYPES, I>
...
+                let should_propose = self.quorum_membership.get_leader(new_view) == self.public_key
+                    && self.consensus.read().await.high_qc.view_number
+                        == proposal.get_view_number();
+
+                if should_propose {
+                    debug!(
+                        "Attempting to publish proposal after voting; now in view: {}",
+                        *new_view
+                    );
+                    let _ = self.publish_proposal(new_view, event_stream.clone()).await;
```
shenkeyao, line 879: I tried to debug the proposal task tests and got a compilation error at this line. I think it's because `publish_proposal` is gated by `not(feature = "dependency-tasks")` but this `QuorumVoteSend` event handling code isn't.

shenkeyao: This can be fixed by adding `#[cfg(not(feature = "dependency-tasks"))]` to `HotShotEvent::QuorumVoteSend(vote) => {`. I don't think this is needed for dependency tasks.

crates/testing/tests/tests_1/quorum_proposal_task.rs
shenkeyao: Btw `test_quorum_proposal_task_quorum_proposal_view_1` and `test_quorum_proposal_task_quorum_proposal_view_gt_1` aren't ignored but also failed, so we need to fix them as well.

crates/task-impls/src/consensus/helpers.rs
jparr721: Hopefully we're going to merge the new tasks soon and this all becomes a distant memory.

crates/testing/tests/tests_1/quorum_proposal_task.rs
jparr721: I'll talk to Brendon today about getting these running.

crates/types/src/consensus.rs
jparr721: Yeah ideally `cur_view` dies sooner rather than later.

crates/testing/tests/tests_1/quorum_proposal_task.rs
jparr721: Wait a second: https://github.com/EspressoSystems/HotShot/blob/main/crates/testing/tests/tests_1/quorum_proposal_task.rs.

I think this PR has a merge issue. I'm not sure how my gates got removed.

shenkeyao: As synced up offline, this test should have a `dependency_tasks` feature gate. But that's probably not the only issue, since other tests with this feature also fail.

crates/task-impls/src/consensus/helpers.rs
bfish713: We can't remove this in the current code just yet, if that's what you are suggesting.  We should be able to get rid of this with the dependency tasks.

crates/task-impls/src/consensus/mod.rs
bfish713: I will remove

crates/task-impls/src/consensus/mod.rs
bfish713: shouldn't need it

crates/task-impls/src/consensus/mod.rs
bfish713: Yes we want to wipe it when we propose.  It is stale for future views, but it is possible another propose task gets spawned after the first proposal (or is already running).  Resetting this could prevent double proposing.  The new variable in the locked state `lastest_proposed_view` is what's really in charge of that, but doing this shouldn't hurt.

crates/task-impls/src/quorum_proposal.rs
bfish713: Yeah it got added in a merge I guess and I thought it was just something in development that would get used (hence the `allow(dead_code)`)

crates/task-impls/src/vid.rs
bfish713: `share` is already owned here and `shares` is being consumed. the `let s = share.clone()` is to get around some type inference issue with the rust  compiler.  For some reason the compiler does not like `let key: <TYPES as NodeType>::SignatureKey = share.recipient_key.clone();`.

crates/testing/tests/tests_5/combined_network.rs
bfish713: This test was failing because it was taking to long on this branch and main so I increased the time to make it less likely to fail

crates/testing/tests/tests_1/proposal_ordering.rs
bfish713: We now need VID to propose so vids[2] is so we can call `Header::new()` with the vid common info to create a proposal.  vids[1] is needed to complete the prior rounds validation for the proposal and updating the state map.

crates/task-impls/src/consensus/mod.rs
ss-es: From my reading of `handle_quorum_proposal_validated`, it doesn't introduce any errors except if `vote_if_able` fails. And I think we'd expect `vote_if_able` to fail at least once or twice per view depending on when it was called, so to me `warn!` feels like it would spam logs in deployment.

crates/testing/src/test_runner.rs
```diff
@@ -311,7 +311,7 @@ where
                         }
                     }
                     Err(e) => {
-                        panic!("Error Joining the test task {:?}", e);
```
ss-es, line 314: I'm not that familiar with async rust, but doesn't this change the semantics?

My understanding is: previously, the main thread would panic if one of the futures panics. But removing this means the main thread will ignore a thread panic and just log to stdout, which doesn't cause a test failure.

(I might be misunderstanding something of course.)

bfish713: It does change the semantics.  The futures are the testing tasks, and I think the issue I was seeing was with cancellation and only on one executor.  I can probably change it back now after after a few fixes

crates/task-impls/src/consensus/mod.rs
bfish713: `vote_if_able` is spawned in a seperate task so if it errors we will not return to this point and log.  I do think anything logged here would be something more unexpected.

crates/testing/src/test_runner.rs
bfish713: Because this is just tokio if we add use tracing::error! it will cause lint errors for async_std.  I think it's easier to just have the fully qualified name here than to gate the imports

jparr721 (approved):

shenkeyao (approved):

PR #3069, opened 2024-04-29
Don't return empty blocks from simple builder

jbearer:


ss-es (approved):

ss-es (approved):
It looks like the tests are down to timing issues/passing on rerun, so I think we can just merge

the worst of the test failures should be mitigated on main now

PR #3070, opened 2024-04-29
Add `all` predicate

ss-es:
No linked issue.

### This PR: 
Adds an `all` predicate, to allow for varying output event orderings in tests.

### DIFF
--- a/crates/testing/src/predicates/event.rs
+++ b/crates/testing/src/predicates/event.rs
@@ -1,5 +1,6 @@
-use std::sync::Arc;
+use std::{collections::HashSet, sync::Arc};
 
+use async_lock::RwLock;
 use async_trait::async_trait;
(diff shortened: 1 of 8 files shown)


crates/testing/src/predicates/event.rs
```diff
@@ -25,6 +26,89 @@ impl<TYPES: NodeType> std::fmt::Debug for EventPredicate<TYPES> {
     }
 }
 
+#[allow(clippy::type_complexity)]
+pub struct TestPredicate<INPUT> {
```
jparr721, line 30: Would `AllPredicates` be a better name?

jparr721 (approved):

crates/testing/src/predicates/event.rs
ss-es: It's a more general type than just the `all` and `all_predicates`; it should accommodate basically any predicate.

PR #3071, opened 2024-04-30
[CX_CLEANUP] - Fix data setting and polling for the dependency task integration

shenkeyao:
Related to #3050. To be merged to #3052.

### This PR: 
* Sets `payload_commitment_and_metadata` for the quorum proposal recv task.
* Adds data polling function for the vote task creation.

### This PR does not: 
* Fix the dependency task tests.


jparr721:
Cherrypicked again due to conflicts. I did not handle the payload commitment and metadata and instead gated that function, it was legacy behavior and we don't want it in the new quorum proposal recv task.

PR #3072, opened 2024-04-30
Cycle through recipients until ready to cancel

### LINKED ISSUE #3026: [LIBP2P] - Improve the loop in the `DelayedRequester::do_vid`

There are two things that require improvement in the `DelayedRequester::do_vid` loop:
1. The loop iterates only once through the list of the recipients. This is not enough as there might be a situation when at the time of sending the requests the recipients do not yet have the data to respond. The requester should not stop sending the requests until the request becomes irrelevant (as indicated in `cancel_vid`).
2. Fix the logical operator in the `DelayedRequester::cancel_vid` method. It should be `||` instead of `&&` as indicated in the correct comment: Returns true if we got the data we wanted, or the view has moved on.

_No response_

_No response_

This is hard to test because it's not deterministic. The most consistent way to reproduce the issue is to apply the following patch:

```
diff --git a/crates/hotshot/src/traits/networking/combined_network.rs b/crates/hotshot/src/traits/networking/combined_network.rs
index 3a56909..0ee74e3 100644
--- a/crates/hotshot/src/traits/networking/combined_network.rs
+++ b/crates/hotshot/src/traits/networking/combined_network.rs
@@ -502,7 +502,7 @@ impl<TYPES: NodeType> ConnectedNetwork<Message<TYPES>, TYPES::SignatureKey>
         });
         // View changed, let's start primary again
         self.primary_down.store(false, Ordering::Relaxed);
-        self.primary_fail_counter.store(0, Ordering::Relaxed);
+        // self.primary_fail_counter.store(0, Ordering::Relaxed);
     }
 
     fn is_primary_down(&self) -> bool {
```

and run the `test_combined_network_cdn_crash` test.

_No response_

lukaszrzasik:
Closes #3026

### This PR: 

* Iterates through recipients in a cycle instead of once in `do_vid`
* Fixes logical operator in `cancel_vid` from `&&` to `||`
* `request_data` now borrows recipient instead of taking ownership
* the tasks are allowed to shut down gracefully by receiving `HotShotEvent::Shutdown` before they are shut down forcibly by using `JoinHandle::cancel` (or tokio `JoinHandle::abort`)

### Key places to review: 

`DelayedRequester::do_vid`

### How to test this PR:

The most consistent way to reproduce the issue is to apply the following patch:

```
diff --git a/crates/hotshot/src/traits/networking/combined_network.rs b/crates/hotshot/src/traits/networking/combined_network.rs
index 3a56909..0ee74e3 100644
--- a/crates/hotshot/src/traits/networking/combined_network.rs
+++ b/crates/hotshot/src/traits/networking/combined_network.rs
@@ -502,7 +502,7 @@ impl<TYPES: NodeType> ConnectedNetwork<Message<TYPES>, TYPES::SignatureKey>
         });
         // View changed, let's start primary again
         self.primary_down.store(false, Ordering::Relaxed);
-        self.primary_fail_counter.store(0, Ordering::Relaxed);
+        // self.primary_fail_counter.store(0, Ordering::Relaxed);
     }
 
     fn is_primary_down(&self) -> bool {
```

and run the `test_combined_network_cdn_crash` test.
All other tests should pass.

### DIFF
--- a/crates/task-impls/src/request.rs
+++ b/crates/task-impls/src/request.rs
@@ -80,12 +80,11 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, Ver: StaticVersionType + 'st
     ) -> Option<Self::Output> {
         match event.as_ref() {
             HotShotEvent::QuorumProposalValidated(proposal, _) => {
-                let sender = task.clone_sender();
-                let state = task.state_mut();
+                let state = task.state();
                 let prop_view = proposal.get_view_number();
                 if prop_view >= state.view {
                     state
-                        .spawn_requests(prop_view, sender, Ver::instance())
+                        .spawn_requests(prop_view, task.clone_sender(), Ver::instance())
                         .await;
                 }
                 None
@@ -128,7 +127,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, Ver: StaticVersionType + 'st
 {
     /// Spawns tasks for a given view to retrieve any data needed.
     async fn spawn_requests(
-        &mut self,
+        &self,
         view: TYPES::Time,
         sender: Sender<Arc<HotShotEvent<TYPES>>>,
         bind_version: Ver,
@@ -156,7 +155,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, Ver: StaticVersionType + 'st
     /// received will be sent over `sender`
     #[instrument(skip_all, fields(id = self.id, view = *self.view), name = "NetworkRequestState run_delay", level = "error")]
     fn run_delay(
-        &mut self,
+        &self,
         request: RequestKind<TYPES>,
         sender: Sender<Arc<HotShotEvent<TYPES>>>,
         view: TYPES::Time,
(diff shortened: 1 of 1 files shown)


crates/task-impls/src/request.rs
```diff
@@ -58,6 +59,8 @@ pub struct NetworkRequestState<
     pub _phantom: PhantomData<fn(&Ver)>,
     /// The node's id
     pub id: u64,
+    /// Vector of shutdown flags for delayed requesters
+    pub shutdown_flags: Vec<Arc<AtomicI8>>,
```
bfish713, line 63: Can this just be 1 atomic bool that every single delayed requester has.  Then in `cancel_vid` we just check if it's true and return true.  That way we don't have to do all the logic with the vector

bfish713 (dismissed):

crates/task-impls/src/request.rs
```diff
@@ -110,14 +117,18 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, Ver: StaticVersionType + 'st
...
+    }
 }
 
 impl<TYPES: NodeType, I: NodeImplementation<TYPES>, Ver: StaticVersionType + 'static>
     NetworkRequestState<TYPES, I, Ver>
 {
     /// Spawns tasks for a given view to retrieve any data needed.
     async fn spawn_requests(
-        &self,
+        &mut self,
```
jparr721, line 131: Maybe I'm just missing something, but, how come this is now mutable?

bfish713 (approved):

PR #3074, opened 2024-04-30
Add duplicate keys checking before registration to Orchestrator

dailinsubjam:
### This PR: 
- Add one api in OrchestratorApi for duplicate keys checking at the very begging when nodes start to post identity
- So that node with same public validator configs ( `my_own_validator_config.get_public_config()` = `public_key`+ `stake_value` + `state_ver_key`) cannot register twice, regardless of their private keys, node index, and p2p config.
- does some nit

### This PR does not: 
- Restructure orchestrator

### DIFF
--- a/crates/orchestrator/src/client.rs
+++ b/crates/orchestrator/src/client.rs
@@ -342,7 +342,7 @@ impl OrchestratorClient {
         skip(self),
         name = "orchestrator check duplicate keys before identity registration"
     )]
-    pub async fn pass_check_duplicate_keys<K: SignatureKey>(
+    pub async fn pass_check_no_dup_keys<K: SignatureKey>(
         &self,
         my_pub_key: PeerConfig<K>,
     ) -> bool {
--- a/crates/orchestrator/src/client.rs
+++ b/crates/orchestrator/src/client.rs
@@ -334,6 +334,37 @@ impl OrchestratorClient {
         self.get_config_after_collection().await
     }
 
+    /// Sends my public key to the orchestrator to check whether it's duplicate
+    /// And get the bool to indicate whether I can register with my identity
+    /// Real public key registration is for later `post_and_wait_all_public_keys`
+    /// Return false if the public key has already been registered
+    #[instrument(
+        skip(self),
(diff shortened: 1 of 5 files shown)


crates/orchestrator/src/client.rs
```diff
@@ -334,6 +334,37 @@ impl OrchestratorClient {
         self.get_config_after_collection().await
     }
 
+    /// Sends my public key to the orchestrator to check whether it's duplicate
+    /// And get the bool to indicate whether I can register with my identity
+    /// Real public key registration is for later `post_and_wait_all_public_keys`
+    /// Return false if the public key has already been registered
```
shenkeyao, line 340: Nit: I find this a bit confusing because, from the function name `check_duplicate`, I would expect it to return true if there's duplication. On the other hand, it also makes sense to return `false` for a `check` function when the check fails.

Would it be better if we rename it to `pass_check_unique_keys`?

shenkeyao (changes requested):
Not very familiar with the orchestrator work but this PR looks straightforward so I took a look! 😃 The changes look good to me, except for a minor comment about a function name.

ss-es:
Question (maybe just for my own understanding): couldn't we just make `post_identity` take an extra `pubkey` argument, and then just store the `node_index` we return for a given public key in a `HashSet`? Then if a node tries to register twice, we give it the same `node_index`.

crates/orchestrator/src/client.rs
dailinsubjam: I changed to 'pass_check_no_dup_keys'

dailinsubjam:
> Question (maybe just for my own understanding): couldn't we just make `post_identity` take an extra `pubkey` argument, and then just store the `node_index` we return for a given public key in a `HashSet`? Then if a node tries to register twice, we give it the same `node_index`.

@ss-es I also agree it could be better if we register identity with public keys, that's also a nice idea by giving it the same `node_index` for same identity, I'll do these changes after your PR merged in.

dailinsubjam:
I will close it without merge when https://github.com/EspressoSystems/HotShot/pull/3055 is in, as dup key checking is covered by it.

PR #3075, opened 2024-04-30
Add timeout for examples in CI

dailinsubjam:
### This PR: 
- Add timeout for examples tests in CI to prevent endless run when anything got stuck, now it's set to 5min for 100 rounds


ss-es:
+1 but could we make it longer, e.g. 20 minutes? I think e.g. `build-release` can take ~13 minutes from scratch, so 20 should be a safe amount of time.

I'm just slightly worried that compiling can take longer than 5 minutes if the cache fails for some reason (e.g. `Cargo.toml` or `Cargo.lock` changes), and we wouldn't want to fail if that happens

ss-es (approved):

PR #3086, opened 2024-05-01
[CX_CLEANUP] - All Dependency Task Tests Passing 

### LINKED ISSUE #3085: [CX_CLEANUP] - All Dependency Task Tests Passing

In order to fully integrate, we need to make sure that the CI suite for the `dependency-tasks` feature flag passes with the same reliability as the non-feature flagged branch.

Essentially, we need to run and debug the failing tests and determine why consensus does not run during outages. This will include the following:
1. Finishing the QuorumProposal task implementation (right now, handling `QuorumProposalRecv` does not include the liveness check).
2. Running and checking for missing behavior in any of the other tasks (quorum_proposal_recv, quorum_vote).
3. Feature gate any tests that depend upon behavior that is to be deprecated.

_No response_

All tests should pass as-is without needing changes.

_No response_

jparr721:
Closes #3085


shenkeyao:
Assigned this to myself so I can get notifications. 😃

PR #3096, opened 2024-05-02
pass instance_state to `from_transactions` by ref

tbro:
This is correcting the mistake of passing `Arc<dyn InstanceState>` to `from_transactions`. If we pass as reference we can use as an opaque type as we do elsewhere. 

Required primarily by https://github.com/EspressoSystems/espresso-sequencer/pull/1382
related to: https://github.com/EspressoSystems/hotshot-builder-core/pull/122, https://github.com/EspressoSystems/hotshot-query-service/pull/565

Closes #<ISSUE_NUMBER>


jparr721 (approved):
LGTM, if you wouldn't mind putting a short description of why you're making a change in the PR, that would greatly help with parsing the "why" of the change.

PR #3098, opened 2024-05-03
[CX_CLEANUP] - Make `high_qc` and `cur_view` private and add functions to avoid accidental overwrite

### LINKED ISSUE #3097: [CX_CLEANUP] - Make `high_qc` and `cur_view` private and add functions to avoid accidental overwrite

We should make `high_qc` and `cur_view` private fields of `Consensus` and add setter and getter functions to avoid accidentally overwriting them to an older one.

_No response_

Zulip discussion: https://espresso.zulipchat.com/#narrow/stream/397846-Consensus-Team/topic/consensus.20logic/near/436758043.

`high_qc` and `cur_view` can only be retrieved and updated using the new functions.

_No response_

shenkeyao:
Closes #3097.

### This PR: 
* Makes the `high_qc` and `cur_view` fields of the `Consensus` struct private.
* Adds getter and setter for `high_qc`.
* Adds a getter for `cur_view` and updates its setter to only accept a newer view.

### This PR does not: 
* Change where to update the high QC or view number, except for restricting to newer views.

### Key places to review: 
* `crates/types/src/consensus.rs`.

### DIFF
--- a/crates/types/src/consensus.rs
+++ b/crates/types/src/consensus.rs
@@ -7,7 +7,7 @@ use std::{
 
 use committable::Commitment;
 use displaydoc::Display;
-use tracing::error;
+use tracing::{debug, error};
 
 pub use crate::utils::{View, ViewInner};
 use crate::{
@@ -53,7 +53,7 @@ pub struct Consensus<TYPES: NodeType> {
     pub saved_da_certs: HashMap<TYPES::Time, DACertificate<TYPES>>,
 
     /// View number that is currently on.
-    pub cur_view: TYPES::Time,
+    cur_view: TYPES::Time,
 
     /// last view had a successful decide event
     pub last_decided_view: TYPES::Time,
@@ -72,7 +72,7 @@ pub struct Consensus<TYPES: NodeType> {
     pub locked_view: TYPES::Time,
 
     /// the highqc per spec
-    pub high_qc: QuorumCertificate<TYPES>,
+    high_qc: QuorumCertificate<TYPES>,
 
     /// A reference to the metrics trait
     pub metrics: Arc<ConsensusMetricsValue>,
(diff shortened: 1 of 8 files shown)


shenkeyao:
Will fix linting errors after the linting in https://github.com/EspressoSystems/HotShot/pull/3099 is passed.

crates/types/src/consensus.rs
```diff
@@ -263,9 +263,57 @@ impl Default for ConsensusMetricsValue {
...
+
+    /// Get the current view.
+    pub fn cur_view(&self) -> TYPES::Time {
+        self.cur_view
+    }
+
     /// Update the current view.
-    pub fn update_view(&mut self, view_number: TYPES::Time) {
-        self.cur_view = view_number;
+    pub fn update_view_if_new(&mut self, view_number: TYPES::Time) {
```
ss-es, line 300: Minor nit: I think the `if_new` is redundant and these could just be `update_view` and `update_high_qc`. This is not a strong opinion, and might just be personal preference so feel free to ignore if you prefer this naming.

crates/examples/Cargo.toml
```diff
@@ -5,7 +5,7 @@ edition = { workspace = true }
 name = "hotshot-examples"
 readme = "README.md"
 version = { workspace = true }
-rust-version = "1.65.0"
+rust-version = "1.70.0"
```
ss-es, line 8: I think we want

```suggestion
rust-version = { workspace = true }
```

Cargo.toml
```diff
@@ -2,7 +2,7 @@
 version = "0.5.43"                                        # same as `hotshot`, but workspace subcrate can also release its own version
 authors = ["Espresso Systems <hello@espressosys.com>"]
 edition = "2021"
-rust-version = "1.65.0"
+rust-version = "1.76.0"
```
ss-es, line 5: I think we want

```suggestion
rust-version = "1.65.0"
```

ss-es (commented):
Looks good to me, apart from the rust version thing!

Cargo.toml
ss-es: nevermind!

ss-es (approved):

ss-es (approved):

crates/types/src/consensus.rs
```diff
@@ -263,9 +263,57 @@ impl Default for ConsensusMetricsValue {
...
+        if view_number > self.cur_view {
+            self.cur_view = view_number;
+        }
+    }
+
+    /// Get the high QC.
+    pub fn high_qc(&self) -> &QuorumCertificate<TYPES> {
```
jparr721, line 307: Can this be `get_high_qc` to be more consistent with our naming?

shenkeyao: I used to debate this and found both `get_something` and `something` in our codebase. It's not consistent anyway, and since I slightly prefer the latter, I didn't add ` get`.

jparr721 (approved):

PR #3099, opened 2024-05-03
Fix linting due to version update

shenkeyao:
Cherry-pick from https://github.com/EspressoSystems/HotShot/pull/3066/commits/c295e43409b1c0532fffa77bc570618c95f00e5b to fix the linting errors due to `rustc` version update.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 version = "0.5.43"                                        # same as `hotshot`, but workspace subcrate can also release its own version
 authors = ["Espresso Systems <hello@espressosys.com>"]
 edition = "2021"
-rust-version = "1.65.0"
+rust-version = "1.76.0"
 homepage = "https://github.com/EspressoSystems/HotShot"
 documentation = "https://hotshot.docs.espressosys.com"
 repository = "https://github.com/EspressoSystems/HotShot"
--- a/crates/examples/Cargo.toml
+++ b/crates/examples/Cargo.toml
@@ -5,7 +5,7 @@ edition = { workspace = true }
 name = "hotshot-examples"
 readme = "README.md"
 version = { workspace = true }
-rust-version = "1.65.0"
+rust-version = "1.70.0"
 
 [features]
 default = ["docs", "doc-images", "hotshot-testing"]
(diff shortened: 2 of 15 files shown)


Cargo.toml
```diff
@@ -2,7 +2,7 @@
 version = "0.5.43"                                        # same as `hotshot`, but workspace subcrate can also release its own version
 authors = ["Espresso Systems <hello@espressosys.com>"]
 edition = "2021"
-rust-version = "1.65.0"
+rust-version = "1.76.0"
```
QuentinI, line 5: Why do we want to update this? As far as I can tell everything still builds with 1.65

ss-es, line 5: I agree, I think this should stay:

```suggestion
rust-version = "1.65.0"
```

crates/examples/Cargo.toml
```diff
@@ -5,7 +5,7 @@ edition = { workspace = true }
 name = "hotshot-examples"
 readme = "README.md"
 version = { workspace = true }
-rust-version = "1.65.0"
+rust-version = "1.70.0"
```
ss-es, line 8: I think this should actually just be

```suggestion
rust-version = { workspace = true }
```

I don't know if all crates do this, but I think they probably should? cc @QuentinI

ss-es (commented):
I think this looks good apart from the rust-version bumps -- thank you for fixing this!

crates/examples/Cargo.toml
QuentinI: Most of our crates don't seem to set `rust-version` at all.
I agree, we should probably inherit MSRV for all crates in workspace.

Cargo.toml
QuentinI: It does not in fact build on 1.65, sorry! These changes are correct.

QuentinI (dismissed):
LGTM, sorry for MSRV confusion
I think we can set `rust-version = { workspace }` for all crates as a separate thing, doesn't really cause any problems as it is now.

crates/examples/Cargo.toml
shenkeyao: I've updated this using `workspace`!

ss-es (approved):

PR #3102, opened 2024-05-04
[Auto Bench] Fix unexpected panic when reporting benchmark results

dailinsubjam:
### This PR: 
- Fix unexpected panic like `Message:  assertion failed: total_num_views <= (failed_num_views + num_successful_commits + 2)`.
- Fix unexpected panic like `Division by 0` when elapsed time is less than 1s for benchmark runs.
- Some nit.


crates/examples/infra/mod.rs
```diff
@@ -560,18 +557,19 @@ pub trait RunDA<
         let consensus_lock = context.hotshot.get_consensus();
         let consensus = consensus_lock.read().await;
         let total_num_views = usize::try_from(consensus.locked_view.get_u64()).unwrap();
-        // When posting to the orchestrator, note that the total number of views also include un-finalized views.
+        let failed_num_views = total_num_views - num_successful_commits; // could include uncommitted views
```
jparr721, line 560: Could we change this from an end comment to an above comment? It can make the lines really long.

jparr721 (dismissed):
One small thing, I can re-approve after.

jparr721 (approved):

PR #3106, opened 2024-05-06
Remove `Default` bound from `InstanceState` trait

tbro:
Instead supply bounds in `testing` crate. This avoids modifying the public API and forcing implementation of  `Default`.

Closes #<ISSUE_NUMBER>

### Key places to review: 
  * `crates/testing/src/block_builder.rs`
  * `crates/types/src/traits.states.rs`


nyospe (approved):
Good, this looks correct.

I would get an approval from someone on the Consensus team as well, but this achieves the objective of containing the `Default` trait bound to the test usage, without being too ugly.

jparr721 (approved):

PR #3107, opened 2024-05-06
Add rust-toolchain.toml

QuentinI:
Closes #0000

### This PR: 
- Adds rust-toolchain.toml
- Makes nix flake use it
- Makes CI use it

This should result in less surprises when toolchain version doesn't match between local environment and CI.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -41,10 +41,15 @@
           "rustfmt"
           "llvm-tools-preview"
         ];
-        fenixStable = fenix.packages.${system}.fromToolchainFile {
-          dir = ./.;
-          sha256 = "sha256-opUgs6ckUQCyDxcB9Wy51pqhd0MPGHUVbwRKKPGiwZU=";
-        };
+        fenixStable = fenix.packages.${system}.combine [
+          (fenix.packages.${system}.latest.withComponents [
+            "rustfmt"
+          ])
+          (fenix.packages.${system}.fromToolchainFile {
+            dir = ./.;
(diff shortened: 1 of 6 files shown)


flake.nix
```diff
@@ -41,20 +41,10 @@
...
-              "clippy"
-              "rust-src"
-              "rustc"
-              "llvm-tools-preview"
-            ]
-          )
-          ( fenix.packages.${system}.latest.withComponents [
-              "rustfmt"
-            ]
-          )
```
ss-es, line 56: I think this means we won't have nightly `rustfmt` in the Nix flake, so the settings in `rustfmt.toml` won't work. Could we do e.g.

```
        fenixStable = fenix.packages.${system}.combine [
        (
        fenix.packages.${system}.latest.withComponents [
          "rustfmt"
        ]
        )
        ( fenix.packages.${system}.fromToolchainFile {
          dir = ./.;
          sha256 = "sha256-opUgs6ckUQCyDxcB9Wy51pqhd0MPGHUVbwRKKPGiwZU=";
        } )

        ];

``` 

(excuse the formatting)

We could also set this in the `rust-toolchain.toml` file if possible, though I don't know exactly how that works

QuentinI: Duh, I forgot why we used that. I don't think you can mix-and-match toolchain components with rustup, but we certainly can retain nightly rustfmt in the flake.

ss-es (approved):

PR #3108, opened 2024-05-06
Update  the CDN

rob-maron:
Does not close a linked issue.

This update changes `topics` so that they are specified application-side instead of CDN-side. This way we can add/remove as we see fit by just changing the enum.

### DIFF
--- a/crates/hotshot/src/traits/networking/push_cdn_network.rs
+++ b/crates/hotshot/src/traits/networking/push_cdn_network.rs
@@ -155,15 +155,12 @@ pub struct PushCdnNetwork<TYPES: NodeType> {
 
 /// The enum for the topics we can subscribe to in the Push CDN
 #[repr(u8)]
-pub enum Topics {
+pub enum Topic {
     /// The global topic
     Global = 0,
     /// The DA topic
     DA = 1,
 }
-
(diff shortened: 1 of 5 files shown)


crates/hotshot/src/traits/networking/push_cdn_network.rs
```diff
@@ -153,6 +153,18 @@ pub struct PushCdnNetwork<TYPES: NodeType> {
...
 }
 
+/// The enum for the topics we can subscribe to in the Push CDN
+#[repr(u8)]
+pub enum Topics {
+    /// The global topic
+    Global = 0,
+    /// The DA topic
+    DA = 1,
+}
```
ss-es, line 163: Very minor nit, but should this be `Topic` (singular)?

ss-es (dismissed):
Looks good to me, just a nit about `pub enum Topic` vs. `pub enum Topics`

crates/hotshot/src/traits/networking/push_cdn_network.rs
rob-maron: Fixed, thanks!

ss-es (approved):

PR #3109, opened 2024-05-06
Fix Tests in bf/vid-common

jparr721:
Closes #<ISSUE_NUMBER>

### This PR: 
Fixes the QuorumProposal task tests within the branch `bf/vid-common`. These tasks now require the vid shares table to be updated, which does not happen in the existing QuorumProposal task logic, resulting in the system failing to send the proposal.

Setting this field was a trip, so I split the test into individual tests to make the state more isolated. 

### This PR does not: 

This PR does not fix the race condition that may cause this to fail in the network tests. This will be done in #3052.

### DIFF
--- a/crates/task-impls/src/consensus/mod.rs
+++ b/crates/task-impls/src/consensus/mod.rs
@@ -35,7 +35,6 @@ use hotshot_types::{traits::storage::Storage, vote::Certificate};
 use jf_primitives::vid::VidScheme;
 #[cfg(async_executor_impl = "tokio")]
 use tokio::task::JoinHandle;
-#[cfg(not(feature = "dependency-tasks"))]
 use tracing::{debug, error, instrument, warn};
 use vbs::version::Version;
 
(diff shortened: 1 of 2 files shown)


crates/task-impls/src/consensus/mod.rs
```diff
@@ -35,7 +35,6 @@ use hotshot_types::{traits::storage::Storage, vote::Certificate};
 use jf_primitives::vid::VidScheme;
 #[cfg(async_executor_impl = "tokio")]
 use tokio::task::JoinHandle;
-#[cfg(not(feature = "dependency-tasks"))]
```
jparr721, line 38: This caused a build failure.

crates/task-impls/src/consensus/mod.rs
```diff
@@ -805,6 +804,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> ConsensusTaskState<TYPES, I>
                     };
                 }
             }
+            #[cfg(not(feature = "dependency-tasks"))]
```
jparr721, line 807: This shouldn't run in the dependency tasks.

PR #3116, opened 2024-05-07
Add versioning in the application interface

### LINKED ISSUE #3212: [Task architecture] - Add versioning to the application interface

We want to ensure that the sequencer instantiates HotShot with the correct version, and this should be done at the type level. In the future, we may constrain the version of parts of HotShot which should raise compile-time errors if there is a mismatch.

_No response_

_No response_

A new `VersionedNode<Base, Upgrade>` type with a base and upgraded version is implemented.

_No response_

ss-es:
Closes #3212 

### This PR: 
This is motivated by a requirement for the sequencer to be able to upgrade the `BlockHeader` and `InstanceState` through a consensus protocol upgrade, to e.g. introduce fees.

To do this, we introduce a `VersionedNode` trait which will be a required bound for the initialization functions in `SystemContext`. Nodes must implement `VersionedNode<BASE_VERSION, UPGRADE_VERSION>` matching the version (at the type-level) that HotShot requires in initialization.

HotShot will propagate the correct `NodeType` implementation to the consensus tasks, and transition between them when the network is upgraded.

### This PR does not:
An alternative implementation would have been to parametrize `NodeType`, i.e.:

    pub trait NodeType<VERSION>
    where
      VERSION: StaticVersionType
    { ... }

In my mind, this is the more correct way to go about this, but the `VERSION` parameter is *extremely* invasive, requiring difficult-to-automate changes in the 600+ places we use `NodeType` in HotShot. I think the current PR provides the same level of type safety in a much smaller change, with a marginal downstream cost.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3233,6 +3233,7 @@ dependencies = [
  "time 0.3.36",
  "tokio",
  "tracing",
+ "vbs",
 ]
 
 [[package]]
--- a/crates/example-types/Cargo.toml
+++ b/crates/example-types/Cargo.toml
@@ -17,6 +17,7 @@ async-compatibility-layer = { workspace = true }
 async-trait = { workspace = true }
 anyhow = { workspace = true }
 sha3 = "^0.10"
+vbs = { workspace = true }
 committable = { workspace = true }
 either = { workspace = true }
 futures = { workspace = true }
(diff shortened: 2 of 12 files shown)


ss-es:
This PR will depend on #3133 for general task restartability.

PR #3118, opened 2024-05-07
Allow nodes to join late if they have already registered

jbearer:
By switching the order of two checks (for manual start and duplicate keys) we can allow nodes that have already posted their keys, but restarted, to successfully get a config even after the network has started.


dailinsubjam (approved):

ss-es (approved):
Looks good to me

PR #3119, opened 2024-05-07
Make BlockPayload::encode infallible

### LINKED ISSUE #3020: [Cleanup] - Remove `BlockPayload::encode` in favor of `EncodeBytes` bound

`EncodeBytes` is meant as a general trait for all data structures that need to be canonicaly encoded.

`BlockPayload::encode` is currently fallible, point of failure being u32->usize conversion of transaction length. This check should be pushed to the point of transaction creation, since non-encodable transactions are basically useless.

_No response_

`BlockPayload::encode` is removed, `BlockPayload: EncodeBytes` bound is added. 

_No response_

QuentinI:
Closes #3020 

### This PR: 
- Removes `BlockPayload::encode`
- Adds `BlockPayload: EncodeBytes` bound instead
- Makes `TestTransaction::encode` infallible
- Makes un-encodable `TestTransaction` un-constructable

### Key places to review: 
`crates/example-types/src/block_types.rs`


crates/example-types/src/block_types.rs
```diff
@@ -17,38 +17,77 @@ use hotshot_types::{
...
+    }
+}
 
 impl TestTransaction {
+    /// Construct new transaction
+    ///
+    /// # Panics
+    /// If `bytes.len()` > `u32::MAX`
+    pub fn new(bytes: Vec<u8>) -> Self {
+        Self::try_new(bytes).expect("Vector too long")
```
QuentinI, line 47: This `expect` may be controversial. My reasoning was
- 4Gb transaction is an absurd notion we shouldn't ever encounter
- Rust stdlib is allowed to panic on allocation failure, this isn't that different of a case
- this is a testing type anyway, no production code will ever call this function

jparr721: That's a reasonable justification.

crates/examples/infra/mod.rs
```diff
@@ -472,7 +472,7 @@ pub trait RunDA<
                                         .get_transactions(leaf.get_block_header().metadata())
                                     {
                                         let restored_timestamp_vec =
-                                            tx.0[tx.0.len() - 8..].to_vec();
+                                            tx.bytes()[tx.bytes().len() - 8..].to_vec();
```
jparr721, line 475: This is so much better. 10/10 change.

crates/example-types/src/block_types.rs
```diff
@@ -17,38 +17,77 @@ use hotshot_types::{
...
+        Self::try_new(value).ok_or(TransactionTooLong)
+    }
+}
 
 impl TestTransaction {
+    /// Construct new transaction
+    ///
+    /// # Panics
+    /// If `bytes.len()` > `u32::MAX`
+    pub fn new(bytes: Vec<u8>) -> Self {
```
jparr721, line 46: Would this be better served as `impl From<Vec<u8>>` or something? That way we could go from bytes to tx and back? Open to your thoughts.

jparr721 (commented):
Great work. I especially like the improved self-documentation of the `bytes` methods.

crates/example-types/src/block_types.rs
QuentinI: It already implements `TryFrom<Vec<u8>>`. Do you mean ditching constructor entirely in favor of that implementation?

jparr721: Yeah that's what I was thinking. But if you prefer to have both, I'm not opposed. Just trying to think of it from a rust-y standpoint.

QuentinI: I think still having a constructor would be more rust-y, see for example all the wrapper types (`Cell`, `Arc`, `Mutex`, etc) in stdlib: they have corresponding `From` implementations, but provide `::new` as well.
Also, constructor is something that you almost definitely will want to call on a type, and to call `TestTransaction::try_from(x)` you'd have import `TryFrom` trait, which is a hurdle.

jparr721 (approved):
Sorry! I thought I approved this.

QuentinI:
@jparr721 please re-approve, resolved conflicts

jparr721 (approved):

PR #3122, opened 2024-05-07
Update the CDN (again)

rob-maron:
No linked issue.

This CDN update lets us specify which topics are valid at the application level. A subscription to an invalid topic will have their connection closed.

### DIFF
--- a/crates/hotshot/src/traits/networking/push_cdn_network.rs
+++ b/crates/hotshot/src/traits/networking/push_cdn_network.rs
@@ -11,7 +11,7 @@ use async_trait::async_trait;
 use bincode::config::Options;
 use cdn_broker::reexports::{
     connection::{protocols::Tcp, NoMiddleware, TrustedMiddleware, UntrustedMiddleware},
-    def::{ConnectionDef, RunDef},
+    def::{ConnectionDef, RunDef, Topic as TopicTrait},
     discovery::{Embedded, Redis},
 };
 #[cfg(feature = "hotshot-testing")]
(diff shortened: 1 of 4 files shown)


crates/hotshot/src/traits/networking/push_cdn_network.rs
```diff
@@ -155,13 +158,18 @@ pub struct PushCdnNetwork<TYPES: NodeType> {
...
 pub enum Topic {
     /// The global topic
     Global = 0,
     /// The DA topic
     DA = 1,
 }
 
+/// Implement the `TopicTrait` for our `Topic` enum. We need this to filter
+/// topics that are not implemented at the application level.
+impl TopicTrait for Topic {}
```
jparr721, line 171: Does this actually implement anything?

jparr721 (approved):
Looks like ci-1 flaked. I re ran it, but it shouldn't be due to your changes.

crates/hotshot/src/traits/networking/push_cdn_network.rs
rob-maron: The trait is an alias that makes sure it implements `TryFrom<u8>`, `Into<u8>`, `Clone`, `PartialEq`, etc to enable conversions to and from a u8

rob-maron: It also uses the default implementation for a function that prunes topics based on whether or not they are convertible to the application-level type

PR #3130, opened 2024-05-08
[CX_CLEANUP] - Make all fields in `Consensus` private to avoid accidental state update

### LINKED ISSUE #3129: [CX_CLEANUP] - Make all fields in `Consensus` private to avoid accidental state update

Similar to #3097, we should make all fields in `Consensus` private and add relevant setters and getters. This will prevent us from accidentally overriding any consensus state and make debugging easier.

_No response_

_No response_

Consensus state can be updated only via the corresponding update function.

_No response_

shenkeyao:
Closes #3129.

### This PR: 
* Makes all fields of the `Consensus` struct private, except `metrics`, to avoid overriding them with an older view.
* Adds setters and/or helpers for these fields.
* Adds loggings when the `update_*` functions fails.

### This PR does not: 
* Change whether/where to update the state.

### Key places to review: 
* `consensus.rs`.
* Places where `update_*` are called.


crates/task-impls/src/da.rs
```diff
@@ -209,9 +209,11 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
                 }
 
                 // Record the payload we have promised to make available.
-                consensus
-                    .saved_payloads
-                    .insert(view, Arc::clone(&proposal.data.encoded_transactions));
+                if let Err(e) = consensus
+                    .update_saved_payloads(view, Arc::clone(&proposal.data.encoded_transactions))
+                {
+                    tracing::trace!("{e:?}");
```
QuentinI, line 215: Wouldn't that be just "()"?

shenkeyao: Good catch! Incorrect pasting.

shenkeyao: Actually I should fix where this is copied from as well.

crates/task-impls/src/quorum_proposal.rs
```diff
@@ -535,8 +535,8 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalTaskState<TYPE
                         }
 
                         let mut consensus = self.consensus.write().await;
-                        if let Err(e) = consensus.update_high_qc(qc.clone()) {
-                            tracing::trace!("{e:?}");
+                        if consensus.update_high_qc(qc.clone()).is_err() {
+                            tracing::trace!("High QC with an equal or higher view exists.");
```
jparr721, line 539: To avoid needing to write this error every time since it looks like it was copied in another spot, we can put this message into the `ensure` statement. Same as the above custom trace messages, then we can just trace print the error. I think we can actually do `trace!(?e)` to avoid string templating.

jparr721 (changes requested):
We should probably centralize the error message. Other than that LGTM.

crates/task-impls/src/quorum_proposal.rs
shenkeyao: Yeah that's better. Done!

jparr721 (approved):

jparr721 (approved):

PR #3131, opened 2024-05-08
Improve view timeout log

jbearer:
* Include pub key of leader for that view
* Put relevant information in attributes so it is easier to parse and query


crates/task-impls/src/view_sync.rs
```diff
@@ -470,9 +470,12 @@ impl<
                 }
 
                 self.num_timeouts_tracked += 1;
+                let leader = self.membership.get_leader(view_number);
                 error!(
-                    "Num timeouts tracked since last view change is {}. View {} timed out",
-                    self.num_timeouts_tracked, *view_number
+                    %leader,
```
ss-es, line 475: Would it be nicer to use `cdn_proto::mnemonic(leader)`? 

This was a function Rob used to display names in the CDN brokers; I used this in the network task trace too

jbearer: I think it would be good to do both, in the sequencer logs and metrics we have the pub key itself

ss-es (approved):

PR #3132, opened 2024-05-08
Enable Broadcast On State Update and Broadcast Validated State Map

jparr721:
No linked issue.

### This PR: 
This PR begins the initial work to broadcast state updates when they occur within the central `Consensus` shared state. Right now, we suffer from race conditions due to the consensus state delaying to update, preventing what would have been an otherwise valid operation from taking place.

In this changeset, we broadcast these messages with the states. This serves two purposes:
1. It moves us gradually toward *fully removing* such state objects like `validated_state_map` from needing to exist whatsoever.
2. Allows for the dependency task tests to pass again since we may now gate on the existence of a particular value without needing to block and wait for the value to actually be inserted by whatever thread is sending it.

Further, the removal of these fields will reduce the lock contention on the consensus shared state.

To facilitate publishing events required a bit of maneuvering within the existing code. In particular, the `types` module needed to be able to broadcast hot shot events. This required moving some of these files around to more appropriate packages, and adding some dependencies to the module cargo.toml file. 

### This PR does not: 
This PR does not remove any fields, it is only additive.

### Key places to review: 
The movement of files and test changes.


bfish713:
Lets hold off on this while we work through Cappucino issues.  It seems like a small change but it does slightly alter current functionality

crates/types/src/lib.rs
```diff
@@ -14,6 +14,7 @@ pub mod constants;
 pub mod data;
 pub mod error;
 pub mod event;
+pub mod events;
```
ss-es, line 17: It looks like we have `event` and `events`. I think we should either condense these or give them different names

jparr721: Renaming to `hotshot_events` seems reasonable. What do you think?

crates/task/src/lib.rs
```diff
@@ -1,8 +1,42 @@
...
+#[allow(clippy::unused_async)]
+pub async fn cancel_task<T>(task: JoinHandle<T>) {
+    #[cfg(async_executor_impl = "async-std")]
+    task.cancel().await;
+    #[cfg(async_executor_impl = "tokio")]
+    task.abort();
+}
+
+/// Helper function to send events and log errors
+pub async fn broadcast_event<E: Clone + std::fmt::Debug>(event: E, sender: &Sender<E>) {
```
shenkeyao, line 26: Why are these two functions moved from `task-impls` to here? Looks like they are still called within `task-impls` only.

crates/types/src/view_sync.rs
```diff
@@ -0,0 +1,14 @@
+//! Types for the View Sync operations
+
+#[derive(PartialEq, PartialOrd, Clone, Debug, Eq, Hash)]
+/// Phases of view sync
+pub enum ViewSyncPhase {
```
shenkeyao, line 5: A similar question here. Why move this enum?

crates/task/src/lib.rs
jparr721: This is now also called in `types`

crates/types/src/view_sync.rs
jparr721: This is needed in `types` due to the use in HotShotEvent

shenkeyao (approved):

PR #3133, opened 2024-05-08
General task architecture refactor

### LINKED ISSUE #3208: [Task architecture] - Make `TaskState` object safe

To make stopping/restarting tasks more uniform, the `TaskState` trait needs to be made object safe.

There are a lot of constraints imposed by existing code (particularly our integration tests), and a large part of the work will be disentangling the varying implementations of the existing `TaskState` trait that are currently spread throughout our code for various different kinds of tasks.

_No response_

`TaskState` is object safe.

_No response_

ss-es:
Closes #3208 

The general goal of this PR is to make it possible to halt tasks, retrieve their state and restart them.

I think this would be nice for a number of reasons, but in particular it would allow us to very easily upgrade `NodeType` at runtime.

### This PR:
- Simplifies the `TaskState` trait: implementations must now only specify an associated `Event` type, the `handle_event` function and a `cancel_subtasks` function to cleanup any spawned subtasks.
- Tasks that receive their events from the network have had their `TaskState` implementations removed, and will be subsumed under a new `NetworkTaskState` in a future PR.
- Each task returns its state as a boxed `TaskState` object via a `JoinHandle`. These `JoinHandle`s are collected in a `ConsensusTaskRegistry`.
- Network tasks return a `JoinHandle<()>` are collected in a `NetworkTaskRegistry`, and will be dealt with in a future PR.
- Test tasks no longer implement `TaskState`, because they fundamentally do not listen on a single receiver in the same way that consensus tasks do.
- `TestTaskState` requires a `check()` function that returns the pass/fail result of the test. 

Minor:
- The `RwLock` and `Arc` indirection in task registries is removed, and these now simply contain a `Vec` of `JoinHandle`s. As a consequence, `SystemContextHandle` is not clonable; this is a mildly dangerous operation anyway, since a HotShot handle is a unique object and cloning it does not create a new one, and makes ownership harder to reason about, especially around mutable operations like `shut_down`. An application that wishes to hold multiple handles should put the handle it receives into a `Arc<RwLock<...>>` and synchronize calls to the handle itself.
- `shut_down` is no longer a `BoxSynFuture`, because shutting down the consensus registry is not `Sync` (it must await on the `JoinHandle`s it holds) 
- Various tasks no longer receive a complete cloned `SystemContextHandle`; they only needed 1-2 fields, so we simply pass these directly to avoid additional the additional `ConsensusApi` constraint.

### This PR does not: 
There should not be any external, observable changes in runtime behavior.

### Key places to review: 
This PR is quite large because of the scope of the changes, but the vast majority of it is rewiring. I think the most significant intentional changes are in:
- `task/src/task.rs`: the main trait changes. I believe it is easier to simply read the final file, rather than the diff.
- `testing/src/test_task.rs`: the test task changes. Note that `TestTask` and the `TestTaskState` trait were previously located in `task/src/task.rs`, but these were essentially rewritten from scratch.
But it's definitely worth looking at other files if you can. If there's anything you suspect might be an unintentional or incorrect logic change, please flag it.


crates/task-impls/src/vote_collection.rs
```diff
@@ -105,36 +104,6 @@ impl<
...
-    async fn handle_event(event: Self::Event, task: &mut Task<Self>) -> Option<Self::Output> {
-        let sender = task.clone_sender();
-        task.state_mut().handle_event(event, &sender).await
-    }
-
-    fn should_shutdown(event: &Self::Event) -> bool {
-        matches!(event.as_ref(), HotShotEvent::Shutdown)
-    }
-}
-
```
ss-es, line 137: I think implementing `TaskState` here (and calling the function `handle_event` below) was misleading, since we were never spawning the vote task as a task.

This may be a `CompletableTaskState` in the future, but that is outside the scope of this PR.

crates/task/src/task.rs
```diff
@@ -1,81 +1,45 @@
...
-    /// The state type
-    type State: TaskState;
-    /// Handle and incoming message and return `Some` if the task is finished
-    fn handle_message(
-        message: Self::Message,
-        id: usize,
-        task: &mut TestTask<Self::State, Self>,
-    ) -> impl Future<Output = Option<Self::Output>> + Send
-    where
-        Self: Sized;
```
ss-es, line 78: `TestTaskState` and `TestTask` are moved to `testing/src/test_task.rs`.

The trait and struct themselves have been somewhat reworked, but should fit into the test runner in the same way.

jparr721 (approved):

jparr721 (approved):

jparr721 (dismissed):

jparr721 (approved):

PR #3137, opened 2024-05-09
Unify committee/da terminology

### LINKED ISSUE #2898: [CLEANUP] - Replace `committee` with `da` in type names

We want to make the names consistent for types that are related to the DA committee, and `da` seems to make more sense than `committee`.

Replace `committee` with `da` in names for the network, membership, vote, etc.

Zulip discussion: https://espresso.zulipchat.com/#narrow/stream/311646-Consensus/topic/Minor.3A.20DA.20vs.2E.20committee.20in.20names/near/430901602.

All DA-related types have `da` in their names.

_No response_

rob-maron:
Closes #2898

### This PR: 

Unifies committee and DA terminology to always be "DA" instead.

### This PR does not: 

Change terminology in any of the election types. This is because "committee" there refers to the members of the _voting_ committee.


crates/example-types/src/node_types.rs
```diff
@@ -70,24 +70,24 @@ pub type StaticMembership = StaticCommittee<TestTypes>;
 
 impl<TYPES: NodeType> NodeImplementation<TYPES> for PushCdnImpl {
     type QuorumNetwork = PushCdnNetwork<TYPES>;
-    type CommitteeNetwork = PushCdnNetwork<TYPES>;
+    type DANetwork = PushCdnNetwork<TYPES>;
```
ss-es, line 73: I think we may want to call this `DaNetwork` based on https://rust-lang.github.io/api-guidelines/naming.html (`Uuid` vs `UUID`).

(I personally dislike this convention also, but I think we agreed on it in Zulip)

crates/task-impls/src/network.rs
```diff
@@ -309,8 +307,8 @@ impl<
                     maybe_action = Some(HotShotAction::DAVote);
                     (
                         vote.get_signing_key(),
-                        MessageKind::<TYPES>::from_consensus_message(SequencingMessage::Committee(
-                            CommitteeConsensusMessage::DAVote(vote.clone()),
+                        MessageKind::<TYPES>::from_consensus_message(SequencingMessage::DA(
+                            DAConsensusMessage::DAVote(vote.clone()),
```
ss-es, line 311: I guess this should be `Da` and `DaConsensusMessage`

ss-es (commented):
Just some very minor naming nitpicks from recent Zulip discussion, but looks good to me otherwise!

shenkeyao (approved):
`Da` looks so strange to me 😅 but thanks for making the changes!

PR #3146, opened 2024-05-10
Maybe fix CDN tests

bfish713:
Closes #<ISSUE_NUMBER>

Should fix the bad state causing unwrap to panic, and IconsistentBlocks error from popping up. 

### This PR: 

This PR checks that we actually have a proposal for the same view before trying to vote after we get our VID share.  Without this check if we got a VID share for one before or any view after the current proposal we have then we might store the state from the current proposal in our maps for the view of the VID share.  In other words there is a condition where we store state for the wrong view when the VID share view doesn't match the current proposals view.

### DIFF
--- a/crates/task-impls/src/consensus/mod.rs
+++ b/crates/task-impls/src/consensus/mod.rs
@@ -249,6 +249,9 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> ConsensusTaskState<TYPES, I>
         let Some(proposal) = self.current_proposal.clone() else {
             return;
         };
+        if proposal.get_view_number() != view {
+            return;
+        }
         let upgrade = self.decided_upgrade_cert.clone();
         let pub_key = self.public_key.clone();
         let priv_key = self.private_key.clone();
--- a/crates/task-impls/src/consensus/mod.rs
+++ b/crates/task-impls/src/consensus/mod.rs
@@ -498,6 +498,12 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> ConsensusTaskState<TYPES, I>
                 if disperse.data.recipient_key != self.public_key {
                     return;
                 }
+                let Some(proposal) = self.current_proposal.clone() else {
+                    return;
+                };
+                if proposal.get_view_number() != view {
+                    return;
+                }
                 self.spawn_vote_task(view, event_stream.clone());
             }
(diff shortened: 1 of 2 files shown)


jparr721 (approved):

crates/task-impls/src/consensus/mod.rs
```diff
@@ -249,6 +249,9 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> ConsensusTaskState<TYPES, I>
         let Some(proposal) = self.current_proposal.clone() else {
             return;
         };
+        if proposal.get_view_number() != view {
+            return;
+        }
```
rob-maron, line 254: Since I'm a bit unfamiliar with the changes to the task system since I last looked at it:
Does returning out of here just throw out the proposal?

bfish713: Ah no, we just have a clone of it here, we only throw out a proposal after we vote (see `QuorumVoteSend` handler)

crates/task-impls/src/consensus/mod.rs
```diff
@@ -498,6 +501,12 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> ConsensusTaskState<TYPES, I>
                 if disperse.data.recipient_key != self.public_key {
                     return;
                 }
+                let Some(proposal) = self.current_proposal.clone() else {
+                    return;
+                };
+                if proposal.get_view_number() != view {
+                    return;
+                }
```
ss-es, line 509: Does this actually do anything? Since we now check at the start of `spawn_vote_task` anyway.

bfish713: No but I don't mind being a bit redundant to reduce the odds of messing this up

bfish713: I'm happy to change it if you think we should

QuentinI: We don't check the view number though

bfish713: I added the view check in spawn_vote as well

QuentinI: Oh, duh, I misread "now", sorry.

ss-es (approved):
Looks good to me.

I wasn't following this issue so I don't really know the reason for the change, but I don't think this creates any problems.

PR #3151, opened 2024-05-11
[Consensus2] - Decouple quorum proposal task

### LINKED ISSUE #3171: [CX_CLEANUP] - Diverge the QuorumProposal task

Right now the quorum proposal task relies on calling the existing consensus functions, but isolates the call site so that way it can exist in an idempotent manner. This has created some issues due to the need to keep backwards compatibility with the consensus task. Instead, we will author our own new logic and only reuse functions that have limited scope.

- [ ] Make a new Quorum Proposal Task module which will hold all relevant logic for that task
- [ ] Fully remove all references to the old `consensus` task unless the functionality will live on in the new tasks
- [ ] Fully test and evaluate all new code in the unit testing suite to ensure that all expected behaviors are understood.
- [ ] Update the design document in notion with the architecture and event diagram.

_No response_

All tests must pass.

_No response_

jparr721:
Closes #3171 

### This PR: 
Right now the quorum proposal task relies on calling the existing consensus functions, but isolates the call site so that way it can exist in an idempotent manner. This has created some issues due to the need to keep backwards compatibility with the consensus task. Instead, we will author our own new logic and only reuse functions that have limited scope. Specifically:
- Makes a new Quorum Proposal Task module which will hold all relevant logic for that task
- Fully removes all references to the old `consensus` task unless the functionality will live on in the new tasks
- Fully tests and evaluate all new code in the unit testing suite to ensure that all expected behaviors are understood.

### This PR does not: 
- Handle upgrades.
- Handle the situation where the parent isn't found in the propose routine.

### Key places to review: 
- Please look at all instances of `ValidatedStateUpdated`, as this method is a gating requirement to be able to propose. Does it make sense to gate on this method? This updates `validated_state_map`, which we use when we propose.
- Please look at my [re-implementation of the leaf traversal ](https://github.com/EspressoSystems/HotShot/pull/3151/files#diff-288d7cb7ee198d9003021372718b10053928706cd6b0ebfefefa62f5eeb824edR91)
- Please look at my unit tests, and answer the following questions to yourself. If ANY of it is unclear, let me know ASAP.
    - Is `QuorumProposalValidated` referencing the right proposal and leaf? 
    - Is the `test_view_1` test evaluating the right genesis views?
    - Do the new events make sense?
    - Are timeouts and view sync handled properly?
- Do [these dependencies make sense?](https://github.com/EspressoSystems/HotShot/pull/3151/files#diff-5b4e77b35ab159e4d24fc0d24b8c8a2f0a47d242148f98bae8b1e0446a349e9bR292)
    - Please see the design document in notion for additional explanation. 
- Does [blocking on the consensus high qc update make sense?](https://github.com/EspressoSystems/HotShot/pull/3151/files#diff-5b4e77b35ab159e4d24fc0d24b8c8a2f0a47d242148f98bae8b1e0446a349e9bR422)
    - Please see the design document in notion for additional explanation.

Also PLEASE look at the design doc in notion! It features a deep-dive of all the changes as well as their motivation.


crates/task-impls/src/quorum_proposal/handlers.rs
```diff
@@ -0,0 +1,332 @@
...
+
+                        // The next leaf in the chain, if there is one, is decided, so this
+                        // leaf's justify_qc would become the QC for the decided chain.
+                        ret.new_decide_qc = Some(leaf.get_justify_qc().clone());
+                    } else if current_chain_length == 3 {
+                        // We've got the 3-chain, which means we can successfully decide on this leaf.
+                        ret.new_decided_view_number = Some(leaf.get_view_number());
+                    }
+                } else {
+                    // Bail out with empty values, but this is not necessarily an error, but we don't have. A
```
shenkeyao, line 172: Nit: "have. A" should be "have a".

crates/task-impls/src/quorum_proposal/handlers.rs
```diff
@@ -0,0 +1,332 @@
...
+
+    // We are in at least a 1-chain, so we start from here.
+    let mut current_chain_length: usize = 1;
+
+    // Get the state so we can traverse the chain to see if we have a 2 or 3 chain.
+    let walk_start_view_number = proposal_parent_view_number;
+
+    // The next view is the next view we're going to traverse, it is the validated state of the
+    // parent of the proposal.
+    let next_view = validated_state_map
```
shenkeyao, line 123: Can we rename `next_view`, `next_leaf` etc. to `parent_*`, to make it clearer that the view number is smaller than the current proposal?

crates/task-impls/src/quorum_proposal/handlers.rs
```diff
@@ -0,0 +1,332 @@
...
+
+    #[allow(clippy::cast_precision_loss)]
+    if let Some(decided_view_number) = new_decided_view_number {
+        // Bring in the cleanup crew. When a new decide is indeed valid, we need to clear out old memory.
+
+        let old_decided_view = consensus_writer.last_decided_view();
+        // TODO - collect garbage.
+
+        // Set the new decided view.
+        consensus_writer.update_last_decided_view(decided_view_number)?;
```
shenkeyao, line 285: Shouldn't we also send `LastDecidedViewUpdated`?

crates/testing/tests/tests_1/quorum_proposal_task.rs
shenkeyao, line 1: General comments on this file:
* I don't quite understand why `QuorumProposalValidated` and `QCFormed` in test inputs aren't referencing the same proposal.
* You mentioned that we should check "Are timeouts and view sync handled properly?", but I don't see many timeout or view sync related changes, so wanted to double check on this note.

crates/task-impls/src/consensus/helpers.rs
```diff
@@ -314,51 +311,42 @@ pub fn validate_proposal_view_and_certs<TYPES: NodeType>(
...
-        // NOTE: This happens on the genesis block
-        debug!(
-            "They don't equal: {:?}   {:?}",
-            leaf_commitment,
-            consensus.high_qc().get_data().leaf_commit
-        );
-    }
-
-    let leaf = consensus
+    let leaf = consensus_reader
```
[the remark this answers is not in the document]
jparr721: It wasn't doing anything, just logging. Also, I found the log to be pretty nondescript, so I had pulled it. Happy to add it back.

crates/task-impls/src/events.rs
```diff
@@ -156,8 +156,17 @@ pub enum HotShotEvent<TYPES: NodeType> {
 
     /* Consensus State Update Events */
     /// A undecided view has been created and added to the validated state storage.
-    ValidatedStateUpdate(TYPES::Time, View<TYPES>),
+    ValidatedStateUpdated(TYPES::Time, View<TYPES>),
```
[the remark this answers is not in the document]
jparr721: It's not emitted in this PR, the emission is added in the quorum proposal recv task.

crates/task-impls/src/quorum_proposal/mod.rs
```diff
@@ -383,12 +251,21 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalTaskState<TYPE
                     timeout_dependency.mark_as_completed(event);
                 }
                 Either::Left(_) => {
-                    qc_dependency.mark_as_completed(event);
+                    // qc_dependency.mark_as_completed(event);
```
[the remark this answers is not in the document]
jparr721: Oof, good catch!

crates/task-impls/src/events.rs
```diff
@@ -156,8 +156,17 @@ pub enum HotShotEvent<TYPES: NodeType> {
...
 
-    /// A new anchor view has been successfully reached by this node.
+    /// A new locked view has been created (2-chain)
+    LockedViewUpdated(TYPES::Time),
+
+    /// A new anchor view has been successfully reached by this node (3-chain).
     LastDecidedViewUpdated(TYPES::Time),
+
+    /// A new high_qc has been reached by this node.
+    HighQcUpdated(QuorumCertificate<TYPES>),
```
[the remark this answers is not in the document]
jparr721: Sure! I think all the names could use some kind of namespace.. it sucks rusts enums don't really allow for a clean way to do that. We should speak offline once we get things cleaned up more about the overall naming scheme.

crates/testing/tests/tests_1/quorum_proposal_task.rs
jparr721: Yeah they're in there and tested, just wanted to verify. Also, for the QCFormed event, since we are sending the justify_qc, wouldn't that mean we need to send it for the next proposal in the tests? Since the justify qc points to a previous state?

crates/task-impls/src/events.rs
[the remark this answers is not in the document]
jparr721: After some consideration, I realized that it doesn't really make sense to update the storage separate from the consensus state. I moved the call, that way any high qc update can be unambiguous.

crates/task-impls/src/quorum_proposal/dependency_handle.rs
```diff
@@ -0,0 +1,274 @@
...
+                        timeout_certificate = Some(timeout.clone());
+                    }
+                    either::Left(_) => {
+                        // Handled by the HighQcUpdated event.
+                    }
+                },
+                HotShotEvent::ViewSyncFinalizeCertificate2Recv(cert) => {
+                    view_sync_finalize_cert = Some(cert.clone());
+                }
+                HotShotEvent::ProposeNow(_, pdd) => {
```
jparr721, line 224: A later PR is going to completely re work this, so you can ignore checking this.

crates/task-impls/src/quorum_proposal/handlers.rs
jparr721: Done.

crates/task-impls/src/quorum_proposal/handlers.rs
jparr721: Fixed.

crates/task-impls/src/quorum_proposal/handlers.rs
jparr721: Correct, fixed.

crates/task-impls/src/quorum_proposal/handlers.rs
```diff
@@ -120,31 +122,38 @@ async fn visit_leaf_chain<TYPES: NodeType, I: NodeImplementation<TYPES>>(
 
     // The next view is the next view we're going to traverse, it is the validated state of the
     // parent of the proposal.
-    let next_view = validated_state_map
+    let parent_view = validated_state_map
```
shenkeyao, line 125: To avoid the check on line 144, would it be better to make this a while loop and move the while loop on line 140 inside here? Something like:
```rust
let mut view_number = proposal_parent_view_number;
while let Some(view) = validated_state_map.get(&view_number) {
  let leaf_commitment = view.leaf_commitment().context(...)?;
  let leaf = saved_leaves.get(&leaf_commitment).context(...)?;
  ...
  // Update view_number
}
```

I think this is clearer but I don't have a strong opinion, so it's up to you!

crates/task-impls/src/quorum_vote.rs
```diff
@@ -145,26 +145,27 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES> + 'static> HandleDepOutput
...
-                self.public_key.clone(),
-                self.consensus,
-                Arc::clone(&self.storage),
-                self.quorum_membership,
-                self.instance_state,
-                PhantomData,
-            )
-            .await;
-        }
+        // TODO
```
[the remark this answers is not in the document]
shenkeyao: Yeah I'm going to replace the `update_state_and_vote_if_able` call here with code to directly update the state.

shenkeyao (approved):
Looks good, just a comment about reordering code.

crates/task-impls/src/quorum_proposal/mod.rs
```diff
@@ -578,21 +447,27 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalTaskState<TYPE
...
+                    event_receiver,
+                    event_sender,
+                    event,
+                );
             }
-            #[cfg(feature = "dependency-tasks")]
             HotShotEvent::QuorumProposalValidated(proposal, _) => {
                 let new_view = proposal.view_number;
 
+                // All nodes get the latest proposed view as a proxy of `cur_view` of olde.
```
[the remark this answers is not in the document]
shenkeyao: Haha TIL!

crates/task-impls/src/events.rs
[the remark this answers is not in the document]
shenkeyao: I agree it makes more sense to update them both at the same place. In this case, though, can we rename this to `UpdateHighQc` to indicate it's not updated yet and should be done when handling the event?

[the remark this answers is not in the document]
jparr721: Done!

crates/task-impls/src/quorum_proposal/handlers.rs
jparr721: I wrote a test to make sure I didn't break anything, added this in! Great suggestion.

shenkeyao (approved):
LGTM

PR #3154, opened 2024-05-14
[CLEANUP] - Remove `get_` from all getters

### LINKED ISSUE #3104: [CLEANUP] - Remove `get_` from all getters

Based on https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter, we should remove `get_` from all getters.

_No response_

_No response_

Getter names no longer start with `get`.

_No response_

shenkeyao:
Closes #3104.

### This PR: 
* Removes `get_` from getters except for:
  * `get_or_calc_vid_share`, because it's either get or calculate.
  * `VidScheme` implementations, because it's defined in jellyfish.
* Removes duplicated functions that differ only in names.

### Key places to review: 
* Whether it makes sense to remove `get` from each name.

### DIFF
--- a/crates/hotshot/src/lib.rs
+++ b/crates/hotshot/src/lib.rs
@@ -146,7 +146,7 @@ pub struct SystemContext<TYPES: NodeType, I: NodeImplementation<TYPES>> {
 
     /// Access to the output event stream.
     #[deprecated(
-        note = "please use the `get_event_stream` method on `SystemContextHandle` instead. This field will be made private in a future release of HotShot"
+        note = "please use the `event_stream` method on `SystemContextHandle` instead. This field will be made private in a future release of HotShot"
     )]
     pub output_event_stream: (Sender<Event<TYPES>>, InactiveReceiver<Event<TYPES>>),
 
@@ -229,14 +229,14 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> SystemContext<TYPES, I> {
         let validated_state = match initializer.validated_state {
             Some(state) => state,
             None => Arc::new(TYPES::ValidatedState::from_header(
-                anchored_leaf.get_block_header(),
+                anchored_leaf.block_header(),
             )),
         };
 
         // Insert the validated state to state map.
         let mut validated_state_map = BTreeMap::default();
         validated_state_map.insert(
-            anchored_leaf.get_view_number(),
+            anchored_leaf.view_number(),
             View {
                 view_inner: ViewInner::Leaf {
                     leaf: anchored_leaf.commit(),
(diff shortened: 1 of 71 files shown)


QuentinI (approved):
Looks great! Thanks for doing this

crates/hotshot/src/lib.rs
```diff
@@ -146,7 +146,7 @@ pub struct SystemContext<TYPES: NodeType, I: NodeImplementation<TYPES>> {
 
     /// Access to the output event stream.
     #[deprecated(
-        note = "please use the `get_event_stream` method on `SystemContextHandle` instead. This field will be made private in a future release of HotShot"
+        note = "please use the `event_stream` method on `SystemContextHandle` instead. This field will be made private in a future release of HotShot"
     )]
     pub output_event_stream: (Sender<Event<TYPES>>, InactiveReceiver<Event<TYPES>>),
```
ss-es, line 151: I think this can probably be removed now

jparr721 (approved):

crates/orchestrator/src/client.rs
```diff
@@ -193,7 +193,7 @@ impl OrchestratorClient {
     /// # Errors
     /// If we were unable to serialize the Libp2p data
     #[allow(clippy::type_complexity)]
-    pub async fn get_config_without_peer<K: SignatureKey>(
+    pub async fn config_without_peer<K: SignatureKey>(
```
ss-es, line 196: I think this should actually be `get_config_without_peer`. It's not a getter method, and I think `get` here is in the sense of `GET` as an HTTP request.

I would probably just revert the changes to the orchestrator, I think all the `get_` functions here are similar.

ss-es (changes requested):
requesting changes for the orchestrator

crates/hotshot/src/types/handle.rs
```diff
@@ -47,7 +47,7 @@ pub struct SystemContextHandle<TYPES: NodeType, I: NodeImplementation<TYPES>> {
 
 impl<TYPES: NodeType, I: NodeImplementation<TYPES> + 'static> SystemContextHandle<TYPES, I> {
     /// obtains a stream to expose to the user
-    pub fn get_event_stream(&self) -> impl Stream<Item = Event<TYPES>> {
+    pub fn event_stream(&self) -> impl Stream<Item = Event<TYPES>> {
```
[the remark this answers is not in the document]
ss-es: (so this is probably fine, was just thinking out loud)

crates/hotshot/src/lib.rs
shenkeyao: Just to double-check my understanding. Did you mean that we can remove `pub` and delete this note, or that we can actually remove this field?

crates/orchestrator/src/client.rs
shenkeyao: Good catch! Yeah, I think it makes more sense to keep the `get` for these functions.

crates/hotshot/src/lib.rs
shenkeyao: Updated assuming it's the former case, since it looks like we can't just remove this function.

ss-es: Oh yes, sorry! I think I had in my head "just delete it", but you're right -- it's a necessary field, so we just need to make it private.

PR #3155, opened 2024-05-14
[CLEANUP] - Capitalize acronyms

### LINKED ISSUE #3105: [CLEANUP] - Capitalize acronyms

According to https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case,
> In UpperCamelCase, acronyms and contractions of compound words count as one word: use Uuid rather than UUID,

so `VID`, `DA`, and `ID` should be `Vid`, `Da`, and `Id`, respectively, in struct and enum names.

We should rename other types if applicable.

_No response_

_No response_

Types are properly renamed following the guideline.

_No response_

shenkeyao:
Closes #3105.

### This PR: 
* Renames `VID`, `DAC`, `QC` etc. to `Vid`, `Dac`, `Qc`.

### DIFF
--- a/crates/types/src/message.rs
+++ b/crates/types/src/message.rs
@@ -12,13 +12,13 @@ use derivative::Derivative;
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 
 use crate::{
-    data::{DAProposal, Leaf, QuorumProposal, UpgradeProposal, VidDisperseShare},
+    data::{DaProposal, Leaf, QuorumProposal, UpgradeProposal, VidDisperseShare},
     simple_certificate::{
-        DACertificate, ViewSyncCommitCertificate2, ViewSyncFinalizeCertificate2,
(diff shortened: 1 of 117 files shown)


crates/types/src/message.rs
```diff
@@ -82,7 +82,7 @@ pub enum MessagePurpose {
     /// Message with a view sync certificate.
     ViewSyncCertificate,
     /// Message with a DAC.
-    DAC,
+    Dac,
```
ss-es, line 85: I was going to suggest `DaCertificate` by analogy with the others, but after a quick search I'm wondering if we even use these. Was this just for the web server?

shenkeyao: Yeah looks like the `Dac` variant isn't used. I'm wondering if we can actually remove the `MessagePurpose` enum and probably some other relevant types.

I just made an issue for this: https://github.com/EspressoSystems/HotShot/issues/3179. In this PR I just renamed it to `DaCertificate` which I think is a good suggestion!

jparr721 (approved):

PR #3158, opened 2024-05-14
Make metrics labels more compatible with Prometheus

jbearer:
### This PR: 

* Add `MetricsFamily` trait to group related metrics that are partitioned by values of their labels. This gives us a way to create metrics that actually use Prometheus labels, a capability we haven't had until now.
* Rename `Label` to `Text` to distinguish it from the Prometheus concept of a label
* Change the way `Text` works. Prometheus doesn't support top-level key-value pairs with text values, so implementing the old `Label` trait for Prometheus, we were forced to hack it by putting the value in a comment. A better approach is to have a "text" metric where the value of the metric is the name (this can be implemented, e.g., as a gauge with a dummy value of 1). Additional key-value pairs can be added using families/labels as described above.
* Remove dead code `Metrics` implementations in consensus and networking

The motivation for this is a request from node operators to make the version metrics populated by the sequencer more parseable. For this, we only need a family with a single instance, whose label values contain information like git revision and whatnot. However, this implementation is more general, and will allow us to fully take advantage of Prometheus labels if we want to in the future.

Examples of both the simple "version label" usage and a more complex use of labels can be found in the doc comments.

### This PR does not: 

Provide a real, prometheus-based implementation of the metrics functionality. As before, the implementation lives in `hotshot-query-service`.

### Key places to review: 

crates/hotshot-types/src/traits/metrics.rs

### DIFF
--- a/crates/hotshot/src/traits/networking.rs
+++ b/crates/hotshot/src/traits/networking.rs
@@ -10,13 +10,9 @@ pub mod libp2p_network;
 pub mod memory_network;
 /// The Push CDN network
 pub mod push_cdn_network;
-use std::{
-    collections::HashMap,
-    sync::{Arc, Mutex},
-};
 
 use custom_debug::Debug;
-use hotshot_types::traits::metrics::{Counter, Gauge, Histogram, Label, Metrics, NoMetrics};
+use hotshot_types::traits::metrics::{Counter, Gauge, Metrics, NoMetrics};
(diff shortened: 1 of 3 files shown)


jparr721 (approved):
Love the doc comments! In the future, please feel free to assign someone from the team in the PR to ensure it gets reviewed promptly.

PR #3159, opened 2024-05-15
Add flakiness script

QuentinI:
Closes #0000

### This PR: 
Adds a script that runs full suite N times, saves the logs and tracks failed iterations.

### DIFF
--- a/scripts/flakiness.sh
+++ b/scripts/flakiness.sh
@@ -1,0 +1,74 @@
+#!/usr/bin/env sh
+
+N_ITERATIONS=100
+OUT_DIR="$(pwd)/flakiness"
+TEST_THREADS=1
+
+# Make script killable with Ctrl+C
+trap "echo; exit" INT
+
+die() {
+    echo "${1}";
+    exit 1;
+}
+
+# Parse arguments
+while [ $# -gt 0 ]; do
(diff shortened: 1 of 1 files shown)


scripts/flakiness.sh
```diff
@@ -0,0 +1,74 @@
...
+
+# Don't clobber previous results
+if [ -d "${OUT_DIR}" ]; then
+  if [ "$(ls -A "${OUT_DIR}")" ]; then
+    die "${OUT_DIR} is not empty"
+  fi
+fi
+
+# Check for nextest
+cargo nextest -V > /dev/null 2>&1 || die "cargo-nextest is not installed"
```
ss-es, line 52: For some reason `nextest` seems to already be installed in the nix-shell, but it isn't explicit so I'm not sure how.

Should we add "nextest" to `rust-toolchain.toml`? (I don't know if it's necessary or not, just wondering.)

bfish713 (approved):

scripts/flakiness.sh
QuentinI: I don't think we can add nextest to `rust-toolchain.toml`, as it's not a standard component. Non-Nix people would need to manually install it no matter what.

PR #3164, opened 2024-05-15
Add default `genesis` to trait

tbro:
Avoids downstream implementations.

-------
We have a TODO in sequencer here: 
https://github.com/EspressoSystems/espresso-sequencer/blob/77484507b7b8a5e94ecbc38dcae356e90d9bbba6/sequencer/src/block.rs#L65

### This PR: 
Adds a default implantation of `genesis` to `BlockPayload`.

### This PR does not: 
It should not change any functionality and only make changes that are required to complete the above.

### Key places to review: 
The intended change is in `BlockPayload` trait in
`crates/types/src/traits/block_contents.rs`. That default
implementation allows us to remove the implementation from
`crates/example-types/src/block_types.rs`.

### DIFF
--- a/crates/types/src/traits/block_contents.rs
+++ b/crates/types/src/traits/block_contents.rs
@@ -88,7 +88,13 @@ pub trait BlockPayload:
     fn from_bytes(encoded_transactions: &[u8], metadata: &Self::Metadata) -> Self;
 
     /// Build the genesis payload and metadata.
-    fn genesis() -> (Self, Self::Metadata);
+    #[must_use]
+    fn genesis() -> (Self, Self::Metadata)
+    where
+        <Self as BlockPayload>::Instance: Default,
+    {
+        Self::from_transactions([], &Default::default()).unwrap()
+    }
 
     /// List of transaction commitments.
     fn transaction_commitments(
--- a/crates/example-types/src/block_types.rs
+++ b/crates/example-types/src/block_types.rs
@@ -198,10 +198,6 @@ impl BlockPayload for TestBlockPayload {
         Self { transactions }
     }
 
-    fn genesis() -> (Self, Self::Metadata) {
-        (Self::genesis(), TestMetadata)
-    }
-
     fn builder_commitment(&self, _metadata: &Self::Metadata) -> BuilderCommitment {
         let mut digest = sha2::Sha256::new();
         for txn in &self.transactions {


dailinsubjam (commented):
I didn't see anything wrong, but might be better to wait for others more familiar with this part to approve it.

crates/types/src/traits/block_contents.rs
```diff
@@ -88,7 +88,13 @@ pub trait BlockPayload:
     fn from_bytes(encoded_transactions: &[u8], metadata: &Self::Metadata) -> Self;
 
     /// Build the genesis payload and metadata.
-    fn genesis() -> (Self, Self::Metadata);
+    #[must_use]
+    fn genesis() -> (Self, Self::Metadata)
+    where
+        <Self as BlockPayload>::Instance: Default,
+    {
+        Self::from_transactions([], &Default::default()).unwrap()
```
ss-es, line 96: I think a default implementation makes sense, though I'm a little uncomfortable with a default implementation that may panic depending on how a non-default (`from_transactions`) is implemented.

It's not a huge deal since I guess this is just a startup panic, but is there any way to avoid this? e.g. are any of our `from_transactions` actually fallible?

tbro: On quick review it appears that the implantation on the sequencer is used everywhere.  The only real failure case I see is [from_namespace_offsets](https://github.com/EspressoSystems/espresso-sequencer/blob/0ab566c81721352e9db3bc0e551a519dbeb8e9b0/sequencer/src/block/payload.rs#L231). I'm assuming that it isn't a huge deal b/c it is itself being unwrapped. But maybe @ggutoski,  @jbearer would have more insight.

ggutoski: My opinion: it's okay because it's just a startup panic, so if there's something wrong then we'll know immediately and there's zero risk of loss. If `from_transactions` fails on an empty list then that's pretty bad and should be fixed. 😆 

Regarding the current implementation details: please keep in mind that everything in `sequencer/src/block` is currently being rewritten, so don't worry to much about what's currently in there.

ggutoski (approved):
Approving because this change seems pretty innocuous to me. See my reply below.

Regarding this explainer comment:
https://github.com/EspressoSystems/espresso-sequencer/blob/77484507b7b8a5e94ecbc38dcae356e90d9bbba6/sequencer/src/block.rs#L67-L70

Has this "future update to hotshot" occurred yet? Seems like it's still a "workaround" and we're merely doing the workaround at the trait level instead of in the underlying implementation. Correct? I see no problem with that. (Indeed, it makes sense to me to take care of it at the trait level.) I just want to clarify status.

tbro:
> Has this "future update to hotshot" occurred yet? Seems like it's still a "workaround" and we're merely doing the workaround at the trait level instead of in the underlying implementation. Correct? I see no problem with that. (Indeed, it makes sense to me to take care of it at the trait level.) I just want to clarify status.

Sorry I missed this comment. True, some discussion occurred with @jbearer where it was clarified that the appropriate action was to define this default function on the trait. I guess you could say that the work around is just being moved, but I'm not seeing a better solution. With this in HotShot we *will* be able to remove that comment and TODO on the sequencer side.

PR #3184, opened 2024-05-17
Don't lock storage

bfish713:
Closes #<ISSUE_NUMBER>

### This PR: 

Fixes more places where we held a lock through some expensive blocks.  

- Don't hold a lock through `BlockHeader::new`
- Don't hold a lock through storing things

### DIFF
--- a/crates/task-impls/src/da.rs
+++ b/crates/task-impls/src/da.rs
@@ -144,7 +144,9 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> + 
                 .await;
             }
             HotShotEvent::DaProposalValidated(proposal, sender) => {
-                if self.consensus.read().await.cur_view() > proposal.data.view_number() + 1 {
+                let curr_view = self.consensus.read().await.cur_view();
+                if curr_view > proposal.data.view_number() + 1 {
+                    tracing::debug!("Validated DA proposal for prior view but it's too old now Current view {:?}, DA Proposal view {:?}", curr_view, proposal.data.view_number());
                     return None;
                 }
                 // Proposal is fresh and valid, notify the application layer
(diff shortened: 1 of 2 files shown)


crates/task-impls/src/da.rs
```diff
@@ -145,6 +144,9 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
                 .await;
             }
             HotShotEvent::DaProposalValidated(proposal, sender) => {
+                if self.consensus.read().await.cur_view() > proposal.data.view_number() + 1 {
+                    return None;
```
jbearer, line 148: Log someething here

crates/task-impls/src/da.rs
```diff
@@ -107,8 +107,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, A: ConsensusApi<TYPES, I> +
                 // the `DaProposalRecv` event. Otherwise, the view number subtraction below will
                 // cause an overflow error.
                 // TODO ED Come back to this - we probably don't need this, but we should also never receive a DAC where this fails, investigate block ready so it doesn't make one for the genesis block
-
-                if self.cur_view != TYPES::Time::genesis() && view < self.cur_view - 1 {
+                if self.consensus.read().await.cur_view() > proposal.data.view_number() + 1 {
```
jbearer, line 110: This seems to be changing the semantics (`self.cur_view` vs `consensus` view). Are we sure this is correct?

bfish713: I'll leave this ass is.  I'm pretty sure it's correct but probably not worth the change right now.  I want to protect agains the DA task falling behind consensus because of VID calculations but I think the other new check covers that

shenkeyao (approved):

PR #3203, opened 2024-05-21
Multiple builders

### LINKED ISSUE #3181: [BUILDER] - Multiple builder support

HotShot should support multiple builders.

- [ ] Multiple builder support
- [ ] Tests for multiple builders

_No response_

- Multiple builder addresses can be specified in config
- HotShot queries all builders for available blocks
- There are tests checking HotShot behaviour w.r.t builders, especially various faulty builder cases

_No response_

QuentinI:
Closes #3181 

### This PR: 
- Adds support for multiple builders
- Adds support for multiple builders in testing harness
- Adds ability to simulate builder failures in tests

### Key places to review:
`crates/task-impls/src/transactions.rs`
`crates/testing/tests/tests_3/test_with_builder_failures.rs`


crates/task-impls/src/transactions.rs
```diff
@@ -312,93 +327,174 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, Ver: StaticVersionType>
...
+                .collect::<FuturesUnordered<_>>();
+            let mut results = Vec::with_capacity(self.builder_clients.len());
+
+            // Instant we start querying builders for available blocks
+            let query_start = Instant::now();
+
+            // First we complete the query to the fastest fraction of the builders
+            let threshold = (self.builder_clients.len() * BUILDER_MAIN_BATCH_THRESHOLD_DIVIDEND)
+                .div_ceil(BUILDER_MAIN_BATCH_THRESHOLD_DIVISOR);
+            let mut tasks = tasks.take(threshold);
```
nyospe, line 358: Oh! Very nice solution!

nyospe (approved):
Yes, I see that this is still a draft, but...

The implementation is both functional and elegant, and assuming the remaining testing holds up, I'm happy with this solution.

crates/examples/infra/mod.rs
```diff
@@ -929,24 +963,26 @@ pub async fn main_entry_point<
                 <RandomBuilderImplementation as TestBuilderImplementation<TYPES>>::start(
                     run_config.config.num_nodes_with_stake.into(),
                     run_config.random_builder.clone().unwrap_or_default(),
+                    HashMap::new(),
                 )
                 .await;
 
-            run_config.config.builder_url = builder_url;
+            run_config.config.builder_urls = vec1::vec1![builder_url];
```
jparr721, line 970: For an example, it seems like vec1 could potentially be overkill? If during initialization we always pass a builder url, and presumably die if one is received, is it necessarily required to have hardline enforcement like this?

crates/task-impls/src/transactions.rs
```diff
@@ -312,93 +327,174 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, Ver: StaticVersionType>
...
-        if !block_info.sender.validate_block_info_signature(
-            &block_info.signature,
-            block_info.block_size,
-            block_info.offered_fee,
-            &block_info.block_hash,
-        ) {
-            bail!("Failed to verify available block info response message signature");
-        }
+        // Create a future for every builder
+        let mut available_blocks = {
```
jparr721, line 331: Could we perhaps get some comments in this code to explain more clearly what's going on? I noticed this method lacks a doc comment. This flow features some non-trivial nesting and chaining, and requires some effort to parse, and it's not abundantly clear, for instance, why we map each block to its index.

crates/task-impls/src/transactions.rs
```diff
@@ -312,93 +327,174 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, Ver: StaticVersionType>
...
+                    client
+                        .available_blocks(
+                            parent_comm,
+                            view_number.u64(),
+                            self.public_key.clone(),
+                            parent_comm_sig,
+                        )
+                        .await
+                        .map(move |blocks| {
+                            blocks.into_iter().map(move |block_info| (block_info, idx))
```
jparr721, line 346: Is the intention always to enumerate everything? If so, is adding `idx` relevant here since we'll always have the top-level context of the associated block?

crates/task-impls/src/transactions.rs
```diff
@@ -312,93 +327,174 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, Ver: StaticVersionType>
...
+            // Then we query the rest, alotting additional `elapsed * BUILDER_ADDITIONAL_TIME_MULTIPLIER`
+            // for them to respond. There's a fixed floor of `BUILDER_MINIMUM_QUERY_TIME` for both
+            // phases
+            let timeout = async_sleep(std::cmp::max(
+                query_start
+                    .elapsed()
+                    .mul_f32(BUILDER_ADDITIONAL_TIME_MULTIPLIER),
+                BUILDER_MINIMUM_QUERY_TIME - query_start.elapsed(),
+            ));
+            futures::pin_mut!(timeout);
```
jparr721, line 375: The `pin_mut` here is not immediately clear at first glance. Can we get a comment explaining why we're using it here?

crates/task-impls/src/transactions.rs
```diff
@@ -312,93 +327,174 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, Ver: StaticVersionType>
...
-        // Verify signature over chosen block.
-        if !block_info.sender.validate_block_info_signature(
-            &block_info.signature,
-            block_info.block_size,
-            block_info.offered_fee,
-            &block_info.block_hash,
-        ) {
-            bail!("Failed to verify available block info response message signature");
-        }
+        // Create a future for every builder
```
jparr721, line 330: This function is quite long, is it possible to pare it down some? It looks like we're doing a lot of logic that's not necessarily tightly related.

crates/testing/src/block_builder/mod.rs
```diff
@@ -0,0 +1,177 @@
...
+    });
+}
+
+/// Helper function to construct all builder data structures from a list of transactions
+async fn build_block<TYPES: NodeType>(
+    transactions: Vec<TYPES::Transaction>,
+    num_storage_nodes: usize,
+    pub_key: TYPES::BuilderSignatureKey,
+    priv_key: <TYPES::BuilderSignatureKey as BuilderSignatureKey>::BuilderPrivateKey,
+) -> (
```
jparr721, line 114: Building a block is pretty crucial, so I think that perhaps it'd be better to have this return type be something other than a tuple. Can we just whip up a struct or use the `BlockEntry` struct that we use elsewhere?

crates/testing/src/block_builder/simple.rs
```diff
@@ -0,0 +1,368 @@
...
+            .transactions
+            .read(|txns| {
+                Box::pin(async {
+                    txns.values()
+                        .filter(|txn| {
+                            // We want transactions that are either unclaimed, or claimed long ago
+                            // and thus probably not included, or they would've been decided on
+                            // already and removed from the queue
+                            txn.claimed
+                                .map(|claim_time| claim_time.elapsed() > Duration::from_secs(30))
```
jparr721, line 160: The magic number of 30 seconds here is a bit unclear. Why specifically 30 seconds? Can we get some clarification in a comment about this?

crates/testing/src/block_builder/simple.rs
```diff
@@ -0,0 +1,368 @@
...
+                                .unwrap_or(true)
+                        })
+                        .cloned()
+                        .map(|txn| txn.transaction)
+                        .collect::<Vec<TYPES::Transaction>>()
+                })
+            })
+            .await;
+
+        if transactions.is_empty() {
```
jparr721, line 170: I understand the comment here, but I don't quite get how returning an empty set of block info implies that we'll timeout if we don't get any info. What would the alternative be which would create the negative case highlighted by this comment? Could we perhaps get an example in the comment of how this could accidentally be misappropriated? From the perspective of someone with no context, I might be fearful of changing this code as the ramifications are not inherently super clear to me.

crates/testing/src/overall_safety_task.rs
```diff
@@ -253,7 +253,7 @@ pub struct RoundResult<TYPES: NodeType> {
     /// block -> # entries decided on that block
     pub block_map: HashMap<VidCommitment, usize>,
 
-    /// node idx -> number transactions
+    /// number transactions -> node idx
```
jparr721, line 256: How come this order changed? Also, nit: "number of transactions"

crates/testing/src/test_builder.rs
```diff
@@ -70,6 +71,21 @@ pub struct TestDescription {
...
+pub enum BuilderChange {
+    Up,
+    Down,
+    FailClaims(bool),
+}
+
+#[derive(Clone, Debug)]
+pub struct BuilderDescription {
+    /// Views to toggle builder on
+    pub changes: HashMap<u64, BuilderChange>,
```
jparr721, line 88: I am not super clear on the name of the variable compared to the doc comment. Is this saying that we have some views where a builder may be on or off? I think perhaps `changes` might not be fully descriptive of what this mechanism does. Furthermore, is this intended to be pre-filled before the test? Or during as a sort of state machine that gets written to and deleted from when done?

crates/testing/src/test_builder.rs
```diff
@@ -70,6 +71,21 @@ pub struct TestDescription {
     pub unreliable_network: Option<Box<dyn NetworkReliability>>,
     /// view sync check task
     pub view_sync_properties: ViewSyncTaskDescription,
+    /// description of builders to run
+    pub builders: Vec1<BuilderDescription>,
+}
+
+#[derive(Clone, Debug)]
+pub enum BuilderChange {
```
jparr721, line 79: Can we get a descriptive comment for what this enum is doing and what each of its fields are doing? It's not immediately clear what `Up` and `Down` etc are doing to an actively running builder, and how that affects the test.

crates/testing/tests/tests_3/test_with_builder_failures.rs
```diff
@@ -0,0 +1,68 @@
...
+        // 5    Up   Up
+        // 6    Down Up
+        // 7    Up   Up
+        // ...
+        //
+        // We give each builder a view of uptime before making it the only available builder so that it
+        // has time to initialize
+
+        // First builder will always respond with available blocks, but will periodically fail claim calls
+        #[allow(clippy::unnecessary_filter_map)] // False positive
```
jparr721, line 42: Is this a false positive? Does clippy not provide a working replacement option? It looks like you could fill the map then just swap the value for key 2 and 0 in this and the below builder without needing to map over it at all.

jparr721 (changes requested):
Some questions, happy to discuss anything further.

crates/examples/infra/mod.rs
QuentinI: We don't always re-assign the builder URL - for example, in case of External builder
My reasoning was:
- To run, HotShot needs at least one builder URL, hard requirement
- Vec1 represents this more accurately
- Vec1 allows for less manual checks for whether builder URLs are non-empty, especially in deserialization of HS config and in examples, where we do some complex manipulations with `run_config`

I'm not particularly attached to Vec1 here if you think there are reasons to avoid it.

crates/testing/src/test_builder.rs
QuentinI: Good catch! Curiously, clippy doesn't catch this

crates/testing/tests/tests_3/test_with_builder_failures.rs
QuentinI: > Is this a false positive?

Oh, wow. The reason I though it was a false positive was because this was supposed to be 
```rust
                0 => Some((view_num, BuilderChange::Down)),
                1 => Some((view_num, BuilderChange::Up)),
                _ => None, // We don't want to send Up when builder is up already, that'd
                                   // restart it
```
and I somehow convinced myself that that's what I wrote... Thank you for the catch!

> It looks like you could fill the map then just swap the value for key 2 and 0 in this and the below builder without needing to map over it at all.

Not sure what you mean by this though - the keys range from 0 to the number of views we're testing for, it's not only 0 and 2.

crates/testing/src/test_builder.rs
QuentinI: Yeah, that's very lazy documentation. Thanks for calling me out on that, I'll make it more sensible.
`changes` is prescriptive to the builder, telling it how to behave (stop, start, continue working but fail claim calls), I named it by analogue with "node changes" mechanism in the harness which does the same for HS nodes. Do you have any suggestions on alternative naming?

crates/testing/src/overall_safety_task.rs
QuentinI: The documentation was just incorrect. That sad, it's still wrong, it's not node index but number of nodes reporting this number of transactions, I'll make amends.

crates/testing/src/block_builder/mod.rs
QuentinI: `BlockEntry` makes sense here, I'll change it

crates/task-impls/src/transactions.rs
QuentinI: I'll add more documentation here as well

crates/task-impls/src/transactions.rs
QuentinI: We want to know what our available blocks are and for each of the blocks record which builder we need to call to claim it.
Would you consider renaming `idx` to `builder_idx` a sufficient improvement?

jparr721: Sure, that seems reasonable.

crates/testing/tests/tests_3/test_with_builder_failures.rs
jparr721: Oh, hah pardon my blindness on the second point. This sounds fine.

crates/testing/src/block_builder/simple.rs
jparr721: Just realized this is a test file. I'm okay with leaving this as-is since we have magic numbers in other tests. Sorry about that!

jparr721 (approved):
Great improvements to the comments. LGTM

nyospe (dismissed):
LGTM

dailinsubjam (commented):
Just a few benchmark-related questions: Is there an instruction on starting multiple builders? And who will run these builders?

dailinsubjam (approved):

PR #3206, opened 2024-05-22
Allow genesis to time out

rob-maron:
No linked issue.

During test deployments, we ran into a case where the genesis block would never time out. If we ever didn't see genesis for any reason (not enough nodes saw it from the CDN, builder has wrong commitment), the network would hang indefinitely. 

### This PR
Fixes that fragility by always requiring a timeout task to be present. 

I tested it upstream on the sequencer and we can indeed time out view 0 now

### DIFF
--- a/crates/hotshot/src/tasks/task_state.rs
+++ b/crates/hotshot/src/tasks/task_state.rs
@@ -8,20 +8,12 @@ use std::{
 use async_compatibility_layer::art::{async_sleep, async_spawn};
 use async_trait::async_trait;
 use hotshot_task_impls::{
-    builder::BuilderClient,
-    consensus::ConsensusTaskState,
-    consensus2::Consensus2TaskState,
-    da::DaTaskState,
-    events::HotShotEvent,
-    helpers::broadcast_event,
-    quorum_proposal::QuorumProposalTaskState,
-    quorum_proposal_recv::QuorumProposalRecvTaskState,
-    quorum_vote::QuorumVoteTaskState,
-    request::NetworkRequestState,
-    transactions::TransactionTaskState,
-    upgrade::UpgradeTaskState,
-    vid::VidTaskState,
-    view_sync::{ViewSyncReplicaTaskState, ViewSyncTaskState},
+    builder::BuilderClient, consensus::ConsensusTaskState, consensus2::Consensus2TaskState,
+    da::DaTaskState, events::HotShotEvent, helpers::broadcast_event,
+    quorum_proposal::QuorumProposalTaskState, quorum_proposal_recv::QuorumProposalRecvTaskState,
+    quorum_vote::QuorumVoteTaskState, request::NetworkRequestState,
+    transactions::TransactionTaskState, upgrade::UpgradeTaskState, vid::VidTaskState,
+    view_sync::ViewSyncTaskState,
 };
 use hotshot_types::traits::{
     consensus_api::ConsensusApi,
@@ -200,7 +192,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> CreateTaskState<TYPES, I>
         let next_view_timeout = handle.hotshot.config.next_view_timeout;
         let timeout_task = async_spawn({
             // Nuance: We timeout on the 1 here because that means that we have
-            // not seen evidence to transition to transition to it.
+            // not seen evidence to transition to it.
             let view_number = 1;
             async move {
                 async_sleep(Duration::from_millis(next_view_timeout)).await;
@@ -280,7 +272,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> CreateTaskState<TYPES, I>
         let next_view_timeout = handle.hotshot.config.next_view_timeout;
         let timeout_task = async_spawn({
             // Nuance: We timeout on the 1 here because that means that we have
-            // not seen evidence to transition to transition to it.
+            // not seen evidence to transition to it.
             let view_number = 1;
             async move {
                 async_sleep(Duration::from_millis(next_view_timeout)).await;
@@ -327,7 +319,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> CreateTaskState<TYPES, I>
         let next_view_timeout = handle.hotshot.config.next_view_timeout;
         let timeout_task = async_spawn({
             // Nuance: We timeout on the 1 here because that means that we have
-            // not seen evidence to transition to transition to it.
+            // not seen evidence to transition to it.
             let view_number = 1;
             async move {
                 async_sleep(Duration::from_millis(next_view_timeout)).await;
@@ -372,7 +364,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> CreateTaskState<TYPES, I>
         let next_view_timeout = handle.hotshot.config.next_view_timeout;
         let timeout_task = async_spawn({
             // Nuance: We timeout on the 1 here because that means that we have
-            // not seen evidence to transition to transition to it.
+            // not seen evidence to transition to it.
             let view_number = 1;
             async move {
                 async_sleep(Duration::from_millis(next_view_timeout)).await;
--- a/crates/hotshot/src/tasks/task_state.rs
+++ b/crates/hotshot/src/tasks/task_state.rs
@@ -2,15 +2,26 @@ use std::{
     collections::{BTreeMap, HashMap},
     marker::PhantomData,
     sync::{atomic::AtomicBool, Arc},
+    time::Duration,
 };
 
+use async_compatibility_layer::art::{async_sleep, async_spawn};
 use async_trait::async_trait;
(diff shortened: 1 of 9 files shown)


crates/hotshot/src/tasks/task_state.rs
```diff
@@ -185,6 +196,22 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> CreateTaskState<TYPES, I>
     async fn create_from(handle: &SystemContextHandle<TYPES, I>) -> ConsensusTaskState<TYPES, I> {
         let consensus = handle.hotshot.consensus();
 
+        let event_stream = handle.internal_event_stream.0.clone();
+        let next_view_timeout = handle.hotshot.config.next_view_timeout;
+        let timeout_task = async_spawn({
+            // Nuance: We timeout on the 1 here because that means that we have
+            // not seen evidence to transition to transition to it.
```
jparr721, line 203:
```suggestion
            // not seen evidence to transition to it.
```

crates/hotshot/src/tasks/task_state.rs
```diff
@@ -185,6 +196,22 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> CreateTaskState<TYPES, I>
     async fn create_from(handle: &SystemContextHandle<TYPES, I>) -> ConsensusTaskState<TYPES, I> {
         let consensus = handle.hotshot.consensus();
 
+        let event_stream = handle.internal_event_stream.0.clone();
```
jparr721, line 199: What happens during restart? Looks like we'd always have a timeout set for view 1, but if we're rebooting from a higher view this would be wrong. Should we check the view number and only add the task then?

crates/hotshot/src/tasks/task_state.rs
```diff
@@ -248,6 +275,23 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> CreateTaskState<TYPES, I>
         handle: &SystemContextHandle<TYPES, I>,
     ) -> QuorumProposalTaskState<TYPES, I> {
         let consensus = handle.hotshot.consensus();
+
+        let event_stream = handle.internal_event_stream.0.clone();
+        let next_view_timeout = handle.hotshot.config.next_view_timeout;
+        let timeout_task = async_spawn({
+            // Nuance: We timeout on the 1 here because that means that we have
+            // not seen evidence to transition to transition to it.
```
jparr721, line 283:
```suggestion
            // not seen evidence to transition to it.
```

crates/hotshot/src/tasks/task_state.rs
```diff
@@ -248,6 +275,23 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> CreateTaskState<TYPES, I>
         handle: &SystemContextHandle<TYPES, I>,
     ) -> QuorumProposalTaskState<TYPES, I> {
         let consensus = handle.hotshot.consensus();
+
+        let event_stream = handle.internal_event_stream.0.clone();
```
jparr721, line 279: Same comment as above, looks like this same block occurs a couple times, so I'll stop commenting this. Perhaps it'd be useful as a helper?

crates/hotshot/src/tasks/task_state.rs
```diff
@@ -278,6 +322,23 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> CreateTaskState<TYPES, I>
         handle: &SystemContextHandle<TYPES, I>,
     ) -> QuorumProposalRecvTaskState<TYPES, I> {
         let consensus = handle.hotshot.consensus();
+
+        let event_stream = handle.internal_event_stream.0.clone();
+        let next_view_timeout = handle.hotshot.config.next_view_timeout;
+        let timeout_task = async_spawn({
+            // Nuance: We timeout on the 1 here because that means that we have
+            // not seen evidence to transition to transition to it.
```
jparr721, line 330:
```suggestion
            // not seen evidence to transition to it.
```

crates/hotshot/src/tasks/task_state.rs
```diff
@@ -307,6 +368,22 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> CreateTaskState<TYPES, I>
     for Consensus2TaskState<TYPES, I>
 {
     async fn create_from(handle: &SystemContextHandle<TYPES, I>) -> Consensus2TaskState<TYPES, I> {
+        let event_stream = handle.internal_event_stream.0.clone();
+        let next_view_timeout = handle.hotshot.config.next_view_timeout;
+        let timeout_task = async_spawn({
+            // Nuance: We timeout on the 1 here because that means that we have
+            // not seen evidence to transition to transition to it.
```
jparr721, line 375:
```suggestion
            // not seen evidence to transition to it.
```

crates/task-impls/src/consensus/view_change.rs
```diff
@@ -91,7 +86,15 @@ pub(crate) async fn update_view<TYPES: NodeType, I: NodeImplementation<TYPES>>(
             )
             .await;
         }
-    }));
+    });
+
+    // Cancel the old timeout task
+    cancel_task(std::mem::replace(
```
jparr721, line 92: Does this cancel by swapping the memory and GCing the dangling reference? I'm not sure how each executor handles killing tasks, I assume it's fine though.

jparr721 (changes requested):
Couple things and questions.

crates/hotshot/src/tasks/task_state.rs
rob-maron: Yeah, good idea, let me add it as a helper somewhere

crates/hotshot/src/tasks/task_state.rs
rob-maron: Right, this makes sense, let me think about it

crates/task-impls/src/consensus/view_change.rs
rob-maron: The `std::mem::replace` atomically replaces the first (dest) argument with the second (src) and returns the first (old dest)

Since the first (dest) contained the old timeout task, we can now run `cancel_task` on it

crates/hotshot/src/tasks/task_state.rs
rob-maron: Updated

crates/hotshot/src/tasks/task_state.rs
```diff
@@ -31,6 +37,29 @@ where
...
+            async_sleep(Duration::from_millis(next_view_timeout)).await;
+            broadcast_event(
+                Arc::new(HotShotEvent::Timeout(start_view + 1)),
+                &event_stream,
+            )
+            .await;
+        }
+    })
+}
+
```
ss-es, line 62: Nit: could this be in `impl SystemContextHandle` as e.g. `startup_timeout_task`?

crates/task-impls/src/consensus/view_change.rs
ss-es: > Does this cancel by swapping the memory and GCing the dangling reference? I'm not sure how each executor handles killing tasks, I assume it's fine though.

It's worth noting there's definitely something to worry about in general: dropping a `JoinHandle` means the task will continue running until it shuts itself down, and we can no longer cancel it. This is basically never what we want.

But this looks right to me.

ss-es (commented):
This looks good to me!

crates/task-impls/src/consensus/view_change.rs
ss-es: > Does this cancel by swapping the memory and GCing the dangling reference? I'm not sure how each executor handles killing tasks, I assume it's fine though.

It's worth noting there's definitely something to worry about in general: dropping a `JoinHandle` means the task will continue running until it shuts itself down, and we can no longer cancel it. This is basically never what we want.

But this looks right to me.

crates/hotshot/src/tasks/task_state.rs
rob-maron: Sure, I thought about putting it there but didn't know if it was relevant to that part of the code

ss-es: I don't actually have a strong opinion -- if you think it shouldn't be there, I'm fine with it being a stand-alone function!

ss-es (approved):

jbearer (approved):
Finally 🙏

rob-maron:
I had to update it to wait for the Libp2p network to be ready first, since it was not fast enough for the first view

crates/hotshot/src/tasks/task_state.rs
jparr721: @rob-maron lmk if you decide to do this, I can approve afterwards.

jparr721 (approved):

ss-es (approved):
Looks good to me!

I actually wonder if the test_runner network changes will help with some CI issues I was having in another PR

PR #3207, opened 2024-05-22
[CX_HARDENING] - Develop A Better Framework for Input Permutation

### LINKED ISSUE #3161: [CX_HARDENING] - Develop A Better Framework for Input Permutation

Inputs are mostly written for the happy path. The existing tests are written with a single stream of inputs, and this has bitten us in the past. However, as we see in the proposal_ordering tests, the permutation with index order is not sufficient and requires the implementor to arduously call a significant number of cases with manually prescribed indices. We can do better. We need a scalable and reliable way to permute inputs (and outputs) in a way that ensures that all possible orderings can be handled by the test.

- [ ] - Define a new set of helper primitives to permute inputs of hot shot events. 
- [ ] - Validate that all cases are exhaustively checked.
- [ ] - Evaluate the time duration of the tests and determine if each run should be a unique permutation, or if all permutations should run every time.

_No response_

All tests passing, new permutation system implemented.

_No response_

jparr721:
Closes #3161

### This PR: 
Inputs are mostly written for the happy path. The existing tests are written with a single stream of inputs, and this has bitten us in the past. However, as we see in the proposal_ordering tests, the permutation with index order is not sufficient and requires the implementor to arduously call a significant number of cases with manually prescribed indices. We can do better. We need a scalable and reliable way to permute inputs (and outputs) in a way that ensures that all possible orderings can be handled by the test.

This PR does just that, it adds a randomization layer over the inputs which can be specified by the new `random` and `serial` macros. These macros make it easier to declare a vec with the context that you're intending for it to be used in without any additional thought.

**The new test script here will eventually completely replace the old one in `script.rs`**

### This PR does not: 

This PR does not rewrite every test, that would make the PR large and any changes would be painful. I have updated one test and will update the rest in a follow on PR


crates/testing/src/script.rs
```diff
@@ -16,8 +17,59 @@ pub struct TestScriptStage<TYPES: NodeType, S: TaskState<Event = Arc<HotShotEven
...
+
+    pub fn iter_mut(&mut self) -> impl Iterator<Item = &mut TestScriptStage<TYPES, S>> {
+        self.stages.iter_mut()
+    }
+
+    pub fn randomize(&mut self) {
+        for stage in self.iter_mut() {
+            stage.inputs.shuffle(&mut rand::thread_rng());
+        }
+    }
```
ss-es, line 63: big +1, very clean.

crates/testing/src/script.rs
```diff
@@ -93,11 +145,14 @@ pub async fn run_test_script<
...
     S: TaskState<Event = Arc<HotShotEvent<TYPES>>> + Send + 'static,
 >(
-    mut script: TestScript<TYPES, S>,
+    // mut script: TestScript<TYPES, S>,
+    script: Vec<TestScriptStage<TYPES, S>>,
     state: S,
 ) where
     TYPES: NodeType,
 {
+    let mut script: TestScript<TYPES, S> = script.into();
```
ss-es, line 154: I really think we should not do this here. The actual execution of the script should be deterministic and shouldn't shuffle its inputs, because we lose the ability to insist on an order.

But I have no objection to adding a wrapper function that randomizes the script before call this and using that everywhere (i.e. have `run_test_script_random` and `run_test_script` or `run_test_script` and `run_test_script_deterministic`).

crates/testing/tests/tests_1/consensus_task.rs
```diff
@@ -68,19 +66,18 @@ async fn test_consensus_task() {
...
         ],
-        outputs: vec![
+        vec![
             exact(ViewChange(ViewNumber::new(1))),
             quorum_proposal_validated(),
             exact(QuorumVoteSend(votes[0].clone())),
         ],
-        asserts: vec![],
-    };
+    );
```
ss-es, line 80: This might just be a personal thing, but I find this somewhat less clear read with `from_inputs_and_outputs` compared to the constructor because the `inputs:` and `outputs:` labels are lost

crates/testing/src/script.rs
jparr721: Yeah this right now is just a hack, I plan to make this much less terrible. I just wanted to get it working before I edited every single file.

crates/testing/tests/tests_1/consensus_task.rs
jparr721: Yeah I think you're right as well, I'll roll this change back.

crates/testing/tests/tests_1/da_task.rs
```diff
@@ -80,25 +80,28 @@ async fn test_da_task() {
                     .unwrap(),
                 precompute,
             ),
-        ],
-        outputs: vec![exact(DaProposalSend(proposals[1].clone(), leaders[1]))],
+        ]),
+        outputs: ordered(vec![exact(DaProposalSend(
+            proposals[1].clone(),
+            leaders[1],
+        ))]),
```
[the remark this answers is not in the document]
jparr721: Spoke offline, this PR is on hold until Salman's change to the macro goes in. Then, I'll update this to require less input-level changes and instead operate over an entire stage.

crates/testing/src/script.rs
```diff
@@ -179,6 +179,15 @@ pub struct Expectations<TYPES: NodeType, S> {
     pub task_state_asserts: Vec<Box<dyn Predicate<S>>>,
 }
 
+impl<TYPES: NodeType, S> Expectations<TYPES, S> {
+    pub fn from_outputs(output_asserts: Vec<Box<dyn Predicate<Arc<HotShotEvent<TYPES>>>>>) -> Self {
```
jparr721, line 183: @ss-es I added this because we only assert in two tests in the entire codebase, so it was easier than writing the same line over and over, and I don't think it hurts readability, though I could be swayed on a different name.

ss-es: Looks good to me! Honestly, you could probably even just call it `outputs`.

crates/testing/tests/tests_1/consensus_task.rs
```diff
@@ -101,17 +98,29 @@ async fn test_consensus_task() {
...
         ],
-        outputs: vec![
+    ];
+
+    let expectations = vec![
+        Expectations::from_outputs(vec![all_predicates(vec![
+            exact(ViewChange(ViewNumber::new(1))),
+            quorum_proposal_validated(),
+            exact(QuorumVoteSend(votes[0].clone())),
+        ])]),
```
ss-es, line 108: This definitely looks a bit cumbersome. I wonder: is it easy to just add a macro like e.g. `all![exact(), quorum_proposal_validated(), exact()]` (though if it takes more than 2 minutes I probably wouldn't bother)

ss-es (approved):
Looks good to me if CI passes

crates/testing/tests/tests_1/consensus_task.rs
jparr721: Let me give 'er a try.

jparr721: Done.

crates/macros/src/lib.rs
```diff
@@ -424,3 +443,82 @@ pub fn test_scripts(input: proc_macro::TokenStream) -> TokenStream {
...
+                            shuffled_inputs.push(events);
+                        },
+                        InputOrder::Serial(events) => {
+                            shuffled_inputs.push(events);
+                        }
+                    }
+                }
+
+                test_scripts![shuffled_inputs, #(#scripts),*].await;
+                tracing::info!("Suite used seed {seed}");
```
ss-es, line 518: It occurred to me -- this might not appear in the same conditions that the panic in `test_scripts` does. 

could we e.g. https://doc.rust-lang.org/std/panic/fn.catch_unwind.html the `test_scripts!` panic and then panic ourselves with the seed? this would avoid situations where the test fails but we don't see the seed because the log level was not high enough etc

ss-es (approved):

crates/macros/src/lib.rs
jparr721: Spoke offline: TL;DR, this is hard.

PR #3135, opened 2024-05-09
[Auto Bench] Integrate push-based CDN to scripts of auto benchmarking on AWS

dailinsubjam:
Closes #2827 .
Sorry for such a large PR, part of it are changes on simple builder which covers https://github.com/EspressoSystems/HotShot/pull/3210 @QuentinI Let me know if there's a way to separate it into another PR.

### This PR: 
- Adds script of running benchmarks with push-based CDN
- Adds a README on running benchmarks on AWS and a README on running examples locally with push cdn
- Update and re-organize old scripts with gpu
- Does some changes to benchmark output structure and input parameters
- Covers the simple builder changes of https://github.com/EspressoSystems/HotShot/pull/3210 and https://github.com/EspressoSystems/HotShot/pull/3324
- Adds some benchmark results 

### This PR does not: 
- update the image tag. It's still `ghcr.io/espressosystems/hotshot/validator-webserver:main-async-std` for now. It seems the new tag will only be created when this runs on main. I will update the tag in the next PR after this one is merged and the image is tested successfully.

### Key places to review: 
Scripts in `scripts/benchmark_scripts`.

### How to test this PR: 
`scripts/benchmarks_results/README.md`

PR #3210, opened 2024-05-23
[DRAFT] Builder only on DA nodes

QuentinI:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/crates/examples/combined/all.rs
+++ b/crates/examples/combined/all.rs
@@ -21,6 +21,7 @@ use hotshot_example_types::state_types::TestTypes;
 use hotshot_orchestrator::client::ValidatorArgs;
 use hotshot_types::traits::node_implementation::NodeType;
 use rand::{rngs::StdRng, RngCore, SeedableRng};
+use surf_disco::Url;
 use tracing::{error, instrument};
 
 use crate::{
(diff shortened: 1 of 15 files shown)


crates/examples/combined/all.rs
```diff
@@ -140,12 +141,19 @@ async fn main() {
             8000 + (u16::try_from(i).expect("failed to create advertise address")),
         );
 
+        let builder_address = Url::parse(&format!(
+            "http://localhost:{}",
+            9000 + (u16::try_from(i).expect("failed to create builder address"))
+        ))
+        .unwrap();
+
```
dailinsubjam, line 149: Looks like instead of `localhost`, we need to extract the machines local ip, so that when nodes are running on different machines they could know each other's address.

crates/examples/libp2p/all.rs
```diff
@@ -46,12 +47,20 @@ async fn main() {
             IpAddr::V4(Ipv4Addr::LOCALHOST),
             8000 + (u16::try_from(i).expect("failed to create advertise address")),
         );
+
+        let builder_address = Url::parse(&format!(
+            "http://localhost:{}",
```
dailinsubjam, line 52: same here

crates/examples/infra/mod.rs
```diff
@@ -921,33 +922,7 @@ pub async fn main_entry_point<
...
-                    (),
-                )
-                .await;
-
-            run_config.config.builder_url = builder_url;
-
-            builder_task
-        }
-    };
+    let builder_task = initialize_builder(&mut run_config, &args, &orchestrator_client).await;
```
dailinsubjam, line 925: Is it possible to learn builder address and assign it to `args.builder_address` before this line?

QuentinI:
In main as part of #3135

PR #3219, opened 2024-05-24
Make BlockPayload::from_transactions async and pass ValidatedState

imabdulbasit:
The `BlockPayload::from_transactions` function takes an `Instance` parameter, which contains `ChainConfig` on the sequencer side. To support updates to ChainConfig, we need to add it to a mutable `ValidatedState`. This requires the function to accept an additional `ValidatedState` parameter. When the chainconfig commitment does not match the instance chainconfig, this function on the sequencer side will do catchup from the peers (GET request) to get the updated ChainConfig so it needs to be async.

This PR:
- adds ValidatedState parameter to BlockPayload::from_transactions
- makes the from_transactions fn async

https://github.com/EspressoSystems/espresso-sequencer/issues/1475
https://github.com/EspressoSystems/espresso-sequencer/pull/1492#issuecomment-2127412305

### DIFF
--- a/crates/task-impls/src/consensus/helpers.rs
+++ b/crates/task-impls/src/consensus/helpers.rs
@@ -395,10 +395,12 @@ pub(crate) async fn publish_proposal_from_upgrade_cert<TYPES: NodeType>(
     )
     .await?;
 
+    let validated_state = consensus.read().await.decided_state();
     // Special case: if we have a decided upgrade certificate AND it does not apply a version to the current view, we MUST propose with a null block.
     ensure!(upgrade_cert.in_interim(cur_view), "Cert is not in interim");
-    let (payload, metadata) = <TYPES::BlockPayload as BlockPayload>::from_transactions(
+    let (payload, metadata) = <TYPES::BlockPayload as BlockPayload<TYPES>>::from_transactions(
         Vec::new(),
+        validated_state.as_ref(),
         instance_state.as_ref(),
     )
     .context("Failed to build null block payload and metadata")?;
@@ -407,9 +409,12 @@ pub(crate) async fn publish_proposal_from_upgrade_cert<TYPES: NodeType>(
     let null_block_commitment = null_block::commitment(quorum_membership.total_nodes())
         .context("Failed to calculate null block commitment")?;
 
-    let null_block_fee =
-        null_block::builder_fee::<TYPES>(quorum_membership.total_nodes(), instance_state.as_ref())
-            .context("Failed to calculate null block fee info")?;
+    let null_block_fee = null_block::builder_fee::<TYPES>(
+        quorum_membership.total_nodes(),
+        validated_state.as_ref(),
+        instance_state.as_ref(),
+    )
+    .context("Failed to calculate null block fee info")?;
 
     Ok(async_spawn(async move {
         create_and_send_proposal(
--- a/crates/task-impls/src/consensus/helpers.rs
+++ b/crates/task-impls/src/consensus/helpers.rs
@@ -410,17 +410,10 @@ pub(crate) async fn publish_proposal_from_upgrade_cert<TYPES: NodeType>(
     let null_block_commitment = null_block::commitment(quorum_membership.total_nodes())
         .context("Failed to calculate null block commitment")?;
 
+    let null_block_fee = null_block::builder_fee::<TYPES>(quorum_membership.total_nodes())
+        .context("Failed to calculate null block fee info")?;
+
     Ok(async_spawn(async move {
-        let Some(null_block_fee) = null_block::builder_fee::<TYPES>(
-            quorum_membership.total_nodes(),
-            validated_state.as_ref(),
-            instance_state.as_ref(),
-        )
-        .await
-        else {
-            error!("Failed to calculate null block fee info");
-            return;
-        };
         create_and_send_proposal(
             public_key,
             private_key,
--- a/crates/task-impls/src/consensus/helpers.rs
+++ b/crates/task-impls/src/consensus/helpers.rs
@@ -403,6 +403,7 @@ pub(crate) async fn publish_proposal_from_upgrade_cert<TYPES: NodeType>(
         validated_state.as_ref(),
         instance_state.as_ref(),
     )
+    .await
     .context("Failed to build null block payload and metadata")?;
 
     let builder_commitment = payload.builder_commitment(&metadata);
@@ -414,6 +415,7 @@ pub(crate) async fn publish_proposal_from_upgrade_cert<TYPES: NodeType>(
         validated_state.as_ref(),
         instance_state.as_ref(),
     )
+    .await
     .context("Failed to calculate null block fee info")?;
 
     Ok(async_spawn(async move {
--- a/crates/task-impls/src/consensus/helpers.rs
+++ b/crates/task-impls/src/consensus/helpers.rs
@@ -410,15 +410,17 @@ pub(crate) async fn publish_proposal_from_upgrade_cert<TYPES: NodeType>(
     let null_block_commitment = null_block::commitment(quorum_membership.total_nodes())
         .context("Failed to calculate null block commitment")?;
 
-    let null_block_fee = null_block::builder_fee::<TYPES>(
-        quorum_membership.total_nodes(),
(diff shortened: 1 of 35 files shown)


crates/task-impls/src/transactions.rs
```diff
@@ -165,20 +165,28 @@ impl<
                         .number_of_empty_blocks_proposed
                         .add(1);
 
+                    let validated_state = self.consensus.read().await.decided_state();
+
                     // Calculate the builder fee for the empty block
                     let Some(builder_fee) = null_block::builder_fee(
                         self.membership.total_nodes(),
+                        validated_state.as_ref(),
```
imabdulbasit, line 173: not sure if we need to pass decided state here

crates/types/src/traits/block_contents.rs
```diff
@@ -79,22 +81,28 @@ pub trait BlockPayload:
     ///
     /// # Errors
     /// If the transaction length conversion fails.
+
     fn from_transactions(
         transactions: impl IntoIterator<Item = Self::Transaction>,
+        validated_state: &Self::ValidatedState,
         instance_state: &Self::Instance,
-    ) -> Result<(Self, Self::Metadata), Self::Error>;
+    ) -> impl Future<Output = Result<(Self, Self::Metadata), Self::Error>> + Send + 'static;
```
jparr721, line 89: Can you please add to this doc comment explaining why the async behavior is needed? From the trait, it's not abundantly clear. 

Could we perhaps just upgrade this trait to `async_trait` since we use it elsewhere in the code anyway?

crates/task-impls/src/transactions.rs
jparr721: I suppose there's nothing explicitly wrong with this, what is your reservation?

crates/testing/tests/tests_1/consensus_task.rs
```diff
@@ -498,6 +510,8 @@ async fn test_view_sync_finalize_vote() {
 /// Makes sure that, when a valid ViewSyncFinalize certificate is available, the consensus task
 /// will NOT vote when the certificate matches a different view number.
 async fn test_view_sync_finalize_vote_fail_view_number() {
+    use futures::StreamExt;
```
jparr721, line 513: Let's move this to the top of the file so we don't need to import it over and over.

crates/task-impls/src/consensus2/mod.rs
```diff
@@ -1,10 +1,10 @@
-use async_broadcast::Sender;
-use hotshot_task::task::Task;
 use std::sync::Arc;
-use tracing::instrument;
 
+use async_broadcast::Sender;
```
jparr721, line 3: Seems like these changes might have been unnecessary?

crates/task-impls/src/consensus2/handlers.rs
```diff
@@ -15,14 +15,13 @@ use hotshot_types::{
 };
 use tracing::debug;
 
+use super::Consensus2TaskState;
```
jparr721, line 18: Seems like this change might have been unnecessary?

jparr721 (dismissed):
This seems fine, though the testing changes should be reviewed by at least one other person. Your justification seems good though for adding this. Just check over my requests and let me know if any of them are unclear. Feel free to disregard if something isn't relevant.

crates/task-impls/src/consensus2/mod.rs
imabdulbasit: I ran `cargo fmt `which formatted the files this way

crates/task-impls/src/transactions.rs
jparr721: Spoke offline, this seems fine if it's the only source of the data. The consensus shared state is going to be hopefully phased out in the next sprint, so as long as this doesn't introduce additional lock contention (which I don't think it does), I don't have any qualms about this.

crates/types/src/traits/block_contents.rs
imabdulbasit: done in dec9dd6de58893f9080dde0b96285d05b2f64038

crates/testing/tests/tests_1/consensus_task.rs
imabdulbasit: done in dec9dd6de58893f9080dde0b96285d05b2f64038

crates/task-impls/src/transactions.rs
```diff
@@ -165,20 +165,28 @@ impl<
...
                         self.instance_state.as_ref(),
-                    ) else {
+                    )
```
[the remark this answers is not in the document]
imabdulbasit: done in 2d78d1e460166d04d8af35cd2671757e032af854

crates/task-impls/src/consensus/helpers.rs
```diff
@@ -395,21 +395,28 @@ pub(crate) async fn publish_proposal_from_upgrade_cert<TYPES: NodeType>(
...
-    let null_block_fee =
-        null_block::builder_fee::<TYPES>(quorum_membership.total_nodes(), instance_state.as_ref())
-            .context("Failed to calculate null block fee info")?;
+    let null_block_fee = null_block::builder_fee::<TYPES>(
+        quorum_membership.total_nodes(),
+        validated_state.as_ref(),
+        instance_state.as_ref(),
+    )
+    .await
+    .context("Failed to calculate null block fee info")?;
```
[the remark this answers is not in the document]
imabdulbasit: done in 2d78d1e460166d04d8af35cd2671757e032af854

crates/task-impls/src/transactions.rs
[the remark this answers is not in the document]
imabdulbasit: where do I keep the handle of this task?

jparr721 (commented):
Looks good, but I'm going to let Brendon approve since he had more specific requests.

crates/task-impls/src/transactions.rs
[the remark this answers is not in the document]
jparr721: You could do something like the consensus task does:
```rust
    pub spawned_tasks: BTreeMap<TYPES::Time, Vec<JoinHandle<()>>>,
```

So we can keep a map around and key on the view or whatever relevant key we need, then we can cancel the task from there. The key here cancels old tasks if we propose for a newer view, so in this case we can probably cancel when we get a new view change? We could cancel with something reminiscent of the following: https://github.com/EspressoSystems/HotShot/blob/main/crates/task-impls/src/consensus/mod.rs#L151 but we might be able to just cancel all tasks older than `view`. Thoughts @bfish713?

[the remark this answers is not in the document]
imabdulbasit: I think this shouldn't be canceled on a view change since we should be able to complete catch up even if there is a view change to keep the updated chain config in ValidatedState.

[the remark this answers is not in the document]
imabdulbasit: could be canceled after a certain time maybe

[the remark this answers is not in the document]
ss-es: I definitely think we shouldn't drop the handle, though I don't think `TaskRegistry` is the right place to put this (it will likely be impossible to put it there very soon). It should definitely be stored in `TransactionTaskState`  somehow, probably either in a `Vec<JoinHandle<()>>` or a `BTreeMap` like Jarred suggested.

I guess the question is: what is the expected behavior if consensus is on view (say) 100 and a calculation for view 50 is still going? do we just keep going indefinitely and risk accumulating these spawned subtasks? or is there a point where we can just give up? (I don't know enough to know what the answer is)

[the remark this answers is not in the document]
imabdulbasit: reverted this change as builder_fee function is synchronous

crates/task-impls/src/consensus/helpers.rs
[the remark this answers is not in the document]
imabdulbasit: reverted

bfish713:
Looks good I like that we change to add `empty` to make this synchronous in consensus.

bfish713:
@jparr721 I'm holding off approval because this would be considered a breaking change during the RC process.  Can you accept this we're ok merging it?

PR #3223, opened 2024-05-24
[CX_HARDENING] - Modernize and Audit Vid Task Tests

### LINKED ISSUE #3165: [CX_HARDENING] - Modernize and Audit Vid Task Tests

The Vid Task has not been modernized to use the new testing framework. It also only has a single test.

- [ ] - Modernize and improve the ergonomics of the Vid task test by switching it to use the new testing framework.
- [ ] - Evaluate the Vid task to determine if enough of the code is covered.
- [ ] - Permute the inputs and verify if the tests still pass.
- [ ] - Clean up and reduce boilerplate within the test file.
- [ ] - Add additional tests for the various cases that were not previously handled.

We should probably have a parallel test that exists alongside the "old" test so that way we know that we aren't accidentally losing on previously covered code, but we can remove the old test before merging.

- [ ] - Test is modernized
- [ ] - New tests added if needed
- [ ] - All tests pass.

_No response_

jparr721:
Closes #3165 

### This PR: 
Modernizes the Vid Task to use the newer test suite, removes the old test, and removes `ConsensusApi` from the task since it's unused.

### DIFF
--- a/crates/hotshot/src/tasks/task_state.rs
+++ b/crates/hotshot/src/tasks/task_state.rs
@@ -80,13 +80,10 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> CreateTaskState<TYPES, I>
 
 #[async_trait]
 impl<TYPES: NodeType, I: NodeImplementation<TYPES>> CreateTaskState<TYPES, I>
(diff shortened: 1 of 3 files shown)


rob-maron (approved):
Changes look good and in line with other tests

PR #3230, opened 2024-05-27
Remove Hash from Recently Edited Types in `HotShotEvent`

jparr721:
Closes #<ISSUE_NUMBER>

### This PR: 
We needed to have `Hash` for `HotShotEvent` to be able to use it in `harness.rs`, but now we do not need that, and we can remove the hash trait from the `View` type and downstream types.

Also updates the `all` predicate to not use the `HashSet`.

### This PR does not: 
This does not evaluate all types who include `Hash` as a result of HotShotEvent requiring it and, instead, it only removes it from the types that I know for sure don't need to implement it.

### DIFF
--- a/crates/testing/src/predicates/event.rs
+++ b/crates/testing/src/predicates/event.rs
@@ -57,28 +57,7 @@ pub fn all<TYPES>(events: Vec<HotShotEvent<TYPES>>) -> Box<TestPredicate<Arc<Hot
 where
     TYPES: NodeType,
 {
-    let info = format!("{:?}", events);
-
-    // Remove duplicate events.
-    let mut events = events;
-    events.dedup();
-
-    let function =
-        move |e: &Arc<HotShotEvent<TYPES>>| match events.iter().find(|x| *x == e.as_ref()) {
-            Some(_) => {
-                if events.is_empty() {
-                    PredicateResult::Pass
-                } else {
-                    PredicateResult::Incomplete
-                }
-            }
-            None => PredicateResult::Fail,
-        };
-
-    Box::new(TestPredicate {
-        function: Arc::new(RwLock::new(function)),
-        info,
-    })
+    all_predicates(events.into_iter().map(exact).collect())
 }
 
 pub fn all_predicates<TYPES: NodeType>(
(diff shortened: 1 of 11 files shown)


crates/testing/src/predicates/event.rs
```diff
@@ -58,18 +58,22 @@ where
...
-                PredicateResult::Pass
-            } else {
-                PredicateResult::Incomplete
+
+    // Remove duplicate events.
+    let mut events = events;
+    events.dedup();
+
+    let function =
+        move |e: &Arc<HotShotEvent<TYPES>>| match events.iter().find(|x| *x == e.as_ref()) {
```
ss-es, line 67: If i understand `find` correctly, I don't think this is right: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find

It looks like it clears the iterator until it finds the first match.

crates/testing/src/predicates/event.rs
```diff
@@ -58,18 +58,22 @@ where
...
-    let function = move |e: &Arc<HotShotEvent<TYPES>>| match set.take(e.as_ref()) {
-        Some(_) => {
-            if set.is_empty() {
-                PredicateResult::Pass
-            } else {
-                PredicateResult::Incomplete
+
+    // Remove duplicate events.
+    let mut events = events;
+    events.dedup();
```
ss-es, line 64: I don't think we want to de-duplicate events (this leads to confusing behavior when calling `all(..)`).

ss-es (changes requested):

ss-es (commented):
Actually, thinking about it: could we make `all(events)` just a wrapper for `all_predicates(events.map(exact))`?

crates/testing/src/predicates/event.rs
jparr721: Sounds good. The HashSet had automatically de-duped so I was trying to find the correct one.

crates/testing/src/predicates/event.rs
jparr721: Fixed by just mapping over the events as suggested below.

ss-es (approved):
Looks good to me!

PR #3231, opened 2024-05-28
Don't send `ResponseMessage::NotFound` right away, wait for txns

### LINKED ISSUE #3046: [LIBP2P] - Delay sending `ResponseMessage::NotFound`

Currently a node immediately responds with `ResponseMessage::NotFound` when it doesn't have necessary data to calculate VID. This behaviour might be problematic in situations when the node hasn't yet received DA proposal but it most probably will receive it soon.

Delay sending `ResponseMessage::NotFound`. Wait for a predetermined timeout in hope that the data required to calculate VID arrives and the node can respond with `ResponseMessage::Found`

_No response_

This is hard to test because it's not deterministic. The most consistent way to reproduce the issue is to apply the following patch:
```
diff --git a/crates/hotshot/src/traits/networking/combined_network.rs b/crates/hotshot/src/traits/networking/combined_network.rs
index 3a56909..0ee74e3 100644
--- a/crates/hotshot/src/traits/networking/combined_network.rs
+++ b/crates/hotshot/src/traits/networking/combined_network.rs
@@ -502,7 +502,7 @@ impl<TYPES: NodeType> ConnectedNetwork<Message<TYPES>, TYPES::SignatureKey>
         });
         // View changed, let's start primary again
         self.primary_down.store(false, Ordering::Relaxed);
-        self.primary_fail_counter.store(0, Ordering::Relaxed);
+        // self.primary_fail_counter.store(0, Ordering::Relaxed);
     }
 
     fn is_primary_down(&self) -> bool {
```
and run the `test_combined_network_cdn_crash` test.

_No response_

lukaszrzasik:
Closes #3046

### This PR: 

* Delays sending `ResponseMessage::NotFound` immediately
* Sleeps for a specified time
* Retries to calculate VID in hope transactions have been received during sleeping

### This PR does not: 

This change delays the response to the requester which blocks the requester from moving to the next possible responder.
This might be viewed as undesirable behaviour depending on the scenario. 

### Key places to review: 

`get_or_calc_vid_share` method

All tests should pass.

### DIFF
--- a/crates/task-impls/src/response.rs
+++ b/crates/task-impls/src/response.rs
@@ -124,7 +124,8 @@ impl<TYPES: NodeType> NetworkResponseState<TYPES> {
         view: TYPES::Time,
         key: &TYPES::SignatureKey,
     ) -> Option<Proposal<TYPES, VidDisperseShare<TYPES>>> {
-        let contained = self.consensus
+        let contained = self
+            .consensus
             .read()
             .await
             .vid_shares()
@@ -150,9 +151,22 @@ impl<TYPES: NodeType> NetworkResponseState<TYPES> {
                 )
                 .await?;
             }
-            return self.consensus.read().await.vid_shares().get(&view)?.get(key).cloned();
+            return self
+                .consensus
+                .read()
+                .await
+                .vid_shares()
+                .get(&view)?
+                .get(key)
+                .cloned();
         }
-        self.consensus.read().await.vid_shares().get(&view)?.get(key).cloned()
+        self.consensus
+            .read()
+            .await
+            .vid_shares()
+            .get(&view)?
+            .get(key)
+            .cloned()
     }
 
     /// Handle the request contained in the message. Returns the response we should send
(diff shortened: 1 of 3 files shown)


crates/task-impls/src/response.rs
```diff
@@ -131,9 +135,20 @@ impl<TYPES: NodeType> NetworkResponseState<TYPES> {
             .is_some_and(|m| m.contains_key(key));
         if !contained {
             let mut consensus = RwLockUpgradableReadGuard::upgrade(consensus).await;
-            consensus
+            if consensus
                 .calculate_and_update_vid(view, Arc::clone(&self.quorum), &self.private_key)
-                .await;
```
bfish713, line 136: I think I missed this in a previous pr but I think we probably want to move the calculation of vid shares back outside of consensus.  This forces a write lock to be held to calculate VID which will basically block everything.

I think just the update VID should need the lock

bfish713 (changes requested):

bfish713 (approved):

PR #3232, opened 2024-05-28
Upgrade fixes

ss-es:
No linked issue

### This PR: 
Fixes some miscellaneous issues related to proposing blank blocks during the upgrade process that seem to have appeared between moving the consensus functionality around and integrating the builder in the transactions task.

This led to some simplifications: notably, the explicit null block override when we are between versions is moved from the proposal logic in consensus into the transaction task, though the voting logic remains unchanged.

### Key places to review:

### DIFF
--- a/crates/task-impls/src/consensus/helpers.rs
+++ b/crates/task-impls/src/consensus/helpers.rs
@@ -14,13 +14,11 @@ use async_std::task::JoinHandle;
 use chrono::Utc;
 use committable::Committable;
 use futures::FutureExt;
-#[cfg(not(feature = "dependency-tasks"))]
-use hotshot_types::simple_vote::QuorumData;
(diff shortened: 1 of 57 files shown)


crates/task-impls/src/consensus/helpers.rs
```diff
@@ -372,76 +372,6 @@ pub(crate) async fn parent_leaf_and_state<TYPES: NodeType>(
     Ok((parent_leaf, Arc::clone(state)))
 }
 
-/// Send a proposal for the view `view` from the latest high_qc given an upgrade cert. This is a special
```
bfish713, line 375: Nice I'm glad we can merge these two functions

bfish713 (approved):

PR #3233, opened 2024-05-28
[Dependency Tasks] Fix Test Build, Update State in Vote Task

jparr721:
Closes #<ISSUE_NUMBER>

### This PR: 
The recent change to the Quorum Vote Task did not include the state updates that happen in the legacy consensus module. This adds them so that way consensus may proceed with proper validated state updates.

### DIFF
--- a/crates/task-impls/src/quorum_proposal/mod.rs
+++ b/crates/task-impls/src/quorum_proposal/mod.rs
@@ -166,7 +166,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalTaskState<TYPE
                     }
                     ProposalDependency::ValidatedState => {
                         if let HotShotEvent::ValidatedStateUpdated(view_number, _) = event {
-                            *view_number + 1
+                            *view_number
                         } else {
                             return false;
                         }
@@ -280,13 +280,15 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalTaskState<TYPE
             secondary_deps.push(AndDependency::from_deps(vec![qc_dependency]));
         }
 
+        let mut primary_deps = vec![payload_commitment_dependency, vid_share_dependency];
+
+        if *view_number > 1 {
+            primary_deps.push(validated_state_update_dependency);
+        }
+
         AndDependency::from_deps(vec![OrDependency::from_deps(vec![
             AndDependency::from_deps(vec![
-                OrDependency::from_deps(vec![AndDependency::from_deps(vec![
-                    payload_commitment_dependency,
-                    vid_share_dependency,
-                    validated_state_update_dependency,
-                ])]),
+                OrDependency::from_deps(vec![AndDependency::from_deps(primary_deps)]),
                 OrDependency::from_deps(secondary_deps),
             ]),
         ])])
@@ -505,7 +507,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalTaskState<TYPE
                     .update_validated_state_map(*view_number, view.clone());
 
                 self.create_dependency_task_if_new(
-                    *view_number + 1,
+                    *view_number,
                     event_receiver,
                     event_sender,
                     Arc::clone(&event),
--- a/crates/task-impls/src/quorum_proposal/mod.rs
+++ b/crates/task-impls/src/quorum_proposal/mod.rs
@@ -166,7 +166,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalTaskState<TYPE
                     }
                     ProposalDependency::ValidatedState => {
                         if let HotShotEvent::ValidatedStateUpdated(view_number, _) = event {
-                            *view_number
+                            *view_number + 1
                         } else {
                             return false;
                         }
@@ -507,7 +507,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalTaskState<TYPE
                     .update_validated_state_map(*view_number, view.clone());
 
                 self.create_dependency_task_if_new(
-                    *view_number,
+                    *view_number + 1,
                     event_receiver,
                     event_sender,
                     Arc::clone(&event),
--- a/crates/task-impls/src/quorum_proposal_recv/handlers.rs
+++ b/crates/task-impls/src/quorum_proposal_recv/handlers.rs
@@ -200,18 +200,17 @@ pub(crate) async fn handle_quorum_proposal_recv<TYPES: NodeType, I: NodeImplemen
         parent
     };
 
-    {
-        let mut consensus_write = task_state.consensus.write().await;
-        if let Err(e) = consensus_write.update_high_qc(justify_qc.clone()) {
-            tracing::trace!("{e:?}");
-        }
-
-        broadcast_event(
-            HotShotEvent::UpdateHighQc(justify_qc.clone()).into(),
-            event_sender,
-        )
-        .await;
+    let mut consensus_write = task_state.consensus.write().await;
+    if let Err(e) = consensus_write.update_high_qc(justify_qc.clone()) {
+        tracing::trace!("{e:?}");
     }
+    drop(consensus_write);
+
+    broadcast_event(
+        HotShotEvent::UpdateHighQc(justify_qc.clone()).into(),
+        event_sender,
+    )
+    .await;
 
     let Some((parent_leaf, _parent_state)) = parent else {
         warn!(
--- a/crates/task-impls/src/quorum_proposal_recv/handlers.rs
+++ b/crates/task-impls/src/quorum_proposal_recv/handlers.rs
@@ -57,6 +57,7 @@ async fn validate_proposal_liveness<TYPES: NodeType, I: NodeImplementation<TYPES
         },
     };
 
+    tracing::error!("Updating validated state map");
     consensus_write.update_validated_state_map(view_number, view.clone());
     consensus_write.update_saved_leaves(leaf.clone());
 
@@ -205,6 +206,12 @@ pub(crate) async fn handle_quorum_proposal_recv<TYPES: NodeType, I: NodeImplemen
         if let Err(e) = consensus_write.update_high_qc(justify_qc.clone()) {
             tracing::trace!("{e:?}");
         }
+
+        broadcast_event(
+            HotShotEvent::UpdateHighQc(justify_qc.clone()).into(),
+            event_sender,
+        )
+        .await;
     }
 
     let Some((parent_leaf, _parent_state)) = parent else {
@@ -216,12 +223,6 @@ pub(crate) async fn handle_quorum_proposal_recv<TYPES: NodeType, I: NodeImplemen
             validate_proposal_liveness(proposal, event_sender, &justify_qc, task_state).await,
         );
     };
-
-    broadcast_event(
-        HotShotEvent::UpdateHighQc(justify_qc.clone()).into(),
-        event_sender,
-    )
-    .await;
 
     // Validate the proposal
     validate_proposal_safety_and_liveness(
--- a/crates/task-impls/src/quorum_proposal_recv/handlers.rs
+++ b/crates/task-impls/src/quorum_proposal_recv/handlers.rs
@@ -57,7 +57,6 @@ async fn validate_proposal_liveness<TYPES: NodeType, I: NodeImplementation<TYPES
         },
     };
 
-    tracing::error!("Updating validated state map");
     consensus_write.update_validated_state_map(view_number, view.clone());
     consensus_write.update_saved_leaves(leaf.clone());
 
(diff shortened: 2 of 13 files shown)


crates/task-impls/src/quorum_vote.rs
```diff
@@ -70,12 +74,168 @@ struct VoteDependencyHandle<TYPES: NodeType, I: NodeImplementation<TYPES>> {
     view_number: TYPES::Time,
     /// Event sender.
     sender: Sender<Arc<HotShotEvent<TYPES>>>,
+    /// The current version of HotShot
+    version: Version,
+}
+
+impl<TYPES: NodeType, I: NodeImplementation<TYPES> + 'static> VoteDependencyHandle<TYPES, I> {
+    async fn update_shared_state(
```
jparr721, line 82: cc @bfish713 and @shenkeyao please check this against the old logic. It *looks* right, but I could have goofed.

crates/task-impls/src/quorum_vote.rs
```diff
@@ -70,12 +74,168 @@ struct VoteDependencyHandle<TYPES: NodeType, I: NodeImplementation<TYPES>> {
...
+
+        // Note: There's a potential inconsistency here between `self.view_number` and `leaf.data.view_number()`.
+        let view = View {
+            view_inner: ViewInner::Leaf {
+                leaf: proposed_leaf.commit(),
+                state: Arc::clone(&state),
+                delta: Some(Arc::clone(&delta)),
+            },
+        };
+        consensus_writer.update_validated_state_map(self.view_number, view.clone());
```
bfish713, line 167: As stated in the comment there is a potential inconsistency.  We should use the view number from proposed leaf instead of `self.view_number`

crates/task-impls/src/quorum_vote.rs
```diff
@@ -70,12 +74,168 @@ struct VoteDependencyHandle<TYPES: NodeType, I: NodeImplementation<TYPES>> {
...
+        // Kick back our updated structures for downstream usage.
+        let new_leaves = consensus_writer.saved_leaves().clone();
+        let new_state = consensus_writer.validated_state_map().clone();
+        drop(consensus_writer);
+
+        // Broadcast now that the lock is dropped.
+        broadcast_event(
+            HotShotEvent::ValidatedStateUpdated(self.view_number, view).into(),
+            &self.sender,
+        )
```
bfish713, line 178: same here for view number

crates/task-impls/src/quorum_vote.rs
jparr721: That makes sense to me, it doesn't do that in the existing logic, so I wanted to check, sorry I meant to highlight that in the PR. Changing!

crates/task-impls/src/quorum_proposal/mod.rs
```diff
@@ -166,7 +166,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalTaskState<TYPE
                     }
                     ProposalDependency::ValidatedState => {
                         if let HotShotEvent::ValidatedStateUpdated(view_number, _) = event {
-                            *view_number + 1
+                            *view_number
```
bfish713, line 169: This seems off to me, the proposal is dependant on the state for the current view?  Shouldn't it be for the last view (i.e the parent)

crates/task-impls/src/quorum_vote.rs
```diff
@@ -70,12 +74,170 @@ struct VoteDependencyHandle<TYPES: NodeType, I: NodeImplementation<TYPES>> {
...
+                vid_share.data.common.clone(),
+                self.version,
+            )
+            .await
+            .context("Block header doesn't extend the proposal!")?;
+
+        let state = Arc::new(validated_state);
+        let delta = Arc::new(state_delta);
+
+        // TODO Upgrades
```
bfish713, line 123: This block isn't for upgrades, this is validating the DA cert, we're still doing that elsewhere right (in the as a dependency maybe?

crates/task-impls/src/quorum_vote.rs
```diff
@@ -143,53 +306,38 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES> + 'static> HandleDepOutput
         )
         .await;
 
-        if !self.quorum_membership.has_stake(&self.public_key) {
-            debug!(
-                "We were not chosen for quorum committee on {:?}",
+        let Some(vid_share) = vid_share else {
```
bfish713, line 308: Same deal here, how can we reach this code with vid_share being `None`

crates/task-impls/src/quorum_vote.rs
```diff
@@ -143,53 +306,38 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES> + 'static> HandleDepOutput
...
-            &self.private_key,
-        ) {
-            GeneralConsensusMessage::<TYPES>::Vote(vote)
-        } else {
-            error!("Unable to sign quorum vote!");
+
+        let Some(cur_proposal) = cur_proposal else {
+            error!("We don't have the Quorum Proposal for this view {:?}, but we should, ecause the vote dependencies have completed.", self.view_number);
             return;
         };
```
bfish713, line 327: Why is cur_proposal even optional here?  We have the deps to vote, that must include the proposal we want to vote on.  Maybe not an issue for this PR but that seems off.

crates/task-impls/src/quorum_vote.rs
```diff
@@ -70,12 +74,170 @@ struct VoteDependencyHandle<TYPES: NodeType, I: NodeImplementation<TYPES>> {
...
+            &self.sender,
+        )
+        .await;
+
+        // Send the new state up to the sequencer.
+        self.storage
+            .write()
+            .await
+            .update_undecided_state(new_leaves, new_state)
+            .await?;
```
bfish713, line 186: I think we might want to do this before broadcasting the event so we only act on shared state updates if it's persisted, but I don't actually think it's that important

crates/task-impls/src/quorum_vote.rs
```diff
@@ -70,12 +74,170 @@ struct VoteDependencyHandle<TYPES: NodeType, I: NodeImplementation<TYPES>> {
...
+    /// The current version of HotShot
+    version: Version,
+}
+
+impl<TYPES: NodeType, I: NodeImplementation<TYPES> + 'static> VoteDependencyHandle<TYPES, I> {
+    /// Updates the shared consensus state with the new voting data.
+    async fn update_shared_state(
+        &self,
+        proposal: &QuorumProposal<TYPES>,
+        proposed_leaf: &Leaf<TYPES>,
```
bfish713, line 86: This seem to basically be duplicates of the same data, can we get away with just passing the leaf?

crates/task-impls/src/quorum_vote.rs
jparr721: We can't get here for this and below the idea is just that we needed some way to get the variables out of the big match statement one at a time. This is an admittedly less glamorous way to do it, but we should be good.

crates/task-impls/src/quorum_vote.rs
jparr721: See above.

crates/task-impls/src/quorum_proposal/mod.rs
shenkeyao: Is this due to the `view_number` parameter in this event has the same name as the argument of this function? I think the `view_number` in the event should be one view older than the `view_number` of the function argument, and here we should set the view number to be the latter.

crates/task-impls/src/quorum_proposal/mod.rs
```diff
@@ -505,7 +507,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalTaskState<TYPE
                     .update_validated_state_map(*view_number, view.clone());
 
                 self.create_dependency_task_if_new(
-                    *view_number + 1,
+                    *view_number,
```
shenkeyao, line 510: Similar to the above comment, shouldn't this also have a `+1`?

shenkeyao (changes requested):

crates/task-impls/src/quorum_vote.rs
```diff
@@ -70,12 +74,169 @@ struct VoteDependencyHandle<TYPES: NodeType, I: NodeImplementation<TYPES>> {
...
+            QuorumData {
+                leaf_commit: leaf.commit(),
+            },
+            self.view_number,
+            &self.public_key,
+            &self.private_key,
+        )
+        .context("Failed to sign vote")?;
+        let message = GeneralConsensusMessage::<TYPES>::Vote(vote);
+        if let GeneralConsensusMessage::Vote(vote) = message {
```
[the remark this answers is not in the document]
jparr721: It's rust's type system being a pain. First we create the signed vote, then we make the vote message, then we basically have to say "hey, I *promise* this is a vote" because we're using an enum.

crates/task-impls/src/quorum_proposal_recv/handlers.rs
```diff
@@ -205,6 +205,12 @@ pub(crate) async fn handle_quorum_proposal_recv<TYPES: NodeType, I: NodeImplemen
         if let Err(e) = consensus_write.update_high_qc(justify_qc.clone()) {
             tracing::trace!("{e:?}");
         }
+
+        broadcast_event(
+            HotShotEvent::UpdateHighQc(justify_qc.clone()).into(),
+            event_sender,
+        )
+        .await;
```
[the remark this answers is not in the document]
jparr721: Ah, meant to do that! Good catch, thanks.

crates/task-impls/src/quorum_vote.rs
jparr721: We get into a chicken and egg situation somewhat since this event is fallible. So our state update goes fine, then this could kill our code path while leaving the state updated on the backend.

crates/task-impls/src/quorum_vote.rs
jparr721: Yep

crates/task-impls/src/quorum_vote.rs
bfish713: I think that situation is ok.  Basically it would mean we won't vote, but we would still have the proposal in our storage if we restart so we could vote on a later proposal.  It's not at all critical though so we can leave as is

crates/task-impls/src/quorum_vote.rs
bfish713: I see, I think it's fine then.

bfish713 (commented):
Still a few comments that I think are not address a couple about view numbers and the commented about DAC validation and Lucas's comment so I'm holding off on approval but I don't see any major issues with state updating

crates/task-impls/src/quorum_proposal/mod.rs
```diff
@@ -280,13 +280,15 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalTaskState<TYPE
...
+        }
+
         AndDependency::from_deps(vec![OrDependency::from_deps(vec![
             AndDependency::from_deps(vec![
-                OrDependency::from_deps(vec![AndDependency::from_deps(vec![
-                    payload_commitment_dependency,
-                    vid_share_dependency,
-                    validated_state_update_dependency,
-                ])]),
+                OrDependency::from_deps(vec![AndDependency::from_deps(primary_deps)]),
```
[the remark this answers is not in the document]
jparr721: Yeah, it's a product of the type system

crates/task-impls/src/quorum_vote.rs
jparr721: Ah totally missed this.

crates/task-impls/src/quorum_proposal/mod.rs
jparr721: Fixed

crates/task-impls/src/quorum_vote.rs
jparr721: Yeah Keyao already took care of this, removed.

crates/task-impls/src/quorum_vote.rs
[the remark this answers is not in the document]
jparr721: No you're right good catch! 

```diff
            &self.private_key,
         )
         .context("Failed to sign vote")?;
-        let message = GeneralConsensusMessage::<TYPES>::Vote(vote);
-        if let GeneralConsensusMessage::Vote(vote) = message {
-            debug!(
-                "Sending vote to next quorum leader {:?}",
-                vote.view_number() + 1
-            );
-            // Add to the storage.
-            self.storage
-                .write()
-                .await
-                .append_vid(&vid_share)
-                .await
-                .context("Failed to store VID share")?;
-            broadcast_event(Arc::new(HotShotEvent::QuorumVoteSend(vote)), &self.sender).await;
-        }
+        debug!(
+            "Sending vote to next quorum leader {:?}",
+            vote.view_number() + 1
+        );
+        // Add to the storage.
+        self.storage
+            .write()
+            .await
+            .append_vid(&vid_share)
+            .await
+            .context("Failed to store VID share")?;
+        broadcast_event(Arc::new(HotShotEvent::QuorumVoteSend(vote)), &self.sender).await;
 
         Ok(())
     }
```

bfish713 (approved):

PR #3235, opened 2024-05-29
[METRICS] - Individual node health

### LINKED ISSUE #3222: [METRICS] - Individual node health

Most of our existing metrics don't tell you much about the health of an individual node. Things like block height or number of timeouts only tell you that there is a problem somewhere in the network, not necessarily which node.

Add these metrics:
* `timeouts_as_leader: Counter`
* `view_time_as_leader: Histogram`
And anything else you think would be useful

_No response_

The above two metrics exist

_No response_

shenkeyao:
Closes #3222.

### This PR: 
* Adds `view_duration_as_leader` and `number_of_timeouts_as_leader` metrics.
* Adds `cur_view_time` field to the task states, which is needed for the `view_duration_as_leader` metric.
* Updates the `update_view` function to take additional arguments to update `view_duration_as_leader` when being a leader.
* Removes unused `NextLeaderViewTimeout` event.

### This PR does not: 
* Modify existing metrics.

### Key places to review: 
* Whether the metrics are updated correctly in:
  * `crates/task-impls/src/consensus/mod.rs`
  * `crates/task-impls/src/consensus/view_change.rs`
  * `crates/task-impls/src/consensus2/handlers.rs`

### DIFF
--- a/crates/hotshot/Cargo.toml
+++ b/crates/hotshot/Cargo.toml
@@ -31,6 +31,7 @@ async-lock = { workspace = true }
 async-trait = { workspace = true }
 bimap = "0.6"
 bincode = { workspace = true }
+chrono = "0.4"
 clap = { workspace = true, optional = true }
 committable = { workspace = true }
 custom_debug = { workspace = true }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3107,6 +3107,7 @@ dependencies = [
  "cdn-broker",
  "cdn-client",
  "cdn-marshal",
+ "chrono",
  "clap",
  "committable",
  "custom_debug",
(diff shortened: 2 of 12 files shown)


crates/hotshot/Cargo.toml
```diff
@@ -31,6 +31,7 @@ async-lock = { workspace = true }
 async-trait = { workspace = true }
 bimap = "0.6"
 bincode = { workspace = true }
+chrono = "0.4"
```
jparr721, line 34: Chrono is imported in `examples` and `task-impls`. To avoid version conflicts, could we perhaps just make the version standard across the workspace? Looks like they're all using 0.4.

jparr721 (commented):
Just one thing otherwise lgtm

dailinsubjam (approved):
LGTM!

bfish713 (approved):

PR #3256, opened 2024-05-31
Minor test fixes

ss-es:
No linked issue.

### This PR: 
Several tests had `num_failed_views` set far too generously, which allowed them to pass when they should not have. This PR fixes that, and causes integration tests to fail we do not fail exactly `num_failed_views` views.

### DIFF
--- a/crates/testing/tests/tests_4/test_with_failures_f.rs
+++ b/crates/testing/tests/tests_4/test_with_failures_f.rs
@@ -15,7 +15,6 @@ cross_tests!(
     Types: [TestTypes],
     Ignore: false,
     Metadata: {
-
         let mut metadata = TestDescription::default_more_nodes();
         metadata.overall_safety_properties.num_failed_views = 6;
         // Make sure we keep committing rounds after the bad leaders, but not the full 50 because of the numerous timeouts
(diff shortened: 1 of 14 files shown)


jparr721 (approved):

crates/testing/tests/tests_4/test_with_failures_f.rs
```diff
@@ -15,9 +15,8 @@ cross_tests!(
     Types: [TestTypes],
     Ignore: false,
     Metadata: {
-
         let mut metadata = TestDescription::default_more_nodes();
-        metadata.overall_safety_properties.num_failed_views = 6;
+        metadata.overall_safety_properties.num_failed_views = 5;
```
bfish713, line 19: hmm this one is a bit weird.  It should be 6 because there are 6 dead nodes.  I'm guessing we don't count one of the view sync views as a failure or something?

bfish713 (approved):
Looks good! Comment is more of a follow up than an issue with this change

crates/testing/tests/tests_4/test_with_failures_f.rs
ss-es: yeah, I don't understand this myself. 

we only get 5 failed views, and it seemed to consistently be the same 5 on all networks: `Results: [NotEnoughFailures { expected: 6, failed_views: {ViewNumber(14), ViewNumber(18), ViewNumber(15), ViewNumber(20), ViewNumber(17)}]`

Views 16 and 19 don't count as failed, for whatever reason -- actually, this was exactly the motivation for me wanting to change the way we counted `failed_views` to use `QC`s instead of/in addition to decides.

PR #3258, opened 2024-05-31
New fallback switching algorithm, first draft

### LINKED ISSUE #3027: [LIBP2P] - Improve network switching algorithm in `CombinedNetworks`

In normal situation `CombinedNetworks` delays sending some messages on the secondary network. If the primary network fails for more than `COMBINED_NETWORK_MIN_PRIMARY_FAILURES`, the primary network is considered down and sending on the secondary is no longer delayed.

The primary network's state is reset and during view update. View updates happen too frequently to fully reset the state.
Instead: keep the fail counter in the inclusive range of 0-6, increase it when the primary fails to deliver, decrease it when the primary successfully delivers.

Delivery failure is when the secondary needs to send after the delay.
Delivery success is when the secondary is cancelled before sending (in `CombinedNetworks::update_view`).

_No response_

_No response_

All the existing tests should pass.

_No response_

lukaszrzasik:
Closes #3027

### This PR: 

* Removes the mechanism that reset the primary network status on every view update.
* Removes the mechanism that cancelled (aborted) delayed tasks for old views during view update.
* Delayed tasks receive a signal via broadcast channel if they should abandon sending  on the secondary network.
* Delayed tasks check for the signal after waking up and increment or decrement `primary_fail_counter` accordingly
* Every time the message is sent on the secondary without delay due to the primary being down, the `no_delay_counter` is incremented. When it reaches preconfigured value (`COMBINED_NETWORK_PRIMARY_CHECK_INTERVAL`) the messages are delayed on the secondary to check whether views progress regardless. If that is true it means the primary is working fine. If the views don't progress, the secondary is quickly reverted to no-delay mode.

### Key places to review: 

`CombinedNetworks::send_both_networks` and `CombinedNetworks::update_view`

### How to test this PR:

All existing tests should pass.

### DIFF
--- a/crates/hotshot/src/traits/networking/combined_network.rs
+++ b/crates/hotshot/src/traits/networking/combined_network.rs
@@ -153,7 +153,7 @@ impl<TYPES: NodeType> CombinedNetworks<TYPES> {
             > COMBINED_NETWORK_MIN_PRIMARY_FAILURES
         {
             // If the primary failed more than `COMBINED_NETWORK_MIN_PRIMARY_FAILURES` times,
-            // we want to delay this message, and from now on we consider the primary as down
+            // we don't want to delay this message, and from now on we consider the primary as down
             warn!(
                 "Primary failed more than {} times and is considered down now",
                 COMBINED_NETWORK_MIN_PRIMARY_FAILURES


crates/hotshot/src/traits/networking/combined_network.rs
```diff
@@ -165,20 +168,63 @@ impl<TYPES: NodeType> CombinedNetworks<TYPES> {
...
-                    info!("Sending on secondary after delay, message possibly has not reached recipient on primary");
-                    primary_fail_counter.fetch_add(1, Ordering::Relaxed);
-                    secondary_future.await
-                }));
+                .or_insert_with(|| {
+                    let (s, r) = broadcast(1);
+                    (s, r.deactivate())
+                })
+                .1
+                .activate_cloned();
```
bfish713, line 183: why do we deactivate the activate the receiver right away?

bfish713: I think the map can just hold the sender side of the channel and the task can hold the receiver

crates/hotshot/src/traits/networking/combined_network.rs
```diff
@@ -165,20 +168,63 @@ impl<TYPES: NodeType> CombinedNetworks<TYPES> {
...
+            async_spawn(async move {
+                async_sleep(duration).await;
+                if receiver.try_recv().is_ok() {
+                    debug!(
+                        "Not sending on secondary after delay, task was canceled in view update"
+                    );
+                    match primary_fail_counter.load(Ordering::Relaxed) {
```
bfish713, line 190: I think I'm finding this logic a little confusing, Maybe it's the nested ifs.  My reading of it is when the primary is up we keep track of failures based on if the view changes in time or not.  Then the next time we try to send we check this number and set the primary as down if it's too high.  Then we proceed to send to the secondary with no delay and after some number of messages we try the primary again.  If it succeeds it will be 1 below the threshold to set the primary down, but when do we set `primary_failed` back to false?  Seems like it's only after trying with a delay until the counter is reduced to 0.  But I don't see how that can happen because with `primary_failed` true we don't hit the code that reduces `primary_fail_counter`.

I'm probably missing something but I think we should try to come up with a way to make this logic more straightforward to read

bfish713: Ok I think I see `primary_down` and `primary_failed` are the same thing.  I think I convinced myself the logic is correct but it's still a bit confusing to me

crates/hotshot/src/traits/networking/combined_network.rs
bfish713: The reason for that in `SystemContextHandle` is that we have to give an active clone to the sequencer but we don't know when the sequencer will ask for it.  Here it seems like the receiver will only have on listener (the spawned task) and if the receiver is closed then the channel should be closed

crates/hotshot/src/traits/networking/combined_network.rs
bfish713: I think just a few comments for now at each if block is completely fine.  I am convinced the logic is correct based on your testing and reading the code

bfish713 (approved):

Issue #3237, opened 2024-05-29
[CATCHUP] - Request Response Messages for Old Proposals


bfish713:
### What is this task and why do we need to work on it?

We need some way to reqeust old proposals.  To do this we need to add message types for it and add the functionality to the request response task

### What work will need to be done to complete this task?

- Create request for proposal struct
- Create response
- Add event to trigger request
- Add logic to handle request and send response

### What are the acceptance criteria to close this issue?

Types and scaffolding in the request and response task to integrate this into consensus

PR #3267, opened 2024-06-03
[CATCHUP] Add Types for Proposal Fetching

bfish713:
Closes: https://github.com/EspressoSystems/HotShot/issues/3237

### This PR: 

Adds the types to request a proposal we are missing as well as the logic in the request task to send the request out.
We can't yet handle the response because we aren't storing the proposals locally yet.  

### This PR does not: 

Doesn't actually trigger the catch-up

### Key places to review: 

The request task

### DIFF
--- a/crates/task-impls/src/request.rs
+++ b/crates/task-impls/src/request.rs
@@ -115,6 +115,15 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, Ver: StaticVersionType + 'st
                 }
                 Ok(())
             }
+            HotShotEvent::QuorumProposalMissing(view) => {
+                self.run_delay(
+                    RequestKind::Proposal(*view),
+                    sender.clone(),
+                    *view,
+                    Ver::instance(),
+                );
+                Ok(())
+            }
             _ => Ok(()),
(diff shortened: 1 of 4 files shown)


crates/task-impls/src/request.rs
```diff
@@ -181,6 +190,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, Ver: StaticVersionType + 'st
             .whole_committee(view)
             .into_iter()
             .collect();
+        let leader = self.da_membership.leader(view);
```
shenkeyao, line 193: Is this always supposed to be the DA leader? Looks like when handling the `QuorumProposalMissing` event, we also call this function, so I'm wondering in that case whether we should use the quorum leader.

shenkeyao (commented):
I have a question about the leader. Everything else looks good.

crates/task-impls/src/request.rs
bfish713: it really should be the quorum leader, but for now it's always the same I think we should actually combine the memberships in the long term since right now they are references to the same thing

PR #3275, opened 2024-06-04
[DEPENDENCY_REFACTOR] - Remove proposal validation from the quorum proposal task

### LINKED ISSUE #3266: [DEPENDENCY_REFACTOR] - Remove proposal validation from the quorum proposal task

Proposals are validated in the quorum vote task and it's unnecessary to validate them again in the proposal task.

_No response_

_No response_

Proposal validation doesn't exist in the quorum proposal task.

_No response_

shenkeyao:
Closes #3266.

### This PR: 
* Replaces the `QuorumProposalValidated` dependency with `QuorumProposalRecv` in the quorum proposal task.
* Moves the `QuorumProposalValidated` handling from the quorum proposal task to the vote task.
* Simplifies the return type of `handle_quorum_proposal_recv` in the quorum proposal recv task.
* Removes the `QuorumProposalLivenessValidated` event.
* Emits `ValidatedStateUpdated` in the quorum proposal recv task for happy path.
* Updates test outputs.
* Adds a command to `justfile`.

### This PR does not: 
* Fix the existing `test_success` failure.
* Fix lints.

### Key places to review: 
* `crates/task-impls/src/quorum_proposal`
* `crates/task-impls/src/quorum_proposal_recv`


crates/task-impls/src/quorum_proposal_recv/handlers.rs
```diff
@@ -72,62 +69,38 @@ async fn validate_proposal_liveness<TYPES: NodeType, I: NodeImplementation<TYPES
...
+/// * `Ok(true)` if the validation succeeds.
+/// * `Ok(false)` if the validation isn't completed due to missing information.
+/// * `Err(..)` if the validation fails.
 pub(crate) async fn handle_quorum_proposal_recv<TYPES: NodeType, I: NodeImplementation<TYPES>>(
     proposal: &Proposal<TYPES, QuorumProposal<TYPES>>,
     sender: &TYPES::SignatureKey,
     event_sender: &Sender<Arc<HotShotEvent<TYPES>>>,
     task_state: &mut QuorumProposalRecvTaskState<TYPES, I>,
-) -> Result<Option<QuorumProposal<TYPES>>> {
+) -> Result<bool> {
```
jparr721, line 103: I think I preferred the other mechanism of returning the proposal. Using a boolean type doesn't convey meaning to the reader in the match statement in `mod.rs` and now I need to go to the source of the code, whereas returning the proposal conveys meaning and I immediately know that if I receive the proposal, I have to use it.

If others disagree, I'm willing to yield on this point, but I think we should at minimum define a descriptive constant such as `LIVENESS_CHECK_SUCCEEDED` or something to that effect.

crates/task-impls/src/quorum_proposal_recv/handlers.rs
```diff
@@ -72,62 +69,38 @@ async fn validate_proposal_liveness<TYPES: NodeType, I: NodeImplementation<TYPES
         warn!("Couldn't store undecided state.  Error: {:?}", e);
     }
 
-    let liveness_check = justify_qc.view_number() > consensus_write.locked_view();
```
jparr721, line 75: How come we're removing this logic? I don't see it replaced anywhere.

crates/task-impls/src/quorum_proposal/mod.rs
```diff
@@ -265,23 +242,15 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalTaskState<TYPE
...
+            // 1. A QcFormed event
+            AndDependency::from_deps(vec![qc_dependency]),
             // 2. A timeout cert was received
             AndDependency::from_deps(vec![timeout_dependency]),
             // 3. A view sync cert was received.
             AndDependency::from_deps(vec![view_sync_dependency]),
         ];
 
-        // 1. A QcFormed event and QuorumProposalValidated event
-        if *view_number > 1 {
```
jparr721, line 276: While I know it's not explicitly required to have received the validated quorum proposal, we do a lot of state updates that *are* used downstream. If we receive a flow of events where the quorum proposal recv was the last one, we'd have already initiated a proposal, potentially proposing before we've have visited the leaf chain, this would break `parent_leaf_an_state`.

jparr721 (changes requested):
I think we need to have passing tests before we can add this change. I am nervous about the ramifications of removing the `QuorumProposalValidated` code path as a blocker to proposing ourselves as it removes the gate against state updates. I think as long as we're using the Consensus shared state, we need to be judicious about disruptions to that flow. Right now, we don't have a reliable network test to validate that this actually works once integrated, so we could be introducing a whole new set of bugs to have to sift through.

crates/task-impls/src/quorum_proposal/mod.rs
```diff
@@ -265,23 +242,15 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalTaskState<TYPE
         };
 
         // We have three cases to consider:
-        let mut secondary_deps = vec![
+        let secondary_deps = vec![
+            // 1. A QcFormed event
+            AndDependency::from_deps(vec![qc_dependency]),
```
dailinsubjam, line 247: Is this line the replace on `ProposalDependency::Proposal`?

crates/task-impls/src/quorum_proposal_recv/handlers.rs
dailinsubjam: I don't have a strong opinion on this. But if we use `Result<bool>`, might need more comment at the place where this function is called (e.g. 'crates/task-impls/src/quorum_proposal_recv/mod.rs').

jparr721: I think that circumstance is precisely what I'd like to avoid. People, by choice or accident, may not comment a block of code. By making the code itself descriptive of what it's doing, it's a lot more clear what is going on.

I think this situation might highlight the necessity for the liveness check on the proposal to perhaps be its own function, handled outside of the quorum proposal recv handler entirely. It can be at the top-level by the match statement, that way we can simply return `Result<()>`, where not being able to find the parent is an error. Anyhow, unfortunately, makes specializing the error not as cut and dry as the baked-in error type, but I think I'd greatly prefer that since, ordinarily, we'd be logging an error if such were to happen (and the liveness check fails).

shenkeyao: Thanks for @jparr721 and @dailinsubjam's comments! Thinking about this more, I'm leaning towards replacing `Result<bool>` with an enum of three variants (similar to the `LIVENESS_CHECK_SUCCEEDED` constant suggested above) so that:
* It's more explicit to the caller that the three variants need to be handled.
* We don't have to duplicate the proposal in the return type, but it's still clear whether the check succeeds.

crates/task-impls/src/quorum_proposal_recv/handlers.rs
shenkeyao: I wanted to move the `QuorumProposalLivenessValidated` emission, but should have kept this liveness check. Restored!

crates/task-impls/src/quorum_proposal/mod.rs
shenkeyao: As discussed on Zulip, I replaced the `QuorumProposalValidated` event handling with `QuorumProposalRecv`.

crates/task-impls/src/quorum_proposal/mod.rs
shenkeyao: Yeah but in the latest commit, I replaced the `QuorumProposalValidated` handling with `QuorumProposalRecv` and kept the `Proposal` dependency.

crates/task-impls/src/quorum_proposal_recv/handlers.rs
```diff
@@ -90,55 +98,41 @@ async fn validate_proposal_liveness<TYPES: NodeType, I: NodeImplementation<TYPES
...
-            )
-            .await;
-        }
-
-        return Some(proposal.data.clone());
+        return QuorumProposalValidity::LivenessValid;
     }
 
-    None
+    QuorumProposalValidity::Failed
```
shenkeyao, line 104: @bfish713 @jparr721 This is the change I mentioned in the pair programming that I'd like a second pair of eyes.

With the previous logic, this should be `QuorumProposalValidity::FullyValid`, but as discussed, I think it makes more sense to mark this as `Failed`, though it probably doesn't matter logically.

shenkeyao:
I'll send another commit to fix the lint errors, as it's turned on recently.

shenkeyao:
> I'll send another commit to fix the lint errors, as it's turned on recently.

Done.

jparr721 (changes requested):
Couple more things.

crates/task-impls/src/quorum_proposal_recv/handlers.rs
shenkeyao: In the happy path where we have the data in the internal state but the liveness fails, we return `Failed`, so here I think we should be consistent.

I think the ultimate question is whether we should call `cancel_tasks` in `handle` in the quorum proposal recv event, or only log the error. If it's the former, we should update both the happy and sad cases to return `FullyValid` if liveness fails (but the happy path should still return `Failed` if other validations fail). If it's the latter, we should return `Failed` in both cases.

crates/task-impls/src/quorum_proposal_recv/handlers.rs
```diff
@@ -90,55 +98,41 @@ async fn validate_proposal_liveness<TYPES: NodeType, I: NodeImplementation<TYPES
     .await;
 
     if liveness_check {
-        let new_view = proposal.data.view_number + 1;
```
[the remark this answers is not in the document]
jparr721: Sure, I guess my question though is: where do we perform this check? It seems like it should still be occurring, right? If not, then when the parent state is not found, we never attempt to propose if the liveness check passes.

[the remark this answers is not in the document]
jparr721: Spoke offline, we propose anyway because the validity path is not needed to be able to submit a proposal anyway, so we don't need the explicit livenesscheck->proposal flow. Thanks for clarifying!

crates/task-impls/src/consensus/helpers.rs
```diff
@@ -66,14 +81,37 @@ pub async fn validate_proposal_safety_and_liveness<TYPES: NodeType>(
...
     );
 
+    let state = Arc::new(
+        <TYPES::ValidatedState as ValidatedState<TYPES>>::from_header(&proposal.data.block_header),
+    );
+    let view = View {
+        view_inner: ViewInner::Leaf {
+            leaf: proposed_leaf.commit(),
+            state,
+            delta: None, // May be updated to `Some` in the vote task.
```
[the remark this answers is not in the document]
shenkeyao: `delta` is an application-specific state delta, and is only returned by `validate_and_apply_block_header` in the vote task, so I think it's fine to set it as `None` first, then update it to `Some` in the vote task, without worrying about the race condition.

crates/task-impls/src/consensus/helpers.rs
```diff
@@ -66,14 +81,37 @@ pub async fn validate_proposal_safety_and_liveness<TYPES: NodeType>(
...
+        },
+    };
+
+    consensus
+        .write()
+        .await
+        .update_validated_state_map(view_number, view.clone());
+
+    // Broadcast that we've updated our consensus state so that other tasks know it's safe to grab.
+    broadcast_event(
```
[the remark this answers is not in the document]
shenkeyao: We use `consensus.write()` directly here before broadcasting so I'm not sure if it's necessary to drop it. If we have `consensus = consensus.write().await` then I think we should drop the write lock.

crates/testing/tests/tests_1/quorum_vote_task.rs
```diff
@@ -195,8 +203,10 @@ async fn test_quorum_vote_task_miss_dependency() {
     ];
 
     let expectations = vec![
-        Expectations::from_outputs(vec![exact(VidShareValidated(vids[1].0[0].clone()))]),
-        Expectations::from_outputs(vec![exact(DaCertificateValidated(dacs[2].clone()))]),
+        Expectations::from_outputs(all_predicates![exact(LockedViewUpdated(ViewNumber::new(1))), exact(VidShareValidated(vids[1].0[0].clone()))]),
```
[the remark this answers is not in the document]
shenkeyao: Formatting with macros really doesn't work well. 😵‍💫

crates/task-impls/src/consensus/helpers.rs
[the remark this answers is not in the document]
shenkeyao: On second thought, there may be a potential race condition if the state update in the vote tasks is later than the update in the current consensus or new proposal recv task.

I made an issue to improve the `update_validated_state_map` API: https://github.com/EspressoSystems/HotShot/issues/3306. Will do this in a separate PR.

crates/task-impls/src/consensus/helpers.rs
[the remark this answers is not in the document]
jparr721: oof, that's a blunder on my part, sorry about that!

crates/task-impls/src/quorum_vote/handlers.rs
```diff
@@ -287,7 +287,7 @@ pub(crate) async fn handle_quorum_proposal_validated<
             .last_decided_view
             .set(usize::try_from(consensus_writer.last_decided_view().u64()).unwrap());
         let cur_number_of_views_per_decide_event =
-            *task_state.latest_proposed_view - consensus_writer.last_decided_view().u64();
+            *proposal.view_number() - consensus_writer.last_decided_view().u64();
```
jparr721, line 290: I'm not sure this is a 1:1 change. The original code was comparing the most recently submitted proposal, but this one is pointing to the latest proposal that we've received, which can be a much different view number (i.e. in the case of timeouts or view sync delays). I don't know if it's wrong necessarily, but I just want to highlight that this is gonna be a different code path running.

Any thoughts @bfish713?

jparr721 (changes requested):
Just one comment about the log, and then one block I'm otherwise concerned about

crates/task-impls/src/quorum_proposal_recv/mod.rs
```diff
@@ -186,10 +187,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalRecvTaskState<
                     )
                     .await;
                 }
-                Ok(None) => {
-                    self.cancel_tasks(proposal.data.view_number()).await;
-                }
-                Err(e) => debug!(?e, "Failed to propose"),
+                Err(_) => debug!("Failed to propose"),
```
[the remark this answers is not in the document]
shenkeyao: Sure, added!

crates/task-impls/src/quorum_vote/handlers.rs
shenkeyao: I agree it's not a 1:1 change, but I think it aligns with the meaning of this metric, i.e., "Number of views that are in-flight since the last anchor view". Since we moved the proposal validation from the proposal task to the vote task, the latest view "in-flight" is just ` proposal.view_number()` for the vote task.

If we specifically want to use the latest proposed view (rather than the latest seen view) in the calculation, then yes, I think this metric update should be moved to the proposal task. But I still feel it's better to keep this in the vote task where we update the decided view.

jparr721: Yeah I  don't think I disagree with this, just wanted to call it out, I assume there's nothing downstream that depends on a certain pattern to the metrics.

jparr721 (approved):

jparr721 (approved):

jparr721 (approved):

PR #3279, opened 2024-06-05
Add Rewind Task for Replaying All Node Events

jparr721:
No Linked Issue

### This PR: 
The rewind task gives us the ability to store all events that a node has seen since startup. This is useful for a number of situations including test and deployment debugging.

The rewind task works very simply, it has the following high level structure:
```rust
#[cfg(not(any(debug_assertions, test)))]
const REWIND_MAX_DEPTH: usize = 1_000;

/// We want the depth to be huge for tests because we don't care that much about a long-running
/// memory leak
#[cfg(any(debug_assertions, test))]
const REWIND_MAX_DEPTH: usize = 100_000;

/// The task state for the `Rewind` task is used to capture all events received
/// by a particular node, in the order they've been received.
pub struct RewindTaskState<TYPES: NodeType> {
    /// All events received by this node since the beginning of time.
    pub events: VecDeque<Arc<HotShotEvent<TYPES>>>,

    /// The id of this node
    pub id: u64,
}

impl<TYPES: NodeType> RewindTaskState<TYPES> {
    /// Handles all events, storing them to the private state
    pub fn handle(&mut self, event: Arc<HotShotEvent<TYPES>>) {
        if self.events.len() == REWIND_MAX_DEPTH {
            self.events.pop_front();
        }
        self.events.push_back(Arc::clone(&event));
    }
}
```
We implement a shallow cache for non debug or test deployments (i.e. production) to prevent resource exhaustion on the node. The cache size of 1000 is somewhat arbitrary, but should give us a decently long history of the node's life. The intention here being that by the time a performance or service degradation is detected, we may have missed the sliding window that contains the erroneous events. This allows us to correlate with the logs. I'm open to another number if this one seems too large, however.

With the recent architecture changes, we no longer need to filter, so this task will receive and save every single event that it receives. Once the task shuts down, it will write out all of its events in a way which never panics. 

Future work on this task would be correlating the events with timestamps, and sliding windows for deployments inside of staging to enable engineers debugging to have an easier time discovering issues.

Lastly, this PR implements `Display` for the `HotShotEvent`. In most debugging situations, we are typically only concerned with the view that we're receiving an event for (outside of a few cases such as inspecting the justify qc). The intention with this implementation is to strip back the irrelevant metadata fields that pollute the output. This is especially helpful as the logging is completely untouched, so when debugging and correlating events with the log files, one could still easily check other fields as they're unchanged in the logs.

The format for the rewind output is simple, `<sequence no>: <event>`. An example of this format can be seen here:
```
0: ViewChange(view_number=ViewNumber(0))
1: ValidatedStateUpdated(view_number=ViewNumber(0))
2: QcFormed(view_number=ViewNumber(0)
3: UpdateHighQc(view_number=ViewNumber(0))
4: TransactionsRecv
5: TransactionsRecv
6: TransactionsRecv
7: DaProposalRecv(view_number=ViewNumber(1))
8: TransactionsRecv
9: DaProposalValidated(view_number=ViewNumber(1))
```

### Key places to review: 

- `events.rs`
- `rewind.rs`
- `hotshot/src/lib.rs` --> This one kills the entire HotShot instance if rewind is on in production (i.e. `--release`), this prevents accidental activation.

To run with the rewind task, turn on any test that spawns the hotshot suite (like `test_success`) and enable `--features rewind`. Outputs are usually in `crates/testing/rewind_*.log` where `*` corresponds to the node ID.

### DIFF
--- a/crates/hotshot/src/lib.rs
+++ b/crates/hotshot/src/lib.rs
@@ -306,9 +306,6 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> SystemContext<TYPES, I> {
         #[cfg(feature = "dependncy-tasks")]
         error!("HotShot is running with the dependency tasks feature enabled!!");
 
-        #[cfg(all(feature = "rewind", not(debug_assertions)))]
-        compile_error!("Cannot run rewind in production builds!");
-
         debug!("Starting Consensus");
         let consensus = self.consensus.read().await;
 
--- a/crates/hotshot/src/lib.rs
+++ b/crates/hotshot/src/lib.rs
@@ -306,6 +306,9 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> SystemContext<TYPES, I> {
         #[cfg(feature = "dependncy-tasks")]
         error!("HotShot is running with the dependency tasks feature enabled!!");
 
+        #[cfg(all(feature = "rewind", not(debug_assertions)))]
+        compile_error!("Cannot run rewind in production builds!");
+
         debug!("Starting Consensus");
         let consensus = self.consensus.read().await;
 
--- a/crates/hotshot/src/lib.rs
+++ b/crates/hotshot/src/lib.rs
@@ -307,7 +307,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> SystemContext<TYPES, I> {
         error!("HotShot is running with the dependency tasks feature enabled!!");
 
         #[cfg(all(feature = "rewind", not(debug_assertions)))]
-        panic!("Cannot run rewind in production builds!");
+        compile_error!("Cannot run rewind in production builds!");
 
         debug!("Starting Consensus");
         let consensus = self.consensus.read().await;
--- a/crates/hotshot/src/lib.rs
+++ b/crates/hotshot/src/lib.rs
@@ -305,6 +305,10 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> SystemContext<TYPES, I> {
     pub async fn start_consensus(&self) {
         #[cfg(feature = "dependncy-tasks")]
         error!("HotShot is running with the dependency tasks feature enabled!!");
+
+        #[cfg(all(feature = "rewind", not(debug_assertions)))]
+        panic!("Cannot run rewind in production builds!");
+
         debug!("Starting Consensus");
         let consensus = self.consensus.read().await;
 
--- a/crates/hotshot/src/tasks/mod.rs
+++ b/crates/hotshot/src/tasks/mod.rs
@@ -7,6 +7,8 @@ use std::{sync::Arc, time::Duration};
 
 use async_compatibility_layer::art::{async_sleep, async_spawn};
 use hotshot_task::task::Task;
+#[cfg(feature = "rewind")]
+use hotshot_task_impls::rewind::RewindTaskState;
 use hotshot_task_impls::{
     consensus::ConsensusTaskState,
     da::DaTaskState,
@@ -168,4 +170,7 @@ pub async fn add_consensus_tasks<
         handle.add_task(QuorumProposalRecvTaskState::<TYPES, I>::create_from(handle).await);
         handle.add_task(Consensus2TaskState::<TYPES, I>::create_from(handle).await);
     }
+
+    #[cfg(feature = "rewind")]
+    handle.add_task(RewindTaskState::<TYPES>::create_from(&handle).await);
 }
--- a/crates/task-impls/src/events.rs
+++ b/crates/task-impls/src/events.rs
@@ -187,6 +187,7 @@ pub enum HotShotEvent<TYPES: NodeType> {
 }
 
 impl<TYPES: NodeType> Display for HotShotEvent<TYPES> {
+    #[allow(clippy::too_many_lines)]
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             HotShotEvent::Shutdown => write!(f, "Shutdown"),
@@ -357,7 +358,7 @@ impl<TYPES: NodeType> Display for HotShotEvent<TYPES> {
             }
             HotShotEvent::LeafDecided(leaves) => {
                 let view_numbers: Vec<<TYPES as NodeType>::Time> =
-                    leaves.iter().map(|leaf| leaf.view_number()).collect();
+                    leaves.iter().map(Leaf::view_number).collect();
                 write!(f, "LeafDecided({view_numbers:?})")
             }
             HotShotEvent::VidDisperseSend(proposal, _) => write!(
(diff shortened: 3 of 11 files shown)


crates/hotshot/src/lib.rs
```diff
@@ -305,6 +305,10 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> SystemContext<TYPES, I> {
     pub async fn start_consensus(&self) {
         #[cfg(feature = "dependncy-tasks")]
         error!("HotShot is running with the dependency tasks feature enabled!!");
+
+        #[cfg(all(feature = "rewind", not(debug_assertions)))]
```
jparr721, line 309: Thoughts?

crates/hotshot/src/tasks/mod.rs
```diff
@@ -168,4 +170,7 @@ pub async fn add_consensus_tasks<
         handle.add_task(QuorumProposalRecvTaskState::<TYPES, I>::create_from(handle).await);
         handle.add_task(Consensus2TaskState::<TYPES, I>::create_from(handle).await);
     }
+
+    #[cfg(feature = "rewind")]
```
jparr721, line 174: This is in the `add_consensus_tasks` function. Since it's most relevant to the consensus tasks, it seems like a fair place to put it. Open to put it anywhere else.

crates/testing/src/test_runner.rs
```diff
@@ -264,16 +264,16 @@ where
...
-            "TEST FAILED! Results: {error_list:?}"
-        );
-
         let mut nodes = handles.write().await;
 
         for node in &mut *nodes {
             node.handle.shut_down().await;
         }
+
+        assert!(
```
jparr721, line 273: If this doesn't move, test failures will not allow the nodes to write out.

crates/hotshot/src/lib.rs
ss-es: Does a `compile_error!` work here?

crates/task-impls/src/events.rs
```diff
@@ -46,15 +46,15 @@ pub enum HotShotEvent<TYPES: NodeType> {
...
+    /// A Da proposal has been received from the network; handled by the Da task
     DaProposalRecv(Proposal<TYPES, DaProposal<TYPES>>, TYPES::SignatureKey),
-    /// A DA proposal has been validated; handled by the DA task and VID task
+    /// A Da proposal has been validated; handled by the Da task and VID task
     DaProposalValidated(Proposal<TYPES, DaProposal<TYPES>>, TYPES::SignatureKey),
-    /// A DA vote has been received by the network; handled by the DA task
+    /// A Da vote has been received by the network; handled by the Da task
     DaVoteRecv(DaVote<TYPES>),
-    /// A Data Availability Certificate (DAC) has been recieved by the network; handled by the consensus task
+    /// A Data Availability Certificate (DaC) has been recieved by the network; handled by the consensus task
```
ss-es, line 55: completely inconsequential nit: I think this should either be `Dac` or `DAC` or `DaCertificate` but not `DaC` (by analogy with `UUID` -> `Uuid`)

ss-es (commented):
Big +1 on the task in general, I'll try to take a closer look later tonight

crates/task-impls/src/events.rs
bfish713: Personally I think acronyms in comments should be the same as in written text, so all capitalized.

crates/task-impls/src/rewind.rs
```diff
@@ -0,0 +1,69 @@
...
+use std::fs::OpenOptions;
+use std::io::Write;
+
+use crate::events::HotShotEvent;
+
+/// The task state for the `Rewind` task is used to capture all events received
+/// by a particular node, in the order they've been received.
+pub struct RewindTaskState<TYPES: NodeType> {
+    /// All events received by this node since the beginning of time.
+    pub events: Vec<Arc<HotShotEvent<TYPES>>>,
```
bfish713, line 17: This is going to be a big memory leak if we run this on staging or similar.  Maybe the task can checkpoint every 100-1000 views to clear the vector and realease the `HotShotEvent` (I assume this task will be the only task holding the `Arc` by then

crates/task-impls/src/events.rs
jparr721: hah! Stupid find/replace... fixed!

crates/task-impls/src/rewind.rs
bfish713: If it's just for integration tests/demo/benchmark runs then feel free to ignore

crates/hotshot/src/tasks/mod.rs
bfish713: Other option is to just call `add_rewind_task` where we add all the tests, but this is fine.

crates/hotshot/src/lib.rs
jparr721: oooh yes it does! Great suggestion. Done.

bfish713 (commented):
This is a great idea!

crates/task-impls/src/rewind.rs
jparr721: It *is* right now, but I'd like to have this be used in deployments as well. I had mentioned in the comment above to add a sliding window for production use (right now it won't let you deploy with a release configuration at all), but I'd be open to discussing an implementation now if you think it's worth tossing in.

crates/task-impls/src/events.rs
jparr721: Sorry @bfish713 I'm not sure how this comment didn't get loaded, I only saw Salman's. I can swap the comments back

crates/task-impls/src/rewind.rs
```diff
@@ -0,0 +1,81 @@
...
+use async_trait::async_trait;
+use hotshot_task::task::TaskState;
+use hotshot_types::traits::node_implementation::NodeType;
+use std::fs::OpenOptions;
+use std::io::Write;
+
+use crate::events::HotShotEvent;
+
+#[cfg(not(any(debug_assertions, test)))]
+const REWIND_MAX_DEPTH: usize = 1_000;
```
jparr721, line 15: Is this depth too deep for production? I think memory wise it's okay, as, even if this was full of our absolutely largest event, we'd still be under 1mb of total memory usage.

crates/task-impls/src/rewind.rs
bfish713: I think we can just check event view numbers and every time they are over a certain threshold we flush to a file.  Since this task isn't really on a critical path it can be as slow as we want.  So basically set a value called `threshold_view` and check some of the events (maybe `QuorumProposalValidated`) view number to be creater than theshold.  If it is set `threshould_view` to `threshold_view + WINDOW_SIZE` and spawn a task to write to a file/db or whatever.

We could also layer on top of that a sliding window where we just write to 10 different files and overwrite the first on the 11th window.  

Basically I don't think it needs to be anything fancy and it can be addressed in another PR

crates/task-impls/src/rewind.rs
bfish713: No this should be totally fine memory wise, I'd actually be more worried it is too small.

crates/task-impls/src/rewind.rs
```diff
@@ -0,0 +1,81 @@
...
+    pub id: u64,
+}
+
+impl<TYPES: NodeType> RewindTaskState<TYPES> {
+    /// Handles all events, storing them to the private state
+    pub fn handle(&mut self, event: Arc<HotShotEvent<TYPES>>) {
+        if self.events.len() == REWIND_MAX_DEPTH {
+            self.events.pop_front();
+        }
+        self.events.push_back(Arc::clone(&event));
```
bfish713, line 38: I fear this will kind of defeat the point in production,  Lets say an issue happens (e.g. a leader doesn't propose) and the critical part was in view 1000.  We probably won't shut down the node for 1000 more views (and a view has many events) so the interesting sequence will just be lost.

I think this comes down to not waiting until shutdown until we write the events to file.

bfish713 (commented):
Some more comments about the memory use, I think it's probably best to just commit this without the sliding window and just not use it for staging/production until we add the actual mitigation for this.

crates/task-impls/src/events.rs
ss-es: I really have no strong preference between any of the other options, except that I thought `DaC` specifically looked weird. I think `DAC` is what I would write too

crates/task-impls/src/rewind.rs
jparr721: Done

bfish713 (approved):

PR #3281, opened 2024-06-06
Bf/proposal fetch

### LINKED ISSUE #3240: [CATCHUP] - Store Proposals we Send Out

The leader needs to keep the proposals it sent out so that it can respond to requests for the data to catch-up.

Before sending out a proposal the leader should store the proposal in local storage, and persist the data in storage

_No response_

All proposals are stored before being sent out.

_No response_

bfish713:
Closes #3240 

### This PR: 

- Stores most recent proposals in memory
- Adds interface for sequencer to store proposals we send
- Calls this interface before sending out a proposal
- Adds the stored proposals to the `HotshotInitializer` so they can be loaded from the sequencer on startup

### Key places to review: 

New interface and init code

### DIFF
--- a/crates/types/src/consensus.rs
+++ b/crates/types/src/consensus.rs
@@ -12,19 +12,20 @@ use tracing::{debug, error};
 
 pub use crate::utils::{View, ViewInner};
 use crate::{
-    data::{Leaf, VidDisperse, VidDisperseShare},
+    data::{Leaf, QuorumProposal, VidDisperse, VidDisperseShare},
     error::HotShotError,
     message::Proposal,
     simple_certificate::{DaCertificate, QuorumCertificate, UpgradeCertificate},
     traits::{
         block_contents::BuilderFee,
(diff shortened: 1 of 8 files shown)


jparr721 (dismissed):

crates/types/src/consensus.rs
```diff
@@ -248,16 +251,24 @@ impl<TYPES: NodeType> Consensus<TYPES> {
...
+    pub fn update_last_proposed_view(
+        &mut self,
+        proposal: Proposal<TYPES, QuorumProposal<TYPES>>,
+    ) -> Result<()> {
         ensure!(
-            view_number > self.last_proposed_view,
+            proposal.data.view_number()
+                > self
+                    .last_proposals
+                    .first_key_value()
```
ss-es, line 266: should this be `last_key_value` for the proposal with the highest key (view number)?

jparr721 (approved):

PR #3284, opened 2024-06-06
Make SimpleBuilder listen on 0.0.0.0

ImJeremyHe:
### This PR: 
For convenience, we usually prefer to run the `SimpleBuilder` in a Docker environment. Listening only to `localhost` makes the Builder API inaccessible from outside the docker container. Although I think this change is unreasonable for HotShot, it is currently the most suitable option.

### This PR does not: 
- introduce any risk since the change is under the `testing` feature

### DIFF
--- a/crates/testing/src/block_builder.rs
+++ b/crates/testing/src/block_builder.rs
@@ -96,7 +96,7 @@ where
         num_storage_nodes: usize,
         config: Self::Config,
     ) -> (Option<Box<dyn BuilderTask<TYPES>>>, Url) {
-        let url = Url::parse(&format!("http://localhost:{0}", config.port)).expect("Valid URL");
+        let url = Url::parse(&format!("http://0.0.0.0:{0}", config.port)).expect("Valid URL");
         let (source, task) = make_simple_builder(num_storage_nodes).await;
 
         let builder_api = hotshot_builder_api::builder::define_api::<


tbro:
Since we already added a `Config` to take in a port, why don't we add `host` there and take it in? If we use default when not provided, there would still not be any changes required on the dev-node. On the other hand, we *could* just add a `Url` param, and change `builder_port` param on dev-node to `builder_url`. Not sure which is better.

ImJeremyHe:
If anyone sees a `url` in the configuration, I think the first thought is that this application will be going to access some other services to fetch back the data. No one would expect that this `url` is what we listening on.
And if anyone sees a `port` in the configuration, I think the most reasonable thought of this would be this `application` is going to run a server on this `port`.
Since the `SimpleBuilder` here is provided to be used as a server, we should not limit its listening address.

tbro:
> If anyone sees a `url` in the configuration, I think the first thought is that this application will be going to access some other services to fetch back the data. No one would expect that this `url` is what we listen on. And if anyone sees a `port` in the configuration, I think the most reasonable thought of this would be that this `application` is going to run a server on this `port`. Since the `SimpleBuilder` here is provided to be used as a server, we should not limit its listening address.

Yea, you are probably right about the expectation of `url`. I can still see potential issues w/ hard-coding a server to bind to all network devices. You often need to assign a server to a specific IP. And adding this to configuration only adds the possibility of doing that, it doesn't prevent anyone from binding to `0.0.0.0`.  If we make `0.0.0.0` the default it's really the same as you have here but w/ some future proofing.

QuentinI (approved):

PR #3285, opened 2024-06-06
Fix Libp2p Success Test

jparr721:
Closes #<ISSUE_NUMBER>

### This PR: 
Removes the gate on the current proposal's view for the libp2p test.

### This PR does not: 
Does not fix the failure test case.

### DIFF
--- a/crates/task-impls/src/quorum_vote.rs
+++ b/crates/task-impls/src/quorum_vote.rs
@@ -52,8 +52,6 @@ enum VoteDependency {
     Vid,
     /// For the `VoteNow` event.
     VoteNow,
-    /// For the `ValidatedStateUpdated` event.
-    ValidatedState,
 }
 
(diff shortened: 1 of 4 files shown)


bfish713 (dismissed):

crates/task-impls/src/quorum_vote.rs
```diff
@@ -418,11 +412,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumVoteTaskState<TYPES, I
             return;
         }
 
-        debug!(
```
jparr721, line 421: I added these two logs when debugging, but the event is mostly `None` so they're just noise.

bfish713 (approved):

bfish713 (changes requested):
Looks like it breaks some unit test

bfish713 (approved):

PR #3287, opened 2024-06-06
Add OuterConsensus for debugging locks

lukaszrzasik:
Closes #<ISSUE_NUMBER>

### This PR: 

Substitutes `Arc<RwLock<Consensus<TYPES>>>` with `OuterConsensus<TYPES>` which is a thin wrapper around the substituted type plus debug traces for debugging deadlocks around `Consensus`.

Additionally adds thin wrappers around lock guards to be able to leave debug traces when the locks are freed.

### DIFF
--- a/crates/task-impls/src/helpers.rs
+++ b/crates/task-impls/src/helpers.rs
@@ -16,9 +16,8 @@ use async_lock::RwLock;
 use async_std::task::JoinHandle;
 use chrono::Utc;
 use committable::{Commitment, Committable};
-use hotshot_types::consensus::{ConsensusUpgradableReadLockGuard, OuterConsensus};
 use hotshot_types::{
-    consensus::Consensus,
+    consensus::{ConsensusUpgradableReadLockGuard, OuterConsensus},
     data::{Leaf, QuorumProposal, ViewChangeEvidence},
     event::{Event, EventType, LeafInfo},
     message::Proposal,
@@ -35,7 +34,7 @@ use hotshot_types::{
 };
 #[cfg(async_executor_impl = "tokio")]
 use tokio::task::JoinHandle;
-use tracing::{debug, error, info, warn};
+use tracing::{debug, error, info, instrument, warn};
 
 /// Trigger a request to the network for a proposal for a view and wait for the response
 pub(crate) async fn fetch_proposal<TYPES: NodeType>(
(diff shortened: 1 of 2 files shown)


crates/task-impls/src/consensus/helpers.rs
```diff
@@ -396,7 +396,10 @@ pub async fn publish_proposal_from_commitment_and_metadata<TYPES: NodeType>(
         view,
         quorum_membership,
         public_key.clone(),
-        Arc::clone(&consensus),
+        OuterConsensus::new(
+            "publish_proposal_from_commitment_and_metadata->parent_leaf_and_state",
```
jparr721, line 400: I think this change is awesome. My only concern is that if this method name changes, we lose the context in all the log spots and, as a result, the duplication of this name creates some cognitive load beyond what the editor could reasonably infer. What are your thoughts here? Perhaps a macro could be used to recover the name? I'm not sure if there's a better option.

jparr721 (approved):
Great work.

shenkeyao (commented):
I really like this work! I have the same concern as @jparr721 and here's the closet discussion I could find online: https://stackoverflow.com/questions/62701543/is-there-way-to-return-the-name-of-a-function-to-a-string-in-rust. The proposed solution is to use `stringify` to get the function name, with the downside that we need to be careful where `stringify` is put to avoid missing the names of the caller and callee. I think this is probably worth considering!

crates/task-impls/src/consensus/helpers.rs
ss-es: There's a crate that (allegedly) captures the call stack when acquiring a lock: https://crates.io/crates/rust-debugging-locks

Though I wasn't able to test it, but if we now have a wrapper struct used everywhere it might be easy to just swap out the underlying `RwLock` for the `RwLockWrapped` it provides and see if it works?

You might also be able to just capture/print a `Backtrace`: https://doc.rust-lang.org/std/backtrace/struct.Backtrace.html (though I don't really know if that does what I think it does)

bfish713: I also think we should look into what the tracing library can do as well.  Specifically if we set up our tracing spans properly this should give us a good idea of who is doing the locking

bfish713 (approved):

PR #3307, opened 2024-06-11
Fix Leaf Ascension

jparr721:
Closes #<ISSUE_NUMBER>

### This PR: 
Fixes a bug in the protocol where the leaf ascension operation would double process the same view, leading to decide events occurring one view earlier than expected, causing a whole host of suspicious downstream behavior.

This PR also removes the re-implemented logic in the quorum proposal task (for now) and replaces it with the working and vetted version. This is in an effort to aid debugging, and the legacy version was updated to be more ergonomic overall.

Lastly, it blows away the old unit testing since we no longer have a bespoke leaf traversal method.


crates/task-impls/src/consensus/helpers.rs
```diff
@@ -717,140 +717,215 @@ pub(crate) async fn handle_quorum_proposal_recv<TYPES: NodeType, I: NodeImplemen
...
+    let mut res = LeafChainTraversalOutcome::default();
+
+    if let Err(e) = consensus_reader.visit_leaf_ancestors(
+        parent_view_number,
+        Terminator::Exclusive(old_anchor_view),
+        true,
+        |leaf, state, delta| {
+            let new_decide_reached = res.new_decided_view_number.is_some();
+
+            if !new_decide_reached {
```
jparr721, line 809: Gonna swap this and remove the above var, just waitng for CI before I push again.

bfish713 (approved):

Issue #3242, opened 2024-05-30
[CATHCUP] - Replicas Have logic to Handle Responses to proposal requests


bfish713:
### What is this task and why do we need to work on it?

The replica need to validate a the response they get and handle the data by updating our storage, or if not found continuing down some other path.

### What work will need to be done to complete this task?

Create functions to handle the case of found an not found proposals according to the spec: https://www.notion.so/espressosys/Fetch-Missing-Data-for-Catch-up-f7befbd80031469aa6dd9805b35c4e55

### What are the acceptance criteria to close this issue?

We have a function that can properly handle the responses

Issue #3241, opened 2024-05-30
[CATCHUP] - Respond With Proposals when Requested


bfish713:
### What is this task and why do we need to work on it?

When some node is trying to catch up and asks us for a proposal we should respond with that proposal if we were the leader that round.  We should look it up from our local storage.

### What are the acceptance criteria to close this issue?

A leader can sends proposals to to other nodes

PR #3310, opened 2024-06-12
[CATCHUP] Fetch Proposal when Missing

### LINKED ISSUE #3243: [CATCHUP] - Integrate Proposal Fetching to Current Impl

We should have all the building blocks where we can request and respond with proposals for prior views, and validate and update our storage.  Now we need to hook up those functions into the implementation.

For the current (not dependency) version of the code we'll need to send the event to look up the proposal when missing, then spawn a task to wait for the response.  On response received we should then call the handling logic and eventually emit a `QuorumProposalValidated` event

_No response_

_No response_

We have a test that shows a node can vote in the first view it comes back on line for

_No response_

bfish713:
Closes #3243, #3242, #3241

### This PR: 

- Adds logic into the request task to handle the response and send it back to the task requesting the data
- Adds a function `fetch_proposal` to trigger the request task to get the proposal, awiats the response and validates the proposal
- When lookup of the parent fails fetches the proposal using `fetch_proposal`

### Key places to review: 
The changes in helper.rs are the most sensitive changes


crates/task-impls/src/events.rs
```diff
@@ -28,6 +29,24 @@ impl<TYPES: NodeType> TaskEvent for HotShotEvent<TYPES> {
     }
 }
 
+/// Wrapper type for the event to notify tasks that a proposal for a view is missing
+/// and the channel to send the event back to
+#[derive(Debug, Clone)]
+pub struct ProposalMissing<TYPES: NodeType> {
```
ss-es, line 35: Would a better name be something like `QuorumProposalRequest`? (Since it has the channel in it)

crates/task-impls/src/consensus/helpers.rs
```diff
@@ -596,14 +651,29 @@ pub(crate) async fn handle_quorum_proposal_recv<TYPES: NodeType, I: NodeImplemen
...
+        .consensus
+        .read()
+        .await
+        .saved_leaves()
+        .get(&justify_qc.date().leaf_commit)
+        .cloned();
+
+    parent_leaf = match parent_leaf {
+        Some(p) => Some(p),
+        None => fetch_proposal(
```
ss-es, line 664: I think this is fine, just wanted to check my understanding here: I guess we can block consensus here for up to `REQUEST_TIMEOUT` if the parent leaf is missing? But it shouldn't be a big deal because post-refactor we will have a separate task loop spawned per view, right?

crates/task-impls/src/consensus/helpers.rs
```diff
@@ -539,6 +546,54 @@ pub async fn publish_proposal_if_able<TYPES: NodeType>(
...
+    consensus_write.update_validated_state_map(
+        view,
+        View {
+            view_inner: ViewInner::Leaf {
+                leaf: leaf.commit(),
+                state,
+                delta: None,
+            },
+        },
+    );
```
ss-es, line 591: I'm not really able to review this part, because I don't know exactly what checks we need to make before calling this function.

In e.g. `update_state_and_vote_if_able` we seem to gate calling this on validating the DAC, so I don't know if that's a requirement or if there are other things we need to be checking for safety.

ss-es (commented):
It looks good to me, but I wanted to leave approving to someone who knows more about `update_validated_state_map` (@shenkeyao?)

crates/task-impls/src/request.rs
```diff
@@ -191,11 +200,39 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> NetworkRequestState<TYPES, I
...
+        };
+        let Ok(data) = bincode::serialize(&request) else {
+            tracing::error!("Failed to serialize request!");
+            return;
+        };
+        let Ok(signature) = TYPES::SignatureKey::sign(&self.private_key, &Sha256::digest(data))
+        else {
+            error!("Failed to sign Data Request");
+            return;
+        };
```
shenkeyao, line 229: Nit: Looks like `do_proposal` will do another serialization on these parameters. Can we put these 4 serializations in one function?

crates/task-impls/src/consensus/helpers.rs
bfish713: Yeah in the new version we can just block "forever" because the task will not block the main task

crates/task-impls/src/request.rs
bfish713: I combined the sign and serialize part into a function used in both code paths.  The serialization in `do_proposal` is different as it's serializing the entire `Message` including the signature to be sent over the wire so it can't be combined with this serialization

crates/hotshot/src/lib.rs
```diff
@@ -602,8 +602,8 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> SystemContext<TYPES, I> {
 
         if let Some(request_receiver) = da_network.spawn_request_receiver_task().await {
             add_response_task(&mut handle, request_receiver).await;
-            add_request_network_task(&mut handle).await;
         }
+        add_request_network_task(&mut handle).await;
```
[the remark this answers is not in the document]
bfish713: Ah it's so that the request for data call can fail and send something back to the caller for fetching proposal.  It simply returns an error if it's not implemented.  Otherwise we ended up waiting forever for a response.  This is a small hack to get it to work in the current impl that can probably be removed since we have the `async_timeout` but I still think it's best to fail faster.  I'll try to move it out so we don't end up getting random tries to request VID shares

PR #3315, opened 2024-06-12
[DEPENDENCY_REFACTOR] - Modify `update_validated_state_map` to not override a view with more info

### LINKED ISSUE #3306: [DEPENDENCY_REFACTOR] - Modify `update_validated_state_map` to not override a view with more info

Add checks to `update_validated_state_map` to avoid:
* Override a `Leaf` view with a `Da` view.
* Override `Some` delta in a `Leaf` view with `None` delta.

Then we can remove the corresponding checks when calling `update_validated_state_map`.

_No response_

_No response_

We only update a validated state if it contains more inforamtion.information

_No response_

shenkeyao:
Closes #3306.

### This PR: 
* Adds checks to `update_validated_state_map` to avoid overriding existing `View`s with new ones with less information.
* Removes a duplicated check in the DA task.
* Fixes tests due to duplicated updates.

### Key places to review: 
* `fn update_validated_state_map`.


crates/types/src/consensus.rs
```diff
@@ -299,7 +299,27 @@ impl<TYPES: NodeType> Consensus<TYPES> {
     }
 
     /// Update the validated state map with a new view_number/view combo.
+    ///
+    /// Skips the update if the new view contains less information than the exisiting view with the
```
jparr721, line 303: What do you think about this returning a result or something instead of just a blank return, and the function could perhaps be renamed to something like `try_update...` to signify the fallible operation? The thing about it right now is that when it's called, it might fail, and the caller may have no idea that it failed, which can make debugging harder.

crates/types/src/consensus.rs
```diff
@@ -299,7 +299,27 @@ impl<TYPES: NodeType> Consensus<TYPES> {
...
+    /// same view number.
     pub fn update_validated_state_map(&mut self, view_number: TYPES::Time, view: View<TYPES>) {
+        if let Some(existing_view) = self.validated_state_map().get(&view_number) {
+            if let ViewInner::Leaf { .. } = existing_view.view_inner {
+                match view.view_inner.clone() {
```
jparr721, line 308: Do we need to clone here?

crates/types/src/consensus.rs
shenkeyao: Yeah how about this -- I can make it similar to `update_high_qc`, i.e., using `ensure` and `Result` to let the caller know something is wrong.

crates/types/src/consensus.rs
shenkeyao: I added this to fix a build error. Let me double-check!

crates/types/src/consensus.rs
jparr721: Perfect

crates/types/src/consensus.rs
shenkeyao: Yeah looks like it's needed. It's due to a partial value move a few lines later on `ViewInner::Leaf { delta, .. }`.

crates/types/src/consensus.rs
```diff
@@ -299,8 +299,32 @@ impl<TYPES: NodeType> Consensus<TYPES> {
...
+        view: View<TYPES>,
+    ) -> Result<()> {
+        if let Some(existing_view) = self.validated_state_map().get(&view_number) {
+            if let ViewInner::Leaf { .. } = existing_view.view_inner {
+                match view.view_inner.clone() {
+                    ViewInner::Leaf { delta, .. } => {
+                        ensure!(
+                            delta.is_some(),
+                            "Skipping the state update to not override a `Leaf` view with `None` state delta."
+                        );
```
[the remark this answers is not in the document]
shenkeyao: Yes! I just realized clippy also suggested `ref`. Done.

PR #3317, opened 2024-06-12
Minor test fixes (again)

ss-es:
No linked issue.

### This PR: 
This is a second attempt at #3256, which was reverted. The tightening of the `num_failed_views` parameter in various tests is retained, but the check that we fail no fewer than `num_failed_views` was removed (since many tests do not reliably fail a specific number of views, this requires more work).

A consistency check was also added, to validate that each leaf chain from each node's decide events is coherent, and that no two nodes disagree on a leaf for any given view.


crates/testing/src/consistency_task.rs
```diff
@@ -0,0 +1,176 @@
...
+
+trait Validatable {
+    fn valid(&self) -> Result<()>;
+}
+
+pub type NodeMap<TYPES> = BTreeMap<<TYPES as NodeType>::Time, Vec<Leaf<TYPES>>>;
+
+pub type NodeMapSanitized<TYPES> = BTreeMap<<TYPES as NodeType>::Time, Leaf<TYPES>>;
+
+pub fn sanitize_node_map<TYPES: NodeType>(
```
jparr721, line 26: Could we get some doc comments explaining what these things are?

crates/testing/src/consistency_task.rs
```diff
@@ -0,0 +1,176 @@
...
+        }
+
+        Ok(())
+    }
+}
+
+pub type NetworkMap<TYPES> = BTreeMap<usize, NodeMap<TYPES>>;
+pub type NetworkMapSanitized<TYPES> = BTreeMap<usize, NodeMapSanitized<TYPES>>;
+
+pub fn sanitize_network_map<TYPES: NodeType>(
```
jparr721, line 78: Can we get doc comments for these?

crates/testing/src/consistency_task.rs
```diff
@@ -0,0 +1,176 @@
...
+) -> Result<NetworkMapSanitized<TYPES>> {
+    let mut result = BTreeMap::new();
+
+    for (node, node_map) in network_map {
+        result.insert(
+            *node,
+            sanitize_node_map(node_map)
+                .context(format!("Node {node} produced inconsistent leaves."))?,
+        );
+    }
```
jparr721, line 89: This could be `BTreeMap::from_iter` if you wanted it to be.

crates/testing/src/consistency_task.rs
```diff
@@ -0,0 +1,176 @@
...
+                view_map.insert(*node_id, leaf.clone());
+            }
+        }
+
+        for (view, view_map) in inverted_map.iter() {
+            let mut leaves: Vec<_> = view_map.iter().collect();
+
+            leaves.dedup_by(|(_node_a, leaf_a), (_node_b, leaf_b)| leaf_a == leaf_b);
+
+            if leaves.len() > 1 {
```
jparr721, line 124: This could be `ensure!` if you want.

crates/testing/src/overall_safety_task.rs
```diff
@@ -225,6 +231,18 @@ impl<TYPES: NodeType, I: TestableNodeImplementation<TYPES>> TestTaskState
                 failed_views: self.ctx.failed_views.clone(),
             }));
         }
+
+        // We should really be able to include a check like this:
```
jparr721, line 235: Should we just remove this and instead file an issue with this context? Just to avoid it getting lost?

jparr721 (dismissed):
Just looking for some more docs to make understanding the tests a bit easier at a glance.

crates/testing/src/consistency_task.rs
ss-es: I initially tried something like that, but I don't think it quite works because you can't get the short-circuiting/early return behavior from the `?` here without an explicit loop

crates/testing/src/overall_safety_task.rs
ss-es: I really like having a comment in the code, but I've also created an issue: https://github.com/EspressoSystems/HotShot/issues/3334

jparr721: Yeah that make sense to me!

bfish713 (dismissed):

jparr721 (approved):

PR #3322, opened 2024-06-13
[DEPENDENCY_REFACTOR] - Fix the ValidatedStateUpdated dependency in the quorum proposal and vote tasks

### LINKED ISSUE #3296: [DEPENDENCY_REFACTOR] - Fix the `ValidatedStateUpdated` dependency in the quorum proposal and vote tasks

The quorum proposal and vote tasks should have no dependency on `ValidatedStateUpdated`. Instead, we should wait for the event before proceeding in `handle_deps`.

_No response_

Zulip discussion: https://espresso.zulipchat.com/#narrow/stream/311646-Consensus/topic/Dependency.20task.20refactor/near/443326357.

The `ValidatedStateUpdated` dependency is removed from both tasks, but we won't handle the dependencies when the event isn't received.

_No response_

shenkeyao:
Closes #3296. **WIP!**


crates/task-impls/src/quorum_proposal/dependency_handle.rs
```diff
@@ -175,6 +172,20 @@ impl<TYPES: NodeType> HandleDepOutput for ProposalDependencyHandle<TYPES> {
...
+            if self
+                .consensus
+                .read()
+                .await
+                .validated_state_map()
+                .contains_key(&high_qc_view_number)
+            {
+                break;
+            }
+        }
```
shenkeyao, line 187: Does this loop make sense? In our original discussion, our conclusion was to wait for the `ValidatedStateUpdated` event, but I found it easier to block on the state map instead. Can we consider them equivalent?

jparr721: I think we can. For the purposes of what we're trying to accomplish, this seems more than adequate.

bfish713: I think this is going to cause really high CPU use as it's quite likely we are almost always going to be able to grab the read lock

crates/task-impls/src/quorum_proposal/dependency_handle.rs
```diff
@@ -175,6 +177,30 @@ impl<TYPES: NodeType> HandleDepOutput for ProposalDependencyHandle<TYPES> {
 
     #[allow(clippy::no_effect_underscore_binding)]
     async fn handle_dep_result(self, res: Self::Output) {
+        let high_qc_view_number = self.consensus.read().await.high_qc().view_number;
```
jparr721, line 180: Followup on an earlier discussion, in all testing, this value, and the provided high qc value were never different, so it seemed like a red herring. I can add the change in though that effectively takes the high qc from the event if it's available, and falls back to the shared state otherwise.

crates/task-impls/src/consensus/helpers.rs
```diff
@@ -105,6 +105,10 @@ pub async fn validate_proposal_safety_and_liveness<TYPES: NodeType>(
     {
         tracing::trace!("{e:?}");
     }
+    consensus
+        .write()
+        .await
+        .update_saved_leaves(proposed_leaf.clone());
```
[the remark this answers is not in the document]
jparr721: Yeah we definitely want to do this. I had originally done it, but the change was not great. I plan to do it as a follow up after evaluating if ViewInner::Da is still relevant, as it made the implementation quite tricky. I’ll tag you in the follow up.

PR #3324, opened 2024-06-13
Benchmarking adjustments

QuentinI:
- Merge main
- Adjust orchestrator to choose from potential builder URLs
- Adjust examples to take advantage of multiple builders


crates/examples/infra/mod.rs
```diff
@@ -1057,48 +1030,64 @@ where
...
+    let mut advertise_urls = Vec::new();
+    let bind_address: Url;
+    match args.builder_address {
+        None => {
+            let port = portpicker::pick_unused_port().expect("Failed to pick an unused port");
+            let possible_urls = local_ip_address::list_afinet_netifas()
+                .expect("Couldn't get list of local IP addresses")
+                .into_iter()
+                .map(|(_name, ip)| ip)
+                .filter(|ip| !ip.is_loopback())
```
dailinsubjam: Could work after adding `.filter(|ip| ip.to_string().starts_with("172.31."))`, but I'm not sure how to do this in a more formal way. Or else there will be `unwrap()` error.

dailinsubjam (approved):

PR #3333, opened 2024-06-18
Increase delays to 5 seconds

### LINKED ISSUE #3199: [Libp2p upgrade] Tune delays and timeouts

We want to tune Libp2p delays and timeouts based on output from the [staging environment](https://github.com/EspressoSystems/HotShot/issues/3194).

This will make the real network more efficient as it uses a sampling of a more realistic deployment scenario.

_No response_

_No response_

- [ ] The obvious performance optimizations have been made

_No response_

lukaszrzasik:
Closes #3199

### This PR: 

Please read [the analysis of the delays and timeouts](https://www.notion.so/espressosys/Combined-networks-timeouts-and-delays-45614652b49e47449d43afba43026ebe?pvs=4).

I've added a file with a staging env config plus the adjusted delays. This is the only change here.

### DIFF
--- a/crates/orchestrator/staging-config.toml
+++ b/crates/orchestrator/staging-config.toml
@@ -61,7 +61,7 @@ num_nodes_without_stake = 0
 staked_da_nodes = 10
 non_staked_da_nodes = 0
 fixed_leader_for_gpuvid = 1
-next_view_timeout = 12_000
+next_view_timeout = 15_000
 timeout_ratio = [ 11, 10 ]
 round_start_delay = 1
 start_delay = 10_000
@@ -69,7 +69,7 @@ num_bootstrap = 5
 builder_urls = [ "https://builder.staging.testnet.espresso.network/" ]
 
 [config.view_sync_timeout]
-secs = 1
+secs = 15
 nanos = 0
 
 [config.builder_timeout]


shenkeyao:
_[Pasting my comment on Zulip]_

Agree that `secondary_network_delay`, or `CombinedNetworks::delay_duration` in the doc, should be in the range of `(3, 12)` and `5` seems to make sense. Also +1 that it's better to adjust via HotShot config.

One thing I'm not sure about is the `data_request_delay` parameter, or `NetworkRequestState::delay` and `DelayedRequester::delay` in the doc. Is it an additional delay (i.e., the total delay is close to `secondary_network_delay + data_request_delay`), or are they more like parallel delays? If it's the former case, I think `data_request_delay` should be a small value, e.g., 200 ms as before, to not make the total delay too close to 12 seconds. If it's the latter case, then I think it makes sense to adjust these two parameters to the same/similar value.

crates/orchestrator/staging-config.toml
```diff
@@ -0,0 +1,95 @@
...
+next_view_timeout = 12_000
+timeout_ratio = [ 11, 10 ]
+round_start_delay = 1
+start_delay = 10_000
+num_bootstrap = 5
+builder_urls = [ "https://builder.staging.testnet.espresso.network/" ]
+
+[config.view_sync_timeout]
+secs = 1
+nanos = 0
```
shenkeyao, line 73: Nit: Are these `nanos` configs necessary if they are all set to 0?

shenkeyao (approved):
Just a nit comment. Everything else looks good!

shenkeyao (approved):

PR #3339, opened 2024-06-19
[CATCHUP] Fetch Proposals in Dependecny Task

### LINKED ISSUE #3244: [CATCHUP] - Inegrate Proposal Fetching Into Dependency Task

Because the dependency tasks are one off tasks, and fetching the parent can be done inline where we need it.  

_No response_

_No response_

Same as #3243 but when running with the new code

_No response_

bfish713:
Closes #3244 

### This PR: 

1. Fetches proposals if we can from the vote task if the parent proposal is missing
2. Fetches proposals from the proposal task if we form a QC for a proposal we don't have
3. Emits the validated state event after updating state in the `fetch_proposal` function

### DIFF
--- a/crates/task-impls/src/quorum_proposal/dependency_handle.rs
+++ b/crates/task-impls/src/quorum_proposal/dependency_handle.rs
@@ -5,7 +5,7 @@ use std::{marker::PhantomData, sync::Arc, time::Duration};
 
 use anyhow::{ensure, Context, Result};
 use async_broadcast::{Receiver, Sender};
-use async_compatibility_layer::art::async_sleep;
+use async_compatibility_layer::art::{async_sleep, async_spawn};
 use async_lock::RwLock;
 use committable::Committable;
 use hotshot_task::{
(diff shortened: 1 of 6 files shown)


crates/task-impls/src/quorum_proposal/dependency_handle.rs
```diff
@@ -185,6 +187,13 @@ impl<TYPES: NodeType> HandleDepOutput for ProposalDependencyHandle<TYPES> {
             .validated_state_map()
             .contains_key(&high_qc_view_number)
         {
+            // The proposal for the high qc view is missing, try to get it asynchronously
+            let memberhsip = Arc::clone(&self.quorum_membership);
+            let sender = self.sender.clone();
+            let consensus = Arc::clone(&self.consensus);
+            async_spawn(async move {
+                fetch_proposal(high_qc_view_number, sender, memberhsip, consensus).await
```
jparr721, line 195: Looks like this acquires a write lock on each invocation. Are we worried about accidentally holding a lock while something else needs it?

jparr721 (approved):
One question but otherwise this looks good.

crates/task-impls/src/quorum_proposal/dependency_handle.rs
jparr721, line 195: How come this is an async spawn instead of just calling? It looks like the method doesn't do anything long-running?

bfish713: The fetch could be potentially very slow (500ms) and this caller isn't expecting the result directly back, in the mean time we might get the `ValidatedStateUpdated` event we need before the request for proposal ever even finishes

PR #3347, opened 2024-06-20
Declutter CI Logs

rob-maron:
No linked issue

### This PR: 

- Removes the instrument that prints every elliptic curve point for all public keys 
- Removes `--verbose` for tests. This appears to just make the build itself verbose, which prints out a bunch of `rustc` data that I don't think we really need. But let me know if I'm missing something
- Makes some miscellaneous changes to satisfy clippy


ss-es (approved):
Looks good to me

PR #3365, opened 2024-06-24
Remove VID precomputation from builder API

### LINKED ISSUE #3183: [BUILDER] - Decouple `claim_block` and `claim_block_header` calls

Currently transaction task waits for both claims before notifying other tasks. `claim_block_header` could take longer because of VID computation on the builder and it's results aren't used by DA task, so DA task should be notified as soon as `claim_block` call succeeds.

_No response_

_No response_

Tasks relying only on information available after `claim_block` call are notified as soon as it succeeds without waiting for `claim_block_header`

_No response_

### LINKED ISSUE #3364: [BUILDER] - Remove VID precomputation from builder API

For marketplace we want to move away from VID computation on the builder.

- Remove `claim_block_header_input` endpoint and all implementations from builder API
- Calculate VID without precompute data

_No response_

- VID precomputation infrastructure removed
- Tests pass

_No response_

QuentinI:
Closes #3364
Closes #3183 as no longer relevant

### This PR: 
- Removes `claim_block_header_input` endpoint from builder API
- Removes all implementations thereof
- Removes usage of VID precompute in favor of full VID computation on the node
- Changes builder's signature over block data to also sign the fee amount 

### Key places to review: 
The most critical place is `crates/types/src/traits/signature_key.rs`, as it changes how the fee is signed. Everything else should be pretty straightforward.


elliedavidson (commented):
I only did a quick glance, but it looks good!  The only thing is that we'll want to enable this change via versioning so that we can switch to this new mode at runtime.  We are still working out how the versioning will work for the marketplace changes.

crates/task-impls/src/consensus/helpers.rs
```diff
@@ -3,14 +3,11 @@ use std::{
...
-use crate::{events::ProposalMissing, request::REQUEST_TIMEOUT};
-use anyhow::bail;
-use anyhow::{ensure, Context, Result};
+use anyhow::{bail, ensure, Context, Result};
 use async_broadcast::{broadcast, Sender};
 use async_compatibility_layer::art::async_timeout;
 use async_lock::RwLock;
-#[cfg(not(feature = "dependency-tasks"))]
-#[cfg(async_executor_impl = "async-std")]
+#[cfg(all(async_executor_impl = "async-std", not(feature = "dependency-tasks")))]
```
jparr721, line 10: Nice

crates/task-impls/src/transactions.rs
```diff
@@ -456,10 +443,14 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, Ver: StaticVersionType>
                 }
             };
 
-            let (block, header_input) = futures::join! {
-                client.claim_block(block_info.block_hash.clone(), view_number.u64(), self.public_key.clone(), &request_signature),
-                client.claim_block_header_input(block_info.block_hash.clone(), view_number.u64(), self.public_key.clone(), &request_signature)
-            };
+            let block = client
+                .claim_block(
```
jparr721, line 447: Something seems up with the spacing here.

jparr721 (commented):
LGTM, but it looks like some tests are failing

crates/task-impls/src/transactions.rs
QuentinI: Yeah, it looks weird, but that's what rustfmt says it should look like for some reason 🤷

QuentinI:
Superseded by #3413

Issue #3378, opened 2024-06-25
[Upgradability] - Avoid tracking the version in task states


shenkeyao:
### What is this task and why do we need to work on it?

[Pasting from @ss-es's comment on Zulip.]

we could probably avoid tracking the version as part of the task state. really, it should just be enough to just write the `decided_upgrade_certificate` `Arc<RwLock<...>>` as soon as we hit a decide on it. then any place that needs a version could calculate it dynamically like we do in `types/src/message.rs` by:
```rust
        let version = match upgrade_certificate {
            Some(ref cert) => {
                if view >= cert.data.new_version_first_view
                    && cert.data.new_version == Upgrade::VERSION
                {
                    Upgrade::VERSION
                } else if view >= cert.data.new_version_first_view
                    && cert.data.new_version != Upgrade::VERSION
                {
                    bail!("The network has upgraded to a new version that we do not support!");
                } else {
                    Base::VERSION
                }
            }
            None => Base::VERSION,
        };
```
(this should probably be its own method)

### What are the acceptance criteria to close this issue?

We aren't rely on the version in task states.

Issue #3357, opened 2024-06-21
[Upgradability] - Remove duplicated fields for the decided upgrade cert in `ConsensusTaskState`


shenkeyao:
### What is this task and why do we need to work on it?

`ConsensusTaskState` has two duplicated fields, `decide_upgrade_certificate` and `decide_upgrade_cert`, the former with `Arc<RwLock<..>>` and the latter without. They are both used but we should only keep one.

### What work will need to be done to complete this task?

Replace the use of `decide_upgrade_cert` with `decide_upgrade_certificate`.

### What are the acceptance criteria to close this issue?

There's only one decided upgrade cert field.

PR #3379, opened 2024-06-25
[DEPENDENCY_REFACTOR] - Integrate upgradability into dependency tasks

### LINKED ISSUE #3356: [DEPENDENCY_REFACTOR] - Integrate upgradability into dependency tasks

Add upgrade cert update in the dependency task logic, to enable upgradability when building with the refactored dependency tasks.

_No response_

_No response_

`RUST_LOG=error,hotshot_task_impls=info just cargo run --example all-push-cdn --features example-upgrade -- --config_file ./crates/orchestrator/run-config.toml` works with `dependency-tasks` feature.

_No response_

shenkeyao:
Closes #3356.

### This PR: 
* Integrates the formed and decided upgrade certificates into dependency tasks.
* Renames functions that take non-`Arc<RwLock<...>>` decided upgrade cert as parameter to `temp_*`, and adds new functions that take `Arc<RwLock<...>>`.
  * The dependency tasks now all use the new functions, but the old consensus task uses `temp_*` functions that will be replaced in a separate PR.

### This PR does not: 
* Make improvements to the existing upgradability code, which will be done later. Relevant issues:
  * #3378
  * #3357

### Key places to review: 
* Whether upgrade certificates are added correctly in the dependency tasks.
* Whether any update is missed.
* (`temp_*` functions may be ignored since they only have the name change.)

### How to test this PR:
* Run `RUST_LOG=error,hotshot_task_impls=info just cargo run --example all-push-cdn --features example-upgrade --features dependency-tasks -- --config_file ./crates/orchestrator/run-config.toml`.
  * Note the `dependency-tasks` feature.
* The upgrade should occur for view 25.

### DIFF
--- a/crates/task-impls/src/consensus/mod.rs
+++ b/crates/task-impls/src/consensus/mod.rs
@@ -435,6 +435,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> ConsensusTaskState<TYPES, I>
                     };
                 }
             },
+            #[cfg(not(feature = "dependency-tasks"))]
             HotShotEvent::UpgradeCertificateFormed(cert) => {
                 debug!(
                     "Upgrade certificate received for view {}!",
--- a/crates/task-impls/src/consensus2/handlers.rs
+++ b/crates/task-impls/src/consensus2/handlers.rs
@@ -14,7 +14,7 @@ use hotshot_types::{
     },
     vote::HasViewNumber,
 };
-use tracing::debug;
+use tracing::{debug, warn};
 
 use super::Consensus2TaskState;
 use crate::{
@@ -132,6 +132,23 @@ pub(crate) async fn handle_view_change<TYPES: NodeType, I: NodeImplementation<TY
 
     // Move this node to the next view
     task_state.cur_view = new_view_number;
+
+    // If we have a decided upgrade certificate, we may need to upgrade the protocol version on a
+    // view change.
+    let decided_upgrade_certificate_read =
+        task_state.decided_upgrade_certificate.read().await.clone();
+    if let Some(cert) = decided_upgrade_certificate_read {
+        if new_view_number == cert.data.new_version_first_view {
+            warn!(
+                "Updating version based on a decided upgrade cert: {:?}",
+                cert
+            );
+            let mut version = task_state.version.write().await;
+            let new_version = cert.data.new_version;
+            *version = new_version;
+            broadcast_event(Arc::new(HotShotEvent::VersionUpgrade(new_version)), sender).await;
+        }
+    }
 
     // Spawn a timeout task if we did actually update view
     let timeout = task_state.timeout;
--- a/crates/task-impls/src/quorum_vote/mod.rs
+++ b/crates/task-impls/src/quorum_vote/mod.rs
@@ -17,6 +17,7 @@ use hotshot_types::{
     data::{Leaf, VidDisperseShare},
     event::Event,
     message::Proposal,
+    simple_certificate::UpgradeCertificate,
     simple_vote::{QuorumData, QuorumVote},
     traits::{
         block_contents::BlockHeader,
@@ -389,6 +390,9 @@ pub struct QuorumVoteTaskState<TYPES: NodeType, I: NodeImplementation<TYPES>> {
 
     /// The curent version of HotShot
     pub version: Version,
+
+    /// An upgrade certificate that has been decided on, if any.
+    pub decided_upgrade_certificate: Arc<RwLock<Option<UpgradeCertificate<TYPES>>>>,
 }
 
 impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumVoteTaskState<TYPES, I> {
--- a/crates/task-impls/src/quorum_vote/mod.rs
+++ b/crates/task-impls/src/quorum_vote/mod.rs
@@ -14,7 +14,7 @@ use hotshot_task::{
 };
 use hotshot_types::{
     consensus::Consensus,
-    data::{Leaf, VidDisperseShare},
+    data::{Leaf, VidDisperseShare, ViewNumber},
     event::Event,
     message::Proposal,
     simple_certificate::UpgradeCertificate,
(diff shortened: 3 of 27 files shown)


jparr721 (commented):
I think this looks good, but I noticed we haven't included any tests. Could we add some for the changed tasks?

shenkeyao:
> I think this looks good, but I noticed we haven't included any tests. Could we add some for the changed tasks?

Sure. I originally planed to add tests after reorganizing consensus2 functions because I found it a bit confusing, but on a second thought, I think it makes sense to add tests now. Will send an update!

shenkeyao:
Converting this to a draft until I turn the upgrade task tests on for the dependency tasks.

shenkeyao:
Tests added. Marking this as ready for review.

crates/testing/tests/tests_1/upgrade_task_with_proposal.rs
```diff
@@ -0,0 +1,238 @@
...
+                task_state_asserts: vec![],
+            },
+            Expectations {
+                output_asserts: vec![],
+                task_state_asserts: vec![],
+            },
+        ],
+    };
+
+    test_scripts![inputs, proposal_script, upgrade_script].await;
```
jparr721, line 237: This is totally a confusing thing, so I apologize for that in advance, but this should actually be using `run_test!` and then we'll just need to change all the inputs from `vec` to `random!` or `serial!`. This will automatically check all input permutations, then you can use `Expectations::from_outputs(all_predicates!` instead of the `vec!` thing. The legacy `test_scripts` macro will likely be removed later.

For an example, check out the quorum proposal task.

crates/testing/tests/tests_1/upgrade_task_with_vote.rs
```diff
@@ -0,0 +1,190 @@
...
+    for view in generator.take(4).collect::<Vec<_>>().await {
+        proposals.push(view.quorum_proposal.clone());
+        votes.push(view.create_quorum_vote(&handle));
+        dacs.push(view.da_certificate.clone());
+        vids.push(view.vid_proposal.clone());
+        leaders.push(view.leader_public_key);
+        leaves.push(view.leaf.clone());
+    }
+
+    let inputs = vec![
```
jparr721, line 96: Same here as above, it should be relatively straightforward to find/replace, then we can guarantee that this works independent of the order.

jparr721 (changes requested):
This looks great! Just want to get those tests to use non deterministic ordering and we're good to go.

crates/testing/tests/tests_1/upgrade_task_with_proposal.rs
shenkeyao: Oh I was using the same pattern in the original upgrade tests, but yeah, I think it’s a good idea to use random ordering!

crates/task-impls/src/consensus/mod.rs
```diff
@@ -435,6 +435,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> ConsensusTaskState<TYPES, I>
                     };
                 }
             },
+            #[cfg(not(feature = "dependency-tasks"))]
```
ss-es, line 438: Is this gate redundant? (I thought we didn't spawn regular consensus  with the dependency tasks enabled, though I might have misunderstood)

crates/task-impls/src/consensus2/handlers.rs
```diff
@@ -133,6 +133,23 @@ pub(crate) async fn handle_view_change<TYPES: NodeType, I: NodeImplementation<TY
...
+    if let Some(cert) = decided_upgrade_certificate_read {
+        if new_view_number == cert.data.new_version_first_view {
+            warn!(
+                "Updating version based on a decided upgrade cert: {:?}",
+                cert
+            );
+            let mut version = task_state.version.write().await;
+            let new_version = cert.data.new_version;
+            *version = new_version;
+            broadcast_event(Arc::new(HotShotEvent::VersionUpgrade(new_version)), sender).await;
```
ss-es, line 149: I think most of this is (hopefully) not used anywhere, except the log which is helpful to see. Ideally we can remove it (possibly in a later PR).

Do you happen to know if the upgrade fails without updating `version`? I can investigate if not.

crates/task-impls/src/quorum_vote/mod.rs
```diff
@@ -133,7 +134,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES> + 'static> VoteDependencyHand
                 &parent,
                 &proposed_leaf.block_header().clone(),
                 vid_share.data.common.clone(),
-                self.version,
+                *self.version.read().await,
```
ss-es, line 137: I think there's a danger we get versions mixed up here, if we use a `version` RwLock rather than calculating it dynamically from the `decided_upgrade_certificate` and the view number -- this would probably only happen around an upgrade view, if there's enough network latency where we receive some messages out of order.

I think this is fine for now, just noting that we should definitely switch to using `decided_upgrade_certificate`  + view number to decide the version.

ss-es (approved):
I haven't fully looked through the new tests yet, but overall it looks good to me! Just left some notes -- I think this is the same as the discussion on Zulip we had earlier and agreed to fix later.

crates/task-impls/src/consensus/mod.rs
shenkeyao: It's actually needed, due to some functions used by both the dependency and non-dependency tasks. But of course, this isn't ideal, so we have a task to separate the shared functions!

crates/task-impls/src/consensus2/handlers.rs
shenkeyao: Removing the version update doesn't break the upgrade test but the `example-upgrade` script does fail, so I'll keep it as is for now. We have this issue to improve the version trakcing: https://github.com/EspressoSystems/HotShot/issues/3378.

jparr721 (approved):
Great work, thanks for putting this together.

PR #3386, opened 2024-06-26
Jp/3348

jparr721:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/crates/hotshot/src/lib.rs
+++ b/crates/hotshot/src/lib.rs
@@ -166,6 +166,9 @@ pub struct SystemContext<TYPES: NodeType, I: NodeImplementation<TYPES>> {
 
     /// a potential upgrade certificate that has been decided on by the consensus tasks.
     pub decided_upgrade_certificate: Arc<RwLock<Option<UpgradeCertificate<TYPES>>>>,
+
+    /// Reference to the AuctionResults type for acquiring solver results.
+    pub auction_results: Arc<RwLock<I::AuctionResults>>,
 }
 impl<TYPES: NodeType, I: NodeImplementation<TYPES>> Clone for SystemContext<TYPES, I> {
     #![allow(deprecated)]
@@ -188,6 +191,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> Clone for SystemContext<TYPE
             id: self.id,
             storage: Arc::clone(&self.storage),
             decided_upgrade_certificate: Arc::clone(&self.decided_upgrade_certificate),
+            auction_results: Arc::clone(&self.auction_results),
         }
     }
 }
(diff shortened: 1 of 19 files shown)


crates/hotshot/src/lib.rs
```diff
@@ -166,6 +166,9 @@ pub struct SystemContext<TYPES: NodeType, I: NodeImplementation<TYPES>> {
 
     /// a potential upgrade certificate that has been decided on by the consensus tasks.
     pub decided_upgrade_certificate: Arc<RwLock<Option<UpgradeCertificate<TYPES>>>>,
+
+    /// Reference to the AuctionResults type for acquiring solver results.
+    pub auction_results: Arc<RwLock<I::AuctionResults>>,
```
jbearer, line 171: I wonder if this can just be `Arc` without the `RwLock`: it feels like all the operations on the auction results should be immutable

crates/types/src/traits/auction_results.rs
```diff
@@ -0,0 +1,31 @@
...
+/// for the builder that HotShot must request from.
+pub trait HasUrl {
+    /// Returns the builer url associated with the datatype
+    fn url(&self) -> Url;
+}
+
+/// The AuctionResults trait is the sole source of Solver-originated allocations, and returns
+/// the results via its own custom type.
+#[async_trait]
+pub trait AuctionResults<TYPES: NodeType>: Send + Sync {
```
jbearer, line 20: I find the name of this trait a little confusing, especially with the associated type `AuctionSolverResult`: it's not totally clear what the difference is. May I suggest calling this something like `AuctionProvider` or `AuctionSolver`, since this trait is really an interface to _get_ the results, note the results themselves? Then we could just call the associated type `AuctionResult`

crates/types/src/traits/auction_results.rs
```diff
@@ -0,0 +1,31 @@
...
+    /// The AuctionSolverResult is a type that holds the data associated with a particular solver
+    /// run, for a particular view.
+    type AuctionSolverResult: HasUrl;
+
+    /// Fetches the auction result for a view. Does not cache the result,
+    /// subsequent calls will invoke additional wasted calls.
+    async fn fetch_auction_result(
+        self,
```
jbearer, line 28:
```suggestion
        &self,
```

jbearer (approved):
Interfaces look good to me, except a minor comment on naming

crates/hotshot/src/lib.rs
jparr721: Make sense, swapping.

crates/types/src/traits/auction_results.rs
jparr721: Done.

crates/types/src/traits/auction_results.rs
jparr721: Done.

jbearer (dismissed):
LGTM

bfish713 (approved):

PR #3387, opened 2024-06-26
Remove combined fail increment

rob-maron:
This removes a spot where we increment the primary network's failure counter when it's not clear if it failed or not. Right now if a view doesn't update for any reason within a specified amount of time, we treat this as a primary network failure. This would force nodes to use Libp2p if something else happens, like we are running into some timeouts


jparr721 (approved):
Are the test failures somehow related to this?

rob-maron:
> Are the test failures somehow related to this?

Unfortunately I don't think so, but it fixes an issue we saw in staging

PR #3390, opened 2024-06-27
[DEPENDENCY_TASK] Fix CDN Crash Tests

### LINKED ISSUE #3368: [DEPENDENCY_REFACTOR] - Fix Combined Network Tests

find the bug

_No response_

_No response_

Green CI

_No response_

bfish713:
Closes #3368 

### This PR: 

Solves the final tests failures for the dependency task refactor.  The failure was caused because for some code paths we can calculate the VID Shares, store them in the vid share map but not emit a `VidShareRecv` event.  This only happens when running with the combined network.  There are 2 places we could calculate the VID shares from a DAProposal.  The first is in the DA task if the primary is down.  The second is if we get a request from another peer for a VID share we'll try to calculate it if we don't have it but do have a DA Proposal.  This led to some nodes not voting when they should be able to because the dependency task was waiting for a VIDShareRecv event that wouldn't come.

Typically what I was seeing in CI was that 6/10 nodes would vote and 4 would not.  I think the reasoning for this is that the VID calculations were triggered by the requests for VID.  What happens was the first nodes to request VID get back a share but the later nodes to request it are already calculating the VID due to the earlier requesters and don't try to request until the calculation finishes and when it does the request task cancels itself.

The sequence that could cause the bug, from the perspective of one were:

1. Primary Network (CDN) is down
2. DA Proposal received
3. DA vote sent
4. QuorumProposalRecv and validated
5. Vid Delayed Request spawned (request not yet sent because we are waiting out the delay)
6. DAC received
7. VID Reqesest Received from another node, VID calculation started
8. VID Caclulation finishes
9. VID delayed request task awakes after waiting the delay and is cancelled because we already have our VID share
10. View Times out because we (and others) didn't vote and the next leader could not form QC.

In this example the VID calculation could also be triggered if primary network was known down by the node. 

The fix for this is to just broadcast that we have the share if we end up cancelling the VID Task because we have the share.  This means there is no way for the vote dependency to not get the event if we got the share or calculated it.


crates/task-impls/src/consensus/helpers.rs
```diff
@@ -686,6 +686,7 @@ pub(crate) async fn fetch_proposal<TYPES: NodeType>(
     quorum_membership: Arc<TYPES::Membership>,
     consensus: Arc<RwLock<Consensus<TYPES>>>,
 ) -> Result<Leaf<TYPES>> {
+    tracing::info!("Fetchign proposal for view {:?}", view);
```
rob-maron, line 689: nit: maybe we should consider logs like these being debug

rob-maron (approved):

jparr721 (dismissed):
I’m getting that Friday feeling!!! So glad these are done…

crates/task-impls/src/consensus/helpers.rs
bfish713: My thinking was that this should be a relatively rare occurrence.  We should only get this if we miss a proposal and are catching back up but I'm not 100%

crates/task-impls/src/consensus/helpers.rs
```diff
@@ -686,6 +686,7 @@ pub(crate) async fn fetch_proposal<TYPES: NodeType>(
     quorum_membership: Arc<TYPES::Membership>,
     consensus: Arc<RwLock<Consensus<TYPES>>>,
 ) -> Result<Leaf<TYPES>> {
+    tracing::debug!("Fetchign proposal for view {:?}", view);
```
ss-es, line 689:
```suggestion
    tracing::debug!("Fetching proposal for view {:?}", view);
```

crates/task-impls/src/request.rs
```diff
@@ -391,15 +394,35 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> DelayedRequester<TYPES, I> {
...
+                .map(|shares| shares.get(&self.public_key).cloned())
+            {
+                broadcast_event(
+                    Arc::new(HotShotEvent::VidShareRecv(vid_share.clone())),
+                    &self.sender,
+                )
+                .await;
+            }
+            tracing::debug!(
+                "Cancleing vid request for view {:?}, cur view is {:?}",
```
ss-es, line 413:
```suggestion
                "Canceling vid request for view {:?}, cur view is {:?}",
```

crates/task-impls/src/request.rs
```diff
@@ -391,15 +394,35 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> DelayedRequester<TYPES, I> {
...
+            );
+        }
+        cancel
     }
 
     /// Transform a response into a `HotShotEvent`
     async fn handle_response_message(&self, message: SequencingMessage<TYPES>) {
         let event = match message {
             SequencingMessage::Da(DaConsensusMessage::VidDisperseMsg(prop)) => {
+                tracing::info!("vid req complete, gor vid {:?}", prop);
```
ss-es, line 425:
```suggestion
                tracing::info!("vid req complete, got vid {:?}", prop);
```

crates/task-impls/src/request.rs
```diff
@@ -253,6 +254,8 @@ struct DelayedRequester<TYPES: NodeType, I: NodeImplementation<TYPES>> {
     pub network: Arc<I::Network>,
     /// Shared state to check if the data go populated
     state: Arc<RwLock<Consensus<TYPES>>>,
```
ss-es, line 256: Nit: not really related to the PR, but could this be called `consensus` rather than `state` for consistency with the other tasks?

jparr721 (approved):

ss-es (approved):
Looks good to me, just some minor comments

crates/task-impls/src/request.rs
bfish713: Yeah it should be I want to follow up on this pr with one that handles the VID stuff a little smarter

ss-es (approved):

crates/task-impls/src/request.rs
ss-es: Sounds good! reapproved

shenkeyao (approved):
Looks great!

PR #3400, opened 2024-06-28
Add time-based configurability for upgrades

ss-es:
No linked issue.

### This PR: 
Allows setting time-based windows for upgrades, in addition to the existing view-based window. Times are specified in seconds since `SystemTime::UNIX_EPOCH`.

Nodes will check both the view *and* time before proposing or voting on an upgrade. To omit either check, the application can set e.g. the start and stop time to 0 and `u64::MAX` respectively for view-based upgrades, or vice-versa.

### DIFF
--- a/crates/task-impls/src/upgrade.rs
+++ b/crates/task-impls/src/upgrade.rs
@@ -1,4 +1,4 @@
-use std::{marker::PhantomData, sync::Arc};
+use std::{marker::PhantomData, sync::Arc, time::SystemTime};
 
 use anyhow::Result;
 use async_broadcast::{Receiver, Sender};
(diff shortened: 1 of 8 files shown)


crates/task-impls/src/upgrade.rs
```diff
@@ -211,35 +231,40 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> UpgradeTaskState<TYPES, I> {
...
+
                 // We try to form a certificate 5 views before we're leader.
-                if *view >= self.start_proposing_view
-                    && *view < self.stop_proposing_view
-                    && self.quorum_membership.leader(view + 5) == self.public_key
+                if view >= self.start_proposing_view
+                    && view < self.stop_proposing_view
+                    && time >= self.start_proposing_time
+                    && time < self.stop_proposing_time
+                    && self.quorum_membership.leader(TYPES::Time::new(view + 5)) == self.public_key
```
jparr721, line 252: Is there any chance this 5, 15, 20, etc could perhaps be constants? I know you have comments, but it looks like the +5 is used in a couple spots where the comment has to be hunted for. It can even just be at the top of this file. Perhaps something like `FORM_CERTIFICATE_OFFSET` or `UPGRADE_OFFSET` or `UPGRADE_INTERVAL`, and commensurately so for the +15 and +20 as something like `UPGRADE_BEGIN_OFFSET` and `UPGRADE_END_OFFSET` and `UPGRADE_DECIDE_BY_OFFSET`.

jparr721 (commented):
One request, if you're willing to throw it in there.

jbearer (dismissed):
LGTM

bfish713 (approved):

PR #3431, opened 2024-07-10
[Upgradability] - Remove duplicated fields for the decided upgrade cert in `ConsensusTaskState`

### LINKED ISSUE #3357: [Upgradability] - Remove duplicated fields for the decided upgrade cert in `ConsensusTaskState`

shenkeyao:
Closes #3357.

### This PR: 
* Removes the non-`Arc<RwLock<...>>` `decided_upgrade_cert` from the consensus task state.
* Replaces code using/updating `decided_upgrade_cert` of the task state with the `Arc<RwLock<...>>` `decided_upgrade_certificate` so that it can be shared globally.
* Deletes `temp_*` functions that were previously added to support the different types of upgrade certs, and replaces the use of them with the corresponding functions with the `Arc<RwLock<...>>` cert type.

### This PR does not: 
* Fix the version upgrade, which is a [separate task](https://github.com/EspressoSystems/HotShot/issues/3378).

### Key places to review: 
* Cert removals and replacements.
* (Testing files may be ignored since they only contain name changes.)

### How to test this PR:
* Run with `--features example-upgrade` and see if the upgrade happens around view 25.
* Upgrade task tests should also pass.


jparr721 (commented):
I think those `dontuse_*` fields are still hanging around, you wanna blow those away too while you're are it?

ss-es (approved):
Looks good to me!

PR #3437, opened 2024-07-12
[Upgradability] - Avoid tracking the version in task states

### LINKED ISSUE #3378: [Upgradability] - Avoid tracking the version in task states

shenkeyao:
Closes #3378.

### This PR: 
* Removes version from task states.
* Adds a helper function to dynamically calculate the version.
* Uses the new function to get the version when it's needed, not relying on the task state.

### This PR does not: 
* Change how the upgrade works except for removing the version field update.

### Key places to review: 
* `fn version` in `crates/types/src/simple_certificate.rs`.
* How `fn version` is called.

### How to test this PR:
* Run with --features example-upgrade and see if the upgrade happens around view 25 for both dependency and non-dependency tasks.
* Upgrade task tests should also pass.


crates/task-impls/src/consensus/handlers.rs
```diff
@@ -720,6 +716,11 @@ pub async fn update_state_and_vote_if_able<TYPES: NodeType, I: NodeImplementatio
         return false;
     };
     drop(read_consnesus);
+
+    let Ok(version) = version(view, &vote_info.1.read().await.clone()) else {
+        error!("Failed to calculate the version.");
```
ss-es, line 721: I think it would be better/more accurate to just match on the result and if it's `Err(e)` print `e` (which should have a more detailed message)

ss-es (approved):
Looks good to me!

bfish713 (approved):

Issue #3425, opened 2024-07-09
[COMBINED NETWORKS] - Fix memory leak


lukaszrzasik:
### What is this task and why do we need to work on it?

The memory leak is in `send_both_networks` in `combined_network.rs`. The secondary (libp2p) future may not be timing out on a failed message.

### What are the acceptance criteria to close this issue?

Best would be to create a test that can consistently reproduce the memory leak and then start working on the fix.

PR #3447, opened 2024-07-15
[Libp2p] Patches

rob-maron:
#3425 
### This PR: 

Includes various bug fixes to current `main`, including:
- Patching the `replication_factor` to be a sane value
- Allowing only a single concurrent lookup for the same key

Misc changes:
- Changes some log levels
- Better `num_failed_messages` counters

Fixes the memory leak on a long-running deployment.

### This PR does not: 

- Include the `try_send` patch, that was already merged to main.

### Key places to review: 

- `dht/mod.rs`

### DIFF
--- a/crates/hotshot/src/traits/networking/libp2p_network.rs
+++ b/crates/hotshot/src/traits/networking/libp2p_network.rs
@@ -800,10 +800,8 @@ impl<K: SignatureKey + 'static> ConnectedNetwork<K> for Libp2pNetwork<K> {
             .inner
             .handle
             .lookup_node(
-                &bincode::serialize(&recipient).map_err(|e| {
-                    self.inner.metrics.num_failed_messages.add(1);
-                    NetworkError::Libp2p { source: e.into() }
-                })?,
+                &bincode::serialize(&recipient)
+                    .map_err(|e| NetworkError::Libp2p { source: e.into() })?,
                 self.inner.dht_timeout,
             )
             .await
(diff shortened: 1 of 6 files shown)


crates/hotshot/src/traits/networking/libp2p_network.rs
```diff
@@ -664,16 +659,20 @@ impl<K: SignatureKey + 'static> Libp2pNetwork<K> {
...
-                info!("WAITING TO CONNECT ON NODE {:?}", id);
+
+                error!("Finished putting Kademlia records");
 
                 // Wait for the network to connect to the required number of peers
                 if let Err(e) = handle.wait_to_connect(4, id).await {
                     error!("Failed to connect to peers: {:?}", e);
                     return Err::<(), NetworkError>(e.into());
                 }
+                error!("Connected to required number of peers");
```
bfish713, line 670: `info!`

crates/libp2p-networking/src/network/node/handle.rs
```diff
@@ -143,27 +143,33 @@ impl NetworkNodeHandle {
...
-            info!(
-                "WAITING TO CONNECT, connected to {} / {} peers ON NODE {}",
-                num_connected, num_peers, node_id
+            if num_connected >= num_required_peers {
+                break;
+            }
+
+            // Log the number of connected peers
+            error!(
+                "Node {} connected to {}/{} peers",
```
bfish713, line 165: reduce from error?

bfish713 (commented):
Looks good, I think maybe there are some error logs that should be lower level.

crates/hotshot/src/traits/networking/libp2p_network.rs
```diff
@@ -664,16 +659,20 @@ impl<K: SignatureKey + 'static> Libp2pNetwork<K> {
                 {
                     async_sleep(Duration::from_secs(1)).await;
                 }
-                // perform connection
-                info!("WAITING TO CONNECT ON NODE {:?}", id);
+
+                error!("Finished putting Kademlia records");
```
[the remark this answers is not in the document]
rob-maron: Not really, but I wanted it to be loud since we had a problem with this. I'll reduce it though

PR #3454, opened 2024-07-16
[CATCHUP] Test all nodes restarting

bfish713:
Closes #<ISSUE_NUMBER>

### This PR: 

Should test that all nodes can go down and come back up without keeping their in memory state

### DIFF
--- a/crates/example-types/src/storage_types.rs
+++ b/crates/example-types/src/storage_types.rs
@@ -10,8 +10,10 @@ use hotshot_types::{
     consensus::CommitmentMap,
     data::{DaProposal, Leaf, QuorumProposal, VidDisperseShare},
     message::Proposal,
+    simple_certificate::QuorumCertificate,
     traits::{node_implementation::NodeType, storage::Storage},
     utils::View,
+    vote::HasViewNumber,
 };
 
 type VidShares<TYPES> = HashMap<
@@ -23,7 +25,8 @@ type VidShares<TYPES> = HashMap<
 pub struct TestStorageState<TYPES: NodeType> {
     vids: VidShares<TYPES>,
     das: HashMap<TYPES::Time, Proposal<TYPES, DaProposal<TYPES>>>,
-    proposals: HashMap<TYPES::Time, Proposal<TYPES, QuorumProposal<TYPES>>>,
+    proposals: BTreeMap<TYPES::Time, Proposal<TYPES, QuorumProposal<TYPES>>>,
+    high_qc: Option<hotshot_types::simple_certificate::QuorumCertificate<TYPES>>,
 }
 
 impl<TYPES: NodeType> Default for TestStorageState<TYPES> {
@@ -31,7 +34,8 @@ impl<TYPES: NodeType> Default for TestStorageState<TYPES> {
         Self {
             vids: HashMap::new(),
             das: HashMap::new(),
-            proposals: HashMap::new(),
+            proposals: BTreeMap::new(),
+            high_qc: None,
         }
     }
 }
@@ -49,6 +53,17 @@ impl<TYPES: NodeType> Default for TestStorage<TYPES> {
             inner: Arc::new(RwLock::new(TestStorageState::default())),
             should_return_err: false,
         }
+    }
+}
+
+impl<TYPES: NodeType> TestStorage<TYPES> {
+    pub async fn proposals_cloned(
+        &self,
+    ) -> BTreeMap<TYPES::Time, Proposal<TYPES, QuorumProposal<TYPES>>> {
+        self.inner.read().await.proposals.clone()
+    }
+    pub async fn high_qc(&self) -> Option<QuorumCertificate<TYPES>> {
+        self.inner.read().await.high_qc.clone()
     }
 }
 
@@ -104,10 +119,18 @@ impl<TYPES: NodeType> Storage<TYPES> for TestStorage<TYPES> {
 
     async fn update_high_qc(
         &self,
-        _high_qc: hotshot_types::simple_certificate::QuorumCertificate<TYPES>,
+        new_high_qc: hotshot_types::simple_certificate::QuorumCertificate<TYPES>,
     ) -> Result<()> {
         if self.should_return_err {
             bail!("Failed to update high qc to storage");
+        }
+        let mut inner = self.inner.write().await;
+        if let Some(ref current_high_qc) = inner.high_qc {
+            if new_high_qc.view_number() > current_high_qc.view_number() {
+                inner.high_qc = Some(new_high_qc);
+            }
+        } else {
+            inner.high_qc = Some(new_high_qc);
         }
         Ok(())
     }
--- a/crates/testing/src/spinning_task.rs
+++ b/crates/testing/src/spinning_task.rs
@@ -171,6 +171,68 @@ where
                                 node.handle.shut_down().await;
                             }
                         }
+                        UpDown::Restart => {
+                            let node_id = idx.try_into().unwrap();
+                            if let Some(node) = self.handles.write().await.get_mut(idx) {
+                                tracing::error!("Node {} shutting down", idx);
+                                node.handle.shut_down().await;
+
+                                let storage = node.handle.storage().clone();
+                                let memberships = node.handle.memberships.clone();
+                                let config = node.handle.hotshot.config.clone();
+                                let auction_results_provider =
+                                    node.handle.hotshot.auction_results_provider.clone();
+                                let read_storage = storage.read().await;
+                                let initializer = HotShotInitializer::<TYPES>::from_reload(
+                                    self.last_decided_leaf.clone(),
+                                    TestInstanceState {},
+                                    None,
+                                    view_number,
+                                    read_storage.proposals_cloned().await,
+                                    read_storage.high_qc().await.unwrap_or(
+                                        QuorumCertificate::genesis(
+                                            &TestValidatedState::default(),
+                                            &TestInstanceState {},
+                                        )
+                                        .await,
+                                    ),
+                                    Vec::new(),
+                                    BTreeMap::new(),
+                                );
+                                // We assign node's public key and stake value rather than read from config file since it's a test
+                                let validator_config = ValidatorConfig::generated_from_seed_indexed(
+                                    [0u8; 32],
+                                    node_id,
+                                    1,
+                                    // For tests, make the node DA based on its index
+                                    node_id < config.da_staked_committee_size as u64,
+                                );
+                                let context = TestRunner::<TYPES, I, N>::add_node_with_config(
+                                    node_id,
+                                    node.network.clone(),
+                                    (*memberships).clone(),
+                                    initializer,
+                                    config,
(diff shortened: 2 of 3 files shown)


crates/hotshot/src/traits/networking/libp2p_network.rs
```diff
@@ -296,9 +295,22 @@ impl<TYPES: NodeType> TestableNetworkingImplementation<TYPES>
...
+                    let mut clear = false;
+                    if write_ids.contains(&node_id) {
+                        write_ids.clear();
+                        clear = true;
+                    }
+                    write_ids.insert(node_id);
+                    drop(write_ids);
+                    if clear {
+                        bootstrap_addrs_ref.write().await.clear();
+                    }
```
shenkeyao, line 313: Can we just move this `clear` call to line 306? Then we don't need the `clear` boolean.

crates/testing/tests/tests_2/catchup.rs
```diff
@@ -286,3 +286,63 @@ async fn test_catchup_reload() {
...
+        idx: i,
+        updown: UpDown::Restart,
+    })};
+
+    metadata.timing_data = timing_data;
+    metadata.start_nodes = 20;
+    metadata.num_nodes_with_stake = 20;
+
+    metadata.spinning_properties = SpinningTaskDescription {
+        // Start the nodes before their leadership.
```
shenkeyao, line 324: Nit: I think we should remove/update this comment.

shenkeyao (approved):
Only nit comments.

crates/hotshot/src/traits/networking/libp2p_network.rs
bfish713: Yeah I think so, was trying to rule out deadlocking by holding both the locks at once, but it wasn't the issue

crates/hotshot/src/traits/networking/libp2p_network.rs
```diff
@@ -251,7 +250,7 @@ impl<TYPES: NodeType> TestableNetworkingImplementation<TYPES>
                     TYPES::SignatureKey::generated_from_seed_indexed([0u8; 32], node_id).1;
                 let pubkey = TYPES::SignatureKey::from_private(&privkey);
                 // we want the majority of peers to have this lying around.
-                let replication_factor = NonZeroUsize::new(2 * expected_node_count / 3).unwrap();
+                let replication_factor = NonZeroUsize::new(expected_node_count).unwrap();
```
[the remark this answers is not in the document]
bfish713: Ah we don't actually need this change now, I was debugging why the libp2p network was not always connecting and thought replicating to all nodes might help, but the fix was to overwrite the bootstrap address

crates/testing/src/spinning_task.rs
```diff
@@ -148,6 +152,9 @@ where
                                         )
                                         .await
                                     }
+                                    LateNodeContext::Restart => {
+                                        panic!("Cannot spin up a node with Restart context")
```
jparr721, line 156: Just for my own understanding: late start is from genesis, right? So a node would delay startup for some period of time then catch up? Is that why restart wouldn't make sense for that node?

crates/testing/src/spinning_task.rs
```diff
@@ -171,6 +178,75 @@ where
                                 node.handle.shut_down().await;
                             }
                         }
+                        UpDown::Restart => {
+                            let node_id = idx.try_into().unwrap();
+                            error!("restarting node");
```
jparr721, line 183: Should this be an error?

crates/testing/tests/tests_2/catchup.rs
```diff
@@ -286,3 +286,63 @@ async fn test_catchup_reload() {
...
+    let mut catchup_nodes = vec![];
+    for i in 1..20 {
+        catchup_nodes.push(ChangeNode {
+        idx: i,
+        updown: UpDown::Restart,
+    })};
+
+    metadata.timing_data = timing_data;
+    metadata.start_nodes = 20;
+    metadata.num_nodes_with_stake = 20;
```
jparr721, line 321: I know we've historically had issues with all the nodes being staked masking issues. Should this number be some value less than the total so we can catch issues with staked nodes flaking out?

crates/testing/tests/tests_2/catchup.rs
```diff
@@ -286,3 +286,63 @@ async fn test_catchup_reload() {
...
+
+    metadata.spinning_properties = SpinningTaskDescription {
+        // Restart all the nodes in view 13
+        node_changes: vec![(13, catchup_nodes)],
+    };
+    metadata.view_sync_properties =
+        hotshot_testing::view_sync_task::ViewSyncTaskDescription::Threshold(0, 21);
```
jparr721, line 328: Any chance we could get a comment? At face value `Threshold(0, 21)` doesn't quite clarify what it's doing.

crates/testing/tests/tests_2/catchup.rs
```diff
@@ -286,3 +286,63 @@ async fn test_catchup_reload() {
...
+
+    async_compatibility_layer::logging::setup_logging();
+    async_compatibility_layer::logging::setup_backtrace();
+    let timing_data = TimingData {
+        next_view_timeout: 2000,
+        ..Default::default()
+    };
+    let mut metadata: TestDescription<TestTypes, CombinedImpl> = TestDescription::default();
+    let mut catchup_nodes = vec![];
+    for i in 1..20 {
```
jparr721, line 313: I notice we use `20` a couple times in this test. Could we perhaps have some constant like `RESTART_TEST_NUM_NODES` so that way if we decide to increase or decrease it, we don't accidentally miss one of the `20`'s?

crates/testing/tests/tests_2/catchup.rs
```diff
@@ -286,3 +286,63 @@ async fn test_catchup_reload() {
...
+
+    metadata.completion_task_description =
+        CompletionTaskDescription::TimeBasedCompletionTaskBuilder(
+            TimeBasedCompletionTaskDescription {
+                duration: Duration::from_secs(60),
+            },
+        );
+    metadata.overall_safety_properties = OverallSafetyPropertiesDescription {
+        // Make sure we keep committing rounds after the catchup, but not the full 50.
+        num_successful_views: 22,
```
jparr721, line 338: Why specifically 22 here?

crates/testing/tests/tests_2/catchup.rs
```diff
@@ -286,3 +286,63 @@ async fn test_catchup_reload() {
...
+        completion_task::{CompletionTaskDescription, TimeBasedCompletionTaskDescription},
+        overall_safety_task::OverallSafetyPropertiesDescription,
+        spinning_task::{ChangeNode, SpinningTaskDescription, UpDown},
+        test_builder::{TestDescription, TimingData},
+    };
+
+    async_compatibility_layer::logging::setup_logging();
+    async_compatibility_layer::logging::setup_backtrace();
+    let timing_data = TimingData {
+        next_view_timeout: 2000,
```
jparr721, line 308: How come we don't use the default 4 seconds?

crates/example-types/src/storage_types.rs
```diff
@@ -52,6 +56,17 @@ impl<TYPES: NodeType> Default for TestStorage<TYPES> {
...
 
+impl<TYPES: NodeType> TestStorage<TYPES> {
+    pub async fn proposals_cloned(
+        &self,
+    ) -> BTreeMap<TYPES::Time, Proposal<TYPES, QuorumProposal<TYPES>>> {
+        self.inner.read().await.proposals.clone()
+    }
+    pub async fn high_qc_cloned(&self) -> Option<QuorumCertificate<TYPES>> {
+        self.inner.read().await.high_qc.clone()
+    }
```
jparr721, line 67: Not necessarily against the naming here, but would the `move`'d return value here not imply that the data was cloned?

jparr721 (commented):
Looks good, just some high level stuff.

crates/hotshot/src/lib.rs
```diff
@@ -191,15 +191,59 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> SystemContext<TYPES, I> {
...
+            external_chan,
+        )
+    }
+
+    /// Creates a new [`Arc<SystemContext>`] with the given configuration options.
+    ///
+    /// To do a full initialization, use `fn init` instead, which will set up background tasks as
+    /// well.
+    ///
+    /// # Errors
```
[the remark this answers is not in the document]
bfish713: This comment was just taken from `new` it's basically the same function, I can add some details

crates/testing/src/spinning_task.rs
bfish713: Yeah late start is a node that never is turned on, restart is a node that is stopped then restarted with only info from `storage`

crates/testing/tests/tests_2/catchup.rs
bfish713: Oh I think it should just be 20, tbh I wasn't 100% sure what it's doing but checked and it meas an amount of nodes form 0-20 can enter view sycn.  I can make a comment

crates/testing/tests/tests_2/catchup.rs
bfish713: It can be any number really but this should be enough that every node is leader 1 time.  Actually since the chain rule fix it maybe should be 23

crates/testing/tests/tests_2/catchup.rs
bfish713: Takes too long haha and I'm just copying from the catchup test

crates/example-types/src/storage_types.rs
bfish713: I guess kind of, I took the naming convention from something I saw in `async_broadcast` but that's pretty flimsy.  I like it here because it tells the caller for sure a real clone is happening.  With this interface I guess it's impossible for much else to happen considering the types are not `Copy` but when you just read `storage.proposals().get()` you might think it's returning a ref

crates/libp2p-networking/src/network/node.rs
```diff
@@ -768,7 +768,7 @@ impl NetworkNode {
                         event = self.swarm.next() => {
                             debug!("peerid {:?}\t\thandling maybe event {:?}", self.peer_id, event);
                             if let Some(event) = event {
-                                info!("peerid {:?}\t\thandling event {:?}", self.peer_id, event);
+                                debug!("peerid {:?}\t\thandling event {:?}", self.peer_id, event);
```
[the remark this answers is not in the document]
bfish713: These are specific libp2p events

crates/testing/src/spinning_task.rs
bfish713: Also here we have to panic because we just don't have the info to create another node

crates/testing/src/spinning_task.rs
bfish713: removing

jparr721 (approved):

Issue #3455, opened 2024-07-16
[Frontend] - Make changes to allow for sequencer `RollCall`


rob-maron:
### What is this task and why do we need to work on it?

@Ayiga would like there to be a way for the sequencer to receive respond to `RollCall` events. This requires some changes in HotShot

### What are the acceptance criteria to close this issue?

The sequencer, with changes, would be able to see `RollCall` events from HotShot and respond to them through the CDN

PR #3457, opened 2024-07-16
[Frontend] External message passthrough

rob-maron:
Related to #3455 .

This PR is the Hotshot integration for an external message passthrough. We need this to allow for the sequencers to communicate about identity information across the networks that are internal to HotShot. It could be used in the future for other things.

# This PR:
- Creates an external message type as a thin wrapper over bytes
- Makes it so that messages received with the external type are directly broadcasted as events to the external stream

# Other options:
If we don't like the route, we could do something else. 

We could also:
- Change the CDN to allow multiple connections from the same key, and then connect the sequencer directly. Could be a bit wasteful since we already have a connection open.
- Allow for filtering of messages directly from `ConnectedNetwork`s. We would then be able to call `recv()` in the sequencer. This will require the most work, we would need message duplication and probably some broadcast queue.

Let me know your thoughts 

# To look at:
- `crates/task-impls/src/network.rs` -> the message passthrough logic

### DIFF
--- a/crates/task-impls/src/network.rs
+++ b/crates/task-impls/src/network.rs
@@ -178,13 +178,12 @@ impl<TYPES: NodeType> NetworkMessageTaskState<TYPES> {
                     }
                 },
 
-                MessageKind::Sequencer(data) => {
-                    // Send the sequencer message to the event stream so the sequencer can
-                    // see it
+                MessageKind::External(data) => {
+                    // Send the external message to the external event stream so it can be processed
                     broadcast_event(
                         Event {
                             view_number: TYPES::Time::new(1),
-                            event: EventType::SequencerMessage(data),
+                            event: EventType::ExternalMessageReceived(data),
                         },
                         &self.external_event_stream,
                     )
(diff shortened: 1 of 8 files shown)


crates/task-impls/src/network.rs
```diff
@@ -174,12 +177,24 @@ impl<TYPES: NodeType> NetworkMessageTaskState<TYPES> {
                         warn!("Request and Response messages should not be received in the NetworkMessage task");
                     }
                 },
+
+                MessageKind::External(data) => {
+                    // Send the external message to the external event stream so it can be processed
+                    broadcast_event(
+                        Event {
+                            view_number: TYPES::Time::new(1),
```
jparr721, line 185: How come this is view 1?

rob-maron: It's arbitrary, it doesn't necessarily need a view number so I set it to one; I followed what `ResponseMessage::Denied` was doing. It might make sense to use the current view but I don't want it to cause confusion

crates/types/src/message.rs
```diff
@@ -172,6 +174,8 @@ pub enum MessageKind<TYPES: NodeType> {
     Consensus(SequencingMessage<TYPES>),
     /// Messages relating to sharing data between nodes
     Data(DataMessage<TYPES>),
+    /// A (still serialized) message to be passed through to external listeners
+    External(Vec<u8>),
```
rob-maron, line 178: This shouldn't break versioning since it is additive but I will have to make sure

PR #3464, opened 2024-07-18
Add `Libp2p` connection limits

rob-maron:
No linked issue.

### This PR: 

Limits the number of simultaneous [pending/established] connections that Libp2p can use.

### This PR does not: 

Break anything at the protocol level. This is all client-side so it should be backwards compatible.

### Key places to review: 

The configured values at `node.rs`

### DIFF
--- a/crates/libp2p-networking/src/network/node.rs
+++ b/crates/libp2p-networking/src/network/node.rs
@@ -20,6 +20,7 @@ use futures::{channel::mpsc, select, FutureExt, SinkExt, StreamExt};
 use hotshot_types::constants::KAD_DEFAULT_REPUB_INTERVAL_SEC;
 use libp2p::{
     autonat,
+    connection_limits::ConnectionLimits,
     core::transport::ListenerId,
     gossipsub::{
         Behaviour as Gossipsub, ConfigBuilder as GossipsubConfigBuilder, Event as GossipEvent,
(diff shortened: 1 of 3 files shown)


crates/libp2p-networking/src/network/node.rs
```diff
@@ -299,13 +300,25 @@ impl NetworkNode {
                 ..Default::default()
             };
 
+            // Limit the number of simultaneous connections
+            let connection_limits = libp2p::connection_limits::Behaviour::new(
+                ConnectionLimits::default()
+                    .with_max_established(Some(100))
+                    .with_max_established_incoming(Some(70)) // To prevent eclipse attacks
+                    .with_max_established_outgoing(Some(100))
+                    .with_max_established_per_peer(Some(1)) // Just reuse the one we already have
```
bfish713, line 309: I'm not sure but we might want this to be 2 or more, I don't know what is considered a connection but it seems like 1 incoming and one outgoing connection will be established at least.  Idk the docs don't make it clear

crates/libp2p-networking/src/network/node.rs
```diff
@@ -299,13 +300,25 @@ impl NetworkNode {
                 ..Default::default()
             };
 
+            // Limit the number of simultaneous connections
+            let connection_limits = libp2p::connection_limits::Behaviour::new(
+                ConnectionLimits::default()
+                    .with_max_established(Some(100))
+                    .with_max_established_incoming(Some(70)) // To prevent eclipse attacks
```
bfish713, line 307: This might be trickier than we expect.  We might actually expect an incoming connection from every participant in the case of voting when the CDN is down.  We'll have to check to make sure that when we block some connection they can eventually get through.

crates/libp2p-networking/src/network/node.rs
bfish713: Looked through the code I think this needs to be 2 so we can establish and incoming and outgoing connection at the same time.

bfish713 (commented):
Because these values are set so high our tests won't catch them, I suggest we test this with the 1000 node benchmark with both libp2p and crashed CDN.  I'm particularly worried about voting not working

crates/libp2p-networking/src/network/node.rs
rob-maron: Yeah, even with the CDN being up this would be a problem

PR #3472, opened 2024-07-22
Add `Bundle` struct and request bundles in transaction task

### LINKED ISSUE #3375: [CX-Marketplace] - Obtain Solver Results and Construct Block with Fees

TODO

_No response_

_No response_

TODO

_No response_

### LINKED ISSUE #3381: [CX-Marketplace] - Add Support for Multiple Version Paths From `AuctionResults`

HotShot needs to be able to accept multiple fees for a block to allow for multiple builders to submit bundles with requisite fees as a result of the Solver’s allocation result. Because of this, we’ll need to extend the flow of the Transactions task to support in-loop web requests depending on which version of the code we’re running. 

The existing flow in the Transactions task is as follows:

```rust
HotShotEvent::ViewChange() => {
...
	let block = {
	    if self
	        .decided_upgrade_certificate
	        .as_ref()
	        .is_some_and(|cert| cert.upgrading_in(block_view))
	    {
	        None
	    } else {
	        self.wait_for_block().await
	    }
	};
	
	if let Some(BuilderResponses {
	    block_data,
	    blocks_initial_info,
	    block_header,
	}) = block
	{
		// Broadcast the block
		...
	} else {
	    // If we couldn't get a block, send an empty block
	    warn!(
	        "Failed to get a block for view {:?}, proposing empty block",
	        view
	    );
			// Construct a null block
			...
	};
...
}
```

This will be one of the first blocks of versioned code that we interact with. In this, we’ll adjust the wait logic to instead query the solver service via the Sequencer by leveraging the versioning check (defined below) to execute the proper code path depending on what version of hotshot we’re interacting with.

```rust
HotShotEvent::ViewChange() => {
...
	let blocks = if self.version >= Marketplace {
			let results = self.auction_results_client.fetch_auction_result(
				block_view
			).await?;
			let bundles = self.query_all_builders(urls).await?;
			
			// Build the block
			Some(bundles.map(|b|b.url()))
	} else {
	    Some(vec1![self.wait_for_block().await])
	};
...
}
```
An easy way to construct blocks is to take the bundles from the Builders as a list of transactions instead of a list of bytes and hotshot could compress this down into bytes and construct the `Arc<[u8]>` type from there.

VID Precompute won’t be already present, we’ll be calculating this ourselves right before we submit the BlockRecv event.
- Need to do that async spawn operation that we do in other spots to ensure that we don’t block out when doing this.
- A HotShot function function exists for this already.

For the remaining fields, we can construct the `BlockRecv` field from the new proposed payload type. So all of the requisite information should be present.

_No response_

_No response_

The block event is completely implemented with all version support and testing.

_No response_

ss-es:
Closes #3381 
Closes #3375 

### This PR: 
- Adds the `Bundle` struct
- Adds a `fetch_bundles` method to the `AuctionResultsProvider` trait that retrieves a `Bundle` from all builders that win a given auction.
- Adds logic in the transactions task to call `fetch_bundles` and construct a `PackedBundle` from multiple builders depending on the version of HotShot in effect.
- Removes the `bid_fees` field from `PackedBundle`.

### Key places to review: 
- `task-impls/src/transactions.rs`
- `types/src/traits/auction_results_provider.rs`

### DIFF
--- a/crates/task-impls/src/transactions.rs
+++ b/crates/task-impls/src/transactions.rs
@@ -233,6 +233,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> TransactionTaskState<TYPES, 
         return None;
     }
 
+    #[allow(clippy::too_many_lines)]
     /// marketplace view change handler
     pub async fn handle_view_change_marketplace(
         &mut self,
@@ -295,7 +296,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> TransactionTaskState<TYPES, 
                     TYPES::BlockPayload::from_transactions(
                         transactions,
                         &validated_state,
-                        &self.instance_state.clone(),
+                        &Arc::clone(&self.instance_state),
                     )
                     .await,
                 ) {
@@ -356,7 +357,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> TransactionTaskState<TYPES, 
         )
         .await;
 
-        return None;
+        None
     }
 
     /// main task event handler
--- a/crates/task-impls/src/transactions.rs
+++ b/crates/task-impls/src/transactions.rs
@@ -77,9 +77,6 @@ pub struct Bundle<TYPES: NodeType> {
     /// The signature over the bundle.
     pub signature: TYPES::SignatureKey,
 
-    /// The fee for submitting a bid.
-    pub bid_fee: BuilderFee<TYPES>,
-
     /// The fee for sequencing
     pub sequencing_fee: BuilderFee<TYPES>,
 }
@@ -170,18 +167,11 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> TransactionTaskState<TYPES, 
             precompute_data,
         }) = block
         {
-            let Some(bid_fee) = null_block::builder_fee(self.membership.total_nodes(), version)
-            else {
-                error!("Failed to get bid fee");
-                return None;
-            };
-
             broadcast_event(
                 Arc::new(HotShotEvent::BlockRecv(PackedBundle::new(
                     block_payload.encode(),
                     metadata,
                     block_view,
-                    vec1::vec1![bid_fee],
                     vec1::vec1![fee],
                     precompute_data,
                 ))),
@@ -221,7 +211,6 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> TransactionTaskState<TYPES, 
                     vec![].into(),
                     metadata,
                     block_view,
-                    vec1::vec1![null_fee.clone()],
                     vec1::vec1![null_fee],
                     Some(precompute_data),
                 ))),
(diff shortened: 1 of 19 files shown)


crates/task-impls/src/transactions.rs
```diff
@@ -120,9 +120,245 @@ pub struct TransactionTaskState<TYPES: NodeType, I: NodeImplementation<TYPES>> {
...
+            {
+                let Some(null_fee) =
+                    null_block::builder_fee(self.membership.total_nodes(), version)
+                else {
+                    error!("Failed to get null fee");
+                    return None;
+                };
+
+                let mut sequencing_fees = Vec::new();
+                let mut bid_fees: Vec<BuilderFee<TYPES>> = Vec::new();
```
elliedavidson, line 279: We'll want to remove the `bid_fee` variable eventually.

crates/task-impls/src/transactions.rs
```diff
@@ -120,9 +120,245 @@ pub struct TransactionTaskState<TYPES: NodeType, I: NodeImplementation<TYPES>> {
...
+                    )
+                    .await;
+
+                    return None;
+                }
+            }
+        }
+
+        // If we couldn't get any bundles, send an empty block
+        warn!(
```
elliedavidson, line 321: Another failure case is when the solver never returns any results.  That should also cause an empty block to be proposed.

crates/types/src/bundle.rs
```diff
@@ -0,0 +1,21 @@
...
+pub struct Bundle<TYPES: NodeType> {
+    /// The bundle transactions sent by the builder.
+    pub transactions: Vec<<TYPES::BlockPayload as BlockPayload<TYPES>>::Transaction>,
+
+    /// The signature over the bundle.
+    pub signature: TYPES::SignatureKey,
+
+    /// The fee for sequencing
+    pub sequencing_fee: BuilderFee<TYPES>,
+}
```
elliedavidson, line 21: This struct looks good!

crates/types/src/traits/auction_results_provider.rs
```diff
@@ -23,9 +24,31 @@ pub trait HasUrls {
 pub trait AuctionResultsProvider<TYPES: NodeType>: Send + Sync + Clone {
     /// The AuctionSolverResult is a type that holds the data associated with a particular solver
     /// run, for a particular view.
-    type AuctionResult: HasUrls;
+    type AuctionResult: HasUrls + Send;
 
     /// Fetches the auction result for a view. Does not cache the result,
     /// subsequent calls will invoke additional wasted calls.
     async fn fetch_auction_result(&self, view_number: TYPES::Time) -> Result<Self::AuctionResult>;
```
elliedavidson, line 31: In both `fetch_auction_result` and `fetch_bundles` we'll want a timeout.  If we get the results we are looking for, we return immediately.  Otherwise, we return whatever results we get by the timeout (for both the auction results and bundles) .  I think that timeout logic still needs to be implemented here.

ss-es: Added timeouts! Added constants for the durations, and the `fetch_bundles` method now has timeouts on both `fetch_auction_result` and the GET request to each builder!

crates/task-impls/src/transactions.rs
ss-es: Clarified the comment -- this was already the case in the logic!

crates/task-impls/src/transactions.rs
ss-es: Removed!

elliedavidson (approved):

PR #3480, opened 2024-07-23
Marketplace header upgrades

elliedavidson:
Closes #<ISSUE_NUMBER>

### This PR: 

This PR adds new functions to the `BlockHeader` trait to accommodate the marketplace.  To enforce versioning the sequencer and HotShot will need to sure they call the appropriate `new()` function.  Two functions were added: `new_marketplace`, which is equivalent to `new_legacy`, and `get_auction_results`. 

### This PR does not: 

This PR does not implement version gating for the marketplace. 

### Key places to review: 

`block_contents.rs`


ss-es (approved):
Still not super familiar with the marketplace spec, but all looks good to me!

PR #3483, opened 2024-07-24
Topic refactor

rob-maron:
No linked issue

### This PR: 

Refactors all networks to broadcast messages through `Topics` instead of a static list of all recipients. This will be needed to allow operation with users dynamically leaving and joining the network.

### This PR does not: 

Make any breaking protocol changes

### Key places to review: 

`libp2p_network.rs`, `memory_network.rs`

### DIFF
--- a/crates/hotshot/src/traits/networking/libp2p_network.rs
+++ b/crates/hotshot/src/traits/networking/libp2p_network.rs
@@ -45,7 +45,7 @@ use hotshot_types::{
     traits::{
         election::Membership,
         metrics::{Counter, Gauge, Metrics, NoMetrics},
-        network::{self, ConnectedNetwork, NetworkError, ResponseMessage},
+        network::{self, ConnectedNetwork, NetworkError, ResponseMessage, Topic},
         node_implementation::{ConsensusTime, NodeType},
         signature_key::SignatureKey,
     },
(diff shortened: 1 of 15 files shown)


crates/hotshot/src/traits/networking/libp2p_network.rs
```diff
@@ -931,20 +932,22 @@ impl<K: SignatureKey + 'static> ConnectedNetwork<K> for Libp2pNetwork<K> {
...
-                    source: Box::new(NetworkNodeHandleError::NoSuchTopic),
-                }
-            })?
-            .clone();
 
-        // gossip doesn't broadcast from itself, so special case
-        if recipients.contains(&self.inner.pk) {
-            // send to self
+        // If the topic existed,
+        if let Some(topic_subscribers) = topic_map.get_by_right(&topic) {
```
bfish713, line 940: Can we flip the topic map now so the topic is the key and recipients are value.  Or better yet can we remove it and just saved the topics we are subscribed to handle the short circuit?

crates/types/src/traits/network.rs
```diff
@@ -274,7 +274,7 @@ pub trait ConnectedNetwork<K: SignatureKey + 'static>: Clone + Send + Sync + 'st
     async fn broadcast_message(
         &self,
         message: Vec<u8>,
-        recipients: BTreeSet<K>,
+        topic: Topic,
         broadcast_delay: BroadcastDelay,
     ) -> Result<(), NetworkError>;
```
bfish713, line 279: Why did broadcast DA not also change?

bfish713 (changes requested):
Requesting changes to have broadcast da also accept a topic. Also given the calls to the network are already split, do we even need to pass a topic in as an argument to the connected network?   Either way `broadcast_message` and `broadcast_da_message` should have the same signature, or be combined again

crates/types/src/traits/network.rs
rob-maron: DA broadcast is a direct send to all recipients instead of an actual broadcast

rob-maron: I think we should remove `da_broadcast_message` and conditionally on the topic have the DA specific logic when needed (as in Libp2p)

bfish713 (approved):

Issue #3462, opened 2024-07-18
[Audit] - Fetch Blocks Properly when Leader Consecutively


bfish713:
### What is this task and why do we need to work on it?

The transactions task has logic to detect missed views and fetch a block for the next view or the current view.  However we could be the leader for both the current view and the next view, in that case we'll only get a block for the first view we lead

### What are the acceptance criteria to close this issue?

Fix the issue and write a test where a node leads twice after a skipped view.

PR #3491, opened 2024-07-25
[Audit] - Fetch Blocks Properly when Leader Consecutively

lukeiannucci:
Closes #[3462](https://github.com/EspressoSystems/HotShot/issues/3462)

### This PR: 

Fixes a bug where a node could be the leader for the current and next view. Currently we only propose block for current view.

### DIFF
--- a/crates/hotshot/src/traits/election/static_committee_leader_two_views.rs
+++ b/crates/hotshot/src/traits/election/static_committee_leader_two_views.rs
@@ -66,7 +66,8 @@ where
     )))]
     /// Index the vector of public keys with the current view number
     fn leader(&self, view_number: TYPES::Time) -> PUBKEY {
-        // this will break other test cases
+        // two connsecutive views will have same index starting with even number.
+        // eg 0->1, 2->3 ... 10->11 etc
         let index = usize::try_from((*view_number / 2) % self.all_nodes_with_stake.len() as u64).unwrap();
         let res = self.all_nodes_with_stake[index].clone();
         TYPES::SignatureKey::public_key(&res)
--- a/crates/hotshot/src/traits/election/static_committee_leader_two_views.rs
+++ b/crates/hotshot/src/traits/election/static_committee_leader_two_views.rs
@@ -1,0 +1,215 @@
+use std::{marker::PhantomData, num::NonZeroU64};
+
+use ethereum_types::U256;
+// use ark_bls12_381::Parameters as Param381;
+use hotshot_types::traits::signature_key::StakeTableEntryType;
+use hotshot_types::{
+    signature_key::BLSPubKey,
+    traits::{election::Membership, node_implementation::NodeType, signature_key::SignatureKey},
+    PeerConfig,
+};
+#[cfg(feature = "randomized-leader-election")]
+use rand::{rngs::StdRng, Rng};
+use tracing::debug;
+
+/// Dummy implementation of [`Membership`]
+
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+pub struct StaticCommitteeLeaderForTwoViews<T, PUBKEY: SignatureKey> {
+    /// All the nodes participating and their stake
+    all_nodes_with_stake: Vec<PUBKEY::StakeTableEntry>,
+    /// The nodes on the static committee and their stake
+    committee_nodes_with_stake: Vec<PUBKEY::StakeTableEntry>,
+    /// builder nodes
+    committee_nodes_without_stake: Vec<PUBKEY>,
+    /// the number of fixed leader for gpuvid
+    fixed_leader_for_gpuvid: usize,
+    /// Node type phantom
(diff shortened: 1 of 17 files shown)


crates/hotshot/src/traits/election/static_committee_leader_two_views.rs
```diff
@@ -0,0 +1,217 @@
...
+{
+    /// Clone the public key and corresponding stake table for current elected committee
+    fn committee_qc_stake_table(&self) -> Vec<PUBKEY::StakeTableEntry> {
+        self.committee_nodes_with_stake.clone()
+    }
+
+    #[cfg(not(any(
+        feature = "randomized-leader-election",
+        feature = "fixed-leader-election"
+    )))]
```
bfish713, line 66: I don't think it makes sense to have these features inside this impl.  I think this should just apply to the entire struct, or not have the feature gating at all

crates/hotshot/src/traits/election/static_committee_leader_two_views.rs
```diff
@@ -0,0 +1,217 @@
...
+    fn leader(&self, view_number: TYPES::Time) -> PUBKEY {
+        // two connsecutive views will have same index starting with even number.
+        // eg 0->1, 2->3 ... 10->11 etc
+        let index =
+            usize::try_from((*view_number / 2) % self.all_nodes_with_stake.len() as u64).unwrap();
+        let res = self.all_nodes_with_stake[index].clone();
+        TYPES::SignatureKey::public_key(&res)
+    }
+
+    #[cfg(feature = "fixed-leader-election")]
```
bfish713, line 77: See above comment, but given this impl is specifically for a certain leader election we should not be implementing the fixed leader version

crates/hotshot/src/traits/election/static_committee_leader_two_views.rs
```diff
@@ -0,0 +1,217 @@
...
+            || self.fixed_leader_for_gpuvid > self.all_nodes_with_stake.len()
+        {
+            panic!("fixed_leader_for_gpuvid is not set correctly.");
+        }
+        let index = usize::try_from(*view_number % self.fixed_leader_for_gpuvid as u64).unwrap();
+        let res = self.all_nodes_with_stake[index].clone();
+        TYPES::SignatureKey::public_key(&res)
+    }
+
+    #[cfg(feature = "randomized-leader-election")]
```
bfish713, line 91: Same as above

crates/task-impls/src/transactions.rs
```diff
@@ -360,30 +360,41 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> TransactionTaskState<TYPES,
...
+                let block_views = match (make_block, next_leader) {
+                    // case where we are next leader and current leader
+                    (true, true) => vec![self.cur_view, next_view],
+                    // current view leader
+                    (true, false) => vec![self.cur_view],
+                    // next view leader
+                    (false, true) => vec![next_view],
+                    (false, false) => {
+                        // return if we aren't the next leader or we skipped last view and aren't the current leader.
+                        debug!("Not next leader for view {:?}", self.cur_view);
```
bfish713, line 374: I'm not sure if it's better but this logic, and the for loop below could be 2 if statements:

```rust
if make_block {
    // make block for self.cur_view
}
if next_leader {
    // make block for next_view
}
```

At the very least I think we can get rid of this match statement and push the just like above

crates/task-impls/src/transactions.rs
```diff
@@ -360,30 +360,41 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> TransactionTaskState<TYPES,
...
+                        }
+                    };
+
+                    if version < MarketplaceVersion::VERSION {
+                        self.handle_view_change_legacy(&event_stream, block_view)
+                            .await;
+                    } else {
+                        self.handle_view_change_marketplace(&event_stream, block_view)
+                            .await;
+                    }
```
bfish713, line 397: If you switch to 2 if statments maybe we can refactor this into a simple helper like `handle_view_change(&self, event_stream: &Sender<..>, view: TYPES::Time)`

bfish713 (commented):
Overall looks good, just had a few suggestions to remove code, and maybe make the logic in the transaction task a bit easier to follow, but that's up to you

crates/task-impls/src/transactions.rs
lukeiannucci: agreed, this was cleaned up recently which i think switch to the above idea works better now.

crates/hotshot/src/traits/election/static_committee_leader_two_views.rs
lukeiannucci: yep i had it before i did some refactoring and making the test build_system_handle generic. now it is not needed

bfish713 (approved):

PR #3493, opened 2024-07-26
[Marketplace] Builder API adjustments

QuentinI:
Closes #0000

### This PR: 
- Adjusts bundle type signature
- Adds /bundle to marketplace builder API
- Removes old endpoints from marketplace builder API
- Removes dead code in transaction task
- Lifts fetching bundles from `AuctionResultsProvider` to transaction task, as it needs to use the builder client, which isn't available in `hotshot-types`

### This PR does not: 
- Implement full bundle-fetching algorithm analogous to legacy block-claiming algorithm
- Implement test builders for new builder API


crates/task-impls/src/transactions.rs
```diff
@@ -240,11 +240,36 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> TransactionTaskState<TYPES,
...
+
+                let mut bundles = Vec::new();
+
+                for bundle in join_all(futures).await {
+                    match bundle {
+                        Ok(Ok(b)) => bundles.push(b),
+                        _ => continue,
+                    }
+                }
+
```
elliedavidson, line 272: This section looks good to me

elliedavidson (approved):

crates/task-impls/src/transactions.rs
ss-es: Looks fine to me too, though I think it would be nice to have this logic be a separate function. We can make an issue for that/do it later though.

QuentinI: Yeah, we can do it when implementing this part of the spec:
> Query all builders in parallel, have a timeout that is constrained by a max system timeout of 1 second, and a minimum set by 2/3 of the builders, then add 10% additional time for the remaining 1/3 to resolve.

QuentinI: Although I'm not opposed to factoring it out now, won't take long.

ss-es (approved):
I'm not particularly familiar with this portion of the code, but as far as I can tell it looks good to me!

PR #3495, opened 2024-07-26
Lr/split validation

### LINKED ISSUE #3494: [DEPENDENCY_REFACTOR] - Split quorum proposal validation

The task is to split the quorum proposal validation. We want to split it into smaller parts, preferably the first part would preliminary checks allowing spoofed proposals to be early discarded.

Analyse validation process and split it in logical parts.

Use the preliminary checked proposal in Quorum Proposal Task in dependency tasks instead of non-validated proposal.

Spoofed proposals should no longer trigger quorum proposal dependency tasks.

_No response_

lukaszrzasik:
Closes #3494

### This PR: 

Adds a new `HotShotEvent::QuorumProposalPreliminarilyValidated`. This event indicates that the received quorum proposal has been preliminarily validated. This means that the quorum proposal 
i. is not for an old view,
ii. has been correctly signed by the leader of the current view,
iii. includes a valid justify QC.
The `HotShotEvent::QuorumProposalPreliminarilyValidated` is received by the `QuorumProposalTaskState` instead of `HotShotEvent::QuorumProposalRecv`. It protects the `QuorumProposalTaskState` against spoofed quorum proposals.

Additionally:
i. Moves the signature check from `validate_proposal_safety_and_liveness` to `validate_proposal_view_and_certs`.
ii. Broadcasts `HighQcUpdated` instead of `UpdateHighQc` at the end of `handle_quorum_proposal_recv` because it is more accurate and saves us from a redundant shared consensus state update.

### This PR does not: 

This PR allows dependency tasks-based consensus to progress when flooded with spoofed quorum proposals but the progression is much slower than normal.

### Key places to review: 

crates/task-impls/src/helpers.rs
crates/task-impls/src/quorum_proposal_recv/handlers.rs
crates/task-impls/src/quorum_proposal/mod.rs

### How to test this PR:

All tests should pass.
The change has been tested with `demo-native` and one of the nodes sending spoofed quorum proposals. The consensus progresses correctly. Only the faulty leader's views timeout.


crates/task-impls/src/helpers.rs
```diff
@@ -500,9 +486,18 @@ pub fn validate_proposal_view_and_certs<TYPES: NodeType>(
...
+    // There is a mistake here originating in the genesis leaf/qc commit. This should be replaced by:
+    //
+    //    proposal.validate_signature(&quorum_membership)?;
+    //
+    // in a future PR.
+    let proposed_leaf = Leaf::from_quorum_proposal(&proposal.data);
     ensure!(
-        view_leader_key == *sender,
-        "Leader key does not match key in proposal"
+        view_leader_key.validate(&proposal.signature, proposed_leaf.commit().as_ref()),
```
jparr721, line 499: This `validate_signature` method exists now. Should we just use that?

crates/task-impls/src/events.rs
```diff
@@ -191,6 +191,13 @@ pub enum HotShotEvent<TYPES: NodeType> {
 
     /// A new high_qc has been updated in `Consensus`.
     HighQcUpdated(QuorumCertificate<TYPES>),
+
+    /// A quorum proposal has been preliminarily validated.
+    /// The preliminary checks include:
+    /// 1. The proposal is not for an old view
+    /// 2. The proposal has been correctly signed by the leader of the current view
+    /// 3. The justify QC is valid
+    QuorumProposalPreliminarilyValidated(Proposal<TYPES, QuorumProposal<TYPES>>),
```
jparr721, line 200: Do you think it'd be helpful to elucidate how this event is different from the `QuorumProposalValidated` event? Unfortunately, it looks like the `QuorumProposalValidated` event's doc comment doesn't give much information right now, so an unfamiliar reader might not be able to trivially differentiate the two events without tracing the validation paths.

jparr721 (commented):
Great work, just a couple things I'd like your opinion on.

bfish713 (dismissed):

crates/task-impls/src/events.rs
```diff
@@ -84,6 +84,11 @@ pub enum HotShotEvent<TYPES: NodeType> {
     /// All dependencies for the quorum vote are validated.
     QuorumVoteDependenciesValidated(TYPES::Time),
     /// A quorum proposal with the given parent leaf is validated.
+    /// The full validation checks include:
```
jparr721, line 87: Love this!

crates/task-impls/src/helpers.rs
```diff
@@ -499,11 +484,14 @@ pub fn validate_proposal_view_and_certs<TYPES: NodeType>(
...
     );
 
-    let view_leader_key = quorum_membership.leader(view);
-    ensure!(
-        view_leader_key == *sender,
-        "Leader key does not match key in proposal"
-    );
+    // Validate the proposal's signature. This should also catch if the leaf_commitment does not equal our calculated parent commitment
+    //
+    // There is a mistake here originating in the genesis leaf/qc commit. This should be replaced by:
```
jparr721, line 489: Looks like we might be able to remove this comment entirely? Looks like we can keep the above, but the "there is a mistake here" part seems to be rectified.

jparr721 (commented):
Looks good, one last little thing

jparr721 (approved):

PR #3508, opened 2024-07-29
[AUDIT][Med/Low Severity] Fix race condition in state update for block building

jparr721:
Closes #<ISSUE_NUMBER>

### This PR: 
Hotshot initiates a block build request when a node is the leader of the next view. This process is handled via the `wait_for_block` function call. However, a race condition exists concerning the timing of reads and writes to the `validated_state_map`.

When `validate_proposal_safety_and_liveness` is called, it can update the `validated_state_map` after the `ViewChange` from the `update_view` function is processed. This results in `wait_for_block` reading stale state as the parent during the while loop lookup. 

This PR fixes the issue by moving the `update_view` call to happen *after* the validated state update has occurred and it also adds another call site to the liveness branch.

### DIFF
--- a/crates/testing/tests/tests_1/quorum_proposal_recv_task.rs
+++ b/crates/testing/tests/tests_1/quorum_proposal_recv_task.rs
@@ -80,7 +80,6 @@ async fn test_quorum_proposal_recv_task() {
     )]];
 
     let expectations = vec![Expectations::from_outputs(vec![
-        exact(ViewChange(ViewNumber::new(2))),
         exact(UpdateHighQc(proposals[1].data.justify_qc.clone())),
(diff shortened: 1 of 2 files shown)


crates/testing/tests/tests_1/quorum_proposal_recv_task.rs
```diff
@@ -95,6 +94,7 @@ async fn test_quorum_proposal_recv_task() {
             proposals[1].data.clone(),
             leaves[0].clone(),
         )),
+        exact(ViewChange(ViewNumber::new(2))),
```
jparr721, line 97: This is a serial output, so this test changed but the liveness test did not need to.

PR #3509, opened 2024-07-29
Fix CDN-only example

rob-maron:
Fixes the CDN-only example by creating a random file in a temporary directory instead of the current.


elliedavidson (approved):
:)

Issue #3515, opened 2024-07-30
[Byzantine testing] - Add Byzantine Test where we have a dishonest leader


lukeiannucci:
### What is this task and why do we need to work on it?

Write a Byzantine integration test where a leader is being dishonest by replacing the QC it is proposing with a QC from an older view 

### What are the acceptance criteria to close this issue?

Integration test that mocks a dishonest leader

PR #3516, opened 2024-07-30
Add Byzantine Test where we have a dishonest leader

lukeiannucci:
Closes #[3515](https://github.com/EspressoSystems/HotShot/issues/3515)

### This PR: 

Uses byzantine trait in testing framework to create a dishonest leader that will replace proposed QC with a QC from a previous view

### Key places to review: 

`tests/tests_1/test_with_failures_2.rs` I added the integration test at the bottom to test dishonest leader

### DIFF
--- a/crates/hotshot/src/tasks/mod.rs
+++ b/crates/hotshot/src/tasks/mod.rs
@@ -25,6 +25,8 @@ use hotshot_task_impls::{
 };
 use hotshot_types::{
     constants::EVENT_CHANNEL_SIZE,
+    data::QuorumProposal,
+    message::Proposal,
     message::{Messages, VersionedMessage},
     traits::{
         network::ConnectedNetwork,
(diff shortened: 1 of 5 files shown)


crates/hotshot/src/tasks/mod.rs
```diff
@@ -364,6 +366,83 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> EventTransformerState<TYPES,
...
+    fn handle_proposal_send_event(
+        &self,
+        event: &HotShotEvent<TYPES>,
+        proposal: &Proposal<TYPES, QuorumProposal<TYPES>>,
+        sender: &TYPES::SignatureKey,
+    ) -> Vec<HotShotEvent<TYPES>> {
+        let dishonest = self.cached_events.len() == 2;
+        if !dishonest {
+            return vec![event.clone()];
+        }
```
bfish713, line 411: IIUC this means this leader will send 1 dishonest proposal on the second time it proposes.

Also you don't need a variable for `dishonest`

lukeiannucci: yes, that is what it currently does. the second time this node node proposes it will send old qc

crates/hotshot/src/tasks/mod.rs
```diff
@@ -408,6 +410,91 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> EventTransformerState<TYPES,
...
+    pub total_proposals_from_node: u64,
+    /// which proposals to be dishonest at
+    pub dishonest_at_proposal_numbers: HashSet<u64>,
+    /// how far back to look for a QC
+    pub view_look_back: usize,
+    /// phantom
+    pub _phantom: std::marker::PhantomData<I>,
+}
+
+/// add trait to handle proposal events for a dishonest leader
```
jparr721, line 428: Could we adjust the punctuation and wording here to more clearly explain this?

crates/hotshot/src/tasks/mod.rs
```diff
@@ -408,6 +410,91 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> EventTransformerState<TYPES,
...
+trait DishonestProposal<TYPES: NodeType, I: NodeImplementation<TYPES>> {
+    /// send dishonest proposal
```
jparr721, line 430: Same here, this comment and function name somewhat clash. It looks like it does more than just send a single dishonest proposal.

crates/hotshot/src/tasks/mod.rs
```diff
@@ -408,6 +410,91 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> EventTransformerState<TYPES,
...
+            .dishonest_at_proposal_numbers
+            .contains(&self.total_proposals_from_node)
+            || length == 0
+        {
+            return vec![event.clone()];
+        }
+
+        // grab qc from specified look back
+        let proposal_from_lookback = if length - 1 < self.view_look_back {
+            // if look back is to far just take the first proposal
```
jparr721, line 459: nit: "too"

crates/hotshot/src/tasks/mod.rs
```diff
@@ -408,6 +410,91 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> EventTransformerState<TYPES,
...
+
+impl<TYPES: NodeType, I: NodeImplementation<TYPES>> DishonestProposal<TYPES, I>
+    for DishonestLeader<TYPES, I>
+{
+    fn handle_proposal_send_event(
+        &self,
+        event: &HotShotEvent<TYPES>,
+        proposal: &Proposal<TYPES, QuorumProposal<TYPES>>,
+        sender: &TYPES::SignatureKey,
+    ) -> Vec<HotShotEvent<TYPES>> {
```
jparr721, line 447: Looks like this function returns a `vec` of HotShotEvents, but both instances we return a vec with a single value in it. Perhaps we can just return the flat event itself?

crates/testing/src/overall_safety_task.rs
```diff
@@ -195,6 +210,16 @@ impl<TYPES: NodeType, I: TestableNodeImplementation<TYPES>> TestTaskState
                 self.error = Some(Box::new(OverallSafetyTaskErr::<TYPES>::TooManyFailures {
                     failed_views: self.ctx.failed_views.clone(),
                 }));
+            } else if !expected_views_to_fail.is_empty()
+                && !expected_views_to_fail.contains(&view_number)
+            {
+                let _ = self.test_sender.broadcast(TestEvent::Shutdown).await;
```
jparr721, line 216: Looks like this exact block is duplicated. Could we perhaps make it a single helper or something? It lowers the refactor burden as someone could miss one of the copies.

jparr721 (dismissed):
Couple of comments

crates/hotshot/src/tasks/mod.rs
lukeiannucci: let me know if this is clearer

crates/testing/src/overall_safety_task.rs
lukeiannucci: refactored

bfish713:
I see that your adding the expected views to fail but I think it's missing the most important part.  The check you have just says it's ok to have a failed view if it's expected.  I think for byzantine tests we want the check to also make sure that we don't succeed in a view we expected to fail.  Maybe improving the safety task with this can be a follow up

jparr721 (commented):
LGTM, I think Brendon has a comment so I'll let him approve

bfish713 (approved):

PR #3518, opened 2024-07-30
[Builder] Fix last_vid_commitment

QuentinI:
Closes #0000

### This PR: 
We shouldn't ignore genesis view when fetching last known commitment.


jparr721 (approved):
When do we terminate in this case?

QuentinI:
We'll still terminate at genesis, either by returning it's commitment or by returning None from `checked_sub`.
The loop shouldn't really iterate too much in most circumstances, but even in some degenerate case where we have a long line of uninterrupted view failures up to genesis and are even somehow missing commitment for genesis itself we're still bounded by checked_sub returning None on target_view 0 and thus terminating the loop.

PR #3521, opened 2024-07-31
Vote and Cert validation

### LINKED ISSUE #3520: [AUDIT] - TIQV-1 Validate DA Certificate

`QuorumVoteTask` does not properly validate a received DA Certifiicate. The issue is described in [the audit results page](https://www.notion.so/espressosys/TIQV-1-3dfb7cdf7f434faa86c0ca3248234e1c?pvs=4)

_No response_

_No response_

All the existing tests must pass.
The POC described in [the audit results page](https://www.notion.so/espressosys/TIQV-1-3dfb7cdf7f434faa86c0ca3248234e1c?pvs=4) should no longer cause issues.

_No response_

lukaszrzasik:
Closes #3520

### This PR: 

This PR tries to improve a vote and a certificate validation. The existing problem is that both a view and a certificate include a view number. These view numbers are not signed in any way and can be easily spoofed by a malicious actor, e.g. a valid certificate can be received by a malicious leader which can then re-send the received certificate with changed view numbers. These will be accepted as valid certificates by the other nodes and can wreck havoc on the consensus.

The approach taken by this PR is following:

Changes in the `Vote` trait:
The `Vote` trait is now `Committable` on its own (not only its data). Explained below for `SimpleVote`.
`create_signed_vote` is now a trait associated function.

Changes to `SimpleVote`:
`vote_commitment` calculates the commitment for the whole vote (data + view number).
There's a new private associated function called `commit` which calculates the commit based on the data and the view number. This method is used in `create_signed_vote` and in the `commit` method.

Changes to `SimpleCertificate`:
The `vote_commitment` field is now private and should not be used in any way. There is a `new` associated function which created a certificate.
`vote_commitment` method calculates and returns the vote's commitment instead of returning the stored the commitment. This means the commitment is calculated from the data and the view number.
`is_valid_cert` now correctly checks the received signature against the locally calculated commitment which includes the vote's data and the view number.

### Key places to review: 

The crucial parts are in:
`is_valid_cert` method and `vote_commitment` method of the `SimpleCertificate`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -89,7 +89,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
 dependencies = [
  "cfg-if",
- "getrandom 0.2.15",
  "once_cell",
  "version_check",
  "zerocopy",
@@ -1318,7 +1317,7 @@ checksum = "41c270e7540d725e65ac7f1b212ac8ce349719624d7bcff99f8e2e488e8cf03f"
 [[package]]
 name = "cdn-broker"
 version = "0.4.0"
-source = "git+https://github.com/EspressoSystems/Push-CDN?tag=0.4.4#398b77fb2f461218a611294e1484b7e0bd571d64"
+source = "git+https://github.com/EspressoSystems/Push-CDN?tag=0.4.5#f6cc7c2fc53eaa52a4901e775d9be7ac820af72c"
 dependencies = [
  "async-std",
  "cdn-proto",
@@ -1342,7 +1341,7 @@ dependencies = [
 [[package]]
 name = "cdn-client"
 version = "0.4.0"
-source = "git+https://github.com/EspressoSystems/Push-CDN?tag=0.4.4#398b77fb2f461218a611294e1484b7e0bd571d64"
+source = "git+https://github.com/EspressoSystems/Push-CDN?tag=0.4.5#f6cc7c2fc53eaa52a4901e775d9be7ac820af72c"
 dependencies = [
  "async-std",
  "cdn-proto",
@@ -1357,7 +1356,7 @@ dependencies = [
 [[package]]
 name = "cdn-marshal"
 version = "0.4.0"
-source = "git+https://github.com/EspressoSystems/Push-CDN?tag=0.4.4#398b77fb2f461218a611294e1484b7e0bd571d64"
+source = "git+https://github.com/EspressoSystems/Push-CDN?tag=0.4.5#f6cc7c2fc53eaa52a4901e775d9be7ac820af72c"
 dependencies = [
  "async-std",
  "cdn-proto",
@@ -1371,7 +1370,7 @@ dependencies = [
 [[package]]
 name = "cdn-proto"
 version = "0.4.0"
-source = "git+https://github.com/EspressoSystems/Push-CDN?tag=0.4.4#398b77fb2f461218a611294e1484b7e0bd571d64"
+source = "git+https://github.com/EspressoSystems/Push-CDN?tag=0.4.5#f6cc7c2fc53eaa52a4901e775d9be7ac820af72c"
 dependencies = [
  "anyhow",
  "ark-serialize",
@@ -2867,9 +2866,9 @@ dependencies = [
 
 [[package]]
 name = "hashlink"
-version = "0.8.4"
+version = "0.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e8094feaf31ff591f651a2664fb9cfd92bba7a60ce3197265e9482ebe753c8f7"
+checksum = "6ba4ff7128dee98c7dc9794b6a411377e1404dba1c97deb8d1a55297bd25d8af"
 dependencies = [
  "hashbrown 0.14.5",
 ]
@@ -2916,9 +2915,6 @@ name = "heck"
 version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "95505c38b4572b2d910cecb0281560f54b440a19336cbbcb27bf6ce6adc6f5a8"
-dependencies = [
- "unicode-segmentation",
-]
 
 [[package]]
 name = "heck"
@@ -4743,9 +4739,9 @@ dependencies = [
 
 [[package]]
 name = "libsqlite3-sys"
-version = "0.27.0"
+version = "0.28.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cf4e226dcd58b4be396f7bd3c20da8fdee2911400705297ba7d2d7cc2c30f716"
+checksum = "0c10584274047cb335c23d3e61bcef8e323adae7c5c8c760540f73610177fc3f"
 dependencies = [
  "cc",
  "pkg-config",
@@ -6929,6 +6925,9 @@ name = "smallvec"
 version = "1.13.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3c5e1a9a646d36c3599cd173a41282daf47c44583ad367b8e6837255952e5c67"
+dependencies = [
+ "serde",
+]
 
 [[package]]
 name = "smartcow"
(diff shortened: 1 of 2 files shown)


crates/types/src/simple_certificate.rs
```diff
@@ -70,7 +70,7 @@ pub struct SimpleCertificate<TYPES: NodeType, VOTEABLE: Voteable, THRESHOLD: Thr
     /// The data this certificate is for.  I.e the thing that was voted on to create this Certificate
     pub data: VOTEABLE,
     /// commitment of all the votes this cert should be signed over
-    pub vote_commitment: Commitment<VOTEABLE>,
+    pub data_commitment: Commitment<VOTEABLE>,
```
ss-es, line 73: We may not be able to change this, but I'm not completely sure. I would lean towards keeping the old name out of caution.

This doesn't affect the bincode serialization of the struct, but I believe it will change the JSON serialization. I'm not sure what repositories use JSON (if any), though I think the sequencer at the very least has a test that might catch this.

jparr721: Yeah I think @jbearer mentioned a few weeks ago that changing the schema will cause downstream changes. Are we still in the "try not to change public interfaces" mode? Or are we past that.

crates/task-impls/src/consensus/mod.rs
```diff
@@ -375,11 +375,11 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> ConsensusTaskState<TYPES, I>
 
                     debug!(
                         "Attempting to publish proposal after forming a TC for view {}",
-                        *qc.view_number
+                        *qc.view_number()
```
jparr721, line 378: You're a hero for making these little fixes!!

bfish713 (commented):
I've been trying to wrap my head around this and I think there is some room to clear things up.  I keep getting confused on the various usages of `vote.commit()`, and `vote_commentment()` as well as where we use the "data commitment".  I think maybe we can remove one or the other, and possibly the `Committable` bound from a few places and that would help clear things up.

crates/types/src/simple_certificate.rs
```diff
@@ -71,11 +71,12 @@ impl<TYPES: NodeType> Threshold<TYPES> for UpgradeThreshold {
...
 #[derive(Serialize, Deserialize, Eq, Hash, PartialEq, Debug, Clone)]
-pub struct SimpleCertificate<TYPES: NodeType, VOTEABLE: Voteable, THRESHOLD: Threshold<TYPES>> {
+pub struct SimpleCertificate<TYPES: NodeType, VOTE: Vote<TYPES>, THRESHOLD: Threshold<TYPES>> {
     /// The data this certificate is for.  I.e the thing that was voted on to create this Certificate
-    pub data: VOTEABLE,
+    pub data: VOTE::Data,
     /// commitment of all the votes this cert should be signed over
-    pub vote_commitment: Commitment<VOTEABLE>,
+    /// Do not use this field, the commitment needs to be calculated instead
+    vote_commitment: Commitment<VOTE>,
```
bfish713, line 79: Can we keep this type the same as it was before, since we don't need it anyway: `Commitment<VOTE::Data>`

crates/types/src/simple_vote.rs
```diff
@@ -138,26 +138,24 @@ impl<TYPES: NodeType, DATA: Voteable + 'static> Vote<TYPES> for SimpleVote<TYPES
...
 
-    fn date(&self) -> &DATA {
+    fn data(&self) -> &DATA {
         &self.data
     }
 
-    fn date_commitment(&self) -> Commitment<DATA> {
-        self.data.commit()
+    fn vote_commitment(&self) -> Commitment<Self> {
+        self.commit()
```
bfish713, line 146: Can we just remove if we can just call commit() anyway?

crates/types/src/vote.rs
```diff
@@ -90,33 +105,39 @@ type SignersMap<COMMITMENT, KEY> = HashMap<
 /// Accumulates votes until a certificate is formed.  This implementation works for all simple vote and certificate pairs
 pub struct VoteAccumulator<
     TYPES: NodeType,
-    VOTE: Vote<TYPES>,
-    CERT: Certificate<TYPES, Voteable = VOTE::Commitment>,
+    VOTE: Vote<TYPES> + Committable,
```
bfish713, line 108: Adding the `Commitable` bound shouldn't be needed

crates/types/src/vote.rs
```diff
@@ -90,33 +105,39 @@ type SignersMap<COMMITMENT, KEY> = HashMap<
...
+        CERT: Certificate<TYPES, Vote = VOTE>,
+    > VoteAccumulator<TYPES, VOTE, CERT>
 {
     /// Add a vote to the total accumulated votes.  Returns the accumulator or the certificate if we
     /// have accumulated enough votes to exceed the threshold for creating a certificate.
     pub fn accumulate(&mut self, vote: &VOTE, membership: &TYPES::Membership) -> Either<(), CERT> {
         let key = vote.signing_key();
 
-        let vote_commitment = vote.date_commitment();
+        let vote_commitment = vote.vote_commitment();
```
bfish713, line 135: `vote.commit()`?

crates/types/src/vote.rs
```diff
@@ -90,33 +105,39 @@ type SignersMap<COMMITMENT, KEY> = HashMap<
...
+    pub signers: SignersMap<Commitment<VOTE>, TYPES::SignatureKey>,
     /// Phantom data to specify the types this accumulator is for
     pub phantom: PhantomData<(TYPES, VOTE, CERT)>,
 }
 
-impl<TYPES: NodeType, VOTE: Vote<TYPES>, CERT: Certificate<TYPES, Voteable = VOTE::Commitment>>
-    VoteAccumulator<TYPES, VOTE, CERT>
+impl<
+        TYPES: NodeType,
+        VOTE: Vote<TYPES> + Committable + Clone,
```
bfish713, line 126: Same do we need `Committable`

crates/types/src/simple_vote.rs
```diff
@@ -138,26 +138,24 @@ impl<TYPES: NodeType, DATA: Voteable + 'static> Vote<TYPES> for SimpleVote<TYPES
...
+    fn create_signed_vote(
         data: DATA,
         view: TYPES::Time,
         pub_key: &TYPES::SignatureKey,
         private_key: &<TYPES::SignatureKey as SignatureKey>::PrivateKey,
     ) -> Result<Self, <TYPES::SignatureKey as SignatureKey>::SignError> {
-        match TYPES::SignatureKey::sign(private_key, data.commit().as_ref()) {
+        match TYPES::SignatureKey::sign(
+            private_key,
+            <SimpleVote<TYPES, DATA> as Vote<TYPES>>::commit(&data, view).as_ref(),
```
bfish713, line 157: Does Self::commit() not work here?

crates/types/src/vote.rs
```diff
@@ -28,19 +28,34 @@ use crate::{
...
 
     /// Get the signature of the vote sender
     fn signature(&self) -> <TYPES::SignatureKey as SignatureKey>::PureAssembledSignatureType;
     /// Gets the data which was voted on by this vote
-    fn date(&self) -> &Self::Commitment;
-    /// Gets the Data commitment of the vote
-    fn date_commitment(&self) -> Commitment<Self::Commitment>;
+    fn data(&self) -> &Self::Data;
+    /// Gets the (Data + view number) commitment of the vote
+    fn vote_commitment(&self) -> Commitment<Self>;
```
bfish713, line 40: Same as other comment if `Vote` must be committable then this is redundant right?

crates/types/src/vote.rs
```diff
@@ -156,7 +177,7 @@ impl<TYPES: NodeType, VOTE: Vote<TYPES>, CERT: Certificate<TYPES, Voteable = VOT
 
         // TODO: Get the stake from the stake table entry.
         *total_stake_casted += stake_table_entry.stake();
-        total_vote_map.insert(key, (vote.signature(), vote.date_commitment()));
+        total_vote_map.insert(key, (vote.signature(), vote.vote_commitment()));
```
bfish713, line 180: I think now that we are calculating a commitment on this call we should not be calling `vote_commitment` repeatedly and just reuse the value already calculated above

crates/types/src/vote.rs
```diff
@@ -173,8 +194,8 @@ impl<TYPES: NodeType, VOTE: Vote<TYPES>, CERT: Certificate<TYPES, Voteable = VOT
             );
 
             let cert = CERT::create_signed_certificate(
-                vote.date_commitment(),
-                vote.date().clone(),
+                vote.commit(),
```
bfish713, line 197: Same here, we already calculated this value, lets not calculated it again

crates/types/src/simple_certificate.rs
```diff
@@ -94,22 +116,21 @@ impl<TYPES: NodeType, VOTEABLE: Voteable + Committable, THRESHOLD: Threshold<TYP
         };
         committable::RawCommitmentBuilder::new("Certificate")
             .field("data", self.data.commit())
-            .field("vote_commitment", self.vote_commitment)
```
ss-es, line 97: Can we leave this in? It would be easier, because we wouldn't have to do any version switch on the certificate

bfish713:
Can we close this now?

PR #3524, opened 2024-07-31
Tj/typos and ordering

0xkato:
### This PR: 
Non-issues from the review.

Correcting typos, variable ordering and removing some superfluous code.


jparr721 (approved):

bfish713 (changes requested):
Looks like there is a lint failure, can you fix that? and then this should be good to do

jparr721 (approved):

bfish713 (approved):

Issue #3463, opened 2024-07-18
[AUDIT] - Investigate Double View Update from View Sync task


bfish713:
### What is this task and why do we need to work on it?

We send two updates from the view sync task in this [code](https://github.com/EspressoSystems/HotShot/blob/4713af7704f88f38a8a69b708acd4677f76d8ff1/crates/task-impls/src/view_sync.rs#L603-L613)

### What work will need to be done to complete this task?

Try to only update to the most recent view from view sync and test what happens, if things break investigate why and fix it.

### What are the acceptance criteria to close this issue?

We only send the `ViewChange` Event for the view that view sync agreed on.

PR #3532, opened 2024-08-02
[AUDIT] - Investigate Double View Update from View Sync task

lukeiannucci:
Closes https://github.com/EspressoSystems/HotShot/issues/3463

### This PR: 

We have added logic to properly handle fetching blocks when a node is the the leader two views in a row (see https://github.com/EspressoSystems/HotShot/issues/3462). Now, we can remove the double update from View Sync and write a integration test to verify when view sync triggers after 2 failures that we fetch both blocks for the next leader who will lead 2 consecutive views

### DIFF
--- a/crates/testing/tests/tests_1/test_with_failures_2.rs
+++ b/crates/testing/tests/tests_1/test_with_failures_2.rs
@@ -10,6 +10,7 @@ use hotshot_testing::{
     completion_task::{CompletionTaskDescription, TimeBasedCompletionTaskDescription},
     spinning_task::{ChangeNode, SpinningTaskDescription, UpDown},
     test_builder::TestDescription,
+    view_sync_task::ViewSyncTaskDescription
 };
 use hotshot_types::data::ViewNumber;
 use hotshot_types::traits::node_implementation::ConsensusTime;
@@ -106,7 +107,6 @@ cross_tests!(
 cross_tests!(
     TestName: test_with_double_leader_failures,
     Impls: [MemoryImpl, Libp2pImpl, PushCdnImpl],
-    // this lets a node be a leader two views in a row
     Types: [TestConsecutiveLeaderTypes],
     Ignore: false,
     Metadata: {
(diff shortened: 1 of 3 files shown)


crates/testing/tests/tests_1/test_with_failures_2.rs
```diff
@@ -52,3 +52,39 @@ cross_tests!(
...
 );
+
+#[cfg(not(feature = "dependency-tasks"))]
+cross_tests!(
+    TestName: test_with_double_leader_failures,
+    Impls: [MemoryImpl, Libp2pImpl, PushCdnImpl],
+    // this lets a node be a leader two views in a row
+    Types: [TestConsecutiveLeaderTypes],
+    Ignore: false,
+    Metadata: {
```
bfish713, line 63: We have a view sync test task that will actually check to see if we hit view sync (or enforce that nodes don't enter view sync), maybe you can add that to this test to show that nodes must hit view sync

lukeiannucci: ok, in test im setting view sync properties threshold now

crates/testing/tests/tests_1/test_with_failures_2.rs
```diff
@@ -101,3 +102,48 @@ cross_tests!(
...
+        let view_spin_node_down = 5;
+        metadata.spinning_properties = SpinningTaskDescription {
+            node_changes: vec![(view_spin_node_down, dead_nodes)]
+        };
+
+        // node 3 is leader twice when we shut down
+        metadata.overall_safety_properties.num_failed_views = 2;
+        metadata.overall_safety_properties.expected_views_to_fail = HashMap::from([
+            // next views after turning node off
+            (ViewNumber::new(view_spin_node_down+1), false),
```
jparr721, line 137: Looks like the formatter didn't run here. Hit this with `rustfmt crates/**/*.rs` and it'll fix this file up.

jparr721 (approved):

Issue #3440, opened 2024-07-12
[Byzantine testing] - Make byzantine integration tests work with `tokio`


ss-es:
### What is this task and why do we need to work on it?

Currently, running byzantine nodes with `tokio` in integration tests causes timing issues during shutdown. While nodes function correctly, after shutdown consensus tasks get stuck in a loop trying to receive events from a closed channel, causing the test to hang.

### What are the acceptance criteria to close this issue?

Byzantine integration tests pass with `tokio`

shamb0:
Hi @ss-es

I am interested to pickup this task, Could you please assign it to me ?

Thanks

ss-es:
@shamb0 sorry, I didn't see this message at all until just now!

marking closed in #3535

shamb0:
> @shamb0 sorry, I didn't see this message at all until just now!

Thanks, `@ss-es`. No worries at all! I completely understand you're on a tight schedule.

PR #3535, opened 2024-08-05
fix:: Implement graceful shutdown for Byzantine integration tests in …

shamb0:
## This PR:

This PR addresses issue #3440 (https://github.com/EspressoSystems/HotShot/issues/3440) by refactoring the `SystemContextHandle` shutdown behavior. The changes implement a graceful shutdown process that works consistently across both `tokio` and `async-std` contexts.

## Key Changes

1. Introduced a new `shutdown_signal` field in `SystemContextHandle` to manage graceful shutdown for:
   - Event relay tasks (both send & receive) between consensus and network layers
   - Network message handler tasks

2. Refactored `EventTransformerState::add_network_tasks` to incorporate the new shutdown mechanism.

3. Modified `NetworkTaskRegistry::shutdown()` to use `join_all()` variant APIs instead of `abort()` variants, ensuring cleaner task exits.

## Validation

The changes have been tested on both tokio and async-std runtimes. Execution traces are available for review:

**tokio ::**

- [01-trace-tokio-byzantine-test-double-propose-vote.md](https://gist.github.com/shamb0/5cdb4496d2d9868c67c144fabd2ba03f)
- [02-trace-tokio-byzantine-test-multiple-bad-proposals.md](https://gist.github.com/shamb0/fc32dcb4d3609cf9115546d9cfbc4e8f)

**async-std ::**

- [03-trace-async-std-byzantine-test-double-propose-vote.md](https://gist.github.com/shamb0/8cf98204898528f05f27d9c86f2f82eb)
- [04-trace-async-std-byzantine-test-multiple-bad-proposals.md](https://gist.github.com/shamb0/00072c2472f151cb757902038f191406)

**This PR does not:**

**Key places to review:**

1. Implementation of the shutdown signal in `SystemContextHandle`
2. Integration of shutdown logic in event relay and network message handler tasks
3. Changes to `NetworkTaskRegistry::shutdown()` method

### DIFF
--- a/crates/hotshot/src/tasks/mod.rs
+++ b/crates/hotshot/src/tasks/mod.rs
@@ -1,3 +1,9 @@
+// Copyright (c) 2021-2024 Espresso Systems (espressosys.com)
+// This file is part of the HotShot repository.
+
+// You should have received a copy of the MIT License
+// along with the HotShot repository. If not, see <https://mit-license.org/>.
+
 //! Provides a number of tasks that run continuously
 
 /// Provides trait to create task states from a `SystemContextHandle`
--- a/crates/hotshot/src/tasks/mod.rs
+++ b/crates/hotshot/src/tasks/mod.rs
@@ -8,6 +8,7 @@ use async_broadcast::broadcast;
 use async_compatibility_layer::art::{async_sleep, async_spawn};
 use async_lock::RwLock;
 use async_trait::async_trait;
+use futures::FutureExt;
 use hotshot_task::task::Task;
 #[cfg(feature = "rewind")]
 use hotshot_task_impls::rewind::RewindTaskState;
@@ -26,8 +27,7 @@ use hotshot_task_impls::{
 use hotshot_types::{
     constants::EVENT_CHANNEL_SIZE,
     data::QuorumProposal,
-    message::Proposal,
-    message::{Messages, VersionedMessage},
+    message::{Messages, Proposal, VersionedMessage},
     traits::{
         network::ConnectedNetwork,
         node_implementation::{ConsensusTime, NodeImplementation, NodeType},
@@ -36,9 +36,10 @@ use hotshot_types::{
 use vbs::version::StaticVersionType;
 
 use crate::{
-    tasks::task_state::CreateTaskState, types::SystemContextHandle, ConsensusApi,
-    ConsensusMetricsValue, ConsensusTaskRegistry, HotShotConfig, HotShotInitializer,
-    MarketplaceConfig, Memberships, NetworkTaskRegistry, SignatureKey, SystemContext,
+    tasks::task_state::CreateTaskState,
+    types::{ShutdownSignal, SystemContextHandle},
+    ConsensusApi, ConsensusMetricsValue, ConsensusTaskRegistry, NetworkTaskRegistry, HotShotConfig,
+    HotShotInitializer, MarketplaceConfig, Memberships, SignatureKey, SystemContext,
 };
 
 /// event for global event stream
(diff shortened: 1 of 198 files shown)


crates/hotshot/src/types/handle.rs
```diff
@@ -170,14 +243,17 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES> + 'static> SystemContextHandl
...
         tracing::error!("Shutting down the network!");
         self.hotshot.network.shut_down().await;
+    
+        tracing::error!("Shutting down network tasks!");
+        self.network_registry.shutdown().await;
 
         tracing::error!("Shutting down consensus!");
         self.consensus_registry.shutdown().await;
+
+        tracing::error!("=== Shutting down node :: {} done!!! ===", self.hotshot.id);
```
jparr721, line 256: We probably don't need to double log the node shutdown.

crates/hotshot/src/types/handle.rs
```diff
@@ -161,6 +230,10 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES> + 'static> SystemContextHandl
 
     /// Shut down the the inner hotshot and wait until all background threads are closed.
     pub async fn shut_down(&mut self) {
+        tracing::error!("=== Shutting down node :: {} ===", self.hotshot.id);
```
jparr721, line 233: We probably don't need to double log the node shutdown.

crates/task/src/task.rs
```diff
@@ -170,14 +170,11 @@ impl NetworkTaskRegistry {
...
     pub async fn shutdown(&mut self) {
-        let handles = &mut self.handles;
-
-        while let Some(handle) = handles.pop() {
-            #[cfg(async_executor_impl = "async-std")]
-            handle.cancel().await;
-            #[cfg(async_executor_impl = "tokio")]
-            handle.abort();
-        }
+        let handles = std::mem::take(&mut self.handles);
```
jparr721, line 173: Nice

crates/hotshot/src/tasks/mod.rs
```diff
@@ -309,36 +326,57 @@ where
...
         // and broadcast the transformed events to the replacement event stream we just created.
+        let shutdown_signal = Arc::clone(&handle.shutdown_signal);
         let send_handle = async_spawn(async move {
+            let mut shutdown = Box::pin(shutdown_signal.wait_for_shutdown().fuse());
             loop {
-                if let Ok(msg) = original_receiver.recv().await {
-                    let mut state = state_out.write().await;
+                futures::select! {
+                    _ = shutdown => {
+                        tracing::error!("Shutting down relay send task !!!");
```
jparr721, line 335: We probably don't need all the exclamations here or in the below log.

crates/hotshot/src/tasks/mod.rs
```diff
@@ -101,42 +102,57 @@ pub fn add_network_message_task<
...
-                                continue;
+            if shutdown_signal.is_shutdown() {
+                tracing::error!("Shutting down network message task !!!");
+                break;
+            }
+            futures::select! {
+                    _ = shutdown => {
+                        tracing::error!("Shutting down network message task !!!");
+                        return;
+                    }
```
jparr721, line 117: It looks like in this particular case, we have to handle the signal event twice, could we not just keep this in the future select?

crates/hotshot/src/types/handle.rs
```diff
@@ -22,6 +29,65 @@ use tracing::instrument;
 
 use crate::{traits::NodeImplementation, types::Event, Memberships, SystemContext};
 
+/// A signal for coordinating shutdown across multiple tasks.
+///
+/// This struct provides methods to initiate a shutdown, check the shutdown status,
+/// and wait for a shutdown signal.
+pub struct ShutdownSignal {
+    event: Arc<OtherEvent>,
+    is_shutdown: Arc<AtomicBool>,
```
jparr721, line 38: Can we get doc comments here?

crates/hotshot/src/types/handle.rs
```diff
@@ -1,12 +1,19 @@
...
+    },
+    time::Duration,
+};
 
 use async_broadcast::{InactiveReceiver, Receiver, Sender};
 use async_compatibility_layer::art::{async_sleep, async_spawn};
 use async_lock::RwLock;
 #[cfg(async_executor_impl = "async-std")]
 use async_std::task::JoinHandle;
+use event_listener::Event as OtherEvent;
```
jparr721, line 16: Can we use a different name for this? Perhaps `EventListenerEvent`, I'd even be okay with the specificity of `ShutdownSignalEvent` since that's all we're using it for.

crates/hotshot/src/types/handle.rs
```diff
@@ -22,6 +29,65 @@ use tracing::instrument;
 
 use crate::{traits::NodeImplementation, types::Event, Memberships, SystemContext};
 
+/// A signal for coordinating shutdown across multiple tasks.
+///
+/// This struct provides methods to initiate a shutdown, check the shutdown status,
+/// and wait for a shutdown signal.
+pub struct ShutdownSignal {
```
jparr721, line 36: I am hesitant about the introduction of a bespoke shutdown signal since we have something like this already in the codebase. Perhaps @ss-es can give context on the byzantine tests specifically, but it seems like we might be able to somehow re-use that behavior? Especially considering that this type is bolted onto the SYstemContextHandle anyway.

jparr721 (changes requested):
Hey there! Thanks for your contribution to the codebase. I have left some cursory comments. At a high level, this change seems good but, I have some overall concerns about the implementation in the context of the rest of the codebase.

I think that, mechanically, I don't mind this decision to use a shutdown signal, but I think that the tricky point here is we now need to arduously copy the same `futures::select` block in every single spot, and it increases the overall burden of replacing or modifying the behavior. 

Since the bulk of the change for the signal handler is in the `mod.rs` file, perhaps we can find a way to wrap this at a higher scope to avoid needless copy-pasted code?

Lastly and, most importantly, however, seems to be that we might be masking a deeper issue by having multiple shutdown handlers. We have all of our registries that are shutting down with some issues in Tokio preventing a graceful termination. It seems almost as though we're fixing the water damage without addressing the leaky roof, metaphorically speaking. Perhaps it's worth a deeper investigation.

shamb0:
Hi @jparr721,

Thank you for your comments and for pointing out the potential duplicates. I have removed "ShutdownSignal" and reused the existing `EventDependency`. Additionally, Updated the "futures::select!" pattern in the Data Relay Task to align with the codebase's standard usage.

Please let me know if any further enhancements or corrections are needed.

crates/hotshot/src/types/handle.rs
ss-es: @shamb0 I think this looks good overall, though I agree with Jarred about the extra shutdown signal.

I think the preferred way to do this would be to listen on the `internal_event_stream` for a `HotShotEvent::Shutdown` (which is the shutdown signal for all tasks). I don't think the network message task has an internal event stream receiver, but you can pass it to the task state on creation in `tasks/mod.rs`.

ss-es: Oh, nevermind -- I see you already made this change! I'll look through the code and add some comments

crates/hotshot/src/tasks/mod.rs
```diff
@@ -207,6 +221,42 @@ pub async fn add_consensus_tasks<TYPES: NodeType, I: NodeImplementation<TYPES>>(
...
+    let event_dep = EventDependency::new(
+        event_stream,
+        Box::new(|e| matches!(e.as_ref(), HotShotEvent::Shutdown)),
+    );
+
+    // Create a future that completes when the EventDependency is satisfied
+    async move {
+        let _ = event_dep.completed().await;
+    }
+    .boxed()
```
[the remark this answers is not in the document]
bfish713: EventDependency is basically just this while loop except it returns the event at the end so the `_` here would alwasy just be `HotShotEvent::Shutdown`   I'd also prefer just doing the loop here since we don't actually need Event returned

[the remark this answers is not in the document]
shamb0: Thanks @ss-es, @bfish713, Implemented a simple manual loop to check for `HotShotEvent::Shutdown`. No longer relying on `EventDependency`.

shamb0:
Thanks @ss-es, Clippy warnings addressed. CI build should now pass. 
Ran cargo clippy checks with both Tokio and Async-std runners.

	```shell
		just async="tokio" cargo clippy --tests --examples
		just async="async-std" cargo clippy --tests --examples
	```

shamb0:
Two quick points:

1. Disregard changes in these files:
   - crates/testing/tests/tests_1/test_with_failures_2.rs
   - crates/testing/tests/tests_1/vote_dependency_handle.rs

2. These files were unintentionally included when merging 
to upstream main. I'll review my git merge configuration on 
my dev machine to prevent this in the future.

jparr721 (approved):

jparr721:
Thanks so much again for your great work!

shamb0:
> Thanks so much again for your great work!

Thankyou @jparr721

ss-es:
It looks like there's some merge conflicts @shamb0 

(if you can't fix them, I'm happy to do it later tonight or tomorrow if you don't mind me pushing to the branch)

shamb0:
> It looks like there's some merge conflicts @shamb0
> 
> (if you can't fix them, I'm happy to do it later tonight or tomorrow if you don't mind me pushing to the branch)

Thanks @ss-es: Branch conflicts have been resolved. The PR should now be ready for the Intake merger review.

ss-es:
merged -- thank you again for the PR!

PR #3538, opened 2024-08-05
[Auto Benchmarks] Precision adjustment of latency 

dailinsubjam:
Closes #<ISSUE_NUMBER>

### This PR: 
- Changes the latency precision from i64 to f64.
- Reduce error logging by increasing the frequency of calling`wait_for_fn_from_orchestrator`


crates/orchestrator/src/lib.rs
```diff
@@ -482,9 +480,9 @@ where
                 // Deal with the bench results from different nodes
                 let cur_metrics = self.bench_results.clone();
                 self.bench_results.avg_latency_in_sec = (metrics.avg_latency_in_sec
-                    * metrics.num_latency
-                    + cur_metrics.avg_latency_in_sec * cur_metrics.num_latency)
-                    / (metrics.num_latency + cur_metrics.num_latency);
+                    * (metrics.num_latency as f64)
+                    + cur_metrics.avg_latency_in_sec * (cur_metrics.num_latency as f64))
+                    / (metrics.num_latency + cur_metrics.num_latency) as f64;
```
jparr721, line 485: All this casting is a bit hard to follow. We first cast num_latency to f64, then cur_metrics' num_latency, then the combined operation of the both of them. Could we perhaps just cast them once as variables to make reading this statement easier? Right now a misplaced closing parenthesis could mess this whole statement up in a not-obvious way.

jparr721 (changes requested):
Some overall comments. At a high level, it seems like casting in each individual spot might create a bit of whack-a-mole whenever we need to update this. Avoiding instances of that type of logic might be a little cleaner.

jparr721:
Closing since I'll be re-doing benchmarks soon anyway, and this PR seems abandoned.

Issue #3526, opened 2024-08-01
 [AUDIT] - QPRM-1 Liveness Branch of Proposing Blocks Voting


bfish713:
### What is this task and why do we need to work on it?

See: https://www.notion.so/espressosys/QPRM-1-05b1081b13ba44c5ac3cd8ea05692581 for full desctiption.

### What work will need to be done to complete this task?

Fix the "Liveness" voting path

### What are the acceptance criteria to close this issue?

Write a test for this path and see that it works

PR #3542, opened 2024-08-06
[AUDIT][MEDIUM][QVM-1] Fix `VoteDependencyData` To Include Proper State

jparr721:
Closes #<ISSUE_NUMBER>

### This PR: 
The `VoteNow` task uses the wrong data in `VoteDependencyData` which uses the parent leaf instead of the proposal leaf, leading to a break in the chain in the vote for the received quorum proposal. 

This PR fixes that by using the proposal to make the leaf, instead of the parent leaf. This also fixes a couple linter errors in the linting on the dependency tasks branch.

### DIFF
--- a/crates/testing/tests/tests_1/vote_dependency_handle.rs
+++ b/crates/testing/tests/tests_1/vote_dependency_handle.rs
@@ -7,18 +7,13 @@ use async_compatibility_layer::art::async_timeout;
 use futures::StreamExt;
 use hotshot::tasks::task_state::CreateTaskState;
 use hotshot_example_types::node_types::{MemoryImpl, TestTypes};
-use hotshot_macros::{run_test, test_scripts};
 use hotshot_task_impls::{
     events::HotShotEvent::*,
     quorum_vote::{QuorumVoteTaskState, VoteDependencyHandle},
 };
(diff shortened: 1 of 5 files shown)


crates/testing/tests/tests_1/vote_dependency_handle.rs
```diff
@@ -99,6 +94,7 @@ async fn test_vote_dependency_handle() {
         let mut event_receiver = handle.internal_event_stream_receiver_known_impl();
         let view_number = ViewNumber::new(node_id);
 
+        #[allow(unused)]
```
jparr721, line 97: This is intentional, we make the dependency task, but it does not implement the typical `handle` method that other tasks have, so it just has to kind of sit here and read from the stream.

bfish713:
This looks fine to me, I think however the entire vote now path can go away.  I have been going through it in regards to [QPRM-1](https://github.com/EspressoSystems/HotShot/issues/3526) and it appears to me that this whole code path is not consistent with the current implementation.  In fact I think it's good that the path was effectively not able to be hit (partially because of what this PR fixes too) because we should not vote for any proposal that we don't have leaf of the prior proposal.

PR #3545, opened 2024-08-07
Consecutive Leader Proposal Fix

lukeiannucci:
Closes #<ISSUE_NUMBER>

### This PR: 

Fix a bug where we would not propose when a node is the leader for two views in a row.

### Key places to review: 

`src/quorum/proposal/mod.rs` in `handle`, removing the return statement if we fail to update the latest proposed view

### DIFF
--- a/crates/task-impls/src/quorum_proposal/mod.rs
+++ b/crates/task-impls/src/quorum_proposal/mod.rs
@@ -345,17 +345,6 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalTaskState<TYPE
             self.latest_proposed_view = new_view;
 
             return true;
-        } else if self.quorum_membership.leader(new_view) == self.public_key
-            && self.quorum_membership.leader(new_view + 1) == self.public_key
-        {
-            // this means we were the leader for the current view and for the next view we will be leader again
-            // there is no dependency task for the current view and we need to create one for the next view
-            // so we return true
-            debug!(
-                "Leader for connsecutive views, latest proposed view {}",
-                *self.latest_proposed_view,
-            );
-            return true;
         }
         false
     }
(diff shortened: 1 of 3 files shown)


crates/task-impls/src/quorum_proposal/mod.rs
```diff
@@ -344,6 +344,17 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> QuorumProposalTaskState<TYPE
...
+        } else if self.quorum_membership.leader(new_view) == self.public_key
+            && self.quorum_membership.leader(new_view + 1) == self.public_key
+        {
+            // this means we were the leader for the current view and for the next view we will be leader again
+            // there is no dependency task for the current view and we need to create one for the next view
+            // so we return true
+            debug!(
+                "Leader for connsecutive views, latest proposed view {}",
+                *self.latest_proposed_view,
+            );
```
bfish713, line 357: Wonder what other opinions on this are but to me it seems like the proposal task should just not short circuit if this function returns false, and the `create_dependency_task_if_new` fn should handle making sure we don't create multiples for the same view.  To me it looks like we already check `self.latest_proposed_view` before creating the dependency task so we probably don't need a special case in this function

jparr721: Yeah I think hard-coding for this could be a code smell. I feel like I'm not clear on the chain here. So if we're the leader for two views in a row, why does that stop us from proposing?

lukeiannucci: yea that logic wasnt needed. thanks for pointing that out i was just trying to make sure it got to the `create_dependency_task_if_new`. but removing the return statement will be enough, even if we dont update latest proposed view, because as brendon pointed out we already have some safety checks in if there are dependency tasks for a view.

jparr721 (approved):

PR #3548, opened 2024-08-07
[Libp2p] DHT overhaul

### LINKED ISSUE #1277: Authenticate the DHT
The DHT should be authenticated. This means that the staking key should sign the data stored. This includes both the public key mapping and the routing table. We want this to prevent eclipse attacks where all a nodes peers will be controlled by other nodes. This is the other half of https://github.com/EspressoSystems/HotShot/issues/678. which forces a more even DHT distribution. To this end:

- Check that this isn't already done by the kademlia implementation in libp2p-rust. I know libp2p-rust exposes a way to extract the peerid of the node(s) with the DHT entry. Maybe we can use that somehow.
- If not already implemented, fork the DHT protocol to add this.

rob-maron:
Closes #1277 

## This PR:

This PR does two things: 
### 1. Introduces namespacing to Libp2p DHT records

This allows for us to better organize the DHT. It also allows us to (in the future) use the authentication condition with different records. Since authenticating a record requires the record's key to be the staking public key, namespacing allows us to serve more than one authenticated record for a particular key. 

For example, our namespace called `Lookup` is responsible for `Staking key -> Libp2p key` lookups. We could just as easily add a second authenticated record type called `Url` that is responsible for mapping a node's `Staking key -> Query node URL`.

### 2. Introduces record authentication

Record authentication is the act of requiring the DHT value to be signed by the key. This is important to prevent DHT attacks where a malicious node could lie about another node's data. For example, a node could overwrite a lookup record like so: `Lookup[other_nodes_staking_key] = my_malicious_libp2p_key`. This could eventually lead to an eclipse attack where some nodes think all other nodes are available at that Libp2p key (which we use for messaging).  

There are now two record types, `Signed` and `Unsigned`. Unsigned records are not checked for validity and are automatically accepted. `Signed` records, when created, generate a signature over the value and the record's namespace (so they cannot be maliciously included in another namespace). 

For clarity, a signed record looks like this:
```
Key: serialized(RecordKey{namespace: "Lookup", key: "my_staking_public_key"})
Value: serialized(RecordValue::Signed{the_real_value: "my_libp2p_peer_id", signature: signature(the_real_value + namespace)}
```  

Record authentication is checked whenever a key is looked up (through `get_record`) or received (through `PutRecord`). This disallows malicious records from being replicated as well as retrieved from the local DHT.

## This PR does not:
- Deal with the routing table. I still have to look into this, I believe it is natively authenticated.


crates/libp2p-networking/src/network/behaviours/dht/mod.rs
```diff
@@ -470,7 +476,39 @@ impl DHTBehaviour {
...
+
+                        // If the record is signed by the correct key,
+                        if record_value.validate(&record_key) {
+                            // Store the record
+                            if let Err(err) = store.put(record.clone()) {
+                                warn!("Failed to store record: {:?}", err);
+                            }
+                        } else {
+                            warn!("Failed to validate record");
+                        }
```
bfish713, line 508: I don't think this is going to do what we want.  I'm fairly certain that the DHT is storing all records automatically when we get a put request.  I think this event is emitted to do something extra on the inbound request but at this point I think the record is likely already stored.

bfish713: Maybe an easy fix for this is to remove from store in any case where validation fails.  I still think we'd need to look at the impl to make sure store is the source we use to give results to get queires

rob-maron: I'm using this: https://docs.rs/libp2p/latest/libp2p/kad/enum.StoreInserts.html#variant.FilterBoth

Relevant code is here: https://github.com/libp2p/rust-libp2p/blob/4725f46f0d0e14f3ef4e88edc085729c1312dff6/protocols/kad/src/behaviour.rs#L1853-L1893

It basically allows for `PutRequest` middleware. I can make a test for rejection

bfish713: Or maybe we can use: https://docs.rs/libp2p/latest/libp2p/kad/enum.StoreInserts.html

bfish713: Oh lol I just linked it, I missed this part in our code, if you are using that code then why do we even need the store code here at all?

rob-maron: Right, it receives the request here, `FilterBoth` moves the onus for storage to the receiver of the event. So basically a rejection is "not adding it" after receiving the message at this step

bfish713: Sorry I should have read the full doc first, I get it now, I thought we'd have to pass a trait object or something into the libp2p impl but it's just a flag that stores nothing by default.  I guess the only thing to confirm is that the `RecordStore` is the source of truth for get queries (I think it is IIRC)

rob-maron: Yeah, the local store is `RecordStore`: https://github.com/EspressoSystems/HotShot/blob/b2f9ee7950087e0ce9d9c0fd2ca0a16deef10637/crates/libp2p-networking/src/network/behaviours/dht/mod.rs#L184

crates/hotshot/src/traits/networking/libp2p_network.rs
```diff
@@ -662,35 +686,19 @@ impl<K: SignatureKey + 'static> Libp2pNetwork<K> {
...
-                {
-                    async_sleep(Duration::from_secs(1)).await;
-                }
-
+                // Map our staking key to our Libp2p Peer ID so we can properly
+                // route direct messages
                 while handle
                     .put_record(
-                        &bincode::serialize(&handle.peer_id()).unwrap(),
-                        &bincode::serialize(&pk).unwrap(),
```
bfish713, line 684: Do you know why we had this reverse mapping in the DHT? Seems like it's fine to get rid of

bfish713 (commented):
Everything Looks good to me, I will go over the new record code one more time tomorrow morning then accept, unless someone else gives it another pass

crates/libp2p-networking/src/network/behaviours/dht/mod.rs
bfish713: This is our code, but I mean if it's not already in our store, does node receiving our get query retrieve the record from the store, or some other place?

crates/hotshot/src/traits/networking/libp2p_network.rs
rob-maron: No idea, which is why I got rid of it. We only ever use the DHT for the routing + lookup in the other direction

crates/libp2p-networking/src/network/behaviours/dht/mod.rs
bfish713: Looks like it uses the store so this is all good: https://docs.rs/libp2p-kad/0.45.3/src/libp2p_kad/behaviour.rs.html#2293

bfish713:
Also I agree that for the routing part of the table (PeerId -> to MultiAddr mapping) we don't need to do any auth.  The PeerId is unique between nodes honest or dishonest and all messages should be signed by the private key related to the PeerId.  I'd be quite shocked if somehow the DHT allowed you to submit records in the routing table with a PeerID that wasn't yours/signed by that PeerId.  So worst a dishonest part could do is add a ton of different routable PIDs that nobody every queries for because there would be no related HotShot key (after this PR).  Thats my understanding at least.

TLDR; we don't need to validate routing entries ourselves

rob-maron:
> Also I agree that for the routing part of the table (PeerId -> to MultiAddr mapping) we don't need to do any auth. The PeerId is unique between nodes honest or dishonest and all messages should be signed by the private key related to the PeerId. I'd be quite shocked if somehow the DHT allowed you to submit records in the routing table with a PeerID that wasn't yours/signed by that PeerId. So worst a dishonest part could do is add a ton of different routable PIDs that nobody every queries for because there would be no related HotShot key (after this PR). Thats my understanding at least.

Yeah, that's what I figured but I wanted to make sure. It's another one of those things that happens implicitly though so while I'm at it I'm going to double check

rob-maron:
Converting to draft while I resolve the CI issues

crates/hotshot/src/traits/networking/libp2p_network.rs
```diff
@@ -271,6 +287,7 @@ impl<TYPES: NodeType> TestableNetworkingImplementation<TYPES>
                             mesh_n: (expected_node_count / 2 + 3),
                         }))
                         .server_mode(true)
+                        .identity(libp2p_keypair)
```
bfish713, line 290: You probably know better, but what is this doing? could it be related to the failures in CI?

rob-maron: The solution is probably to cache the validation result so we don't re-run it for every `get_record` call

bfish713 (approved):

PR #3549, opened 2024-08-07
Add builder_commitment to new_marketplace

QuentinI:
Closes #0000

### This PR: 
Adds builder_commitment to `new_marketplace` of the `BlockHeader` trait. Builder uses the commitment to correlate quorum and DA proposals and there's no way to calculate it having just the header.

### DIFF
--- a/crates/example-types/src/block_types.rs
+++ b/crates/example-types/src/block_types.rs
@@ -245,6 +245,29 @@ pub struct TestBlockHeader {
     pub timestamp: u64,
 }
 
+impl TestBlockHeader {
+    fn new<TYPES: NodeType<BlockHeader = Self>>(
+        parent_leaf: &Leaf<TYPES>,
+        payload_commitment: VidCommitment,
+        builder_commitment: BuilderCommitment,
+    ) -> Self {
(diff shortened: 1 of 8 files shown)


crates/example-types/src/block_types.rs
```diff
@@ -245,6 +245,29 @@ pub struct TestBlockHeader {
     pub timestamp: u64,
 }
 
+impl TestBlockHeader {
+    fn new<TYPES: NodeType<BlockHeader = Self>>(
```
shenkeyao, line 249: Can we just put the implementation inside `new_legacy` as in the current code? I think in this way it's clearer whether the caller is using the legacy or marketplace version, and makes it less possible to misuse them.

shenkeyao (changes requested):
The change looks good. The comment is more of a code-style suggestion.

shenkeyao (approved):

crates/example-types/src/block_types.rs
shenkeyao: As discussed on Zulip this is actually good.

Issue #3456, opened 2024-07-16
[Testing] - Adding delays to trait functions the sequencer implements


dailinsubjam:
### What is this task and why do we need to work on it?

The delay is to simulate the asynchronous behaviour of the sequencer. In our tests the functions return instantly but when running with the sequencer they do various things, like store to a DB, or make a network request.

### What work will need to be done to complete this task?

Adding a random configurable delay to functions in: TestStorage, TestValidatedState, and maybe more places and adding a few tests that have nonzero delays. The configuration can just be part of the TestMetadata

### What are the acceptance criteria to close this issue?

Delay time is configurable in the testing.

PR #3556, opened 2024-08-08
Adding delays to trait functions the sequencer implements

lukeiannucci:
Closes https://github.com/EspressoSystems/HotShot/issues/3456

### This PR: 

Add support to add delays to functions that the sequencer implements, method calls currently return immediately as of now. We want to mock behavior where one of these trait functions may take more time to complete

### DIFF
--- a/crates/example-types/src/block_types.rs
+++ b/crates/example-types/src/block_types.rs
@@ -4,12 +4,10 @@
 // You should have received a copy of the MIT License
 // along with the HotShot repository. If not, see <https://mit-license.org/>.
 
-use async_compatibility_layer::art::async_sleep;
 use std::{
     fmt::{Debug, Display},
     mem::size_of,
     sync::Arc,
-    time::Duration,
 };
 
 use async_trait::async_trait;
@@ -24,7 +22,6 @@ use hotshot_types::{
     utils::BuilderCommitment,
     vid::{VidCommitment, VidCommon},
 };
-use rand::Rng;
 use serde::{Deserialize, Serialize};
 use sha3::{Digest, Keccak256};
 use snafu::Snafu;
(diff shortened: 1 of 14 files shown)


crates/example-types/src/block_types.rs
```diff
@@ -365,3 +376,24 @@ impl Committable for TestBlockHeader {
...
+                    let sleep_in_millis = rand::thread_rng().gen_range(
+                        settings.min_time_in_milliseconds..=settings.max_time_in_milliseconds,
+                    );
+
+                    async_sleep(Duration::from_millis(sleep_in_millis)).await;
+                }
+            }
+        }
+    }
+}
```
ss-es, line 399: It looks like most of this code is duplicated for the various types. If it's the same thing, I think `handle_async_delay` could have a default implementation in the trait, and you could have each type just implement e.g. `settings(DelayConfig) -> Option<DelaySettings>`

lukeiannucci: yes it is duplicated, i thought the same thing, but left it like this incase in the future we have more specific tests

lukeiannucci: thanks for pointing this out! i realized the code was duplicated but this default method will help clean it up!

crates/example-types/src/testable_delay.rs
```diff
@@ -0,0 +1,132 @@
...
+            2 => Some(SupportedTraitTypesForAsyncDelay::BlockHeader),
+            _ => {
+                assert_eq!(self.index, 3, "Need to ensure that newly added or removed `SupportedTraitTypesForAsyncDelay` enum is handled in iterator");
+                return None;
+            }
+        };
+        self.index += 1;
+        supported_type
+    }
+}
```
bfish713, line 132: What is the point of implementing iterator?

lukeiannucci: add the same settings for each type

lukeiannucci: basically so you dont have to do:
config.add_setting(SupportedTraitTypesForAsyncDelay::Storage, &delay_settings);
config.add_setting(SupportedTraitTypesForAsyncDelay::BlockHeader, &delay_settings);
...

bfish713 (approved):

PR #3566, opened 2024-08-12
[Libp2p] Cleanup for PushCDN Request Response Implementation

jparr721:
Closes #<ISSUE_NUMBER>

### This PR: 
Sets the stage for the pushcdn request/response feature. It refactors some areas where the type granularity was overly general, which caused problems with the downstream usage of the `request_data` method in pushcdn due to the collision of `TYPES` in the upper scope, and on the trait method implementation.

This PR swaps `TYPES` for `SignatureKey` within pushcdn, and moves around some types to make them more available as, initially, they were only defined in the `libp2p` module.

### This PR does not: 
No logic changes, just moving things around and specializing the pushcdn types.

### Key places to review: 
- `types/src/request_response.rs`
- `push_cdn_network.rs`
Everything else is just propagating the changes.


rob-maron (approved):
Looks good to me

Issue #3160, opened 2024-05-15
[CX_HARDENING] - Audit DA Task Tests


jparr721:
### What is this task and why do we need to work on it?

The Consensus tests do not always cover all of the code paths and permutations of inputs for the events that they cover.

### What work will need to be done to complete this task?

[ ] - Evaluate the DA task to determine if enough of the code is covered.
[ ] - Permute the inputs and verify if the tests still pass.
[ ] - Clean up and reduce boilerplate within the test file.
[ ] - Add additional tests for the various cases that were not previously handled.

### What are the acceptance criteria to close this issue?

[ ] - New tests are added
[ ] - All tests pass

shamb0:
Hi @jparr721, I am interested to work on this task, Could you please assign it to me ?

PR #3568, opened 2024-08-13
[CX_HARDENING] Enhance DA Task Test Coverage

shamb0:
## This PR:

This PR partially addresses issue [#3160](https://github.com/EspressoSystems/HotShot/issues/3160) by adding test cases for the following scenarios:

- Handling outdated proposals.
- Detecting duplicate votes.
- Validating vote collection and processing.
- Ensuring non-leader nodes correctly handle DA votes.

## Key Changes

1. Added a new test file: `crates/testing/tests/tests_cx_hardening/cxh_da_task.rs`.
2. Introduced the `create_lean_test_handle()` function to create a lightweight system handle for task state testing.
3. Fixed a bug in the `build_da_cert<>` function located in `HotShot/crates/testing/src/helpers.rs`.
4. Enhanced test functionality to verify external events on `output_event_stream`.

**This PR does not:**

- Introduce any other changes beyond the mentioned test cases and bug fixes.

**Key Areas to Review:**

1. All files committed as part of this PR.

### DIFF
--- a/crates/hotshot/src/lib.rs
+++ b/crates/hotshot/src/lib.rs
@@ -639,6 +639,31 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> SystemContext<T
     }
 }
 
+#[cfg(feature = "hotshot-testing")]
+impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> SystemContext<TYPES, I, V> {
+    /// Creates a lightweight version of the system handle for task state testing.
+    ///
+    /// This method provides a minimal context for task state tests, omitting the full
+    /// consensus and network task launches. It results in fewer traces and simplifies debugging.
+    ///
+    /// # Returns
+    /// A `SystemContextHandle<TYPES, I, V>` with minimal setup for task state testing.
+    pub fn create_lean_test_handle(&self) -> SystemContextHandle<TYPES, I, V> {
+        let (internal_sender, internal_receiver) = broadcast(EVENT_CHANNEL_SIZE);
+
+        SystemContextHandle {
+            consensus_registry: ConsensusTaskRegistry::new(),
+            network_registry: NetworkTaskRegistry::new(),
+            output_event_stream: self.external_event_stream.clone(),
+            internal_event_stream: (internal_sender, internal_receiver.deactivate()),
+            hotshot: self.clone().into(),
+            storage: Arc::clone(&self.storage),
+            network: Arc::clone(&self.network),
+            memberships: Arc::clone(&self.memberships),
+        }
+    }
+}
+
 /// An async broadcast channel
 type Channel<S> = (Sender<Arc<S>>, Receiver<Arc<S>>);
 
--- a/crates/testing/src/helpers.rs
+++ b/crates/testing/src/helpers.rs
@@ -5,15 +5,17 @@
 // along with the HotShot repository. If not, see <https://mit-license.org/>.
 
 #![allow(clippy::panic)]
-use std::{fmt::Debug, hash::Hash, marker::PhantomData, sync::Arc};
+use std::{fmt::Debug, hash::Hash, marker::PhantomData, sync::Arc, time::Duration};
 
 use async_broadcast::{Receiver, Sender};
+use async_compatibility_layer::art::async_timeout;
 use bitvec::bitvec;
 use committable::Committable;
 use ethereum_types::U256;
+use futures::StreamExt;
 use hotshot::{
     traits::{NodeImplementation, TestableNodeImplementation},
-    types::{BLSPubKey, SignatureKey, SystemContextHandle},
+    types::{BLSPubKey, Event, SignatureKey, SystemContextHandle},
     HotShotInitializer, Memberships, SystemContext,
 };
 use hotshot_example_types::{
@@ -40,11 +42,15 @@ use hotshot_types::{
     utils::{View, ViewInner},
     vid::{vid_scheme, VidCommitment, VidSchemeType},
     vote::{Certificate, HasViewNumber, Vote},
+    PeerConfig,
 };
 use jf_vid::VidScheme;
 use serde::Serialize;
 
-use crate::test_builder::TestDescription;
+use crate::{
+    predicates::PredicateResult, script::ExternalEventsExpectations, test_builder::TestDescription,
+    test_launcher::TestLauncher,
+};
 
 /// create the [`SystemContextHandle`] from a node id
 /// # Panics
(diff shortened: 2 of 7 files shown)


crates/hotshot/src/lib.rs
```diff
@@ -639,6 +639,31 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> SystemContext<T
     }
 }
 
+#[cfg(feature = "hotshot-testing")]
+impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> SystemContext<TYPES, I, V> {
```
jparr721, line 643: I think we can just add this to the existing trait impl and enable the method when testing is enabled instead of doing separate bespoke impl.

crates/testing/src/helpers.rs
```diff
@@ -128,6 +134,105 @@ pub async fn build_system_handle<
     .expect("Could not init hotshot")
 }
 
+/// create the [`SystemContextHandle`] from a launcher
+/// # Panics
+/// if cannot create a [`HotShotInitializer`]
+pub async fn build_system_handle_from_launcher<
```
jparr721, line 140: I'm hesitant to make such a specific change for the tests in this case, especially since you have them working. While it's not ideal to spin up the entire SystemContext for unit tests, this method unfortunately creates a rift in what is the accepted testing framework, and what is done here. If we made this change, we'd want to propagate it to *all* tests, lest we end up in the situation where specific knowledge of this specific suite rests solely with someone who is not on the team. 

I am not opposed to the outright removal of this in theory, though I think for this PR, we should stick to the existing spawner.

crates/testing/src/helpers.rs
```diff
@@ -393,3 +538,40 @@ pub fn build_fake_view_with_leaf_and_state(
         },
     }
 }
+
+pub async fn check_external_events<TYPES: NodeType, S: StreamExt<Item = Event<TYPES>> + Unpin>(
```
jparr721, line 542: Curious as to what this is doing that isn't handled by the testing macro?

crates/testing/tests/tests_cx_hardening.rs
```diff
@@ -0,0 +1,3 @@
+mod tests_cx_hardening {
+    automod::dir!("tests/tests_cx_hardening");
```
jparr721, line 2: I think `tests_cx_hardening` would make a task namespace into a first-class feature of the codebase, muddying the circumstances for when we should change them. Instead, let's just add these under the existing da task tests and remove this special file.

crates/testing/tests/tests_cx_hardening/cxh_da_task.rs
```diff
@@ -0,0 +1,564 @@
...
+        },
+    })])];
+
+    // Create DA task state and script for the test
+    let da_state = DaTaskState::<TestTypes, MemoryImpl>::create_from(&handle).await;
+    let mut da_script = TaskScript {
+        timeout: Duration::from_millis(100),
+        state: da_state,
+        expectations,
+    };
```
jparr721, line 140: I don't think we need these external event handlers as the `DaProposal` event is transformed into `DaProposalRecv` in the downstream handlers.

crates/types/src/event.rs
```diff
@@ -173,6 +173,70 @@ pub enum EventType<TYPES: NodeType> {
     /// A message destined for external listeners was received
     ExternalMessageReceived(Vec<u8>),
 }
+
+// Implement Eq as well, since EventType should have reflexive equality
+impl<TYPES: NodeType> Eq for EventType<TYPES> {}
+
+impl<TYPES: NodeType> PartialEq for EventType<TYPES> {
```
jparr721, line 180: This might be a bit overkill if the above external event changes are removed. If we cannot derive Eq and PartialEq here, perhaps we should consider fixing the types within EventType to do so instead, that way we can avoid arduously maintaining this for a yet-unneeded change.

jparr721 (changes requested):
Great first step, and thank you for doing this! Big ticket items are the following:
1. Let's just put all the da tests in the existing test file instead of its own module
2. Let's avoid spawning a custom handler for this test until, as a team, we're ready to refactor the entire test suite to use a more lean handler. If you want a more lean test handler, just change the `build_system_handler` helper method to call `SystemContext::new` instead of `init`, and that should get us 90% of what you've done here.
3. Let's try to avoid handling external events since they nearly-always get turned into internal events anyway. A good place to look is `network.rs` or, broadly, the network task.

crates/testing/tests/tests_cx_hardening.rs
ss-es: +1 to just throw this into one of the test groups

crates/types/src/event.rs
ss-es: +1 this should be a `#derive`. I think you might need to derive `Eq` and/or `PartialEq` for a few types, but I don't think that should be an issue

crates/testing/src/helpers.rs
shamb0: @jparr721, I understand your concerns about consistency in our testing framework. To clarify, the existing `hotshot-testing::helpers::build_system_handle()` function remains unchanged, and all current test cases continue to work as expected.

The new API, `hotshot-testing::helpers::build_system_handle_from_launcher()`, was added to address issue #3160, which requires 'Permute the inputs and verify if the tests still pass.'. While `build_system_handle()` uses default test configuration parameters, the new function provides flexibility by allowing modifications to the setup.

I suggest we review its usage in the test scenarios. If it doesn’t add significant value, we can consider removing it. Specifically, 

- Validating vote collection and processing.
- Ensuring non-leader nodes correctly handle DA votes.

crates/types/src/event.rs
shamb0: @jparr721, @ss-es, I agree that it’s not an ideal solution. My initial approach was to add `#[derive(Eq, PartialEq)]` annotations to the dependent inner types. However, this would have required modifying a significant number of files, so I considered handling it in a separate PR to reduce the review effort.

Please let me know if you prefer that I include these changes in this PR or if it would be better to address them in a separate PR.

crates/testing/tests/tests_cx_hardening/cxh_da_task.rs
shamb0: Related to the comment, https://github.com/EspressoSystems/HotShot/pull/3568#discussion_r1715774905

crates/testing/src/helpers.rs
jparr721: Why not just use the `random!` macro to get permuted inputs in that case?

crates/testing/tests/tests_cx_hardening/cxh_da_task.rs
jparr721: I think I still feel strongly that we don't necessarily need to process the external events. If the event is transformed successfully into a, for example, `DaProposalRecv` event, then we *know* with certainty that the external event was received, so we don't need to handle this stream in that case. To avoid bespoke changes, I'd prefer to remove it since the checks are redundant under this circumstance.

crates/testing/src/helpers.rs
jparr721: See remark below. This is not necessary since the network task will transform the raw event into a HotShot event, so the checks will be redundant.

crates/hotshot/src/lib.rs
ss-es: I think this should actually just go in `testing/src/helpers.rs` (ungated), next to `build_system_handle` (which is what we use currently in the tests). I'd call it `build_inactive_handle` or something (I think "lean" is slightly misleading, because it's just a dummy handle that isn't running anything).

I made an issue a while ago to do something like this for the non-integration tests (all the task-specific tests in `tests_1`), so I think this is good to have in general

crates/types/src/event.rs
ss-es: I think I'm just strongly opposed to a custom `Eq` or `PartialEq` implementation.

But after looking over the PR some more, I'm not sure if any of this is really that necessary; I think I agree with @jparr721 that the external event checking could just be removed altogether.

We might have to discuss internally about whether we want to test this (and to what extent), but I think I would just split this off from the PR

crates/hotshot/src/lib.rs
shamb0: Good One @ss-es, I wanted to acknowledge the suggestion to place the test-specific helper function in `testing/src/helpers.rs`. During my implementation, I encountered issues due to the private and public(crate) visibility of dependent fields in pub struct SystemContext<> and pub struct SystemContextHandle<>. To avoid modifying the existing visibility of these fields, I opted for an alternative approach.

crates/testing/src/helpers.rs
shamb0: Thank you, @jparr721, for your suggestion. I agree that utilizing random!() could be an effective way to introduce fuzziness in test configuration parameters. I will explore this approach and provide an update on the outcome.

PR #3569, opened 2024-08-13
Add marketplace integration test

ss-es:
### This PR: 
- Allows integration tests to be configured to perform an upgrade at a given view.
- Reworks the `consistency_task` to ensure that an upgrade certificate is actually attached to some leaf.
- Adds a `test_success_marketplace` to test that an upgrade to the marketplace completes successfully and continues to run without any view failures.

### Key places to review: 
You can run the test on this branch with `RUST_LOG=hotshot_task_impls=debug just cargo test test_success_marketplace -- --nocapture`

### DIFF
--- a/crates/example-types/src/block_types.rs
+++ b/crates/example-types/src/block_types.rs
@@ -29,6 +29,7 @@ use time::OffsetDateTime;
 use vbs::version::Version;
 
 use crate::{
+    auction_results_provider_types::TestAuctionResult,
     node_types::TestTypes,
     state_types::{TestInstanceState, TestValidatedState},
     testable_delay::{DelayConfig, SupportedTraitTypesForAsyncDelay, TestableDelay},
@@ -274,6 +275,7 @@ impl<
             BlockHeader = Self,
             BlockPayload = TestBlockPayload,
             InstanceState = TestInstanceState,
+            AuctionResult = TestAuctionResult,
         >,
     > BlockHeader<TYPES> for TestBlockHeader
 {
@@ -349,7 +351,7 @@ impl<
     }
 
     fn get_auction_results(&self) -> Option<TYPES::AuctionResult> {
-        unimplemented!()
+        Some(TYPES::AuctionResult { urls: vec![] })
     }
 }
 
--- a/crates/task-impls/src/events.rs
+++ b/crates/task-impls/src/events.rs
@@ -183,8 +183,6 @@ pub enum HotShotEvent<TYPES: NodeType> {
     UpgradeVoteSend(UpgradeVote<TYPES>),
     /// Upgrade certificate has been sent to the network
     UpgradeCertificateFormed(UpgradeCertificate<TYPES>),
-    /// A HotShot upgrade was decided
-    UpgradeDecided(UpgradeCertificate<TYPES>),
 
     /* Consensus State Update Events */
     /// A undecided view has been created and added to the validated state storage.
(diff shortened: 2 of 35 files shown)


crates/testing/tests/tests_1/test_success.rs
```diff
@@ -39,6 +42,31 @@ cross_tests!(
...
+                                             TimeBasedCompletionTaskDescription {
+                                                 duration: Duration::from_secs(60),
+                                             },
+                                         ),
+            upgrade_view: Some(5),
+            ..TestDescription::default()
+        };
+
+
+        metadata.overall_safety_properties.num_failed_views = 10;
```
bfish713, line 64: Why does it need to allow 10 failed views?

bfish713 (commented):
This looks reasonable to me, just one question and maybe somebody working on the marketplace can confirm this is testing what you need

crates/example-types/src/block_types.rs
```diff
@@ -349,7 +351,7 @@ impl<
     }
 
     fn get_auction_results(&self) -> Option<TYPES::AuctionResult> {
-        unimplemented!()
+        Some(TYPES::AuctionResult { urls: vec![] })
```
elliedavidson, line 354: Can we make an issue to update this once the simple builder supports the marketplace?

crates/task-impls/src/events.rs
```diff
@@ -183,8 +183,6 @@ pub enum HotShotEvent<TYPES: NodeType> {
     UpgradeVoteSend(UpgradeVote<TYPES>),
     /// Upgrade certificate has been sent to the network
     UpgradeCertificateFormed(UpgradeCertificate<TYPES>),
-    /// A HotShot upgrade was decided
-    UpgradeDecided(UpgradeCertificate<TYPES>),
```
QuentinI, line 187: Why do we want to remove this?

QuentinI (approved):
LGTM with some questions

crates/testing/tests/tests_1/test_success.rs
```diff
@@ -39,6 +42,31 @@ cross_tests!(
     },
 );
 
+cross_tests!(
```
QuentinI, line 45: Probably would make sense to also add test for 0.1 -> 0.2 upgrade

crates/example-types/src/block_types.rs
QuentinI: We should also enable transaction generation and checking that blocks contain transactions in tests when simple builder is updated, so that the test check the builder interaction as well.

crates/testing/tests/tests_1/test_success.rs
ss-es: oops -- fixed!

crates/task-impls/src/events.rs
ss-es: It's an old event; it's not used for anything and shouldn't be used for anything, since the correct way to access version information is via the `upgrade_lock` passed around to the tasks.

(this was just some cleanup, not really related to the PR)

crates/testing/tests/tests_1/test_success.rs
ss-es: Nothing in consensus is gated on the fee upgrade (or even has access to a constant representing its version number), so I think the marketplace test passing without any view failures already covers general upgrades too.

crates/example-types/src/block_types.rs
ss-es: Done! https://github.com/EspressoSystems/HotShot/issues/3583

crates/testing/tests/tests_1/test_success.rs
QuentinI: Makes sense, thanks

PR #3573, opened 2024-08-14
Update builder marketplace API

ss-es:
### This PR: 
Changes the builder marketplace API to include a parent view and parent commitment.

This made the fallback logic for returning an empty block even more unwieldy, so this PR also includes some changes to the transactions task marketplace logic.

### DIFF
--- a/crates/task-impls/src/transactions.rs
+++ b/crates/task-impls/src/transactions.rs
@@ -9,7 +9,7 @@ use std::{
     time::{Duration, Instant},
 };
 
-use anyhow::{bail, Result};
+use anyhow::{bail, ensure, Context, Result};
 use async_broadcast::{Receiver, Sender};
 use async_compatibility_layer::art::{async_sleep, async_timeout};
 use async_trait::async_trait;
@@ -34,7 +34,7 @@ use hotshot_types::{
 };
 use tracing::{debug, error, instrument, warn};
 use url::Url;
-use vbs::version::StaticVersionType;
+use vbs::version::{StaticVersionType, Version};
 use vec1::Vec1;
 
 use crate::{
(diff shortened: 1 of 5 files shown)


elliedavidson (commented):
This looks good to me

crates/task-impls/src/transactions.rs
```diff
@@ -232,114 +232,102 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> TransactionTask
...
 
-    #[allow(clippy::too_many_lines)]
-    /// marketplace view change handler
-    pub async fn handle_view_change_marketplace(
+    /// Produce a block by fetching auction results from the solver and bundles from builders.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the solver cannot be contacted, or if none of the builders respond.
+    pub async fn produce_block_marketplace(
```
QuentinI, line 240: Nit: doesn't need to be pub

crates/task-impls/src/transactions.rs
```diff
@@ -232,114 +232,102 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> TransactionTask
...
-                for bundle in join_all(futures).await {
-                    match bundle {
-                        Ok(Ok(b)) => bundles.push(b),
-                        _ => continue,
-                    }
-                }
+        for bundle in join_all(futures).await {
+            match bundle {
+                Ok(Ok(b)) => bundles.push(b),
+                _ => continue,
```
QuentinI, line 291: Out of scope for this, so feel free to ignore, but it'd be nice to have some logging of the timeout or error here.

crates/task-impls/src/transactions.rs
```diff
@@ -232,114 +232,102 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> TransactionTask
...
+        ))
+    }
 
+    /// Produce a null block
+    pub async fn null_block(
+        &mut self,
+        block_view: TYPES::Time,
+        version: Version,
+    ) -> Option<PackedBundle<TYPES>> {
         // Increment the metric for number of empty blocks proposed
```
QuentinI, line 331: Nit: I'd make this function sync and move the metric increment to the caller to make this one side-effect free

crates/task-impls/src/transactions.rs
```diff
@@ -361,16 +349,51 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> TransactionTask
...
+    ) -> Option<HotShotTaskCompleted> {
+        let version = match self.upgrade_lock.version(block_view).await {
+            Ok(v) => v,
+            Err(err) => {
+                error!("Upgrade certificate requires unsupported version, refusing to request blocks: {}", err);
+                return None;
+            }
+        };
+
+        let (parent_view, parent_hash) = self.last_vid_commitment(block_view).await?;
```
QuentinI, line 377: Can we add a retry loop on this similar to what we have in legacy? It's not uncommon for the leader not to have the relevant VID right away.
Also, if we fail to find it, we should still produce an empty block.

ss-es: Updated: I created a wrapper `last_vid_commitment_retry` for the loop, and used that in both places.

QuentinI (approved):
LGTM!

PR #3575, opened 2024-08-14
[LibP2P] - Implement Request/Response For The PushCDN Network

### LINKED ISSUE #3565: [LibP2P] - Implement Request/Response For The PushCDN Network

Request/Response is not yet available for the PushCDN network layer. We need to implement this functionality to be able to route such requests through PushCDN. In particular, we need to implement the following specific trait methods:
- `spawn_request_receiver_task`
- `request_data`

These methods enable the processing of messages where arbitrary data is being fetched from a peer.

### `spawn_request_receiver_task`
This will require the creation of a broadcast channel for receiving messages from `NetworkResponseState` and processing. This `mpsc` channel will exist within the pushcdn type and be made available for processing requests forwarded to the background task.

### `request_data`
We'll use this method to actually facilitate requesting the designated peer.

This work will also remove the dependency on `TYPES` within the `request_data` method and instead put it into the implementor's domain. This is to get around an issue like the following:
```rust
#[async_trait]
impl<TYPES: NodeType> ConnectedNetwork<TYPES::SignatureKey> for PushCdnNetwork<TYPES> {
    async fn request_data<TYPES2: NodeType>(
        &self,
        request: Vec<u8>,
        recipient: TYPES2::SignatureKey,
    ) -> Result<Vec<u8>, NetworkError> {
        self.client.send_direct_message(recipient, request).await;

        Ok(vec![])
    }

    async fn spawn_request_receiver_task(
        &self,
    ) -> Option<mpsc::Receiver<(Vec<u8>, ResponseChannel<Vec<u8>>)>> {
        None
    }
```

Right now, a network backend that uses `TYPES` natively will not be able to cleanly implement the trait method as we need to perform disambiguation on the types. The alternative is to make `TYPES` a member of `ConnectedNetwork`, however, this requires tedious changes to the sequencer and, as a result, is less ideal.

_No response_

- New implementation methods are added
- This backend is added to the test suite
- New tests are added for the new behavior

_No response_

jparr721:
Closes #3565 

### This PR: 
Adds broad support for request/response for Quorum Proposals across the entire network layer. It accomplishes this by introducing new HotShot events that are able to interpret requests for data. It leverages the existing `NetworkRequestTask` to read events and respond if data exists.

This PR also updates the `NetworkTask` to support the new events, and supports direct messaging between the requesting and responding node.

### DIFF
--- a/crates/hotshot/src/traits/networking/push_cdn_network.rs
+++ b/crates/hotshot/src/traits/networking/push_cdn_network.rs
@@ -33,6 +33,7 @@ use cdn_client::{
 };
 #[cfg(feature = "hotshot-testing")]
 use cdn_marshal::{Config as MarshalConfig, Marshal};
+use futures::channel::mpsc;
 #[cfg(feature = "hotshot-testing")]
 use hotshot_types::traits::network::{
     AsyncGenerator, NetworkReliability, TestableNetworkingImplementation,
@@ -40,6 +41,7 @@ use hotshot_types::traits::network::{
 use hotshot_types::{
     boxed_sync,
     data::ViewNumber,
+    request_response::NetworkMsgResponseChannel,
     traits::{
         metrics::{Counter, Metrics, NoMetrics},
         network::{BroadcastDelay, ConnectedNetwork, PushCdnNetworkError, Topic as HotShotTopic},
@@ -434,21 +436,20 @@ impl<TYPES: NodeType> TestableNetworkingImplementation<TYPES>
 
 #[async_trait]
 impl<K: SignatureKey + 'static> ConnectedNetwork<K> for PushCdnNetwork<K> {
-    // async fn request_data<ReqDataTYPES: NodeType>(
-    //     &self,
-    //     request: Vec<u8>,
-    //     recipient: ReqDataK,
-    // ) -> Result<Vec<u8>, NetworkError> {
-    //     self.client.send_direct_message(recipient, request).await;
+    async fn request_data<TYPES: NodeType>(
+        &self,
+        _request: Vec<u8>,
+        _recipient: &K,
+    ) -> Result<Vec<u8>, NetworkError> {
+        Ok(vec![])
+    }
 
-    //     Ok(vec![])
-    // }
-
-    // async fn spawn_request_receiver_task(
-    //     &self,
-    // ) -> Option<mpsc::Receiver<(Vec<u8>, NetworkMsgResponseChannel<Vec<u8>>)>> {
-    //     None
-    // }
+    async fn spawn_request_receiver_task(
+        &self,
+    ) -> Option<mpsc::Receiver<(Vec<u8>, NetworkMsgResponseChannel<Vec<u8>>)>> {
+        let (mut _tx, rx) = mpsc::channel(1);
+        Some(rx)
+    }
 
     /// Pause sending and receiving on the PushCDN network.
     fn pause(&self) {
--- a/crates/task-impls/src/helpers.rs
+++ b/crates/task-impls/src/helpers.rs
@@ -40,7 +40,7 @@ use tracing::{debug, error, info, instrument, warn};
 
 use crate::{events::HotShotEvent, request::REQUEST_TIMEOUT};
 
-/// Trigger a request to the network for a proposal for a view and wait for the response
+/// Trigger a request to the network for a proposal for a view and wait for the response or timeout.
 #[instrument(skip_all)]
 pub(crate) async fn fetch_proposal<TYPES: NodeType>(
     view_number: TYPES::Time,
@@ -57,9 +57,11 @@ pub(crate) async fn fetch_proposal<TYPES: NodeType>(
     )
     .await;
 
-    // Hard-block waiting for the arrival of the event
+    // Make a background task to await the arrival of the event data.
     let Ok(Some(proposal)) = async_spawn(async move {
+        // We want to explicitly timeout here so we aren't waiting around for the data.
         async_timeout(REQUEST_TIMEOUT, async move {
+            // First, capture the output from the event dependency
             let event = EventDependency::new(
                 event_receiver.clone(),
                 Box::new(move |event| {
@@ -78,6 +80,7 @@ pub(crate) async fn fetch_proposal<TYPES: NodeType>(
             .completed()
             .await;
 
+            // Then, if it's `Some`, make sure that the data is correct
             if let Some(hs_event) = event.as_ref() {
                 if let HotShotEvent::QuorumProposalResponseRecv(_view_number, quorum_proposal) =
                     hs_event.as_ref()
--- a/crates/task-impls/src/helpers.rs
+++ b/crates/task-impls/src/helpers.rs
@@ -58,7 +58,7 @@ pub(crate) async fn fetch_proposal<TYPES: NodeType>(
     .await;
 
     // Make a background task to await the arrival of the event data.
-    let Ok(Some(proposal)) = async_spawn(async move {
+    let Ok(Some(proposal)) =
         // We want to explicitly timeout here so we aren't waiting around for the data.
         async_timeout(REQUEST_TIMEOUT, async move {
             // First, capture the output from the event dependency
@@ -92,8 +92,6 @@ pub(crate) async fn fetch_proposal<TYPES: NodeType>(
             None
         })
         .await
-    })
-    .await
     else {
         bail!("Request for proposal failed");
     };
--- a/crates/task-impls/src/helpers.rs
+++ b/crates/task-impls/src/helpers.rs
@@ -11,13 +11,14 @@ use std::{
 };
 
 use anyhow::{bail, ensure, Context, Result};
-use async_broadcast::{broadcast, SendError, Sender};
+use async_broadcast::{Receiver, SendError, Sender};
 use async_compatibility_layer::art::{async_sleep, async_spawn, async_timeout};
 use async_lock::RwLock;
 #[cfg(async_executor_impl = "async-std")]
 use async_std::task::JoinHandle;
 use chrono::Utc;
 use committable::{Commitment, Committable};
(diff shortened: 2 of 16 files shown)


crates/types/src/message.rs
```diff
@@ -162,6 +162,12 @@ pub enum GeneralConsensusMessage<TYPES: NodeType> {
     /// Message with a quorum proposal.
     Proposal(Proposal<TYPES, QuorumProposal<TYPES>>),
 
+    /// A peer node needs a proposal from the leader.
+    ProposalRequested(TYPES::Time, TYPES::SignatureKey),
+
+    /// The leader has responded with a valid proposal.
+    LeaderProposalAvailable(Proposal<TYPES, QuorumProposal<TYPES>>),
```
rob-maron, line 169: This message is needed to "finish" the request right? Is there any downside to sending-and-forgetting the request and getting a proposal, and then doing normal consensus operations with it?

crates/testing/tests/tests_1/quorum_proposal_recv_task.rs
```diff
@@ -210,7 +210,10 @@ async fn test_quorum_proposal_recv_task_liveness_check() {
                 ),
             ),
         )),
-        quorum_proposal_missing(),
+        exact(QuorumProposalRequestSend(
+            ViewNumber::new(2),
+            handle.public_key()
+        )),
```
rob-maron, line 216: Not as familiar with the integration tests - why do we expect a request for view 2 here?

crates/task-impls/src/network.rs
```diff
@@ -292,6 +303,20 @@ impl<
                         TransmitType::Direct(membership.leader(vote.view_number() + 1)),
                     )
                 }
+                HotShotEvent::QuorumProposalRequestSend(view_number, sender_key) => (
+                    sender_key.clone(),
+                    MessageKind::<TYPES>::from_consensus_message(SequencingMessage::General(
+                        GeneralConsensusMessage::ProposalRequested(view_number, sender_key),
+                    )),
+                    TransmitType::Direct(membership.leader(view_number)),
+                ),
```
rob-maron, line 312: This is a future change - but I believe a malicious node could put a `sender_key` that's not them and amplify traffic to them, as they would get the response proposal

Perhaps we could sign over this

crates/task-impls/src/helpers.rs
```diff
@@ -37,32 +38,64 @@ use hotshot_types::{
...
+                    ) = event
+                    {
+                        *qpr_view_number == view_number
+                    } else {
+                        false
+                    }
+                }),
+            )
+            .completed()
+            .await;
```
rob-maron, line 81: Similar to the comment in `message.rs` - we could get rid of the need for an extra task if we send and forget. But not sure how nicely that plays with the dependency stuff

crates/hotshot/src/traits/networking/push_cdn_network.rs
```diff
@@ -434,21 +436,20 @@ impl<TYPES: NodeType> TestableNetworkingImplementation<TYPES>
...
-    // ) -> Option<mpsc::Receiver<(Vec<u8>, NetworkMsgResponseChannel<Vec<u8>>)>> {
-    //     None
-    // }
+    async fn request_data<TYPES: NodeType>(
+        &self,
+        _request: Vec<u8>,
+        _recipient: &K,
+    ) -> Result<Vec<u8>, NetworkError> {
+        Ok(vec![])
+    }
```
rob-maron, line 445: Are we getting rid of these entirely in the network function signatures? Do we need them?

rob-maron (commented):
Looks good, just some high-level questions for my understanding

crates/testing/tests/tests_1/quorum_proposal_recv_task.rs
jparr721: That proposal is missing in the parent's state. Looking here
```rust
    let inputs = vec![serial![QuorumProposalRecv(
        proposals[2].clone(),
        leaders[2]
    )]];
```

We were sent a proposal for view 3, but the proposal for view 2 (i.e. the `justify_qc`) does *not* exist in memory due to the setup of the test, so we broadcast a request for the proposal from the leader to get the proposal so we know that we can safely branch off of this node.

crates/task-impls/src/network.rs
jparr721: Yeah I know we had chatted about this before. Tobias and I had talked about it as well but the crux of the matter is that anyone at any time could spam a single node by adding a new internal event that always broadcasts and roasts the leader. I think we'll have to just penalize to avoid spamming, even if we sign it to ensure validity I can still get bogus data through the network layer to whomever I want, even if they end up dropping the message.

crates/task-impls/src/helpers.rs
jparr721: The downside of send-and-forget is that it disallows us the ability to block on the data coming back. So, in this case, the proposal's arrival begets a block before we vote/propose if we're missing that prior state. Because of that, I opted to just block (as the prior version did as well) and timeout in some short interval if we don't hear back from the leader in time. I think in general though, I agree that this type of blocking should *not* happen. I'd like to do a more earnest look at the R/R tasks but I didn't want to make this a massive PR.

crates/testing/tests/tests_1/quorum_proposal_recv_task.rs
rob-maron: Awesome, that's what I suspected, thanks!

crates/hotshot/src/traits/networking/push_cdn_network.rs
jparr721: Unfortunately yes, but I'd like to delete these methods entirely. If you look in the `request.rs` file, the vid request uses the `DelayedRequester` which calls `request_data`. So, it'll fail for the PushCDN, but our backstop from this is that combined network *only* uses the LibP2p layer as-is anyway, so I don't think this'll cause problems. Admittedly, this particular solution is not my favorite. I plan to do essentially this exact same changeset, but for the VID data request, in a follow-on PR. I think we could wait to merge this and I can branch from this and do the VID change, merge that one into this one, then ship it all as one macro-feature if we think that this approach could cause breakage. Admittedly I was not sure.

crates/task-impls/src/network.rs
rob-maron: Right, but I guess in this case you could essentially have all _other_ nodes DoS this one node, which might be harder to deal with

But long-term thing if any, not significant for this PR

crates/task-impls/src/helpers.rs
rob-maron: Gotcha, yeah. Maybe down the line we could utilize the dependency framework to get it to block - although admittedly I am not the most familiar. For this PR it is fine though

crates/hotshot/src/traits/networking/push_cdn_network.rs
rob-maron: Sounds good!

rob-maron (approved):

crates/types/src/message.rs
jparr721: Yeah so basically it is the receiving node getting a `QuorumProposalResponseSend` event, and transforming it into a `QuorumProposalResponseRecv` event. This could be done any number of ways, I just opted to stratify things a little bit here because I prefer to have explicit events for each stage of the request chain. This GeneralConsensusMessage type honestly only exists to fulfill the need for there to be a consensus message that will not 
A. Be confused with a genuine proposal.
B. Be confused with a request for a proposal.

I'd happily name this anything else if there's any opinions there.

On your send-and-forget remark, I think generally I would like this, but it comes with two caveats given our async-focused design in HotShot

1. Right now, if I need to **guarantee** that data is available in my local in-memory state, I can *only* spin lock in a background thread (with some arbitrary delay).
2. I need a way to be able to halt a critical operation until my data has arrived.

As a result, the event dependency made the most sense. So, yes, *technically* we for sure could send and forget to get the proposal, the problem is we'd end up invariably waiting anyway, so by making the events explicitly say "I want a proposal right now for view x and I will wait until it is found" we can specialize what they're doing.

bfish713 (approved):

crates/task-impls/src/network.rs
bfish713: Actually previously the DataRequest message type was signed by the sender so this type of amplification/DDoS attack would not be possible, this PR is actually removing the signing.  We should added it back soon than later

crates/task-impls/src/request.rs
```diff
@@ -115,12 +108,22 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> TaskState for NetworkRequest
...
+                            *view_number,
+                            sender_key.clone(),
+                            quorum_proposal.clone(),
+                        )
+                        .into(),
+                        sender,
+                    )
+                    .await;
+                }
+
```
bfish713, line 126: It's not really wrong to have this here, but it logically belongs in the "Response" task which handles responding to requests where the request task is handling sending out requests and waiting for the response

crates/task-impls/src/network.rs
```diff
@@ -292,6 +303,20 @@ impl<
                         TransmitType::Direct(membership.leader(vote.view_number() + 1)),
                     )
                 }
+                HotShotEvent::QuorumProposalRequestSend(view_number, sender_key) => (
+                    sender_key.clone(),
+                    MessageKind::<TYPES>::from_consensus_message(SequencingMessage::General(
+                        GeneralConsensusMessage::ProposalRequested(view_number, sender_key),
+                    )),
+                    TransmitType::Direct(membership.leader(view_number)),
```
bfish713, line 311: Just FYI the next task related to this is to make it so we can request proposals from the DA committee as well so this logic will have to get significantly more complex.  Basically the logic of do_vid in the request task will need to live here now, or we can just handle this event there like before

Issue #3528, opened 2024-08-01
[LOGGING] - Log Errors or Panic When Critical Tasks exit


bfish713:
### What is this task and why do we need to work on it?

We found a bug where the network task could exit if serialization failed.  This halted consensus.  We need an easy way to detect this kind of thing

### What work will need to be done to complete this task?

Add a new task that will broadcast a health event to all tasks, if health check doesnt receive a response in specified amount of time add logging.

### What are the acceptance criteria to close this issue?

All points where a critical long running task stops, aside from shutdown are logged at error level

PR #3577, opened 2024-08-15
[LOGGING] - Log Errors or Panic When Critical Tasks exit

lukeiannucci:
Closes [#3528 ](https://github.com/EspressoSystems/HotShot/issues/3528)


crates/task/src/task.rs
```diff
@@ -75,38 +92,102 @@ impl<S: TaskState + Send + 'static> Task<S> {
         Box::new(self.state) as Box<dyn TaskState<Event = S::Event>>
     }
 
+    #[cfg(async_executor_impl = "async-std")]
+    /// Periodic delay
+    pub fn get_periodic_interval_in_secs(
+        periodic_delay: u64,
```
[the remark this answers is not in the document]
lukeiannucci: yea i can make it a const

crates/task-impls/src/network.rs
```diff
@@ -248,6 +248,14 @@ impl<
     }
 
     async fn cancel_subtasks(&mut self) {}
+
+    async fn periodic_task(&self, task_id: String, sender: &Sender<Arc<Self::Event>>) {
+        broadcast_event(Arc::new(HotShotEvent::HeartBeat(task_id)), sender).await;
+    }
+
+    fn get_task_name(&self) -> String {
+        "NetworkEventTask".to_string()
```
jparr721, line 257: Hard-coding this is a dangerous game since it'll need to be synced with the name of the struct, and would not error on typos or name mismatches.. Could we potentially pre-define the trait method using something like [type_name](https://doc.rust-lang.org/std/any/fn.type_name.html)? Not sure if this is the best way to do it, but if it works it'd reduce the manual bit here a lot.

crates/task-impls/src/network.rs
```diff
@@ -248,6 +248,14 @@ impl<
     }
 
     async fn cancel_subtasks(&mut self) {}
+
+    async fn periodic_task(&self, task_id: String, sender: &Sender<Arc<Self::Event>>) {
```
jparr721, line 252: Why don't we just define these in the trait? Looks like all instances of this method do the same thing?

jparr721 (commented):
Couple of high level comments.

crates/task-impls/src/network.rs
lukeiannucci: only health check task behaves differently, as that will have logic for the logging if heartbeat missed for some time. i can make a trait and have this as default implementation

crates/task/src/task.rs
```diff
@@ -75,38 +92,102 @@ impl<S: TaskState + Send + 'static> Task<S> {
...
                 }
             }
-        })
+        });
+        ConsensusHandle { handle, task_id }
     }
 }
 
+/// Wrapper around handle and task id so we can map
+pub struct ConsensusHandle<EVENT> {
```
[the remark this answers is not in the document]
lukeiannucci: good idea!

crates/task-impls/src/network.rs
lukeiannucci: yes i agree hardcoding isnt the best. let me try out your ideas. thanks

crates/task/src/task.rs
```diff
@@ -43,6 +46,12 @@ pub trait TaskState: Send {
         _sender: &Sender<Arc<Self::Event>>,
         _receiver: &Receiver<Arc<Self::Event>>,
     ) -> Result<()>;
+
+    /// Runs a specified job in the main task every so often
+    async fn periodic_task(&self, task_id: String, sender: &Sender<Arc<Self::Event>>);
```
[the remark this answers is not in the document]
lukeiannucci: ok i remember why i did it this way the broadcast_event is in crate hotshot-task-impl in the helpers. this code is in hotshot-task crate. i was getting a circular dependency as hotshot-task-impl has a dependency on hotshot-task when i added it in cargo.toml. ill see if i can find another way to do this

crates/task-impls/src/network.rs
lukeiannucci: let me know what you think about it now, i used type_name as suggested

crates/task-impls/src/network.rs
lukeiannucci: let me know what you think about this new approach

crates/task/src/task.rs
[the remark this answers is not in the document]
QuentinI: I think inlining broadcast_event here makes sense.

QuentinI (approved):
LGTM!

PR #3581, opened 2024-08-15
Add random assignees workflow

QuentinI:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/.github/workflows/assign-reviewers.yml
+++ b/.github/workflows/assign-reviewers.yml
@@ -1,0 +1,21 @@
+name: Assign reviewers
+on:
+  pull_request:
+    types: [opened, ready_for_review, reopened]
+
+jobs:
+  test:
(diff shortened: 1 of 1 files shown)


.github/workflows/assign-reviewers.yml
```diff
@@ -0,0 +1,21 @@
+name: Assign reviewers
```
shenkeyao, line 1: Nit: maybe call this something like "Automate assignees", i.e., remove "reviewers" from the name to avoid confusion?

shenkeyao (approved):
Just a nit suggestion.

.github/workflows/assign-reviewers.yml
QuentinI: Renamed to "Choose random assignees"

shenkeyao (approved):

PR #3590, opened 2024-08-20
Add test for verify_share return type

elliedavidson:
Closes #<ISSUE_NUMBER>

### This PR: 
This PR fixes a bug where an invalid vid share could be marked as valid.  Specifically, the `verify_share` function from jellyfish returns a nested `Result`.  Previously we were only checking the outer result.  This means that a call to `verify_share` that returned a result of `Ok(Err)` would be incorrectly return `true`.  This PR checks both the inner and outer results.  

### Key places to review: 
`task_impls/src/consensus/mod.rs`
`testing/test-1/consensus_task.rs`

 ### How to test this PR: 
To test this PR, check out this branch.  Next, check out the following branch of jellyfish locally, `309fb5ba80a2b401dc7d5c355ff5cecece8fa412`.  This jellyfish branch contains a function to create a corrupt VID share in the exact way we were not covering before.  This commit in HotShot contains the `test_invalid_vid_disperse` test, but it is currently ignored.  To test that the test is doing what it is supposed to, revert the change in the `consensus/mod.rs` file to what it was previously.  The test will now fail.  

To run the test use the following command: 
`just tokio test test_invalid_vid_disperse --features=test-srs`

NOTE: This test is ignored in the final commit because the corresponding function in jellyfish will likely not be merged soon, if at all.  The test is gated by the `test-srs` feature to ensure it does not compile during normal testing.  The point is just to demonstrate the old bug and that it was fixed.  It might make sense to only merge the bug fixes and not the tests. 

Tagging @jparr721 in case we want to look for bugs of a similar class.

### DIFF
--- a/crates/task-impls/src/consensus/mod.rs
+++ b/crates/task-impls/src/consensus/mod.rs
@@ -177,19 +177,16 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> ConsensusTaskSt
         }
 
         // Validate the VID share.
-        if vid_scheme(self.quorum_membership.total_nodes())
-            .verify_share(
-                &disperse.data.share,
-                &disperse.data.common,
-                &payload_commitment,
-            )
-            .is_err()
-        {
-            debug!("Invalid VID share.");
-            return false;
+        // NOTE: `verify_share` returns a nested `Result`, so we must check both the inner
+        // and outer results
+        match vid_scheme(self.quorum_membership.total_nodes()).verify_share(
+            &disperse.data.share,
+            &disperse.data.common,
+            &payload_commitment,
+        ) {
+            Ok(inner) => inner.is_ok(),
+            Err(_) => false,
         }
-
-        true
     }
 
     /// Publishes a proposal
--- a/crates/task-impls/src/consensus/mod.rs
+++ b/crates/task-impls/src/consensus/mod.rs
@@ -184,8 +184,8 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> ConsensusTaskSt
             &disperse.data.common,
             &payload_commitment,
         ) {
-            Ok(inner) => inner.is_ok(),
-            Err(_) => false,
+            Ok(Ok(_)) => true,
+            _ => false,
         }
     }
 
--- a/crates/task-impls/src/consensus/mod.rs
+++ b/crates/task-impls/src/consensus/mod.rs
@@ -179,14 +179,14 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> ConsensusTaskSt
         // Validate the VID share.
         // NOTE: `verify_share` returns a nested `Result`, so we must check both the inner
         // and outer results
-        match vid_scheme(self.quorum_membership.total_nodes()).verify_share(
-            &disperse.data.share,
-            &disperse.data.common,
-            &payload_commitment,
-        ) {
-            Ok(Ok(_)) => true,
-            _ => false,
-        }
+        matches!(
+            vid_scheme(self.quorum_membership.total_nodes()).verify_share(
+                &disperse.data.share,
+                &disperse.data.common,
+                &payload_commitment,
+            ),
+            Ok(Ok(()))
+        )
     }
 
     /// Publishes a proposal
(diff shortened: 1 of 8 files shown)


jparr721 (approved):
Great find!!

crates/task-impls/src/quorum_vote/mod.rs
```diff
@@ -605,16 +605,19 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> QuorumVoteTaskS
...
+                    &disperse.data.share,
+                    &disperse.data.common,
+                    &payload_commitment,
+                ) {
+                    Ok(inner) => {
+                        if inner.is_err() {
+                            return;
+                        }
+                    }
+                    Err(_) => return,
```
ss-es, line 620: Can we simplify this to `Ok(Err(_)) | Err(_) => { return; }`?

crates/task-impls/src/consensus/mod.rs
```diff
@@ -177,19 +177,16 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> ConsensusTaskSt
...
-            return false;
+        // NOTE: `verify_share` returns a nested `Result`, so we must check both the inner
+        // and outer results
+        match vid_scheme(self.quorum_membership.total_nodes()).verify_share(
+            &disperse.data.share,
+            &disperse.data.common,
+            &payload_commitment,
+        ) {
+            Ok(inner) => inner.is_ok(),
+            Err(_) => false,
```
ss-es, line 188: Similar to below this could be `Ok(Ok(_)) => true` and ` _ => false`, but I think this one is also fine as written

crates/types/src/vid.rs
```diff
@@ -231,6 +231,12 @@ impl VidScheme for VidSchemeType {
     fn get_multiplicity(common: &Self::Common) -> u32 {
         <Advz as VidScheme>::get_multiplicity(common)
     }
+
+    /// Helper function for testing only
+    #[cfg(feature = "test-srs")]
```
ss-es, line 236: I don't think we ever run tests with `test-srs` -- it doesn't seem to even be in the justfile. Is this just for downstream users like the sequencer?

crates/testing/tests/tests_1/consensus_task.rs
```diff
@@ -602,3 +602,79 @@ async fn test_vid_disperse_storage_failure() {
 
     run_test![inputs, consensus_script].await;
 }
+
+/// Tests that VID shares that return validation with an Ok(Err) result
+/// are correctly rejected
+#[cfg(test)]
+#[cfg_attr(async_executor_impl = "tokio", tokio::test(flavor = "multi_thread"))]
+#[cfg_attr(async_executor_impl = "async-std", async_std::test)]
+#[cfg(feature = "test-srs")]
```
ss-es, line 611: Is it possible to write this test without the feature gate so that it runs in CI? We can run it with the feature gate, but it would take more work and we'd have to add it separately

ss-es (commented):
Looks good to me overall! 

Just some minor comments + questions about the feature gate

crates/types/src/vid.rs
ss-es: Oops, ignore this one -- I see it was used in the test which was also feature gated

crates/testing/tests/tests_1/consensus_task.rs
elliedavidson: I feature gated it because it relies on changes in Jellyfish that may not be merged for a while.  So I don't think it makes sense to run it in CI until jellyfish is updated.

crates/types/src/vid.rs
elliedavidson: Yes, that's intentional that we never run with this feature.  It is a feature to match the 'test-srs' feature in jellyfish.

crates/testing/tests/tests_1/consensus_task.rs
elliedavidson: Also, as I suggested above, it may make sense just to merge the fix for now, and wait to merge the test until later.

crates/task-impls/src/quorum_vote/mod.rs
elliedavidson: Done!

ss-es (approved):
Looks good to me!

Issue #3608, opened 2024-08-23
[Byzantine testing] - View Lagging node


lukeiannucci:
### What is this task and why do we need to work on it?

Write a test where we have one node that is constantly behind by 1 view or more.  When node recieves an event for current view return the lagged event instead

### What are the acceptance criteria to close this issue?

Integration test where a node is constantly lagging behind

PR #3609, opened 2024-08-23
[Byzantine testing] - View Lagging node

lukeiannucci:
Closes (https://github.com/EspressoSystems/HotShot/issues/3608)

### This PR: 

Adds a byzantine test where we have one node that is constantly behind by a a configurable set of views on the receive handle. When node receives an event for current view it will return an event that is cached.
Also this PR cleans up where we keep the Byzantine tests.

### Key places to review: 

`ViewDelay` struct and how it implements `EventTransformerState` to see the logic of how we handle delayed views. Also `/crates/testing/src/byzantine` this is where i moved the byzantine logic to. And `crates/testing/tests/tests_2` is where byzantine tests are placed.


crates/testing/tests/tests_2/byzantine_tests.rs
```diff
@@ -0,0 +1,199 @@
...
+        let behaviour = Rc::new(|node_id| {
+                let view_delay = ViewDelay {
+                    number_of_views_to_delay: 2,
+                    received_events: HashMap::new(),
+                    vote_rcv_count: HashMap::new(),
+                    num_nodes_with_stake: 10,
+                    stop_view_delay_at_view_number: 50,
+                };
+                match node_id {
+                    8 => Behaviour::Byzantine(Box::new(view_delay)),
```
bfish713, line 169: I think in this case we want to test with many nodes having this (maybe with different delays as well)

crates/testing/src/byzantine/byzantine_behaviour.rs
```diff
@@ -0,0 +1,314 @@
...
+                            event_name: hotshot_event_name.to_string(),
+                            view_number: view_num,
+                            cached_event: event.clone(),
+                        });
+                    }
+                }
+            }
+        }
+        None
+    }
```
bfish713, line 237: I don't think this should be doing any string stuff at all, we should just pattern match on the events that have views or else this is going to be very brittle

crates/testing/src/byzantine/byzantine_behaviour.rs
```diff
@@ -0,0 +1,314 @@
...
+
+    fn compare_views(current_view: u64, cached_view: u64, number_of_views_to_delay: u64) -> bool {
+        if current_view <= cached_view {
+            return false;
+        }
+
+        if current_view - cached_view > number_of_views_to_delay {
+            return true;
+        }
```
bfish713, line 246: I think if you use saturating_sub instead of just `-` this function simplifies to just this inequality

crates/testing/src/byzantine/byzantine_behaviour.rs
```diff
@@ -0,0 +1,314 @@
...
+                    }
+                    return cached_event;
+                }
+                Entry::Vacant(vacant) => {
+                    vacant.insert(event_info);
+                }
+            }
+        }
+        event.clone()
+    }
```
bfish713, line 300: Something feels off here.  I think we basically just storing all events from previous views and then sending them after we've delayed enough views.  Why do we have to do all this logic with the votes?  

Maybe I'm missing something but I feel like the logic can simply be, have a map from view to vec of events.  if the event has a view number push it to the vec for that view, return all events from the view - num views to delay.  If it has no view number, either add it to the highest views vec, or just return that event.

If we did that we can get rid of `HotShotEventInfo` (which I think we should do anyway)

crates/testing/src/byzantine/byzantine_behaviour.rs
```diff
@@ -0,0 +1,314 @@
...
+
+        if current_view - cached_view > number_of_views_to_delay {
+            return true;
+        }
+
+        false
+    }
+
+    async fn handle_event(&mut self, event: &HotShotEvent<TYPES>) -> HotShotEvent<TYPES> {
+        if let Some(event_info) = self.extract_hotshot_event_info(event) {
```
bfish713, line 252: All this really does is get the view number for the `event` I think it's probably best to just implement a `fn view_number(&self) -> Option<TYPES::Time>` on the `HotshotEvent` enum

I'll looks a lot like `fmt`

bfish713 (changes requested):

bfish713:
I'm not sure this does what it's supposed to do, my read of the logic is that is delays each event type by a number of views.  So if we get a proposal for view 2 with a delay of 1 we'll actually broadcast the proposal for 1 but we wouldn't broadcast a DAC or VID share for view 1. With votes this is really awkward because it means at minimum the voting is delayed until our next leadership because that's the next time we'll get a vote

crates/testing/src/byzantine/byzantine_behaviour.rs
```diff
@@ -0,0 +1,232 @@
...
+                return correct_event;
+            }
+
+            // add current view or push event to the map if view number has been added
+            let events_for_current_view = self.events_for_view.entry(view_number).or_default();
+            events_for_current_view.push(event.clone());
+
+            // ensure we are actually able to lookback enough views
+            let view_diff = (*view_number).saturating_sub(self.number_of_views_to_delay);
+            if view_diff >= self.number_of_views_to_delay {
```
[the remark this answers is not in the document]
lukeiannucci: yes 100% good catch

crates/testing/src/byzantine/byzantine_behaviour.rs
```diff
@@ -0,0 +1,232 @@
...
+            // add current view or push event to the map if view number has been added
+            let events_for_current_view = self.events_for_view.entry(view_number).or_default();
+            events_for_current_view.push(event.clone());
+
+            // ensure we are actually able to lookback enough views
+            let view_diff = (*view_number).saturating_sub(self.number_of_views_to_delay);
+            if view_diff > 0 {
+                return match self
+                    .events_for_view
+                    .get(&<TYPES as NodeType>::Time::new(view_diff))
```
ss-es, line 218: I feel like this should be a `remove` not a `get`, shouldn't it? Otherwise you'll return all events for a given view multiple times

ss-es (commented):
Not sure if there was anything specific you wanted me to check, but looks fine to me overall - just one minor comment about `remove` (not 100% sure if that's what the method is) vs. `get`.

crates/testing/src/byzantine/byzantine_behaviour.rs
lukeiannucci: yes good point, we should only need to return these events once

lukeiannucci: if we remove then the next event on the same view will return "correct_event" which we dont want

lukeiannucci: actually, it was flaky returning all the events for every receive, so i added above logic to return remove all events for a given view

crates/testing/src/byzantine/byzantine_behaviour.rs
```diff
@@ -0,0 +1,233 @@
...
+                return correct_event;
+            }
+
+            // add current view or push event to the map if view number has been added
+            let events_for_current_view = self.events_for_view.entry(view_number).or_default();
+            events_for_current_view.push(event.clone());
+
+            // ensure we are actually able to lookback enough views
+            let view_diff = (*view_number).saturating_sub(self.number_of_views_to_delay);
+            if view_diff > 0 {
```
bfish713, line 215: Do we ever send view 0 events?

bfish713 (approved):

crates/testing/src/byzantine/byzantine_behaviour.rs
lukeiannucci: afaik we do not

PR #3613, opened 2024-08-26
Update sqlx again

rob-maron:
Gets rid of audit warning for good this time, as `0.8.1` is finally out


bfish713 (approved):

bfish713:
Looks like it's still pulling in 0.8.0?

rob-maron:
> Looks like it's still pulling in 0.8.0?

Yep, didn't realize I had to update the lock in HS too. Done now

PR #3614, opened 2024-08-26
Change gossip params

rob-maron:
This setup means we will gossip about the last 30s of messages. In Cappuccino it is much higher (100), but realistically we will only want to gossip about the last view.


bfish713:
Does setting this to a factor of the configured view timeout make sense?

rob-maron:
> Does setting this to a factor of the configured view timeout make sense?

Sure, what do you think about double?

PR #3615, opened 2024-08-27
Lr/vote collector

### LINKED ISSUE #3564: [AUDIT] - Old Consensus - fix vote collector spawning

The task is to fix how we spawn the vote collectors. Currently we have only one vote collector per vote type. Whenever we receive a vote with a higher view number than the current vote collector, we overwrite the existing vote collector and spawn a new one. E.g. for quorum vote:

```
                let mut collector = self.vote_collector.write().await;

                if collector.is_none() || vote_view_number > collector.as_ref().unwrap().view {
                    (...)
                    *collector = create_vote_accumulator::<
                        TYPES,
                        QuorumVote<TYPES>,
                        QuorumCertificate<TYPES>,
                    >(&info, event, &event_stream)
                    .await;
                } else {
                    (...)
                }
```

_No response_

_No response_

Each view should have its own vote collector. The vote collectors should process votes for each view independently. The vote collector should be garbage collected.

_No response_

lukaszrzasik:
Closes #3564

### This PR: 

* Adds a `dishonest_voting` test that shows the original issue.
* Adds a wrapper `NetworkEventTaskStateModifier` which can change `NetworkEventTaskState`'s behaviour for testing purposes.
* Old consensus, new consensus, DA and Upgrade tasks now store a BTreeMap of vote collector tasks. When a vote with a higher view number arrives, it doesn't overwrite the existing vote collector for a different view. Those collectors can accumulate votes for different views in parallel. When the threshold is reached the collector is removed and all the collector for the older views are removed as well.
* Creates a helper function `handle_vote` to avoid code duplication

### This PR does not: 

Does not change vote collecting for ViewSync tasks as these differ in the vote collecting logic. As far as I understand they are not susceptible to the original issue as well.

### Key places to review: 

Changes in the `crates/task-impls/src/vote_collection.rs` file. Especially the garbage collecting logic. I'm not sure about it.
`crates/task-impls/src/network.rs` file for the new testing wrapper.

### How to test this PR:

All CI tests should pass, including the new test.


crates/task-impls/src/consensus/mod.rs
```diff
@@ -91,13 +85,12 @@ pub struct ConsensusTaskState<TYPES: NodeType, I: NodeImplementation<TYPES>, V:
     /// Membership for DA committee Votes/certs
     pub da_membership: Arc<TYPES::Membership>,
 
-    /// Current Vote collection task, with it's view.
-    pub vote_collector:
-        RwLock<VoteCollectorOption<TYPES, QuorumVote<TYPES>, QuorumCertificate<TYPES>, V>>,
+    /// A map of `QuorumVote` collector tasks.
+    pub vote_collectors: VoteCollectorsMap<TYPES, QuorumVote<TYPES>, QuorumCertificate<TYPES>, V>,
```
lukeiannucci, line 89: why are locks no longer needed?

crates/task-impls/src/network.rs
```diff
@@ -286,156 +286,262 @@ impl<
...
-                        )),
-                        TransmitType::Broadcast,
-                    )
-                }
+        let Some((sender, message_kind, transmit)) =
+            self.parse_event(event, &mut maybe_action, membership).await
+        else {
+            return;
+        };
+        self.spawn_transmit_task(message_kind, membership, maybe_action, transmit, sender);
```
lukeiannucci, line 294: nit to get rid of else
```suggestion
        if let Some((sender, message_kind, transmit)) =
            self.parse_event(event, &mut maybe_action, membership).await {
            self.spawn_transmit_task(message_kind, membership, maybe_action, transmit, sender);
        };
```

crates/task-impls/src/network.rs
```diff
@@ -501,79 +607,123 @@ impl<
...
+        > NetworkEventTaskStateModifier<TYPES, V, COMMCHANNEL, S>
+    {
+        /// Handles the received event modifying it before sending on the network.
+        pub async fn handle(
+            &mut self,
+            event: Arc<HotShotEvent<TYPES>>,
+            membership: &TYPES::Membership,
+        ) {
+            let mut maybe_action = None;
+            let Some((mut sender, mut message_kind, mut transmit)) =
```
lukeiannucci, line 659: same as above comment to just remove the else

crates/task-impls/src/network.rs
```diff
@@ -501,79 +607,123 @@ impl<
...
-    async fn handle_vid_disperse_proposal(
-        &self,
-        vid_proposal: Proposal<TYPES, VidDisperse<TYPES>>,
-        sender: &<TYPES as NodeType>::SignatureKey,
-    ) -> Option<HotShotTaskCompleted> {
-        let view = vid_proposal.data.view_number;
-        let vid_share_proposals = VidDisperseShare::to_vid_share_proposals(vid_proposal);
-        let mut messages = HashMap::new();
+/// A module with test helpers
+pub mod test {
```
lukeiannucci, line 613: i dont know how much of a pain itll be / if it makes sense to somehow move this outside this file.

crates/hotshot/src/tasks/mod.rs
```diff
@@ -479,6 +501,57 @@ where
     fn get_task_name(&self) -> &'static str {
         std::any::type_name::<dyn EventTransformerState<TYPES, I, V>>()
     }
+
+    /// Adds the `NetworkEventTaskState` tasks possibly modifying them as well.
+    fn add_network_event_tasks(&self, handle: &mut SystemContextHandle<TYPES, I, V>) {
```
lukeiannucci, line 506: is there any reason we cant use the existing add_network_event_tasks method?

lukeiannucci (approved):
looks good to me

PR #3618, opened 2024-08-27
Clean up Libp2p

rob-maron:
#### 1. Gossipsub configuration
- Changes the parameters to use sane defaults
- Moves specification from the global network file to the function caller. This allows us to specify downstream in the sequencer.

#### 2. Miscellaneous
- Removes `NetworkNodeType` as we are treating most nodes as "bootstrap" nodes
- Removes a bunch of unused configuration parameters
- Changes terminology of `identity` to `keypair` to make it easier to understand
- Not sure why we have `Libp2pConfigFile`, but it was not pulling in the right `bootstrap_nodes`

### This PR does **not**:
Address any changes in the sequencer.

### Testing:

I have tested that the gossip changes will not break config deserialization, as they are purely subtractive

### DIFF
--- a/crates/hotshot/src/traits/networking/libp2p_network.rs
+++ b/crates/hotshot/src/traits/networking/libp2p_network.rs
@@ -70,7 +70,7 @@ use libp2p_networking::{
         MeshParams,
         NetworkEvent::{self, DirectRequest, DirectResponse, GossipMsg},
         NetworkNodeConfig, NetworkNodeConfigBuilder, NetworkNodeHandle, NetworkNodeReceiver,
-        NetworkNodeType, DEFAULT_REPLICATION_FACTOR,
+        DEFAULT_REPLICATION_FACTOR,
     },
     reexport::{Multiaddr, ResponseChannel},
 };
@@ -281,10 +281,8 @@ impl<TYPES: NodeType> TestableNetworkingImplementation<TYPES>
                             // the worst case of 7/2+3 > 5
                             mesh_n: (expected_node_count / 2 + 3),
                         }))
-                        .server_mode(true)
-                        .identity(libp2p_keypair)
+                        .keypair(libp2p_keypair)
                         .replication_factor(replication_factor)
-                        .node_type(NetworkNodeType::Bootstrap)
                         .bound_addr(Some(addr))
                         .to_connect_addrs(HashSet::default())
                         // setting to sane defaults
@@ -303,10 +301,8 @@ impl<TYPES: NodeType> TestableNetworkingImplementation<TYPES>
                             mesh_outbound_min: 4,
                             mesh_n: 8,
                         }))
-                        .server_mode(true)
-                        .identity(libp2p_keypair)
+                        .keypair(libp2p_keypair)
                         .replication_factor(replication_factor)
-                        .node_type(NetworkNodeType::Regular)
                         .bound_addr(Some(addr))
                         .to_connect_addrs(HashSet::default())
                         // setting to sane defaults
@@ -464,8 +460,7 @@ impl<K: SignatureKey + 'static> Libp2pNetwork<K> {
         .with_context(|| "Failed to sign DHT lookup record")?;
 
         config_builder
-            .server_mode(libp2p_config.server_mode)
-            .identity(keypair)
+            .keypair(keypair)
             .replication_factor(replication_factor)
             .bound_addr(Some(bind_address.clone()))
             .mesh_params(Some(MeshParams {
@@ -555,17 +550,11 @@ impl<K: SignatureKey + 'static> Libp2pNetwork<K> {
         let (mut rx, network_handle) = spawn_network_node::<K>(config.clone(), id)
             .await
             .map_err(Into::<NetworkError>::into)?;
-        // Make bootstrap mappings known
-        if matches!(
-            network_handle.config().node_type,
-            NetworkNodeType::Bootstrap
-        ) {
-            let addr = network_handle.listen_addr();
-            let pid = network_handle.peer_id();
-            let mut bs_cp = bootstrap_addrs.write().await;
-            bs_cp.push((pid, addr));
-            drop(bs_cp);
-        }
+
+        // Add our own address to the bootstrap addresses
+        let addr = network_handle.listen_addr();
+        let pid = network_handle.peer_id();
+        bootstrap_addrs.write().await.push((pid, addr));
 
         let mut pubkey_pid_map = BiHashMap::new();
         pubkey_pid_map.insert(pk.clone(), network_handle.peer_id());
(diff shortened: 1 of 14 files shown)


crates/hotshot/src/traits/networking/libp2p_network.rs
```diff
@@ -268,53 +267,19 @@ impl<TYPES: NodeType> TestableNetworkingImplementation<TYPES>
...
                 let replication_factor = NonZeroUsize::new(2 * expected_node_count / 3).unwrap();
-                let config = if node_id < num_bootstrap as u64 {
-                    NetworkNodeConfigBuilder::default()
-                        // NOTICE the implicit assumption that bootstrap is less
-                        // than half the network. This seems reasonable.
-                        .mesh_params(Some(MeshParams {
-                            mesh_n_high: expected_node_count,
-                            mesh_n_low: 5,
-                            mesh_outbound_min: 3,
-                            // the worst case of 7/2+3 > 5
```
rob-maron, line 281: This spot removes bootstrap node logic when starting tests. We don't need it because we have been treating every node as potentially "bootstrap"

crates/hotshot/src/traits/networking/libp2p_network.rs
```diff
@@ -555,17 +513,11 @@ impl<K: SignatureKey + 'static> Libp2pNetwork<K> {
...
-        ) {
-            let addr = network_handle.listen_addr();
-            let pid = network_handle.peer_id();
-            let mut bs_cp = bootstrap_addrs.write().await;
-            bs_cp.push((pid, addr));
-            drop(bs_cp);
-        }
+
+        // Add our own address to the bootstrap addresses
+        let addr = network_handle.listen_addr();
```
rob-maron, line 518: Same as above

crates/hotshot/src/traits/networking/libp2p_network.rs
```diff
@@ -268,53 +267,19 @@ impl<TYPES: NodeType> TestableNetworkingImplementation<TYPES>
                 )
                 .expect("Failed to sign DHT lookup record");
 
-                // we want the majority of peers to have this lying around.
+                // We want 2/3 of the nodes to have any given record in the DHT
                 let replication_factor = NonZeroUsize::new(2 * expected_node_count / 3).unwrap();
```
QuentinI, line 271: Should this be `div_ceil` maybe? To never fall on the side of < 2/3

crates/libp2p-networking/src/network/node.rs
```diff
@@ -177,74 +176,45 @@ impl<K: SignatureKey + 'static> NetworkNode<K> {
     #[instrument]
     pub async fn new(config: NetworkNodeConfig<K>) -> Result<Self, NetworkError> {
         // Generate a random `KeyPair` if one is not specified
-        let identity = if let Some(ref kp) = config.identity {
+        let keypair = if let Some(ref kp) = config.keypair {
```
QuentinI, line 179:
```suggestion
        let keypair = config.keypair.clone().unwrap_or_else(Keypair::generate_ed25519);
```

crates/libp2p-networking/src/network/node.rs
```diff
@@ -177,74 +176,45 @@ impl<K: SignatureKey + 'static> NetworkNode<K> {
...
-                    NetworkNodeType::Conductor => MeshParams {
-                        mesh_n_high: 21,
-                        mesh_n_low: 8,
-                        mesh_outbound_min: 4,
-                        mesh_n: 12,
-                    },
-                }
-            };
+            // Use the default mesh parameters if none are provided
+            let mesh_params = config.clone().mesh_params.unwrap_or_default();
```
QuentinI, line 205:
```suggestion
            let mesh_params = config.mesh_params.clone().unwrap_or_default();
```

crates/libp2p-networking/tests/common/mod.rs
```diff
@@ -190,78 +186,34 @@ pub async fn print_connections<K: SignatureKey + 'static>(handles: &[Arc<Network
...
+            Multiaddr::from_str(format!("/ip4/127.0.0.1/udp/{port}/quic-v1").as_str()).unwrap();
 
-    for j in 0..(num_of_nodes - num_bootstrap) {
-        let addr = Multiaddr::from_str("/ip4/127.0.0.1/udp/0/quic-v1").unwrap();
-        // NOTE use this if testing locally and want human readable ports
-        // let addr = Multiaddr::from_str(&format!(
-        //     "/ip4/127.0.0.1/udp/{}/quic-v1",
-        //     start_port + num_bootstrap + j
-        // )).unwrap();
         let regular_node_config = NetworkNodeConfigBuilder::default()
```
QuentinI, line 203: Can we rename it to just conifg if all nodes are the same now?

QuentinI (commented):
Nitpicked a bunch, but overall looks good!
No libp2p experience, so can't evaluate how good new parameters are though.

crates/hotshot/src/traits/networking/libp2p_network.rs
rob-maron: Sure, good idea

crates/libp2p-networking/tests/common/mod.rs
```diff
@@ -190,77 +186,33 @@ pub async fn print_connections<K: SignatureKey + 'static>(handles: &[Arc<Network
 pub async fn spin_up_swarms<S: Debug + Default + Send, K: SignatureKey + 'static>(
     num_of_nodes: usize,
     timeout_len: Duration,
-    num_bootstrap: usize,
 ) -> Result<Vec<(HandleWithState<S, K>, NetworkNodeReceiver)>, TestError<S>> {
     let mut handles = Vec::new();
     let mut bootstrap_addrs = Vec::<(PeerId, Multiaddr)>::new();
```
lukeiannucci, line 191: if all nodes are bootstrap now, does this make sense for a name? instead of `node_addrs` or something

lukeiannucci (commented):
this seems fine to me, ill leave for someone else to approve who can look closer at the parameter changes. just one small question/nit i left

crates/libp2p-networking/src/network/node.rs
```diff
@@ -177,74 +176,42 @@ impl<K: SignatureKey + 'static> NetworkNode<K> {
...
-                        mesh_n_low: 8,
-                        mesh_outbound_min: 4,
-                        mesh_n: 12,
-                    },
-                }
-            };
-
-            // Create a custom gossipsub
+            // Derive a `Gossipsub` config from our gossip config
             let gossipsub_config = GossipsubConfigBuilder::default()
```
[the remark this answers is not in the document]
rob-maron: Right, so the (hotshot) `GossipConfig` is meant to represent values that we _probably_ want to configure. I can't pass a built (libp2p) `GossipsubConfig` around because we do some fundamental (possibly breaking) changes like the `Blake3` override for the message ID. And I can't pass around just the (libp2p) `GossipsubConfigBuilder` because it isn't cloneable, so we can't build off of it

So we take from the default (libp2p) `GossipsubConfig` and audit the defaults with the user-configurable stuff, if that makes sense

bfish713 (approved):

PR #3619, opened 2024-08-27
Allow Proposal requesting from any DA node

### LINKED ISSUE #3603: [Libp2p] - Allow Proposal requesting from any DA node

We want to be able to make requests for proposals from any node in the DA committee. We need to update the existing Proposal request messages to broadcast to the DA committee instead of the view's leader.

_No response_

_No response_

New broadcast is added and working. Tests passing.

_No response_

jparr721:
Closes #3603 

### This PR: 
Lets the requester of a proposal broadcast to the entire DA committee to receive a proposal on a first-come, first-served basis.

### DIFF
--- a/crates/task-impls/src/consensus/handlers.rs
+++ b/crates/task-impls/src/consensus/handlers.rs
@@ -137,11 +137,6 @@ pub async fn create_and_send_proposal<TYPES: NodeType, V: Versions>(
         proposed_leaf.view_number(),
     );
 
-    consensus
-        .write()
-        .await
-        .update_last_proposed_view(message.clone())?;
-
     async_sleep(Duration::from_millis(round_start_delay)).await;
 
     broadcast_event(
--- a/crates/task-impls/src/consensus/handlers.rs
+++ b/crates/task-impls/src/consensus/handlers.rs
@@ -518,7 +518,7 @@ pub(crate) async fn handle_quorum_proposal_recv<
         .entry(proposal.data.view_number())
         .or_default()
         .push(async_spawn(
-            validate_proposal_safety_and_liveness(
+            validate_proposal_safety_and_liveness::<TYPES, I, V>(
                 proposal.clone(),
                 parent_leaf,
                 OuterConsensus::new(Arc::clone(&task_state.consensus.inner_consensus)),
@@ -529,6 +529,7 @@ pub(crate) async fn handle_quorum_proposal_recv<
                 task_state.output_event_stream.clone(),
                 task_state.id,
                 task_state.upgrade_lock.clone(),
+                Arc::clone(&task_state.storage),
             )
             .map(AnyhowTracing::err_as_debug),
         ));
--- a/crates/task-impls/src/consensus/mod.rs
+++ b/crates/task-impls/src/consensus/mod.rs
@@ -420,7 +420,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> ConsensusTaskSt
                     }
 
                     if let Err(e) = self.consensus.write().await.update_high_qc(qc.clone()) {
-                        tracing::error!("{e:?}");
+                        tracing::trace!("{e:?}");
                     }
                     debug!(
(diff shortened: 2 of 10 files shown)


crates/task-impls/src/consensus/mod.rs
```diff
@@ -420,7 +420,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> ConsensusTaskSt
                     }
 
                     if let Err(e) = self.consensus.write().await.update_high_qc(qc.clone()) {
-                        tracing::error!("{e:?}");
+                        tracing::trace!("{e:?}");
```
jparr721, line 423: This was intentional. This is a useless log and noisy as heck.

crates/task-impls/src/helpers.rs
```diff
@@ -505,37 +507,39 @@ pub async fn validate_proposal_safety_and_liveness<TYPES: NodeType, V: Versions>
...
-            broadcast_event(
-                Event {
-                    view_number,
-                    event: EventType::Error { error: Arc::new(e) },
-                },
-                &event_sender,
-            )
-            .await;
-        }
+    {
```
jparr721, line 510: This bit me below, I just wrapped it so that way the consensus handles in this file go out of scope. It's on the roadmap to refactor a lot of this during the dependency refactor rollout.

crates/task-impls/src/consensus/handlers.rs
```diff
@@ -137,11 +137,6 @@ pub async fn create_and_send_proposal<TYPES: NodeType, V: Versions>(
         proposed_leaf.view_number(),
     );
 
-    consensus
-        .write()
-        .await
-        .update_last_proposed_view(message.clone())?;
-
```
rob-maron, line 144: Out of curiosity, what's the reason for removing this line? Was it not doing anything?

crates/task-impls/src/quorum_proposal/handlers.rs
```diff
@@ -229,10 +229,6 @@ impl<TYPES: NodeType, V: Versions> ProposalDependencyHandle<TYPES, V> {
             proposed_leaf.view_number(),
         );
 
-        self.consensus
-            .write()
-            .await
-            .update_last_proposed_view(message.clone())?;
```
rob-maron, line 235: Same question as above

rob-maron (approved):
Looks good, one small question

bfish713 (changes requested):
I think there is a critical missing piece, the leader is the only one that stores the proposal to disc.  This means if all nodes were to restart and the high QC points to a proposal where the leader was not in DA we will be stuck permanently.  We need the DA nodes to store (`storage.append_proposal`) the proposal message as well.  Also the request should go to the leader of the view as well as the DA committee.

The restart tests work here because everyone is on the DA committee so the leader can always respond.

crates/task-impls/src/consensus/handlers.rs
bfish713: I think it's just being moved to to being stored when we receive the proposal

jparr721: Yeah I just centralized where it happens instead of having it happen in two separate spots. Also, this would only get stored when the node sends a proposal, so no one else would have access to it. By moving it here, everyone can now store it.

crates/types/src/traits/network.rs
```diff
@@ -84,6 +84,8 @@ pub enum TransmitType<TYPES: NodeType> {
     Broadcast,
     /// broadcast to DA committee
     DaCommitteeBroadcast,
+    /// broadcast to the leader and the DA
+    DaCommitteeBroadcastAndLeader(TYPES::SignatureKey),
```
jparr721, line 88: Thoughts?

rob-maron: `DaCommitteeAndLeaderBroadcast`?

bfish713 (approved):

PR #3620, opened 2024-08-28
Update simple builder for marketplace

ss-es:
### This PR: 
Updates the simple builder to produce bundles for the marketplace


crates/example-types/src/node_types.rs
```diff
@@ -137,8 +137,8 @@ impl<TYPES: NodeType> NodeImplementation<TYPES> for Libp2pImpl {
 pub struct TestVersions {}
 
 impl Versions for TestVersions {
-    type Base = StaticVersion<0, 1>;
-    type Upgrade = StaticVersion<0, 2>;
+    type Base = StaticVersion<0, 3>;
```
elliedavidson, line 140: It would be nice to test the upgrade from v2 to v3 here :)

elliedavidson (commented):
LGTM!  Just one minor comment.

crates/task-impls/src/builder.rs
```diff
@@ -78,7 +78,7 @@ impl<TYPES: NodeType, Ver: StaticVersionType> BuilderClient<TYPES, Ver> {
     /// If the URL is malformed.
     pub fn new(base_url: impl Into<Url>) -> Self {
         Self {
-            inner: Client::builder(base_url.into().join("block_info").unwrap())
+            inner: Client::builder(base_url.into().join("bundle_info").unwrap())
```
QuentinI, line 81: This will make Client incompatible with legacy builders. Real marketplace builder also mounts its API on `block_info` actually, so that too.
Is it possible to mount two APIs on the same path in SimpleBuilderSource?

QuentinI (dismissed):

crates/example-types/src/node_types.rs
ss-es: This was just for testing -- I wanted to make it easy to run the examples off the branch. I'll be reverting this before merging, and relying on integration tests (rather than the examples)

crates/task-impls/src/builder.rs
ss-es: Yeah, this is the part I need to figure out how to do properly -- we can't put two on the same endpoint in a single builder.

But if the real builder is serving at `block_info`, maybe it would be better to just have multiple simple builders -- one for each version? I'll look at this again and continue on Zulip

crates/testing/src/block_builder/mod.rs
```diff
@@ -62,19 +65,19 @@ struct BlockEntry<TYPES: NodeType> {
     header_input: Option<AvailableBlockHeaderInput<TYPES>>,
 }
 
-/// Construct a tide disco app that mocks the builder API.
+/// Construct a tide disco app that mocks the builder API 0.1.
 ///
 /// # Panics
 /// If constructing and launching the builder fails for any reason
-pub fn run_builder_source<TYPES, Source>(
+pub fn run_builder_source_0_1<TYPES, Source>(
```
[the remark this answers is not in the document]
ss-es: This is just for the random builder -- I'd have to implement `v0_3::data_source::BuilderDataSource` for `RandomBuilderSource`, but didn't think we cared about the random builder (but I can try to do that if we want it)

crates/testing/src/block_builder/simple.rs
```diff
@@ -126,7 +131,67 @@ impl<TYPES: NodeType> ReadState for SimpleBuilderSource<TYPES> {
...
+                fee_amount,
+            )
+            .expect("Failed to sign fee!"),
+        };
+
+        let signature =
+            TYPES::BuilderSignatureKey::sign_bundle::<TYPES>(&self.priv_key, &transactions)
+                .unwrap();
+
+        Ok(Bundle {
```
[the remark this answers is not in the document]
ss-es: Done!

crates/task-impls/src/builder.rs
ss-es: Updated: we now have multiple clients at different module paths, which are defined as constants in `types/src/constants.rs`

crates/examples/infra/mod.rs
```diff
@@ -407,7 +407,7 @@ pub trait RunDa<
         let marketplace_config = MarketplaceConfig {
             auction_results_provider: TestAuctionResultsProvider::<TYPES>::default().into(),
             // TODO: we need to pass a valid fallback builder url here somehow
-            fallback_builder_url: url::Url::parse("http://localhost").unwrap(),
+            fallback_builder_url: config.config.builder_urls.first().clone(),
```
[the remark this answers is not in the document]
ss-es: I see, yes that makes sense! I'll take a look at this and try to get a follow-up PR by today

elliedavidson (approved):

elliedavidson:
Do we know why one of the tests failed?

ss-es:
> Do we know why one of the tests failed?

Oh good catch, I thought it was just flaky and reran it but forgot to look at the result! I'll fix it before merging

elliedavidson (commented):
This looks good to merge, but I see there is still a test failing.  I leave it up to @bfish713 whether it is okay to merge with this failure or not.  (It doesn't seem like the failure is related to these code changes).

PR #3621, opened 2024-08-28
[AUDIT][MEDIUM][O-1] Orchestrator slots may be trivially exhausted

jparr721:
Closes #<ISSUE_NUMBER>

### This PR: 
Ochestrator is susceptible to resource exhaustion on startup for a malicious attacker. Within the orchestrator’s post_ready method, there is a limited number of available slots (also, it logs using println, which is not good) and a motivated attacker with even a modest internet connection can easily absorb all of the slots as the increment for the connected nodes is unauthenticated and has no validation. This can only occur when the server is in manual mode.

This PR fixes the problem by doing three critical steps
1. Establishing a list of trusted public keys that are allowed to authenticate.
2. Refusing to register certain keys that are not in the list of public keys.
3. Loading the keys from a config file on startup.


crates/orchestrator/run-config.toml
```diff
@@ -39,22 +39,299 @@ seed = [
 ]
 start_delay_seconds = 0
 cdn_marshal_address = "127.0.0.1:9000"
+public_keys = [
+    [
```
ss-es, line 43: could we format this to a single line? (especially if we might add more keys to this list)

happy to do the formatting myself!

crates/orchestrator/src/lib.rs
```diff
@@ -83,6 +83,8 @@ struct OrchestratorState<KEY: SignatureKey> {
     peer_pub_ready: bool,
     /// A map from public keys to `(node_index, is_da)`.
     pub_posted: HashMap<Vec<u8>, (u64, bool)>,
+    /// A collection for storing `ready` public keys.
+    ready_posted: Vec<Vec<u8>>,
```
lukeiannucci, line 87: does it make sense to make this and public_keys hashset<vec<u8>>?

crates/orchestrator/src/lib.rs
```diff
@@ -416,8 +436,33 @@ where
...
+            .contains(&staked_pubkey.stake_table_entry.public_key())
+        {
+            return Err(ServerError {
+                status: tide_disco::StatusCode::FORBIDDEN,
+                message: "You are unauthorized to register with the orchestrator".to_string(),
+            });
+        }
+
+        // Have they already connected?
+        if self.ready_posted.contains(pubkey) {
```
lukeiannucci, line 456: can ready_posted be a hashset and we do something like:
```suggestion
        if !self.ready_posted.insert(pubkey) {
        ...
        }
```

then we also wouldnt need to push onto the vector after this

jparr721: Done

crates/orchestrator/src/lib.rs
```diff
@@ -636,7 +681,23 @@ where
     })?
     .post(
         "post_ready",
-        |_req, state: &mut <State as ReadState>::State| async move { state.post_ready() }.boxed(),
+        |req, state: &mut <State as ReadState>::State| {
+            async move {
+                let mut body_bytes = req.body_bytes();
```
ss-es, line 686: I think that this requires a change in `orchestrator/src/client.rs:445` to change how we form the request (since we now need to include our public key)

jparr721: Done!

crates/orchestrator/src/config.rs
```diff
@@ -491,10 +497,20 @@ pub struct NetworkConfigFile<KEY: SignatureKey> {
     /// random builder configuration
     #[serde(default)]
     pub random_builder: Option<RandomBuilderConfig>,
+    /// The list of public keys that are allowed to connect to the orchestrator
+    #[serde(default)]
+    pub public_keys: Vec<KEY>,
```
lukeiannucci, line 502: can this also be a hashset?

jparr721: Oof good catch

jparr721: I got the other one too

crates/orchestrator/src/config.rs
```diff
@@ -491,10 +498,20 @@ pub struct NetworkConfigFile<KEY: SignatureKey> {
...
+    #[serde(default)]
+    pub disable_registration_verification: bool,
 }
 
 impl<K: SignatureKey> From<NetworkConfigFile<K>> for NetworkConfig<K> {
     fn from(val: NetworkConfigFile<K>) -> Self {
+        if val.disable_registration_verification {
+            tracing::error!("REGISTRATION VERIFICATION IS TURNED OFF");
+        }
+
```
ss-es, line 514: I think this should probably go in `run_orchestrator` in `orchestrator/src/lib.rs`, but that's pretty minor

crates/orchestrator/src/lib.rs
```diff
@@ -317,7 +323,7 @@ where
             return Ok((*node_index, *is_da));
         }
 
-        if !self.accepting_new_keys {
+        if !self.config.disable_registration_verification && !self.accepting_new_keys {
```
ss-es, line 326: Was the `disable_registration_verification` check meant to go in the other conditional? (I don't think it should be here)

crates/orchestrator/src/lib.rs
```diff
@@ -328,7 +334,21 @@ where
...
         let node_index = self.pub_posted.len() as u64;
 
+        // Deserialize the public key
         let staked_pubkey = PeerConfig::<KEY>::from_bytes(pubkey).unwrap();
+
+        // Check if the node is allowed to connect
+        if !self
+            .config
+            .public_keys
+            .contains(&staked_pubkey.stake_table_entry.public_key())
```
ss-es, line 344: I think this should be `if !self.config.disable_verification && !self.config.public_keys.contains(..)`

(though the double negatives are confusing, maybe it should be `enable_verification`? that's very minor though)

crates/orchestrator/src/lib.rs
```diff
@@ -371,7 +391,7 @@ where
             }
         }
 
-        println!("Posted public key for node_index {node_index}");
+        tracing::info!("Posted public key for node_index {node_index}");
```
ss-es, line 394: can this be `error`? it's a start-up only log and it's very useful

crates/orchestrator/src/lib.rs
```diff
@@ -416,11 +436,29 @@ where
     }
 
     // Assumes nodes do not post 'ready' twice
-    // TODO ED Add a map to verify which nodes have posted they're ready
-    fn post_ready(&mut self) -> Result<(), ServerError> {
+    fn post_ready(&mut self, pubkey: &KEY) -> Result<(), ServerError> {
+        // If we have not disabled registration verification.
+        if !self.config.disable_registration_verification {
+            // Is this node allowed to connect?
+            if !self.config.public_keys.contains(pubkey) {
```
ss-es, line 443: I think we can replace both of these and just check `known_nodes_with_stake.contains(pubkey)` (a node has to register before posting ready)

ss-es (changes requested):
a few comments, mostly minor.

main change I'm requesting to fix is the disable verification check (if I'm understanding correctly)

crates/orchestrator/src/lib.rs
jparr721: This one was painful 😅 but we got it.

crates/orchestrator/src/lib.rs
```diff
@@ -416,11 +434,26 @@ where
     }
 
     // Assumes nodes do not post 'ready' twice
-    // TODO ED Add a map to verify which nodes have posted they're ready
-    fn post_ready(&mut self) -> Result<(), ServerError> {
+    fn post_ready(&mut self, peer_config: &PeerConfig<KEY>) -> Result<(), ServerError> {
+        // If we have not disabled registration verification.
+        if self.config.enable_registration_verification {
```
ss-es, line 439: I think we can skip the `enable_registration_verification` check here too -- nodes shouldn't be able to post ready unless they're on the `known_nodes_with_stake` list (this is the one we update internally when they register with us)

ss-es (approved):
just one minor comment, but overall looks great! thank you again for doing this!

PR #3648, opened 2024-09-04
[CATCHUP] Init Consensus with 2 views to prevent Double Vote/Propose

### LINKED ISSUE #3626: [CATCHUP] - Different Parameters for Start View and Last Voted View

See: https://github.com/EspressoSystems/espresso-sequencer/issues/1931

Basically we need to prevent double voting by having a last voted view where a node will not vote for this view, or previous views, but also start the node in the same view it was shutdown in so that it can enter view sync for the same view.

Change the parameters of `HotShotInitializer` and update the sequencer
Add a variable to the consensus task/s for last_voted_view and check it before voting
Update `TestStorage` to be the same as the sequener for `record_action` and use what's stored there for restart.

_No response_

Restart tests using the same initializer values the sequencer would have

_No response_

bfish713:
Closes #3626 

### This PR: 

This PR adds a new start parameter to consensus which is the view we last "actioned" in.  An action is voting or proposing for DA/Quorum.  The PR also enforces that we don't send double actions by checking in the network task if we've already sent something for the view we are about to transmit a vote/proposal to or a higher view.  This means if we propose in view 10 we can't then propose for view 9.  We should never want to do this.

The PR also adds the consensus shared state to the byzantine behaviors so that they can reset the last actioned view so their byzantine sends are not blocked.
### This PR does not: 
This PR doesn't apply to DA votes because after view sync we can often legitimately get a proposal for view n and n+1 in either order.  If we handle the n+1 view first we don't want to not vote for view n

### Key places to review: 
-Changes to the `HotshotInitializer`
- Changes in `network.rs`
- Changes to `Consensus`

### DIFF
--- a/crates/types/src/consensus.rs
+++ b/crates/types/src/consensus.rs
@@ -244,6 +244,17 @@ struct HotShotActionViews<T: ConsensusTime> {
     da_vote: T,
 }
 
+impl<T: ConsensusTime> Default for HotShotActionViews<T> {
+    fn default() -> Self {
+        let genesis = T::genesis();
+        Self {
+            proposed: genesis,
+            voted: genesis,
+            da_proposed: genesis,
+            da_vote: genesis,
+        }
+    }
+}
 impl<T: ConsensusTime> HotShotActionViews<T> {
     /// Create HotShotActionViews from a view number
     fn from_view(view: T) -> Self {
@@ -490,6 +501,11 @@ impl<TYPES: NodeType> Consensus<TYPES> {
             return true;
         }
         false
+    }
+
+    /// reset last actions to genesis so we can resend events in tests
+    pub fn reset_actions(&mut self) {
+        self.last_actions = HotShotActionViews::default();
     }
 
     /// Update the last proposal.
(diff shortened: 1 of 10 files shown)


crates/types/src/consensus.rs
```diff
@@ -433,11 +476,43 @@ impl<TYPES: NodeType> Consensus<TYPES> {
...
+        };
+        if view > *old_view {
+            *old_view = view;
+            return true;
+        }
+        false
+    }
+
+    /// reset last actions to genesis so we can resend events in tests
+    pub fn reset_actions(&mut self) {
```
lukeiannucci, line 507: is there a way to gate this to tests only? or do we want to potentially use it elsewhere in the future

jparr721 (approved):
LGTM

crates/types/src/consensus.rs
bfish713: I tried to gate this for test but it causes a whole bunch of issues because lots of the byzantine code isn't gated for test etc.  I'll follow up with the gating because we should gate this

PR #3652, opened 2024-09-05
Enable Dependency Tasks by Default

jparr721:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/crates/hotshot/Cargo.toml
+++ b/crates/hotshot/Cargo.toml
@@ -8,7 +8,7 @@ version = { workspace = true }
 rust-version = { workspace = true }
 
 [features]
-default = ["docs", "doc-images"]
+default = ["docs", "doc-images", "dependency-tasks"]
 example-upgrade = ["hotshot-task-impls/example-upgrade"]
 gpu-vid = ["hotshot-task-impls/gpu-vid"]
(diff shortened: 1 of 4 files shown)


crates/hotshot/Cargo.toml
```diff
@@ -8,7 +8,7 @@ version = { workspace = true }
 rust-version = { workspace = true }
 
 [features]
-default = ["docs", "doc-images"]
+default = ["docs", "doc-images", "dependency-tasks"]
```
ss-es, line 11: Not a strong opinion but could we just remove the feature-gating entirely? Even without the feature, I think if we ever want to revert in the future we'd only have to change which consensus tasks we spawn in `hotshot/src/tasks/mod.rs`

bfish713: I think the issue would be that we can't keep running the legacy tests while it's still deployed in prod and we might need to hotfix it

lukeiannucci (commented):
just one question, are we going to have any double tests now? as i see there wont be any old consensus tests running (maybe im missing something) but in CI we still see for example
Build and Test / test (async-std, test-ci-3)
and the same one for dependency tasks wont these now be the same exact test?

PR #3653, opened 2024-09-05
Refactor memberships

rob-maron:


crates/example-types/src/node_types.rs
```diff
@@ -52,7 +52,7 @@ impl NodeType for TestTypes {
     type Transaction = TestTransaction;
     type ValidatedState = TestValidatedState;
     type InstanceState = TestInstanceState;
-    type Membership = GeneralStaticCommittee<TestTypes, Self::SignatureKey>;
+    type Membership = GeneralStaticCommittee<TestTypes>;
```
rob-maron, line 55: The signature type is already in `TestTypes`

crates/hotshot/src/lib.rs
```diff
@@ -91,14 +91,10 @@ pub struct MarketplaceConfig<TYPES: NodeType, I: NodeImplementation<TYPES>> {
...
-    /// Quorum Membership
+    /// The entire quorum
     pub quorum_membership: TYPES::Membership,
-    /// DA
+    /// The DA nodes
     pub da_membership: TYPES::Membership,
-    /// VID
-    pub vid_membership: TYPES::Membership,
-    /// View Sync
-    pub view_sync_membership: TYPES::Membership,
```
rob-maron, line 101: VID == view sync == quorum membership

crates/hotshot/src/traits/election/static_committee.rs
```diff
@@ -4,78 +4,128 @@
...
+    /// NOTE: This is currently a hack because the DA leader needs to be the quurm
+    /// leader but without voting rights.
+    eligible_leaders: Vec<<T::SignatureKey as SignatureKey>::StakeTableEntry>,
+
+    /// The nodes on the committee and their stake
+    stake_table: Vec<<T::SignatureKey as SignatureKey>::StakeTableEntry>,
+
+    /// The nodes on the committee and their stake, indexed by public key
+    indexed_stake_table:
+        BTreeMap<T::SignatureKey, <T::SignatureKey as SignatureKey>::StakeTableEntry>,
```
rob-maron, line 37: These are mostly renamed, but I added an index by key which we didn't have before

crates/hotshot/src/traits/election/static_committee.rs
```diff
@@ -4,78 +4,128 @@
...
+                .collect();
+
+        // Index the stake table by public key
+        let indexed_stake_table: BTreeMap<
+            TYPES::SignatureKey,
+            <TYPES::SignatureKey as SignatureKey>::StakeTableEntry,
+        > = members
+            .iter()
+            .map(|entry| (TYPES::SignatureKey::public_key(entry), entry.clone()))
+            .collect();
```
rob-maron, line 81: This (previously `create_election`) should be much simpler now

crates/hotshot/src/traits/election/static_committee.rs
```diff
@@ -4,78 +4,128 @@
...
-{
-    /// Clone the public key and corresponding stake table for current elected committee
-    fn committee_qc_stake_table(&self) -> Vec<PUBKEY::StakeTableEntry> {
-        self.committee_nodes_with_stake.clone()
+    /// Get the stake table for the current view
+    fn get_stake_table(
+        &self,
+    ) -> Vec<<<TYPES as NodeType>::SignatureKey as SignatureKey>::StakeTableEntry> {
+        self.stake_table.clone()
+    }
```
rob-maron, line 98: Renamed from `committee_qc_stake_table`

crates/hotshot/src/traits/election/static_committee.rs
```diff
@@ -4,78 +4,128 @@
...
+
+    /// Get all members of the committee for the current view
+    fn get_committee_members(
+        &self,
+        _view_number: <TYPES as NodeType>::Time,
```
rob-maron, line 103: Was previously `staked_committee/whole_committee`

crates/hotshot/src/traits/election/static_committee.rs
```diff
@@ -84,152 +134,53 @@ where
         feature = "fixed-leader-election"
     )))]
     /// Index the vector of public keys with the current view number
-    fn leader(&self, view_number: TYPES::Time) -> PUBKEY {
-        let index = usize::try_from(*view_number % self.all_nodes_with_stake.len() as u64).unwrap();
-        let res = self.all_nodes_with_stake[index].clone();
+    fn get_leader(&self, view_number: TYPES::Time) -> TYPES::SignatureKey {
+        let index = usize::try_from(*view_number % self.eligible_leaders.len() as u64).unwrap();
+        let res = self.eligible_leaders[index].clone();
```
rob-maron, line 139: More verbose with pulling from `eligible_leaders`

crates/hotshot/src/traits/election/static_committee.rs
```diff
@@ -84,152 +134,53 @@ where
...
-    #[allow(clippy::must_use_candidate)]
-    /// get the non-staked builder nodes
-    pub fn non_staked_nodes_count(&self) -> usize {
-        self.committee_nodes_without_stake.len()
-    }
-    #[allow(clippy::must_use_candidate)]
-    /// get all the non-staked nodes
-    pub fn non_staked_nodes(&self) -> Vec<PUBKEY> {
-        self.committee_nodes_without_stake.clone()
+        NonZeroU64::new(((self.stake_table.len() as u64 * 9) / 10) + 1).unwrap()
```
rob-maron, line 184: This is the same as it was before, changed it to use `self.stake_table`

crates/hotshot/src/traits/election/static_committee_leader_two_views.rs
```diff
@@ -1,202 +1,157 @@
-use std::{marker::PhantomData, num::NonZeroU64};
+// Copyright (c) 2021-2024 Espresso Systems (espressosys.com)
+// This file is part of the HotShot repository.
+
+// You should have received a copy of the MIT License
+// along with the HotShot repository. If not, see <https://mit-license.org/>.
+
```
rob-maron, line 6: This whole file is the same except for the `get_leader` function

crates/testing/src/helpers.rs
```diff
@@ -79,35 +79,25 @@ pub async fn build_system_handle<
...
             Topic::Global,
+            #[cfg(feature = "fixed-leader-election")]
             config.fixed_leader_for_gpuvid,
         ),
-        da_membership: TYPES::Membership::create_election(
-            known_nodes_with_stake.clone(),
-            config.known_da_nodes.clone(),
+        da_membership: TYPES::Membership::new(
+            all_nodes,
+            da_nodes,
```
rob-maron, line 98: Eligible leaders = all nodes, but DA nodes are the only ones with voting power

crates/testing/src/test_runner.rs
```diff
@@ -411,29 +410,22 @@ where
...
                     Topic::Global,
+                    #[cfg(feature = "fixed-leader-election")]
                     config.fixed_leader_for_gpuvid,
                 ),
-                da_membership: <TYPES as NodeType>::Membership::create_election(
-                    known_nodes_with_stake.clone(),
-                    config.known_da_nodes.clone(),
+                da_membership: <TYPES as NodeType>::Membership::new(
+                    all_nodes,
+                    da_nodes,
```
rob-maron, line 426: Eligible nodes to be leader == all nodes, but DA nodes are the only ones with voting power

crates/orchestrator/run-config.toml
```diff
@@ -56,10 +56,8 @@ enable_registration_verification = true
 
 [config]
 num_nodes_with_stake = 10
-num_nodes_without_stake = 0
 start_threshold = [8, 10]
 staked_da_nodes = 10
-non_staked_da_nodes = 0
```
rob-maron, line 62: These are purely subtractive and will not break backwards-compatibility

crates/hotshot/src/traits/election/static_committee_leader_two_views.rs
```diff
@@ -1,202 +1,159 @@
...
+    /// Get the total number of nodes in the committee
     fn total_nodes(&self) -> usize {
-        self.committee_nodes_with_stake.len()
+        self.stake_table.len()
     }
 
+    /// Get the voting success threshold for the committee
     fn success_threshold(&self) -> NonZeroU64 {
-        NonZeroU64::new(((self.committee_nodes_with_stake.len() as u64 * 2) / 3) + 1).unwrap()
+        NonZeroU64::new(((self.stake_table.len() as u64 * 2) / 3) + 1).unwrap()
```
jparr721, line 147: Looks like you didn't add them originally, but since it's touched here I might ask: what are these magic numbers for? And can we perhaps do named constants instead? I feel like if I had to change this code I'd end up spending 20 minutes just working out what these mean

crates/hotshot/src/traits/election/static_committee_leader_two_views.rs
```diff
@@ -1,202 +1,159 @@
...
-                .map(|entry| entry.stake_table_entry)
-                .collect(),
-            committee_nodes_with_stake,
-            committee_nodes_without_stake,
-            fixed_leader_for_gpuvid,
-            _type_phantom: PhantomData,
-            committee_topic,
-        }
+    /// Index the vector of public keys with the current view number
+    fn get_leader(&self, view_number: TYPES::Time) -> TYPES::SignatureKey {
```
jparr721, line 133: IIRC I think we're using names for trivial getters, not `get_name`

crates/hotshot/src/traits/election/static_committee_leader_two_views.rs
```diff
@@ -1,202 +1,159 @@
...
+    ) -> std::collections::BTreeSet<<TYPES as NodeType>::SignatureKey> {
+        self.stake_table
+            .iter()
+            .map(TYPES::SignatureKey::public_key)
+            .collect()
     }
 
-    fn stake(
+    /// Get the stake table entry for a public key
+    fn get_stake(
```
jparr721, line 112: Same remark here as below about trivial getters, and everywhere else in this file.

jparr721 (commented):
couple of things

crates/hotshot/src/traits/election/static_committee_leader_two_views.rs
bfish713: Idk what I would name the numbers here, it's just 2/3 + 1

bfish713:
Since you are already in here I want to pose some questions, why do we even have a DA membership at all?  Why do we even have a DA leader concept at all, the leader is just the leader of a view.  For DA even we theoretically could just send votes to whoever sent the proposal regardless of view/leadership.  So can we just combine the memberships to just be 1 singular membership that is used everywhere and a DA committee is just a part of it

rob-maron:
> Since you are already in here I want to pose some questions, why do we even have a DA membership at all? Why do we even have a DA leader concept at all, the leader is just the leader of a view. For DA even we theoretically could just send votes to whoever sent the proposal regardless of view/leadership. So can we just combine the memberships to just be 1 singular membership that is used everywhere and a DA committee is just a part of it

Good question, I also thought this. I actually had gotten it to be almost removed in this PR, but didn't want to touch voting logic too much. I am 100% in favor of combining them though

bfish713 (approved):

PR #3655, opened 2024-09-06
Light client contract related changes

philippecamacho:
Part of https://github.com/EspressoSystems/espresso-sequencer/issues/1938

### This PR: 
The purpose of this PR is to update the GenericLightClientState struct (removing 4 fields) and introduce a new struct 
`GenericStakeTableState` to represent the stake table state (which is fixed). These changes are required to account for a refactoring of the Light Client contract which purpose is to remove all logic related to epochs. See https://github.com/EspressoSystems/espresso-sequencer/issues/1938 for more context.

### Key places to review: 
Only `crates/types/src/light_client.rs` as other changes are due to running `cargo fmt`.
 

As usual (see README.md)


bfish713 (approved):

crates/types/src/light_client.rs
```diff
@@ -174,57 +207,51 @@ impl<F: PrimeField> From<Vec<F>> for GenericPublicInput<F> {
...
     #[must_use]
     pub fn stake_amount_comm(&self) -> F {
-        self.0[7]
+        self.0[5]
+    }
+
+    /// Return the threshold
+    #[must_use]
+    pub fn threshold(&self) -> F {
+        self.0[6]
```
jparr721, line 255: All of these magic numbers are kind of tricky to reason about, can we just use a named  struct?

jparr721 (changes requested):

crates/types/src/light_client.rs
philippecamacho: @alxiong @alysiahuggins do you remember if there is any particular reason why we don't use a named struct?

alxiong: this is the decision made by @mrain and i recall the public input right now is simple enough to keep track of? 
i also personally prefer a named struct, less error prone. 

but i tried to make minimal changes for this task, so kept the legacy way. Can update in the future as tech debt.

PR #3657, opened 2024-09-06
[Byzantine Testing] - Byzantine coordination attack

lukeiannucci:
Closes [#3661](https://github.com/EspressoSystems/HotShot/issues/3661)

### This PR: 

In this PR we have a test with 3 byzantine nodes, one being a dishonest leader and the other 2 are being a dishonest voters. How it works is the following:
1.) Dishonest Leader will send a `justify_qc` from a previous view, after a defined amount of qc proposals.
2.) A byzantine replica will receive the proposal and still send a vote to the next honest leader
3.) The byzantine replica will also not send out a timeout for the proposal.

This will cause the next view to fail as other honest replicas will send their timeout votes to the honest leader

### DIFF
--- a/crates/testing/src/byzantine/byzantine_behaviour.rs
+++ b/crates/testing/src/byzantine/byzantine_behaviour.rs
@@ -114,6 +114,8 @@ pub struct DishonestLeader<TYPES: NodeType> {
     pub dishonest_at_proposal_numbers: HashSet<u64>,
     /// How far back to look for a QC
     pub view_look_back: usize,
+    /// Shared state flag of all view numbers we send bad proposal at
+    pub dishonest_proposal_view_numbers: Arc<RwLock<HashSet<TYPES::Time>>>,
 }
 
 /// Add method that will handle `QuorumProposalSend` events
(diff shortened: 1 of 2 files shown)


crates/testing/src/byzantine/byzantine_behaviour.rs
```diff
@@ -358,3 +358,73 @@ impl<TYPES: NodeType> std::fmt::Debug for DishonestVoting<TYPES> {
...
+        public_key: &TYPES::SignatureKey,
+        private_key: &<TYPES::SignatureKey as SignatureKey>::PrivateKey,
+        upgrade_lock: &UpgradeLock<TYPES, V>,
+        _consensus: Arc<RwLock<Consensus<TYPES>>>,
+    ) -> Vec<HotShotEvent<TYPES>> {
+        match event {
+            HotShotEvent::QuorumProposalRecv(_proposal, _sender) => {
+                if self
+                    .dishonest_at_vote_numbers
+                    .contains(&self.total_votes_from_node)
```
ss-es, line 393: I think this should work, but it feels a bit fragile to me -- is it possible to use the view number of the proposal instead? I'm worried about e.g. a node sending out a double proposal throwing the test off if things change down the line

alternatively, for this specific scenario we could probably even share state with a cloned `RwLock` e.g. leader writes its bad proposals to an `RwLock`, follower always votes on proposals in that lock

lukeiannucci: i like the idea of using an RwLock, view number also feels a bit fragile to me.

lukeiannucci: i added a shared hashset between all Byzantine replicas that will store any view we send a dishonest proposal on. The Byzantine voting replicas will just check the current view in the set to see if they should send a qc vote and dont send a timeout vote.

ss-es (approved):
Looks good to me

PR #3662, opened 2024-09-09
More marketplace testing

ss-es:
### This PR: 
- Adds integration tests for various configurations of the auction solver and builder not starting. 
- `TestBlockHeader` now contains the number of transactions in the block via `TestMetadata`, which allows us to verify that a nonzero number of transactions were committed in a leaf or across the entire test.
- Additionally, `TestBlockHeader` now contains a `random: u64` field, which is a number with no inherent meaning that is randomly generated on each invocation of `TestBlockHeader::new_legacy` or `TestBlockHeader::new_marketplace`. This is not directly related to marketplace testing, but should ensure CI fails if we send out multiple proposals in a given view.
- The consistency task now performs a validation check on the number of transactions committed in each view.
- A `nonempty_block_threshold` function is added to ensure we meet a threshold of nonempty blocks in the marketplace tests. Similarly, `nonempty_block_limit` is used to ensure that we propose with empty blocks where required (when the solver + fallback builder are both down, we want to make sure we're not taking bundles from the regular builders).

### Key places to review: 
- The added tests in `testing/tests/tests_3/test_marketplace.rs`
- Changes to `TestBlockHeader` and `TestMetadata` in `example-types/src/block_types.rs`.
- The additional transaction check in `testing/src/consistency_task.rs`.
- The `nonempty_block_threshold` and `nonempty_block_limit` functions in `testing/src/test_builder.rs` that we use to ensure tests end with a sufficient number of nonempty blocks.

### DIFF
--- a/crates/testing/tests/tests_1/quorum_proposal_task.rs
+++ b/crates/testing/tests/tests_1/quorum_proposal_task.rs
@@ -80,7 +80,11 @@ async fn test_quorum_proposal_task_quorum_proposal_view_1() {
         // We don't have a `QuorumProposalRecv` task handler, so we'll just manually insert the proposals
         // to make sure they show up during tests.
         consensus_writer
-            .update_saved_leaves(Leaf::from_quorum_proposal(&view.quorum_proposal.data));
+            .update_saved_leaves(
+                Leaf::from_quorum_proposal(&view.quorum_proposal.data),
+                &handle.hotshot.upgrade_lock,
+            )
+            .await;
     }
 
     // We must send the genesis cert here to initialize hotshot successfully.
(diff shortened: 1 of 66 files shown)


crates/testing/tests/tests_1/quorum_proposal_task.rs
```diff
@@ -576,7 +592,9 @@ async fn test_quorum_proposal_task_liveness_check() {
             SendPayloadCommitmentAndMetadata(
                 build_payload_commitment::<TestTypes>(&quorum_membership, ViewNumber::new(1)),
                 builder_commitment.clone(),
-                TestMetadata,
+                TestMetadata {
```
lukeiannucci, line 595: one concern i have here is a lot of places we set num_transactions to 0. if we add another field to the `TestMetadata` down the line and it also has some default value we will have to change all these files again. I dont know if its worth implementing default and changing everywhere we set to 0.

ss-es: I'm a little wary of `default` here because I don't think there's necessarily a "default" for what the metadata should be, especially in this case since it's actually tied to something else.

But I do think these tests should really try to take their fields from a block header (since everything is coming from a view generator we should have access to one for each view) instead of manually constructing it; let me see if I can fix that

bfish713 (approved):

PR #3664, opened 2024-09-09
[CDN] Implement hooks for DDoS protection

rob-maron:
## Reviews
The only file that *actually* needs review is `message_hook.rs`. Everything else is either removing unused code or moving it around.

## This PR:
Allows for "hooking" incoming messages from both brokers and users to implement custom logic. This will be the main source of DDoS protection measures.

- Breaks out `push_cdn_network.rs` into more relevant, smaller files
- Removes some unused/confusing `generator` arguments

### The current hooking functionality:
- Deduplicates messages to the same [topic or sender] (double votes, etc)

**Uses an SMA algorithm**. If the connection's average bytes per second (separate for direct/broadcast) is larger than the global average bytes per second (within a sliding window of 1,000), the user is disconnected.

## Testing
This PR has unit tests for expected functionality, and has been ran on the local demo to make sure it doesn't cause any issues.

### DIFF
--- a/crates/hotshot/src/traits/networking/push_cdn_network/definition/message_hook.rs
+++ b/crates/hotshot/src/traits/networking/push_cdn_network/definition/message_hook.rs
@@ -1,0 +1,96 @@
+use std::hash::Hasher;
+use std::marker::PhantomData;
+use std::num::NonZeroUsize;
+
+use anyhow::{Context, Result};
+use cdn_broker::reexports::def::hook::{HookResult, MessageHookDef};
+use cdn_broker::reexports::message::{Broadcast, Direct, Message as PushCdnMessage};
+use hotshot_types::traits::node_implementation::NodeType;
+use lru::LruCache;
+use twox_hash::xxh3::Hash64;
+
+#[derive(Clone)]
+/// The message hook for `HotShot` messages. Each user has a unique message hook.
+pub struct HotShotMessageHook<T: NodeType> {
+    /// The cache for message hashes. We use this to deduplicate a sliding window of
+    /// 100 messages.
+    message_hash_cache: LruCache<u64, ()>,
+
+    /// The phantom data for the node type
+    pd: PhantomData<T>,
+}
+
+impl<T: NodeType> HotShotMessageHook<T> {
+    /// Create a new `HotShotMessageHook`
+    pub fn new() -> Self {
+        Self {
+            message_hash_cache: LruCache::new(NonZeroUsize::new(100).unwrap()),
+            pd: PhantomData,
+        }
+    }
+
+    /// Process incoming broadcast messages from the user
+    fn process_broadcast_message(&mut self, broadcast: &mut Broadcast) -> Result<HookResult> {
+        // Calculate the hash of the message
+        let mut hasher = Hash64::default();
+        hasher.write(&broadcast.message);
+        for topic in &broadcast.topics {
+            hasher.write(&[*topic]);
+        }
+
+        // Make sure we have not already seen it
+        if self.message_hash_cache.put(hasher.finish(), ()).is_some() {
+            return Ok(HookResult::SkipMessage);
+        }
+
+        // TODO: Deserialize the message
+        // let message = Self::deserialize_message(&direct.message)?;
+
+        Ok(HookResult::ProcessMessage)
+    }
+
+    fn process_direct_message(&mut self, direct: &mut Direct) -> Result<HookResult> {
+        // Calculate the hash of the message
+        let mut hasher = Hash64::default();
+        hasher.write(&direct.message);
+        hasher.write(&direct.recipient);
+
+        // Make sure we have not already seen it
+        if self.message_hash_cache.put(hasher.finish(), ()).is_some() {
+            return Ok(HookResult::SkipMessage);
+        }
+
+        // TODO: Deserialize the message
+        // let message = Self::deserialize_message(&direct.message)?;
+
+        Ok(HookResult::ProcessMessage)
+    }
+
+    // TODO
+    // fn deserialize_message(message: &[u8]) -> Result<Message<T>> {
+    //     // Hack off the version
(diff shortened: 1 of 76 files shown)


crates/hotshot/src/traits/networking/push_cdn_network/definition/message_hook.rs
```diff
@@ -0,0 +1,482 @@
...
+use std::num::NonZeroUsize;
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::Arc;
+use std::time::Instant;
+
+use anyhow::{Context, Result};
+use cdn_broker::reexports::def::hook::{HookResult, MessageHookDef};
+use cdn_broker::reexports::message::{Broadcast, Direct, Message as PushCdnMessage};
+use lru::LruCache;
+use parking_lot::Mutex;
```
lukeiannucci, line 12: is this better performance than async_std mutex? just curious why we use this mutex

crates/hotshot/src/traits/networking/push_cdn_network/definition/message_hook.rs
```diff
@@ -0,0 +1,482 @@
...
+
+        // Calculate the hash of the message
+        let mut hasher = Hash64::default();
+        hasher.write(&direct.message);
+        hasher.write(&direct.recipient);
+
+        // Make sure we have not already seen it
+        if self.message_hash_cache.put(hasher.finish(), ()).is_some() {
+            return Ok(HookResult::SkipMessage);
+        }
```
lukeiannucci, line 274: nit: we are doing same hashing logic in `process_broadcast_message` and `process_direct_message` can we extract this to another method?

crates/hotshot/src/traits/networking/push_cdn_network/definition/message_hook.rs
```diff
@@ -0,0 +1,482 @@
...
+    //         Version::deserialize(&message).with_context(|| "failed to deserialize message")?;
+
+    //     // Deserialize the message
+    //     let message = Serializer::<StaticVersion<0, 1>>::deserialize_no_version(&message)
+    //         .with_context(|| "failed to deserialize message")?;
+
+    //     // Return the version and message
+    //     Ok((message, version))
+    // }
+}
```
lukeiannucci, line 294: i see some stuff commented out will this be used in the future or can we clean up?

crates/hotshot/src/traits/networking/push_cdn_network/definition/message_hook.rs
lukeiannucci: not sure if itll be easy since we are doing hash on Broadcast and Direct structs, wonder if we can make them params somehow if not no worries

crates/hotshot/src/traits/networking/push_cdn_network/definition/message_hook.rs
rob-maron: Yeah, related comment [here](https://github.com/EspressoSystems/HotShot/pull/3671#discussion_r1757538068). I can just go with the `std` mutex if we need

crates/hotshot/src/traits/networking/push_cdn_network/definition/message_hook.rs
rob-maron: Good idea, let me see if I can abstract this out

crates/hotshot/src/traits/networking/push_cdn_network/definition/message_hook.rs
lukeiannucci: no need more out of curiosity

crates/hotshot/src/traits/networking/push_cdn_network/definition/message_hook.rs
rob-maron: Yeah - we may want to use it in the future.

Right now the hooks are completely agnostic to HotShot because of serialization/versioning. If we want to know what kind of HotShot message the user sent, we have to basically run consensus to figure out what version we should be asking for

In the future this may become more important (e.g. only let a node vote once per view or something) but we're not doing it quite yet

rob-maron:
Drafting while I fix up

crates/hotshot/src/traits/networking/push_cdn_network/definition/message_hook.rs
```diff
@@ -0,0 +1,536 @@
...
+                &mut self.global_broadcast_bps,
+            ),
+            MessageType::Direct => (&mut self.local_direct_bps, &mut self.global_direct_bps),
+        };
+
+        // Get the current time
+        let now = Instant::now();
+
+        // Commit the sample if that interval has elapsed
+        if now.duration_since(sample.last_committed_time) >= self.sample_commit_interval {
```
lukeiannucci, line 204: even if we are in cool we still add to global sma? if we are in cooldown for a while wont bytes_sent for sample always be 0?

lukeiannucci: also seems below we have this logic probably want to remove it as we reset sample and second one shouldnt get hit

rob-maron: My reasoning for adding to the global while still in cooldown was that it should count for _something_ since the CDN still is burdened with receiving it. But now that you mention it, it does probably make more sense to increment it after

rob-maron: For adding to the global during cooldown: I wanted to avoid the chance that everyone is in cooldown and nobody is committing anything. Even if you send a really big message once and get put in jail I still think it should count

rob-maron: Good catch on the duplicate logic, will remove it

PR #3668, opened 2024-09-11
Bf/restart tune

bfish713:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/crates/testing/src/overall_safety_task.rs
+++ b/crates/testing/src/overall_safety_task.rs
@@ -249,9 +249,12 @@ impl<TYPES: NodeType, I: TestableNodeImplementation<TYPES>, V: Versions> TestTas
             expected_views_to_fail,
         }: OverallSafetyPropertiesDescription<TYPES> = self.properties.clone();
 
-        let num_incomplete_views = self.ctx.round_results.len()
-            - self.ctx.successful_views.len()
-            - self.ctx.failed_views.len();
+        let num_incomplete_views = self
+            .ctx
+            .round_results
+            .len()
+            .saturating_sub(self.ctx.successful_views.len())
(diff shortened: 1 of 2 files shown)


crates/testing/src/overall_safety_task.rs
```diff
@@ -249,9 +249,12 @@ impl<TYPES: NodeType, I: TestableNodeImplementation<TYPES>, V: Versions> TestTas
...
 
-        let num_incomplete_views = self.ctx.round_results.len()
-            - self.ctx.successful_views.len()
-            - self.ctx.failed_views.len();
+        let num_incomplete_views = self
+            .ctx
+            .round_results
+            .len()
+            .saturating_sub(self.ctx.successful_views.len())
+            .saturating_sub(self.ctx.failed_views.len());
```
ss-es, line 257: I'm not necessarily opposed to this but I think the fact that we need to make this change means we have something wrong in the test logic (I think it's almost better to fail the test with a panic here)

It looks like we can do `self.ctx.successful_views.insert()` and `self.ctx.failed_views.insert()` independently for the same view. Do we still need the `saturating_sub` if we e.g. check that a view is not in `failed_views` before inserting into `successful_views`, and remove it from `successful_views` before inserting into `failed_views`?

bfish713: The problem comes down to a node can timeout a view, then later decide the view.  That's were it gets inserted into these maps.  I think @lukeiannucci is going to update the logic to fix this so that if a view later succeeds we can remove it from the failed views map

lukeiannucci: yes im going to fix that in another pr

lukeiannucci (approved):
Looks good to me

Issue #3651, opened 2024-09-05
[TESTING] - Investigate Restart Test Failures


bfish713:
### What is this task and why do we need to work on it?

test_all_restart_one_da has been very flaky on CI
What we know so far:
1.) There is some timing issue on startup, causing some nodes to timeout on view 1 at first then eventually view passes
2.) When view 1 is timing out initially this causes some of our logic to fail and we get a subtraction overflow
3.) Lowering the next_view_timeout causes some random views to fail, see if we can find the bottleneck or bug

### What work will need to be done to complete this task?

Investigate the 3 issues above, for the subtraction overflow it is a pretty easy fix to get the test from crashing but still want to know why we are getting into this scenario with shorter view timeouts

### What are the acceptance criteria to close this issue?

No flaky restart tests

lukeiannucci:
pr1: some cleanup to prevent overflow
https://github.com/EspressoSystems/HotShot/pull/3668#pullrequestreview-2298091366

PR #3669, opened 2024-09-11
[TESTING] - Fix Startup / Restart Test Failures

lukeiannucci:
Closes (https://github.com/EspressoSystems/HotShot/issues/3651)

### This PR: 

We noticed with certain flaky catchup tests that a node may come up before other nodes and wait for a proposal. The node wont receive the proposal in time then timeout the view, and add it to `failed_views` in our test harness. Later, the node will receive the proposal and decide upon the view, but in our test harness we never remove the previously timed out view from `failed_views`. So this PR will add some logic to do so.

### DIFF
--- a/crates/testing/src/overall_safety_task.rs
+++ b/crates/testing/src/overall_safety_task.rs
@@ -165,8 +165,19 @@ impl<TYPES: NodeType, I: TestableNodeImplementation<TYPES>, V: Versions> TestTas
                 let paired_up = (leaf_chain.to_vec(), (*qc).clone());
                 match self.ctx.round_results.entry(view_number) {
                     Entry::Occupied(mut o) => {
-                        o.get_mut()
-                            .insert_into_result(id, paired_up, maybe_block_size)
+                        let entry = o.get_mut();
+                        let leaf = entry.insert_into_result(id, paired_up, maybe_block_size);
+
+                        // Here we noticed is a node may start up and time out waiting for a proposal
+                        // So we add the timeout to failed_views, but eventually the proposal is received we decide on the view
+                        // If we do indeed have a view timeout for the node at this point we want to remove it
+                        entry.cleanup_previous_timeouts_on_view(
+                            &mut self.ctx.failed_views,
+                            &view_number,
+                            &(id as u64),
+                        );
+
+                        leaf
                     }
                     Entry::Vacant(v) => {
                         let mut round_result = RoundResult::default();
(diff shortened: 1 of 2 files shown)


crates/testing/src/overall_safety_task.rs
```diff
@@ -249,12 +260,9 @@ impl<TYPES: NodeType, I: TestableNodeImplementation<TYPES>, V: Versions> TestTas
             expected_views_to_fail,
         }: OverallSafetyPropertiesDescription<TYPES> = self.properties.clone();
 
-        let num_incomplete_views = self
-            .ctx
-            .round_results
-            .len()
-            .saturating_sub(self.ctx.successful_views.len())
-            .saturating_sub(self.ctx.failed_views.len());
+        let num_incomplete_views = self.ctx.round_results.len()
```
lukeiannucci, line 263: i think we want to set this back, also i wonder if i should do a check to see if we are hitting a subtraction overflow and fail the test. this way maybe next time it wont be as hard to debug as we couldnt see error in CI previously

ss-es (commented):
I think this looks right to me overall

one question: if I understand correctly, I think the failure check for a round needs to be fixed too. it seems like we currently fail a view as soon as 1/3 of the quorum says it fails. but that's probably not correct if nodes can send a quorum vote, then a timeout vote, but have the view eventually pass. (maybe the call to `view.update_status` just needs to lag 3 views behind?)

bfish713:
> I think this looks right to me overall
> 
> one question: if I understand correctly, I think the failure check for a round needs to be fixed too. it seems like we currently fail a view as soon as 1/3 of the quorum says it fails. but that's probably not correct if nodes can send a quorum vote, then a timeout vote, but have the view eventually pass. (maybe the call to `view.update_status` just needs to lag 3 views behind?)



> I think this looks right to me overall
> 
> one question: if I understand correctly, I think the failure check for a round needs to be fixed too. it seems like we currently fail a view as soon as 1/3 of the quorum says it fails. but that's probably not correct if nodes can send a quorum vote, then a timeout vote, but have the view eventually pass. (maybe the call to `view.update_status` just needs to lag 3 views behind?)

I don't think a node should ever timeout and vote for a view.  The only scenario where a node fails a view, then succeeds (by the testing logic) is if the other nodes vote for the view.  if 1/3 of the nodes timeout then the view will actually fail and never be decided (unless there is a bug)

ss-es:
Ah yes, I was mixing up the views -- agree with @bfish713 that this should be fine!

PR #3671, opened 2024-09-12
[TECH DEBT] Improve `ConnectedNetwork` - `recv_message`

### LINKED ISSUE #2558: [TECH_DEBT] - Don't Sleep When Waiting for Message

The network task does this:
```
    let direct_handle = async_spawn(async move {
        loop {
            let msgs = match network.recv_msgs(TransmitType::Direct).await {
                Ok(msgs) => Messages(msgs),
                Err(err) => {
                    error!("failed to receive direct messages: {err}");

                    // return zero messages so we sleep and try again
                    Messages(vec![])
                }
            };
            if msgs.0.is_empty() {
                async_sleep(Duration::from_millis(100)).await;
            } else {
                state.handle_messages(msgs.0).await;
            }
        }
    });
```
Ideally `recv_messages` would not be able to return an empty vector.
Sadly the web server implementation uses locked vectors for it's message queues so we can't simply `await` for the next message, we have to periodically check the queue.  

Either simply remove the sleep case and just await in a loop when we get rid of the webserver, or fix the webserver impl to use channels where it should.

_No response_

No sleeps in this code

_No response_

rob-maron:
Is part of #3672
Closes #2558 

### 1. Changes `recv_msgs` to `recv_message`, allowing us to receive a single message for each call. 

This simplifies the receiving layer and allows us to remove the `async_sleep` that we were using.

### 2. Improves the message receiving loop

Doesn't do the unfold/pin/stream anymore. Instead, selects between the shutdown stream and our `recv_messages` future. 

### 3. Improves performance of the `combined_network` loop by:

Changing it to use a synchronous lock. It will also now keep going until it receives something deserializable, instead of returning an empty message

## This PR does not: 

Make any further improvements to `ConnectedNetwork`

## Key places to review: 

`combined_network.rs`
`network.rs` for `handle_message()`

## Testing
Running the examples locally I get a 10x speedup (only tested with `combined` so far) due to the rewrite of the loop and the removal of the sleep. Now, the `combined_network` loop will keep receiving until it receives an uncached message instead of sleeping

### DIFF
--- a/crates/hotshot/src/traits/networking/combined_network.rs
+++ b/crates/hotshot/src/traits/networking/combined_network.rs
@@ -18,6 +18,8 @@ use std::{
     time::Duration,
 };
 
+use parking_lot::RwLock as PlRwLock;
+
 use async_broadcast::{broadcast, InactiveReceiver, Sender};
 use async_compatibility_layer::{
     art::{async_sleep, async_spawn},
@@ -68,7 +70,7 @@ pub struct CombinedNetworks<TYPES: NodeType> {
     networks: Arc<UnderlyingCombinedNetworks<TYPES>>,
 
     /// Last n seen messages to prevent processing duplicates
-    message_cache: Arc<RwLock<LruCache<u64, ()>>>,
+    message_cache: Arc<PlRwLock<LruCache<u64, ()>>>,
 
     /// How many times primary failed to deliver
     primary_fail_counter: Arc<AtomicU64>,
@@ -106,7 +108,7 @@ impl<TYPES: NodeType> CombinedNetworks<TYPES> {
 
         Self {
             networks,
-            message_cache: Arc::new(RwLock::new(LruCache::new(
+            message_cache: Arc::new(PlRwLock::new(LruCache::new(
                 NonZeroUsize::new(COMBINED_NETWORK_CACHE_SIZE).unwrap(),
             ))),
             primary_fail_counter: Arc::new(AtomicU64::new(0)),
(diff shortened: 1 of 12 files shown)


crates/task-impls/src/network.rs
```diff
@@ -102,109 +102,106 @@ pub struct NetworkMessageTaskState<TYPES: NodeType> {
...
+            MessageKind::Data(message) => match message {
+                DataMessage::SubmitTransaction(transaction, _) => {
                     broadcast_event(
-                        Event {
-                            view_number: TYPES::Time::new(1),
-                            event: EventType::ExternalMessageReceived(data),
-                        },
-                        &self.external_event_stream,
+                        Arc::new(HotShotEvent::TransactionsRecv(vec![transaction])),
+                        &self.internal_event_stream,
```
bfish713, line 184: I guess this was the whole reason for the batching of messages, to collect transactions.  Doesn't seem worth keeping to me

crates/hotshot/src/traits/networking/combined_network.rs
```diff
@@ -106,7 +108,7 @@ impl<TYPES: NodeType> CombinedNetworks<TYPES> {
 
         Self {
             networks,
-            message_cache: Arc::new(RwLock::new(LruCache::new(
+            message_cache: Arc::new(PlRwLock::new(LruCache::new(
```
bfish713, line 111: Why the switch to parking lot library?

bfish713 (approved):

crates/hotshot/src/traits/networking/combined_network.rs
rob-maron: The original was an async lock, which is slower and we would really only need if we were holding across `.await` points. 

`parking_lot` is kind of std++ when it comes to `RwLocks/Mutexes`, it performs better in most cases and is my goto for a sync lock

crates/task-impls/src/network.rs
rob-maron: I think it was also to potentially increase throughput by decreasing context switching, but we were not really making use of it in the way we could have. If we end up having better facilities in later for batching (e.g. a message type that is a vec of messages), I might consider reintroducing some form of batching

PR #3676, opened 2024-09-13
Lower upgrade threshold

ss-es:


shenkeyao (approved):
Approving this assuming it should be "upper" not "lower" in the PR title.

ss-es:
It should be lowering it in most cases! Note that it's now `floor(9n/10)` rather than `floor(9n/10) + 1`, unless the regular quorum threshold is higher 

This just makes it so that we don't need every single node when `n <= 10`

PR #3685, opened 2024-09-18
Use pre-defined stake-table in orchestrator

ss-es:
### This PR: 
Updates the orchestrator's verification mode to now use the given stake table and pre-defined DA committee. In this mode, `run-config.toml` must provide:

- the stake table key
- the light client `state_ver_key`
- the node's stake
- whether the node is a DA node

before startup. Note that this replaces the previous verification mode, meaning the previous behaviour of allowing nodes from a whitelist (but not necessarily requiring the whole list) no longer exists.

One (welcome?) side-effect is that each node's index is now just its position in `run-config.toml`.

### DIFF
--- a/crates/orchestrator/run-config.toml
+++ b/crates/orchestrator/run-config.toml
@@ -3,40 +3,7 @@ indexed_da = true
 transactions_per_round = 10
 transaction_size = 1000
 node_index = 0
-seed = [
-    0,
-    0,
-    0,
-    0,
-    0,
-    0,
-    0,
-    0,
-    0,
-    0,
-    0,
-    0,
-    0,
-    0,
-    0,
-    0,
-    0,
-    0,
-    0,
-    0,
-    0,
-    0,
-    0,
-    0,
-    0,
-    0,
-    0,
-    0,
-    0,
-    0,
-    0,
-    0,
-]
+seed = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
 start_delay_seconds = 0
 cdn_marshal_address = "127.0.0.1:9000"
 public_keys = [
@@ -60,7 +27,6 @@ public_keys = [
 
     { stake_table_key = "BLS_VER_KEY~PAAQNgOYfj3GiVX7LxSlkXfOCDSnNKZDqPVYQ_jBMxKzOCn0PXbqQ62kKPenWOmCxiCE7X158s-VenBna6MjHJgf61eBAO-3-OyTP5NWVx49RTgHhQf2iMTKk2iqK2gjnjZimBU135YU4lQFtrG-ZgRezwqkC5vy8V-q46fschIG", state_ver_key = "SCHNORR_VER_KEY~APKGX39-mOmApq6jMdIEiuuyddJ_k8xFeIwU1Zs2zShH1rI--eZR180Us5vqNWmK3zAidScvVcW4bAsOMHB3LPg", stake = 1, da = true }
 ]
-enable_registration_verification = true
 
 [config]
 num_nodes_with_stake = 10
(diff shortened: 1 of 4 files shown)


ss-es:
note: there are still some clippy errors. I'll handle them before merging (and after merging main in)

crates/orchestrator/run-config.toml
```diff
@@ -40,17 +40,26 @@ seed = [
...
-    "BLS_VER_KEY~PAAQNgOYfj3GiVX7LxSlkXfOCDSnNKZDqPVYQ_jBMxKzOCn0PXbqQ62kKPenWOmCxiCE7X158s-VenBna6MjHJgf61eBAO-3-OyTP5NWVx49RTgHhQf2iMTKk2iqK2gjnjZimBU135YU4lQFtrG-ZgRezwqkC5vy8V-q46fschIG",
-    "BLS_VER_KEY~96hAcdFZxQT8CEHcyV8j2ILJRsXagquENPkc9AwLSx3u6AE_uMupIKGbNJRiM99oFneK2vI5g1u61HidWeuTLRPM2537xAXeaO8e-wJYx4FaPKw_xTcLPrIm0OZT7SsLAMwFuqfMbDdKM71-RyrLwhff5517xXBKEk5Tg9iT9Qrr",
-    "BLS_VER_KEY~-pVi7j6TEBeG7ABata4uWWDRM2SrY8wWotWsGnTpIhnOVYJI_lNWyig6VJUuFmBsMS8rLMU7nDxDm8SbObxyA-SLFcr_jCkZqsbx8GcVQrnBAfjNRWuPZP0xcTDMu2IkQqtc3L0OpzbMEgGRGE8Wj09pNqouzl-xhPoYjTmD06Bw",
-    "BLS_VER_KEY~IUPSdnsNUHgNx_74ZhBPrICcDZ9Bp_DAt-6kFz8vSwJES2Vy1Ws8NJ1mxb9XGE1u13sw0FRe8kn5Ib3p2stbEtR_1Qgbuif6aoLrGaSUzy0MvwrO58u9kHZk3rXIuSAN7n4ok3-KKk2CmnBfx7fchFoqT56FXCd1EJ7XRrYj8wTh",
+    { stake_table_key = "BLS_VER_KEY~bQszS-QKYvUij2g20VqS8asttGSb95NrTu2PUj0uMh1CBUxNy1FqyPDjZqB29M7ZbjWqj79QkEOWkpga84AmDYUeTuWmy-0P1AdKHD3ehc-dKvei78BDj5USwXPJiDUlCxvYs_9rWYhagaq-5_LXENr78xel17spftNd5MA1Mw5U", state_ver_key = "SCHNORR_VER_KEY~lJqDaVZyM0hWP2Br52IX5FeE-dCAIC-dPX7bL5-qUx-vjbunwe-ENOeZxj6FuOyvDCFzoGeP7yZ0fM995qF-CRE", stake = 1, da = true },
```
rob-maron, line 43: This looks good to me, no way around having to supply all of this

crates/orchestrator/src/lib.rs
```diff
@@ -246,70 +252,11 @@ pub trait OrchestratorApi<KEY: SignatureKey> {
...
-                message: "Node index getter for key pair generation has reached capacity"
-                    .to_string(),
-            });
-        }
-        Ok(tmp_node_index)
-    }
-
-    #[allow(clippy::cast_possible_truncation)]
-    fn register_public_key(
+    fn register_unknown(
```
rob-maron, line 259: Any reason for the name change here?

crates/orchestrator/src/lib.rs
```diff
@@ -400,6 +334,137 @@ where
         Ok((node_index, added_to_da))
     }
 
+    fn register_from_list(
+        &mut self,
+        pubkey: &mut Vec<u8>,
+        _da_requested: bool,
```
rob-maron, line 340: I think we should consider checking if or if not the node should be DA, and only let them connect if it matches their request

crates/orchestrator/src/lib.rs
```diff
@@ -400,6 +334,137 @@ where
...
+
+        if usize::from(tmp_node_index) >= self.config.config.num_nodes_with_stake.get() {
+            return Err(ServerError {
+                status: tide_disco::StatusCode::BAD_REQUEST,
+                message: "Node index getter for key pair generation has reached capacity"
+                    .to_string(),
+            });
+        }
+        Ok(tmp_node_index)
+    }
```
rob-maron, line 451: Assuming this is all unchanged?

crates/orchestrator/src/lib.rs
```diff
@@ -768,8 +833,29 @@ where
         network_config.manual_start_password = env_password.ok();
     }
 
-    network_config.config.known_nodes_with_stake = vec![];
-    network_config.config.known_da_nodes = vec![];
+    if network_config.enable_registration_verification {
```
rob-maron, line 836: I know this wasn't introduced in this PR, but:

What do you think about enabling registration verification automatically if the list is nonzero? And having it be off if it's zero?

rob-maron (approved):
Thanks for doing this, approved but with a couple questions

crates/orchestrator/src/lib.rs
ss-es: oh yes, great idea!

I'll update the PR and remove the `enable_registration_verification` field altogether

crates/orchestrator/src/lib.rs
ss-es: there are some minor changes in `register_from_list`

we previously had two logic paths in `register_public_key`, depending on whether we were validating against the key list we had. these have now been split into `register_unknown` and `register_from_list`.

one difference is that `register_from_list` is essentially a no-op on the internal orchestrator state now. previously we would add nodes to the stake table as they register, but now we're just using our own list as the immutable stake table. so `register_from_list` just tells nodes their `node_id` and DA status, and then sends them straight to `post_ready` (which still waits for a threshold of nodes)

crates/orchestrator/src/lib.rs
ss-es: done!

note that this may require some (mild?) testing, just to make sure that the `is_da` flag is being correctly set in the sequencer. I think it may not have mattered previously

crates/orchestrator/src/lib.rs
ss-es: left a longer comment in another response that should explain this!

Issue #3617, opened 2024-08-27
[Libp2p] - Refactor VID Delayed Requester To Use Message Passing


jparr721:
### What is this task and why do we need to work on it?

The Request/Response behavior for proposals uses HotShotEvent message passing to get the proposals sent to and from the leader. We should do the same for VID. The key differentiation here, however, is the need for the VID request to include retries.

### What are the acceptance criteria to close this issue?

New code and tests are added.

Issue #3599, opened 2024-08-22
[CLEANUP] - Don't iterate Entire DA Committee to verify VID share


bfish713:
### What is this task and why do we need to work on it?

When we request the VID share we can received it from the DA committee but we don't put that as the sender.  The leader is always the sender, so to verify the message signature we check if the leader signed it or any of the DA committee without really knowing who did.

### What are the acceptance criteria to close this issue?

Remove the loop

PR #3688, opened 2024-09-20
[Libp2p] - Refactor VID Delayed Requester To Use Message Passing

lukeiannucci:
Closes https://github.com/EspressoSystems/HotShot/issues/3617 and https://github.com/EspressoSystems/HotShot/issues/3599

### This PR: 

Implements the HotShotEvent Request / Response framework for VID shares. We recently did this to retrieve QCs and now we are making it a standard. In the Request and Response task we now emit/receive new `HotShotEvent`s, those being:

`VidRequestSend`: This is when a node sends a request to the DA committee to retrieve a VID share for a specified view.
`VidRequestRecv`: This event is for members of the DA committee, they will receive the request and respond with the VID share for the view.
`VidResponseSend`: This event emits the VID share to the network task so we can send it back to the requesting node.
`VidResponseRecv`: The requesting node will wait in a background task for this response and then validate the VID share.

Previously we used a separate channel to request / receive VID shares. We also cleanup the old logic in this PR.

Another thing we do is remove the loop where we iterate over all DA committee members for a given view `VidShareRecv` events to verify . We now add the sender public key and use that so no need to check every member.

### Key places to review: 

I do most of the refactoring in `request.rs` and `response.rs`. I am also not using the other channel anymore and relying on the main network task to send these messages now.

Also on HS Event `VidShareRecv` make sure the logic looks correct where I remove the loop

### How to test this PR:

The `combined_network` tests will test this new logic

### DIFF
--- a/crates/hotshot/src/tasks/mod.rs
+++ b/crates/hotshot/src/tasks/mod.rs
@@ -36,7 +36,6 @@ use hotshot_types::{
     consensus::Consensus,
     constants::EVENT_CHANNEL_SIZE,
     message::{Message, UpgradeLock},
-    request_response::RequestReceiver,
     traits::{
         network::ConnectedNetwork,
         node_implementation::{ConsensusTime, NodeImplementation, NodeType},
@@ -80,11 +79,9 @@ pub async fn add_request_network_task<
 /// Add a task which responds to requests on the network.
 pub fn add_response_task<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions>(
     handle: &mut SystemContextHandle<TYPES, I, V>,
-    request_receiver: RequestReceiver,
 ) {
     let state = NetworkResponseState::<TYPES>::new(
         handle.hotshot.consensus(),
-        request_receiver,
         handle.hotshot.memberships.quorum_membership.clone().into(),
         handle.public_key().clone(),
         handle.private_key().clone(),
@@ -93,6 +90,7 @@ pub fn add_response_task<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versi
     handle.network_registry.register(run_response_task::<TYPES>(
         state,
         handle.internal_event_stream.1.activate_cloned(),
+        handle.internal_event_stream.0.clone(),
     ));
 }
 
@@ -563,9 +561,9 @@ pub async fn add_network_message_and_request_receiver_tasks<
     add_network_message_task(handle, &network);
     add_network_message_task(handle, &network);
 
-    if let Some(request_receiver) = network.spawn_request_receiver_task().await {
+    if let Some(_request_receiver) = network.spawn_request_receiver_task().await {
         add_request_network_task(handle).await;
-        add_response_task(handle, request_receiver);
+        add_response_task(handle);
     }
 }
 
--- a/crates/hotshot/src/tasks/mod.rs
+++ b/crates/hotshot/src/tasks/mod.rs
@@ -561,10 +561,8 @@ pub async fn add_network_message_and_request_receiver_tasks<
     add_network_message_task(handle, &network);
     add_network_message_task(handle, &network);
 
-    if let Some(_request_receiver) = network.spawn_request_receiver_task().await {
-        add_request_network_task(handle).await;
-        add_response_task(handle);
-    }
+    add_request_network_task(handle).await;
+    add_response_task(handle);
 }
 
 /// Adds the `NetworkEventTaskState` tasks.
--- a/crates/hotshot/src/tasks/mod.rs
+++ b/crates/hotshot/src/tasks/mod.rs
@@ -83,6 +83,7 @@ pub fn add_response_task<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versi
     let state = NetworkResponseState::<TYPES>::new(
         handle.hotshot.consensus(),
         handle.hotshot.memberships.quorum_membership.clone().into(),
+        handle.public_key().clone(),
         handle.private_key().clone(),
         handle.hotshot.id,
     );
--- a/crates/hotshot/src/tasks/mod.rs
+++ b/crates/hotshot/src/tasks/mod.rs
@@ -83,7 +83,6 @@ pub fn add_response_task<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versi
     let state = NetworkResponseState::<TYPES>::new(
         handle.hotshot.consensus(),
         handle.hotshot.memberships.quorum_membership.clone().into(),
-        handle.public_key().clone(),
         handle.private_key().clone(),
         handle.hotshot.id,
     );
--- a/crates/task-impls/src/network.rs
+++ b/crates/task-impls/src/network.rs
@@ -160,7 +160,6 @@ impl<TYPES: NodeType> NetworkMessageTaskState<TYPES> {
                             HotShotEvent::UpgradeVoteRecv(message)
                         }
                         GeneralConsensusMessage::VidRequested(message, sender) => {
-                            // error!("received");
                             HotShotEvent::VidRequestRecv(message, sender)
                         }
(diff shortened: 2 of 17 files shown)


crates/hotshot/src/tasks/mod.rs
```diff
@@ -546,9 +543,9 @@ pub async fn add_network_message_and_request_receiver_tasks<
     add_network_message_task(handle, &network);
     add_network_message_task(handle, &network);
 
-    if let Some(request_receiver) = network.spawn_request_receiver_task().await {
+    if let Some(_request_receiver) = network.spawn_request_receiver_task().await {
```
lukeiannucci, line 546: i dont think we will need this at all anymore

crates/task-impls/src/request.rs
```diff
@@ -69,8 +63,6 @@ pub struct NetworkRequestState<TYPES: NodeType, I: NodeImplementation<TYPES>> {
     /// The node's id
     pub id: u64,
     /// A flag indicating that `HotShotEvent::Shutdown` has been received
-    pub shutdown_flag: Arc<AtomicBool>,
-    /// A flag indicating that `HotShotEvent::Shutdown` has been received
     pub spawned_tasks: BTreeMap<TYPES::Time, Vec<JoinHandle<()>>>,
```
lukeiannucci, line 66: i dont think we will need this or shutdown flag, im spawning short lived tasks now

crates/task-impls/src/network.rs
```diff
@@ -522,6 +531,20 @@ impl<
                     .await;
                 None
             }
+            HotShotEvent::VidRequestSend(req, signature) => Some((
+                req.key.clone(),
+                MessageKind::<TYPES>::from_consensus_message(SequencingMessage::General(
+                    GeneralConsensusMessage::VidRequested(req.clone(), signature),
+                )),
+                TransmitType::DaCommitteeBroadcast,
```
lukeiannucci, line 539: i dont know if we want to broadcast to whole committee or do it like we did before, choose a member of the committee for a view, if no response and it times out try another member

crates/task-impls/src/request.rs
```diff
@@ -160,237 +159,106 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> TaskState for NetworkRequest
...
-                }
-                Err(_) => {
-                    warn!("Request to other node timed out");
-                }
-            }
-        }
-    }
-    /// Returns true if we got the data we wanted, or the view has moved on.
-    #[instrument(skip_all, target = "DelayedRequester", fields(id = self.id, view = *req.0))]
-    async fn cancel_vid(&self, req: &VidRequest<TYPES>) -> bool {
```
bfish713, line 341: I don't think we want to get rid of this, your changes basically means a task is spawned right away every time we get a proposal and never canceled until a node actually sends us the VID or it times out.  In the prior version we first delay, then check if we even need to send the request (and check again after each timeout) this means more often than not we don't request VID on the happy path.

crates/task-impls/src/network.rs
jparr721: For now, I don't necessarily see an issue with sending to the entire committee. I think doing a round-robin request-response the right way will require more work than is probably needed for this PR.

crates/task-impls/src/request.rs
```diff
@@ -160,237 +159,106 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> TaskState for NetworkRequest
...
+                HotShotEvent::VidRequestSend(request, signature).into(),
+                &sender,
+            )
+            .await;
 
-                return;
-            }
-        };
+            // start waiting
+            for _ in 0..4 {
```
jparr721, line 209: Why do we have 0..4 here? We could instead just wrap this in a `loop` statement or `while` like we do with the proposal request, then wrap *that* loop in an `async_timeout` that way it'll just be reaped if it's not done in time.

lukeiannucci: yea this was just testing, thats a good idea!

crates/task-impls/src/request.rs
```diff
@@ -160,237 +159,106 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> TaskState for NetworkRequest
...
-    };
-    Message {
-        sender: req.1.clone(),
-        kind: MessageKind::Data(DataMessage::RequestData(data_request)),
+    /// Sign the serialized version of the request
+    fn serialize_and_sign(
+        &self,
+        request: &ProposalRequestPayload<TYPES>,
+    ) -> Option<Signature<TYPES>> {
+        let Ok(data) = bincode::serialize(&request) else {
```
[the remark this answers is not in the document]
lukeiannucci: thanks for pointing this out! good idea

crates/task-impls/src/request.rs
```diff
@@ -219,140 +218,117 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> NetworkRequestState<TYPES, I
...
+                            false
                         }
-                        Err(e) => {
-                            error!("Failed to deserialize response: {e}");
+                    }),
+                )
+                .completed()
+                .await;
+
+                if let Some(hs_event) = event.as_ref() {
```
bfish713, line 273: Why can't this check be part of the callback in the event dependency above.  If it were you could just create the dependency and await it and get rid of the while loop.

crates/task-impls/src/response.rs
```diff
@@ -58,83 +48,63 @@ impl<TYPES: NodeType> NetworkResponseState<TYPES> {
...
-        match req.kind {
-            MessageKind::Data(DataMessage::RequestData(request)) => {
-                if !self.valid_sender(&sender) || !valid_signature::<TYPES>(&request, &sender) {
-                    let serialized_msg = match bincode::serialize(
-                        &self.make_msg(ResponseMessage::Denied),
-                    ) {
-                        Ok(serialized) => serialized,
-                        Err(e) => {
-                            tracing::error!("Failed to serialize outgoing message: this should never happen. Error: {e}");
+            match receiver.recv_direct().await {
```
bfish713, line 72: Shouldn't this loop also handle the proposal requests as well now?

crates/task-impls/src/network.rs
```diff
@@ -182,8 +187,31 @@ impl<TYPES: NodeType> NetworkMessageTaskState<TYPES> {
...
+                            SequencingMessage::Da(da_message) => {
+                                if let DaConsensusMessage::VidDisperseMsg(proposal) = da_message {
+                                    broadcast_event(
+                                        Arc::new(HotShotEvent::VidResponseRecv(sender, proposal)),
+                                        &self.internal_event_stream,
+                                    )
+                                    .await;
+                                }
+                            }
+                            SequencingMessage::General(_) => {}
```
bfish713, line 202: Should this also include responses for proposal? Or is that handled by a different message type?

crates/hotshot/src/tasks/mod.rs
bfish713: Yes as a follow up task to this we should clean this up.  We should completely remove `spawn_request_receiver_task` and the associated stuff in libp2p, as well as `request_data`

crates/task-impls/src/response.rs
lukeiannucci: yes i agree with this. i can move it here

crates/task-impls/src/network.rs
lukeiannucci: we have those `GeneralConsensusMessage`. Im not sure how easy it would be to refactor them into here. I need to think a little more on this

lukeiannucci: If we want to move them we can probably make a separate task

crates/hotshot/src/tasks/mod.rs
lukeiannucci: i will make a separate task for this

PR #3693, opened 2024-09-23
Randomize leader election

### LINKED ISSUE #3684: [DRB] - Add seeded randomized leader rotation

As a first step towards a full DRB-based randomized leader rotation, we should have an implementation of a seeded randomized leader rotation

_No response_

_No response_

Leader rotation is randomized with all tests passing, gated on a version upgrade

_No response_

ss-es:
Closes #3684

### This PR:
- Adds randomized (seeded by view) leader rotation as a separate type implementing the `Membership` trait. This logic previously existed but was gated behind a feature flag.
- Mostly removes feature flag gating around leader rotation logic. The `randomized-leader-election` feature is completely removed, while `fixed-leader-election` is now exclusively a feature in the `examples` crate.
- `test_success` and `test_all_restart` now also run with randomized leader rotations.

### Key places to review: 
- Are there other tests that should be run with randomized leader election?


crates/orchestrator/src/lib.rs
```diff
@@ -173,7 +152,6 @@ impl<KEY: SignatureKey + 'static> OrchestratorState<KEY> {
             transactions_per_round: self.config.transactions_per_round,
             transaction_size: self.bench_results.transaction_size_in_bytes,
             rounds: self.config.rounds,
-            leader_election_type: OrchestratorState::<KEY>::election_type(),
```
dailinsubjam, line 176: It would be better to leave this as it will show which `election_type` we use for benchmarking, but not an urgent thing if you'll add that back in with some small config changes.

dailinsubjam (commented):
Others LGTM! Copy-pasted from Zullip for future sequencer-benchmarking reference:

https://github.com/EspressoSystems/espresso-sequencer/blob/8151b2b54c36802d13ef7262a6d364176ae13509/sequencer/src/context.rs#L106-L110

the first argument to new is just the nodes that you want to allow to be leader.

crates/orchestrator/src/lib.rs
ss-es: added back in in a slightly different form (taking the information from the nodes directly, rather than at compile-time in the orchestrator)

crates/hotshot/src/traits/election/randomized_committee.rs
```diff
@@ -0,0 +1,166 @@
...
+    fn committee_topic(&self) -> Topic {
+        self.committee_topic.clone()
+    }
+
+    /// Index the vector of public keys with the current view number
+    fn leader(&self, view_number: TYPES::Time) -> TYPES::SignatureKey {
+        let mut rng: StdRng = rand::SeedableRng::seed_from_u64(*view_number);
+        let randomized_view_number: usize = rng.gen();
+        let index = randomized_view_number % self.eligible_leaders.len();
+        let res = self.eligible_leaders[index].clone();
```
rob-maron, line 139: Question for my understanding: Is there a downside to doing `rng.gen_range` here?

crates/examples/infra/mod.rs
```diff
@@ -380,27 +380,29 @@ pub trait RunDa<
...
         // as the known da nodes
-        let da_membership = <TYPES as NodeType>::Membership::new(
-            all_nodes.clone(),
-            da_nodes,
-            Topic::Da,
-            #[cfg(feature = "fixed-leader-election")]
-            config.config.fixed_leader_for_gpuvid,
-        );
+        let da_membership =
+            <TYPES as NodeType>::Membership::new(all_nodes.clone(), da_nodes, Topic::Da);
```
rob-maron, line 405: Nice

rob-maron (approved):

crates/hotshot/src/traits/election/randomized_committee.rs
ss-es: Oh good catch.

It looks like `gen` doesn't have any guarantees about the distribution it uses, so I think `gen_range` is the correct thing to use. updated

crates/hotshot/src/traits/election/randomized_committee.rs
```diff
@@ -0,0 +1,170 @@
...
+        signature_key::{SignatureKey, StakeTableEntryType},
+    },
+    PeerConfig,
+};
+use rand::{rngs::StdRng, Rng};
+
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+
+/// The static committee election
+pub struct GeneralStaticCommittee<T: NodeType> {
```
bfish713, line 24: Why not just name this GeneralRandomCommittee? or similar so we don't have 2 structs named the exact same thing, I could see this causing some weird headache in the future.  Unless this is some common code style thing I wasn't aware of

crates/types/src/traits/election.rs
```diff
@@ -30,6 +29,9 @@ pub trait Membership<TYPES: NodeType>:
     /// Get all participants in the committee for a specific view
     fn committee_members(&self, view_number: TYPES::Time) -> BTreeSet<TYPES::SignatureKey>;
 
+    /// Get all leaders in the committee for a specific view
+    fn committee_leaders(&self, view_number: TYPES::Time) -> BTreeSet<TYPES::SignatureKey>;
+
     /// Get the stake table entry for a public key, returns `None` if the
```
bfish713, line 35: Is this function just so we can run the static committee with a single leader, and know that leader?

crates/hotshot/src/traits/election/static_committee.rs
```diff
@@ -35,12 +33,6 @@ pub struct GeneralStaticCommittee<T: NodeType> {
     indexed_stake_table:
         BTreeMap<T::SignatureKey, <T::SignatureKey as SignatureKey>::StakeTableEntry>,
 
-    // /// The members of the committee
-    // committee_members: BTreeSet<T::SignatureKey>,
-    #[cfg(feature = "fixed-leader-election")]
-    /// The number of fixed leaders for gpuvid
-    fixed_leader_for_gpuvid: usize,
-
```
bfish713, line 43: Where did the fixed leader logic go to? I thought you mentioned we'd keep it, but I am missing where

crates/hotshot/src/traits/election/static_committee.rs
```diff
@@ -130,37 +130,10 @@ impl<TYPES: NodeType> Membership<TYPES> for GeneralStaticCommittee<TYPES> {
...
-        let res = self.eligible_leaders[index].clone();
-        TYPES::SignatureKey::public_key(&res)
-    }
-
-    #[cfg(feature = "randomized-leader-election")]
-    /// Index the vector of public keys with a random number generated using the current view number as a seed
-    fn leader(&self, view_number: TYPES::Time) -> TYPES::SignatureKey {
-        let mut rng: StdRng = rand::SeedableRng::seed_from_u64(*view_number);
-        let randomized_view_number: usize = rng.gen();
-        let index = randomized_view_number % self.eligible_leaders.len();
```
bfish713, line 163: Very happy to get rid of all this

crates/types/src/traits/election.rs
rob-maron: I saw it was just used to print the number of eligible leaders on creation

crates/hotshot/src/traits/election/static_committee.rs
ss-es: to recreate e.g. the single fixed leader, we'd just set `eligible_leaders` to that single node in `Memberships::new()`

crates/types/src/traits/election.rs
ss-es: yeah, this is just used to get the number of eligible leaders into `BenchResults` for benchmarking.

crates/hotshot/src/traits/election/randomized_committee.rs
ss-es: I was just following the style from before, but yeah I agree -- renamed!

PR #3695, opened 2024-09-23
[AUDIT] Fix `post_identity` endpoint to not allow slot exhaustion

jparr721:
Closes N/A

### This PR: 
Changes the `post_identity` method to utilize a new state variable to record the received (valid) node identifiers, and uses this map as the source of truth for the node_index field. It does not remove `tmp_node_index` as this is for testing and does not correlate with `node_index`.

### DIFF
--- a/crates/orchestrator/src/lib.rs
+++ b/crates/orchestrator/src/lib.rs
@@ -90,7 +90,7 @@ pub struct OrchestratorState<KEY: SignatureKey> {
     /// The nodes that have posted they are ready to start
     nodes_connected: HashSet<PeerConfig<KEY>>,
     /// The nodes that have posted that they are ready to receive a configuration
-    nodes_identified: HashSet<(Multiaddr, PeerId)>,
+    nodes_identified: HashMap<(Multiaddr, PeerId), u16>,
     /// The results of the benchmarks
     bench_results: BenchResults,
     /// The number of nodes that have posted their results
@@ -131,7 +131,7 @@ impl<KEY: SignatureKey + 'static> OrchestratorState<KEY> {
             peer_pub_ready,
             pub_posted: HashMap::new(),
             nodes_connected: HashSet::new(),
-            nodes_identified: HashSet::new(),
+            nodes_identified: HashMap::new(),
             start: false,
             bench_results: BenchResults::default(),
             nodes_post_results: 0,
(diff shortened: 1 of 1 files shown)


crates/orchestrator/src/lib.rs
```diff
@@ -89,8 +87,10 @@ struct OrchestratorState<KEY: SignatureKey> {
     /// Whether nodes should start their HotShot instances
     /// Will be set to true once all nodes post they are ready to start
     start: bool,
-    /// The total nodes that have posted they are ready to start
+    /// The nodes that have posted they are ready to start
     nodes_connected: HashSet<PeerConfig<KEY>>,
+    /// The nodes that have posted that they are ready to receive a configuration
+    nodes_identified: HashSet<(Multiaddr, PeerId)>,
```
ss-es, line 93: I think we'd want this to be a `HashMap` from `(Multiaddr, PeerId)` to `node_index`?

crates/orchestrator/src/lib.rs
```diff
@@ -421,10 +425,9 @@ where
         libp2p_address: Option<Multiaddr>,
         libp2p_public_key: Option<PeerId>,
     ) -> Result<u16, ServerError> {
-        let node_index = self.latest_index;
-        self.latest_index += 1;
+        let node_index = self.nodes_identified.len();
 
-        if usize::from(node_index) >= self.config.config.num_nodes_with_stake.get() {
+        if node_index >= self.config.config.num_nodes_with_stake.get() {
```
ss-es, line 430: I think we'd want this to look up the `node_index` in the existing list and return that if it exists, so that a node restarting after `post_identity` still receives the same `node_index` it did the first time

ss-es (commented):
Caveat on my review: I don't actually know where/why `post_identity` is currently used, though I think if we're changing it we want it to be resilient to nodes restarting

crates/orchestrator/src/lib.rs
jparr721: Hmm, the index is a u16 in this case, what list stores the indices? I see `pub_posted`, but that takes a public key, which doesn't seem to be either of the posted fields here?

crates/orchestrator/src/lib.rs
jparr721: Yeah that makes sense

PR #3696, opened 2024-09-23
Generic-array should not be an explicit dependency

jparr721:
Closes #<ISSUE_NUMBER>

### This PR: 
From the ticket:

> Currently generic-array is an explicit direct dependency. All our use of generic-array is inherited from upstream RustCrypto crates such as blake3, sha3, digest, etc. Thus, the version of generic-array that we use should be whatever's used by these upstream crates. Thus, we should not list it as an explicit direct dependency.
> 
> Unfortunately, we need serde for generic-array and I don't know how to get it without listing generic-array as an explicit dependency. This is an insidious hack because it means that we must take care to always keep the version of generic-array we list in Cargo.toml consistent with whatever's used upstream.

This PR removes the `generic-array` dependency and instead gets it from `digest` which re-exports it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3426,7 +3426,6 @@ dependencies = [
  "espresso-systems-common",
  "ethereum-types",
  "futures",
- "generic-array",
  "jf-pcs",
  "jf-signature",
  "jf-utils",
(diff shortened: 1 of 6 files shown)


ggutoski (approved):
Thanks! Looks like there's a CI failure to resolve but I'm happy.

PR #3697, opened 2024-09-23
Fix Hanging CI

bfish713:
### This PR: 

Finally fix the test hanging issue.  The main issue was that the test task implementation could infinite loop.  The issue was exposed one of the test tasks takes the internal event streams.  The task is just a while loop that selects the next event from any event receiver it has.  Because in the case of the restart test we create new internal senders for the new nodes and kill all of senders of the of the old nodes, the receivers held by the test tasks are all closed.  This means that each time we select one of the receivers next event we immediately get an error and the loop just becomes a hot loop that constantly is processing the error and never awaiting.  This spin loop then starves the executor from running so none of the async code can run.   It was easy to reproduce by limiting the async executor threads to 1 on tokio.  

In the process of debugging this I found a few other issues that I fixed:

- In the spinning test task I now spawn all the new nodes in parallel so they should start a roughly the same time
- I noticed that when we create consensus and before we call `start consensus` we spawn the first timeout tasks, if it takes awhile to call `start_consensus` we could get a timeout before even getting the event for genesis.  I fixed that by starting the first timeout task when we start consensus
- We remove a failed view from our failed view list if all nodes who failed it eventually succeed it, but in some tests the nodes might go down or not decided it before the test completes so I made it so that we remove the failed view if a quorum of nodes succeed it.  (this scenario should not be possible but I think it's something that might have been fixed with the timeout change
- The completion task was trying to shutdown nodes but that logic is already handled when the test finishes so it is duplicated.  
- Rebalances the time of the CI jobs, they should be all about 10 minutes now
- Update the justfile to log all logs at info.  That's the level we log in production so it shouldn't be too spammy.  Also the previous justfile would only show logs of consensus starting up and was not usually helpful.

### This PR does not: 

Fix the underlying issue that some test tasks are holding dead receivers when all nodes are restarted.  I think it's just the view sync task and we should probably not spawn it for the restart tests.

### Key places to review: 

Changes to `spinning_task`, the new initial timeout logic, changes to `test_task`


crates/testing/src/overall_safety_task.rs
```diff
@@ -222,6 +222,8 @@ impl<TYPES: NodeType, I: TestableNodeImplementation<TYPES>, V: Versions> TestTas
             match view.status.clone() {
                 ViewStatus::Ok => {
                     self.ctx.successful_views.insert(view_number);
+                    // if a view succeeds remove it from the failed views
+                    self.ctx.failed_views.remove(&view_number);
```
lukeiannucci, line 226: i think you can get rid of `cleanup_previous_timeouts_on_view`

bfish713:
This has gone through at least 5 or 6 consecutive CI successes without a failure, and more like 10 if you ignore the marketplace test failures which I believe were not related

lukeiannucci (approved):
lgtm

crates/testing/src/test_task.rs
```diff
@@ -97,13 +97,18 @@ impl<S: TestTaskState + Send + 'static> TestTask<S> {
...
-                        .inspect_err(|e| tracing::error!("{e}"));
-                }
+                match async_timeout(Duration::from_millis(2500), select_all(messages)).await {
+                    Ok((Ok(input), id, _)) => {
+                        let _ = S::handle_event(&mut self.state, (input, id))
+                            .await
+                            .inspect_err(|e| tracing::error!("{e}"));
+                    }
+                    Ok((Err(e), _id, _)) => {
+                        error!("Error from one channel in test task {:?}", e);
```
ss-es, line 107: not sure if this would be better, but a thought: could we match and drop the receiver here? e.g. `Ok((Err(RecvError::Closed), id, _)) => { self.receivers.remove(id) }`

(this might require also breaking at the start if `self.receivers.is_empty()`)

ss-es (commented):
can't imagine how annoying this must've been to track down! left just a few comments, I'll try to take a closer look tomorrow

justfile
```diff
@@ -53,31 +53,31 @@ test *ARGS:
 
 test-ci *ARGS:
   echo Testing {{ARGS}}
-  RUST_LOG=error,hotshot=debug,libp2p-networking=debug cargo test --lib --bins --tests --benches --workspace --no-fail-fast {{ARGS}} -- --test-threads=1
```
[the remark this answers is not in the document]
bfish713: Nope it should be info like the rest,

crates/testing/src/test_task.rs
bfish713: yeah I want to do that, but it would change the idx of the receiver and might lead to some weird behaviours.  I plan to follow up with this with a few more cleanups and will store the receivers with their node_id somehow in this task (or replace with empty or new receivers if possible)

justfile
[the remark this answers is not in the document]
bfish713: Sorry ignore the now delete comment.  This change was intentional.  The issue was that before we'd a ton of logs from network and hotshot startup and no logs from the actual consensus impl once the nodes started (I think "Starting HothShot" as usually the last log) so it wasn't very helpful.  There is probably some better version we can do, but I do think info level for everything is a good balance of info and noise to debug.

crates/task-impls/src/request.rs
```diff
@@ -110,33 +109,6 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> TaskState for NetworkRequest
...
-                            quorum_proposal.clone(),
-                        )
-                        .into(),
-                        sender,
-                    )
-                    .await;
-                }
-
-                Ok(())
-            }
```
rob-maron, line 139: Here

PR #3714, opened 2024-09-30
[Libp2p] Use native Libp2p multiaddresses

rob-maron:
### This PR: 

Changes Libp2p to use native multiaddressing. This allows us to use DNS and IPv6 entries directly instead of (for DNS) resolving it at runtime and using the resolved address.

This prevents the node from failing to start if the DNS entry does not yet exist, and also prevents the local resolution of a cloud address causing issues (e.g. `espresso.node` resolving to some AWS-local IP). It also allows for node addresses to be automatically updated through just changing the DNS entry.

To do this, we introduce the function `derive_libp2p_multiaddress` which derives the proper Multiaddress from a string. It supports IPv4, IPv6, and FQDNs. In the event that the configured FQDN does not exist, it will warn the user but still allow the node to start. It also includes testing for expected functionality.

Also does some cleanup, like removing the need for an orchestrator client to be created from `ValidatorArgs`

### This PR does not: 

Make any protocol-breaking changes 

### Key places to review: 

`derive_libp2p_multiaddress`

### How to test this PR:

I have already tested this in the sequencer and it works as expected

### DIFF
--- a/crates/examples/combined/all.rs
+++ b/crates/examples/combined/all.rs
@@ -151,7 +151,7 @@ async fn main() {
             infra::main_entry_point::<TestTypes, Network, NodeImpl, TestVersions, ThisRun>(
                 ValidatorArgs {
                     url: orchestrator_url,
-                    advertise_address: Some(advertise_address),
+                    advertise_address: Some(advertise_address.to_string()),
                     builder_address: Some(builder_address),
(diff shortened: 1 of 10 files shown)


ss-es (approved):
looks fine to me

bfish713 (approved):

PR #3719, opened 2024-10-02
[CLEANUP] Remove Duplicate Network Task Spawns

bfish713:
### This PR: 

Just some cleanup.  We used to spawn a bunch of network event tasks for every group of messages.  Instead we collapse that into one task with both DA and Quorum membership and let the task decide which to use.

This also gets rid of an extra network message task that was spawned for no reason at all.  

It gets rid of the short circuit logic for DA leader and committee broadcast.  If the leader isn't reachable we shouldn't give up on the request.  Worst case if our network is down then we'll just error twice quickly. 

Finally I got rid of all the extra filters we don't need anymore and renamed most places we used channel (as a holdover from communication channel) to network.


crates/task-impls/src/network.rs
```diff
@@ -38,63 +38,6 @@ use crate::{
...
-            | HotShotEvent::UpgradeVoteSend(_)
-            | HotShotEvent::ViewChange(_)
-    )
-}
-
-/// DA filter
-pub fn da_filter<TYPES: NodeType>(event: &Arc<HotShotEvent<TYPES>>) -> bool {
-    !matches!(
-        event.as_ref(),
-        HotShotEvent::DaProposalSend(_, _)
```
ss-es, line 67: just to check that I understand correctly, was `DaProposalSend` the only thing that needed to be broadcast to the DA committee (as opposed to broadcasting to the quorum or sent to an individual node directly?)

ss-es (approved):
looks good to me! big fan of removing all the filters

rob-maron (approved):

crates/task-impls/src/network.rs
bfish713: Yeah that's correct

PR #3737, opened 2024-10-07
Adjust some log levels

rob-maron:
Changes log levels for some logs to be more accurate.

### DIFF
--- a/crates/libp2p-networking/src/network/behaviours/dht/mod.rs
+++ b/crates/libp2p-networking/src/network/behaviours/dht/mod.rs
@@ -190,7 +190,7 @@ impl<K: SignatureKey + 'static> DHTBehaviour<K> {
         if let Some(entry) = kad.store_mut().get(&key.clone().into()) {
             // The key already exists in the cache
             if chan.send(entry.value.clone()).is_err() {
-                error!("Get DHT: channel closed before get record request result could be sent");
+                warn!("Get DHT: channel closed before get record request result could be sent");
             }
         } else {
             // Check if the key is already being queried
@@ -277,7 +277,7 @@ impl<K: SignatureKey + 'static> DHTBehaviour<K> {
                     }
                 },
                 Err(err) => {
-                    error!("GOT ERROR IN KAD QUERY {:?}", err);
+                    warn!("Error in Kademlia query: {:?}", err);
                     0
                 }
             },
@@ -330,7 +330,7 @@ impl<K: SignatureKey + 'static> DHTBehaviour<K> {
                     // Only return the record if we can store it (validation passed)
                     if store.put(record).is_ok() {
                         if notify.send(r).is_err() {
-                            error!("Get DHT: channel closed before get record request result could be sent");
+                            warn!("Get DHT: channel closed before get record request result could be sent");
                         }
                     } else {
                         error!("Failed to store record in local store");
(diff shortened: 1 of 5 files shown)


crates/libp2p-networking/src/network/behaviours/dht/mod.rs
```diff
@@ -190,7 +190,7 @@ impl<K: SignatureKey + 'static> DHTBehaviour<K> {
         if let Some(entry) = kad.store_mut().get(&key.clone().into()) {
             // The key already exists in the cache
             if chan.send(entry.value.clone()).is_err() {
-                error!("Get DHT: channel closed before get record request result could be sent");
```
bfish713, line 193: This one seems like a real error.  If this chan is closed I think the network node is shutdown

crates/libp2p-networking/src/network/behaviours/dht/mod.rs
```diff
@@ -277,7 +277,7 @@ impl<K: SignatureKey + 'static> DHTBehaviour<K> {
                     }
                 },
                 Err(err) => {
-                    error!("GOT ERROR IN KAD QUERY {:?}", err);
```
bfish713, line 280: Do we get this a lot? I'd be ok lowering even to info

crates/libp2p-networking/src/network/behaviours/dht/mod.rs
```diff
@@ -330,7 +330,7 @@ impl<K: SignatureKey + 'static> DHTBehaviour<K> {
                     // Only return the record if we can store it (validation passed)
                     if store.put(record).is_ok() {
                         if notify.send(r).is_err() {
-                            error!("Get DHT: channel closed before get record request result could be sent");
+                            warn!("Get DHT: channel closed before get record request result could be sent");
```
bfish713, line 333: Same comment this probably is a real error if the chan is closed although this one I could see happening if the store query is just slow and whoever sent it timedout and dropped the rx side of the chan

crates/task-impls/src/transactions.rs
```diff
@@ -445,7 +445,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> TransactionTask
 
                 let mut make_block = false;
                 if *view - *self.cur_view > 1 {
-                    error!("View changed by more than 1 going to view {:?}", view);
+                    warn!("View changed by more than 1 going to view {:?}", view);
```
bfish713, line 448: Maybe make this info? It seems to happen in a lot of cases

bfish713 (approved):
Approving because this clearly an improvement and I trust your judgement to change some log levels base on my comments

crates/libp2p-networking/src/network/behaviours/dht/mod.rs
rob-maron: I believe we see this on `Canceled` requests when we do the outbound query deduplication but I'll make sure before merging

crates/libp2p-networking/src/network/behaviours/dht/mod.rs
rob-maron: Yeah we get it whenever a lookup actually fails

crates/libp2p-networking/src/network/behaviours/dht/mod.rs
rob-maron: Yeah same as above, we see it when we insta-cancel our own requests if we already have an outbound one for the same key

crates/task-impls/src/transactions.rs
rob-maron: Sounds good to me

PR #3738, opened 2024-10-07
Make `Membership::leader()` return a `Result<_>`

ss-es:
### This PR: 
Makes the `leader` function fallible, returning an `anyhow::Result`. As a consequence of this, a lot of task logic was refactored to also return an `anyhow::Result`.

### This PR does not: 
This is not a complete refactor towards returning `anyhow::Result` from all tasks, but it moves us most of the way towards that.

### Key places to review: 
It's worth checking that each `ensure!` is equivalent (or in some cases, more correct) than the original conditional everywhere these were replaced.


crates/task-impls/src/da.rs
```diff
@@ -108,35 +108,31 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> DaTaskState<TYP
                 // the `DaProposalRecv` event. Otherwise, the view number subtraction below will
                 // cause an overflow error.
                 // TODO ED Come back to this - we probably don't need this, but we should also never receive a DAC where this fails, investigate block ready so it doesn't make one for the genesis block
```
bfish713, line 110: Can we resolve this TODO comment?  I don't think receiving a proposal has much to do with the block ready event.  A bad leader could send us proposals for old views so we need this check to discard them early.

Longer term follow up, what if we get a valid DAProposal late, is there any value in backfilling the leaf with the full block data?

crates/task-impls/src/da.rs
```diff
@@ -108,35 +108,31 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> DaTaskState<TYP
...
-                {
-                    warn!("Received DA proposal for view {:?} but we already have a payload for that view.  Throwing it away", view);
-                    return None;
-                }
+                    .contains_key(&view),
+                    format!("Received DA proposal for view {:?} but we already have a payload for that view.  Throwing it away", view));
 
                 let encoded_transactions_hash = Sha256::digest(&proposal.data.encoded_transactions);
 
                 // ED Is this the right leader?
```
bfish713, line 126: It is the right leader, we can remove the comment

crates/task-impls/src/da.rs
```diff
@@ -108,35 +108,31 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> DaTaskState<TYP
...
+                );
 
-                if !view_leader_key.validate(&proposal.signature, &encoded_transactions_hash) {
-                    error!("Could not verify proposal.");
-                    return None;
-                }
+                ensure!(
+                    view_leader_key.validate(&proposal.signature, &encoded_transactions_hash),
+                    "Could not verify proposal."
+                );
```
bfish713, line 135: This is downgrading the error to info log level.  I think this one is at least a warn level event (someone is sending us malformed proposals)

crates/task-impls/src/da.rs
```diff
@@ -146,10 +142,9 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> DaTaskState<TYP
             }
             HotShotEvent::DaProposalValidated(proposal, sender) => {
                 let curr_view = self.consensus.read().await.cur_view();
-                if curr_view > proposal.data.view_number() + 1 {
-                    tracing::debug!("Validated DA proposal for prior view but it's too old now Current view {:?}, DA Proposal view {:?}", curr_view, proposal.data.view_number());
-                    return None;
-                }
+                ensure!(curr_view <= proposal.data.view_number() + 1,
+                    format!("Validated DA proposal for prior view but it's too old now Current view {:?}, DA Proposal view {:?}", curr_view, proposal.data.view_number()));
+
```
bfish713, line 147: This is now promoted to info level and will show up in data dog.  I think we don't want that

crates/task-impls/src/da.rs
```diff
@@ -163,20 +158,16 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> DaTaskState<TYP
...
-                    error!(
-                        "Failed to store DA Proposal with error {:?}, aborting vote",
-                        e
-                    );
-                    return None;
-                }
+                    )
+                );
+
+                self.storage.write().await.append_da(proposal).await?;
```
bfish713, line 169: error -> info level

crates/task-impls/src/da.rs
```diff
@@ -195,11 +186,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> DaTaskState<TYP
                     &self.private_key,
                     &self.upgrade_lock,
                 )
-                .await
-                else {
-                    error!("Failed to sign DA Vote!");
-                    return None;
-                };
+                .await?;
```
bfish713, line 189: error ->info

bfish713 (commented):
Started reviewing, changing membership and using ensure seems like a good change, but I think we need to resolve the issue with log levels.

Also this pr can get a bit smaller after: https://github.com/EspressoSystems/HotShot/pull/3755 merges

crates/task-impls/src/da.rs
ss-es: Set it to warn!

crates/task-impls/src/da.rs
ss-es: now at debug!

crates/task-impls/src/da.rs
ss-es: all storage methods should now log at error!

crates/task-impls/src/da.rs
ss-es: `create_signed_vote` now logs at error!

bfish713 (approved):

PR #3743, opened 2024-10-09
Seed leader rotation from QC

ss-es:
### This PR: 
- makes `leader` an async function
- adds a method `add_seed` to the `Membership` trait to add a seed at a given block height
- adds a new committee type for statically-seeded leader rotation, with the old `RandomizedCommittee` used for dynamically-seeded leader rotation


ss-es:
outdated

PR #3745, opened 2024-10-09
Remove Dependency on Orchestrator from HotShot Types Module

jparr721:
No Linked Issue

### This PR: 
Removes the dependency on `hotshot-orchestrator` from `hotshot-types`. This is to enable an easier time when working with libraries which import the `libp2p-network`. Right now, anyone looking to implement connected network needs to not only include all of `hotshot-types` and `hotshot`, but also `libp2p-network` and `hotshot-orchestrator`. Orchestrator, in particular, is quite pernicious as it builds binaries and other noise that we don't want in a project that isn't using it. However, at a nearly fractal level of granularity, the orchestrator config seemingly has infinite depth in relations and, as a result, I had to do a bunch of spelunking to get everyone moved into an agreeable configuration.

### This PR does not: 
Change any logic. It just moves some of the structs around.

### Key places to review: 
Mostly just the names and locations of the files. I tried to split them logically to avoid one super-config file as we had before.


crates/orchestrator/src/client.rs
```diff
@@ -156,6 +161,33 @@ pub struct MultiValidatorArgs {
     pub network_config_file: Option<String>,
 }
 
+/// Asynchronously retrieves a `NetworkConfig` from an orchestrator.
+/// The retrieved one includes correct `node_index` and peer's public config.
+///
+/// # Errors
+/// If we are unable to get the configuration from the orchestrator
+pub async fn get_complete_config<K: SignatureKey>(
```
jparr721, line 169: I needed to decouple `OrchestratorClient`. This method didn't really *need* to live on `NetworkConfig` anyway since it creates a network config, and has a side effect output with the source value.

crates/types/Cargo.toml
```diff
@@ -28,7 +28,7 @@ ethereum-types = { workspace = true }
 futures = { workspace = true }
 cdn-proto = { workspace = true }
 reqwest = { workspace = true }
-
+serde-inline-default = "0.1"
```
ss-es, line 31: can we move this to the workspace (maybe with the latest version, `= "0.2"` unless that breaks something) and do

```suggestion
serde-inline-default = { workspace = true }
```

crates/types/src/validator_config.rs
```diff
@@ -0,0 +1,72 @@
...
+                    }
+                };
+                data
+            }
+            // Handle the `error` case.
+            Err(e) => {
+                // Write `msg` to `stderr`.
+                error!("Could not read file `{}`: {}", filename, e);
+                ValidatorConfigFile::default()
+            }
```
ss-es, line 62: @rob-maron might have a stronger opinion on this, but I feel like I'd want this to panic if we passed it a nonempty file location and it couldn't get the file

crates/types/src/validator_config.rs
```diff
@@ -0,0 +1,72 @@
...
+        }
+    }
+}
+
+impl<K: SignatureKey> From<ValidatorConfigFile> for ValidatorConfig<K> {
+    fn from(val: ValidatorConfigFile) -> Self {
+        // here stake_value is set to 1, since we don't input stake_value from ValidatorConfigFile for now
```
ss-es, line 69: can we add it and just set it to 1 in the file?

crates/types/src/hotshot_config_file.rs
```diff
@@ -0,0 +1,158 @@
...
+    }
+}
+
+impl<KEY: SignatureKey> From<ValidatorConfigFile> for HotShotConfig<KEY> {
+    fn from(value: ValidatorConfigFile) -> Self {
+        let mut config: HotShotConfig<KEY> = HotShotConfigFile::default().into();
+        config.my_own_validator_config = value.into();
+        config
+    }
+}
```
ss-es, line 112: I'm uneasy about this because I feel like `From` is for conversions that are "obvious" (whereas this is defaulting the HotShotConfig to a completely arbitrary value)

crates/types/src/hotshot_config_file.rs
```diff
@@ -0,0 +1,158 @@
...
+
+impl<KEY: SignatureKey> Default for HotShotConfigFile<KEY> {
```
ss-es, line 114: I never liked the `Default` implementation here (as opposed to calling it something more descriptive like e.g. `hotshot_config_5_10`)

crates/types/src/network.rs
```diff
@@ -4,29 +4,27 @@
...
 
-use crate::client::OrchestratorClient;
+/// default number of rounds to run
+pub const ORCHESTRATOR_DEFAULT_NUM_ROUNDS: usize = 100;
+/// default number of transactions per round
+pub const ORCHESTRATOR_DEFAULT_TRANSACTIONS_PER_ROUND: usize = 10;
+/// default size of transactions
+pub const ORCHESTRATOR_DEFAULT_TRANSACTION_SIZE: usize = 100;
+/// default delay before beginning consensus
+pub const ORCHESTRATOR_DEFAULT_START_DELAY_SECONDS: u64 = 60;
```
ss-es, line 27: can these live in `types/src/constants.rs`?

ss-es (approved):
just a few comments, though I don't think you introduced any of the things I mentioned so you don't necessarily need to address them

I think @rob-maron or someone else should also get a chance to look at it, but I think this is a strictly positive change

crates/types/src/network.rs
jparr721: Done

crates/types/src/hotshot_config_file.rs
jparr721: Done

crates/types/src/hotshot_config_file.rs
jparr721: This isn't even used. Nuked.

crates/types/Cargo.toml
jparr721: Oh, that's embarassing, didn't realize I duplicated the import in two spots!

crates/types/src/validator_config.rs
jparr721: I agree. Will check with Rob today.

PR #3748, opened 2024-10-10
Expose Requesting Proposal

bfish713:
Closes #<ISSUE_NUMBER>


crates/hotshot/src/types/handle.rs
```diff
@@ -77,6 +87,76 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES> + 'static, V: Versions>
...
+            if let Some(hs_event) = event.as_ref() {
+                if let HotShotEvent::QuorumProposalResponseRecv(quorum_proposal) = hs_event.as_ref()
+                {
+                    // Make sure that the quorum_proposal is valid
+                    if quorum_proposal
+                        .validate_signature(mem, upgrade_lock)
+                        .await
+                        .is_err()
+                    {
+                        continue;
```
jbearer, line 148: Log something here

crates/hotshot/src/types/handle.rs
```diff
@@ -77,6 +87,76 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES> + 'static, V: Versions>
...
+                    }
+                    let proposed_leaf = Leaf::from_quorum_proposal(&quorum_proposal.data);
+                    let commit = proposed_leaf.commit(upgrade_lock).await;
+                    if commit == leaf_commitment {
+                        return Ok(quorum_proposal.data.clone());
```
jbearer, line 153: Log something if they are not equal

jbearer (approved):
Looks good but could use some more logging

Issue #3728, opened 2024-10-03
[EPOCH TRANSITION] - a new config field for epoch height


lukaszrzasik:
### What is this task and why do we need to work on it?

A new config field needs to be added to specify the height of an epoch. Epoch's height is measured in number of blocks. A value of zero should be interpreted as: there are no epochs, there is only one static stake table.

### What are the acceptance criteria to close this issue?

A new config field is added to that specified the height of an epoch. The default value should be set to zero so that the current code can run without problems.

ss-es:
Completed in #3751

PR #3751, opened 2024-10-10
Add epoch parameter in `Membership` trait's methods

### LINKED ISSUE #3725: [EPOCH TRANSITION] - `Membership` trait's methods take epoch parameter

This is required to implement a variable stake table. Stake table can change while transitioning from one epoch to the next. Looking at it from a different angle: there are multiple stake tables, one per epoch. Most of the method's in the `Membership` trait need take an additional parameter: epoch.

Epoch transition details can be found here: https://www.notion.so/espressosys/Epoch-transition-without-extra-types-and-with-last-votes-broadcast-1112431b68e98092a47aea71b41bec8c

_No response_

_No response_

The `Membership` trait's method take an additional parameter `epoch` where applicable. This task alone doesn't include the epoch transition so the functionality should be mocked.

_No response_

lukaszrzasik:
Closes #3725 #3728

### This PR: 

* Adds `epoch` parameter to most methods of the `Membership` trait
* Adds `epoch_height` config field so that the number of blocks in an epoch can be configurable.

### This PR does not: 

This PR does not implement epoch transition. There is no new logic related to epoch transition. This will be added in separate PRs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3245,6 +3245,7 @@ name = "hotshot-fakeapi"
 version = "0.5.77"
 dependencies = [
  "anyhow",
+ "async-compatibility-layer",
  "async-lock 2.8.0",
  "async-trait",
  "futures",
--- a/crates/fakeapi/Cargo.toml
+++ b/crates/fakeapi/Cargo.toml
@@ -12,6 +12,7 @@ repository.workspace = true
 toml = { workspace = true }
 tide-disco = { workspace = true }
 anyhow = { workspace = true }
+async-compatibility-layer = { workspace = true }
 hotshot-types = { path = "../types" }
 vbs = { workspace = true }
 serde = { workspace = true }
--- a/crates/fakeapi/src/fake_solver.rs
+++ b/crates/fakeapi/src/fake_solver.rs
@@ -1,9 +1,10 @@
 use std::{
     io::{self, ErrorKind},
-    thread, time,
+    time,
 };
 
 use anyhow::Result;
+use async_compatibility_layer::art::async_sleep;
 use async_lock::RwLock;
 use futures::FutureExt;
 use hotshot_example_types::auction_results_provider_types::TestAuctionResult;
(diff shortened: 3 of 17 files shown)


crates/types/src/traits/election.rs
```diff
@@ -23,33 +23,45 @@ pub trait Membership<TYPES: NodeType>:
         committee_topic: Topic,
     ) -> Self;
 
-    /// Get all participants in the committee (including their stake)
-    fn stake_table(&self) -> Vec<<TYPES::SignatureKey as SignatureKey>::StakeTableEntry>;
+    /// Get all participants in the committee (including their stake) for a specific epoch
+    fn stake_table(
+        &self,
+        epoch: TYPES::EpochTime,
```
ss-es, line 29: I guess all of these functions will now have to return a `Result`, because we may not have the stake table for a given (future) epoch?

(since this was quite an involved change, I can do it in #3738 after merging this -- just asking for clarification)

crates/types/src/lib.rs
```diff
@@ -232,6 +232,8 @@ pub struct HotShotConfig<KEY: SignatureKey> {
     pub start_voting_time: u64,
     /// Unix time in seconds at which we stop voting on an upgrade. To prevent voting on an upgrade, set stop_voting_time <= start_voting_time.
     pub stop_voting_time: u64,
+    /// Number of blocks in an epoch, zero means there are no epochs
+    pub epoch_height: u64,
```
ss-es, line 236: this might be personal preference, but I feel like this is a protocol-level constant and might be better as an associated constant in the `NodeType` trait (rather than a configuration value)

ss-es (commented):
left some comments. I think I looked at all the files that had meaningful changes, and looks good overall

crates/types/src/lib.rs
bfish713: yeah this one is weird, view timeout is also a parameter of the protocol but is specified in this config.  If we were to move this to associated constant, we need to do so for all protocol values (which is at least a few of these values)

ss-es: I guess the distinction in my mind was:

nodes could set different view timeouts, round_start_delay etc (within reason) and consensus would probably be fine. nodes could even update it dynamically between views (within reason) and I don't think anything would break

but every single node needs to have exactly the same epoch height, forever, and if we change this at all without an upgrade consensus would break

ss-es (approved):

bfish713 (approved):

PR #3755, opened 2024-10-11
remove consensus, extraneous parameter for update_view, repoint deps

### LINKED ISSUE #3640: [CX_HARDENING] - Delete Legacy Consensus

This is straightforward. Go in and delete the `consensus` module and clean up the compiler errors. This should include a check of the duplicated functions, however, before removal. In particular

- `update_view`
- `handle_quorum_proposal_validated`
- `create_and_send_proposal`

The purpose of these checks is to make sure that things like upgrade logic etc is included in the new code paths. We also need to rename `consensus2` to `consensus`.

_No response_

_No response_

- HotShot only has one `consensus` module.
- The dependency tasks are stable and deployed in all relevant environments.

_No response_

pls148:
Closes #3640

### This PR: 

This PR removes the consensus module (deprecated in favor of consensus2) and removes an extra parameter for update_view. Also removes references to the "dependency-tasks" feature. 

### This PR does not: 

This does NOT rename the consensus2 module to consensus, nor change any other types. That will come in a later PR.


bfish713:
This looks good.  However I think now the CI is redundant.  The ci jobs for test and test-dependency-tasks are running the same tests now I think.  In a follow up can you confirm that the sets of tests are the same and remove one of them?

Future follow up is to pick out the now unused or misused types/functions and types in task-impl helpers (and other places)

bfish713 (approved):

ss-es (commented):
I think the feature also appears in a few `Cargo.toml` files, and should be removed from there

For CI, you can also just delete the block called `test-dependency-tasks` in `.github/workflows/build-and-test.yml` (keeping the regular tests, which now always run the dependency tasks)

crates/testing/tests/tests_1/upgrade_task_with_proposal.rs
```diff
@@ -4,7 +4,6 @@
 // You should have received a copy of the MIT License
 // along with the HotShot repository. If not, see <https://mit-license.org/>.
 
-#![cfg(feature = "dependency-tasks")]
 // TODO: Remove after integration of dependency-tasks
```
ss-es, line 7: I think you can remove this TODO

crates/testing/tests/tests_1/upgrade_task_with_vote.rs
```diff
@@ -4,7 +4,6 @@
 // You should have received a copy of the MIT License
 // along with the HotShot repository. If not, see <https://mit-license.org/>.
 
-#![cfg(feature = "dependency-tasks")]
 // TODO: Remove after integration of dependency-tasks
```
ss-es, line 7: this one too

ss-es (approved):
looks good to me!

I didn't look at the original PR in depth, but I think Brendon did so should be good to merge

PR #3756, opened 2024-10-11
Allow for different builders in cross-tests

QuentinI:
Closes #0000

### This PR: 
Adds an optional argument to cross-tests macro to allow specifying builder implementation(s)

### This PR does not: 
Change builder implementation in existing cross-tests

### Key places to review: 
Everything


bfish713 (approved):
I like this change, cc: @ss-es to double check the macro stuff

crates/macros/src/lib.rs
```diff
@@ -19,6 +19,9 @@ use syn::{
 struct CrossTestData {
     /// imlementations
     impls: ExprArray,
+    /// builder impl
+    #[builder(default = "syn::parse_str(\"[SimpleBuilderImplementation]\").unwrap()")]
+    builder_impls: ExprArray,
```
bfish713, line 24: I assume this will eventually be used for some marketplace builder tests?

QuentinI: Yes, both here and in marketplace repos.

crates/macros/src/lib.rs
```diff
@@ -19,6 +19,9 @@ use syn::{
 struct CrossTestData {
     /// imlementations
     impls: ExprArray,
+    /// builder impl
+    #[builder(default = "syn::parse_str(\"[SimpleBuilderImplementation]\").unwrap()")]
```
ss-es, line 23: I slightly prefer not having a default implementation (and just explicitly writing `SimpleBuilderImplementation` in all the integration tests), but I think it's fine either way and we could just change later if we want

ss-es (approved):
looks good to me!

PR #3758, opened 2024-10-14
push task_state as QuorumProposalRecv task state into helper function…

### LINKED ISSUE #3640: [CX_HARDENING] - Delete Legacy Consensus

pls148:
…s, rename consensus2

Closes #3640 

### This PR: 

Finishes the rest of #3640 by pushing task_state into helper functions instead of copying the component parts; this is no longer necessary as consensus has been removed in lieu of consensus2. Also renames consensus2->consensus, and cleans up some other bits of code.

### Key places to review: 

In particular, check over the functions update_view and handle_quorum_proposal_recv. In at least one case, I was worried that a local copy of a mutable reference var was being copied locally to allow changing the pointed-to value while being able to access the previous value. I believe I got this correct but want an extra set of eyes to check it.

### DIFF
--- a/crates/testing/src/view_generator.rs
+++ b/crates/testing/src/view_generator.rs
@@ -63,7 +63,7 @@ pub struct TestView {
     pub da_certificate: DaCertificate<TestTypes>,
     pub transactions: Vec<TestTransaction>,
     upgrade_data: Option<UpgradeProposalData<TestTypes>>,
-    pub formed_upgrade_certificate: Option<UpgradeCertificate<TestTypes>>,
+    formed_upgrade_certificate: Option<UpgradeCertificate<TestTypes>>,
     view_sync_finalize_data: Option<ViewSyncFinalizeData<TestTypes>>,
     timeout_cert_data: Option<TimeoutData<TestTypes>>,
     upgrade_lock: UpgradeLock<TestTypes, TestVersions>,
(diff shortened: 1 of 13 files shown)


crates/testing/src/view_generator.rs
```diff
@@ -63,7 +63,7 @@ pub struct TestView {
     pub da_certificate: DaCertificate<TestTypes>,
     pub transactions: Vec<TestTransaction>,
     upgrade_data: Option<UpgradeProposalData<TestTypes>>,
-    pub formed_upgrade_certificate: Option<UpgradeCertificate<TestTypes>>,
+    formed_upgrade_certificate: Option<UpgradeCertificate<TestTypes>>,
```
ss-es, line 66: was this intentional? (not that it matters either way)

ss-es (approved):
looks good to me.

as far as I can tell, the `cur_view: &mut ...` -> `task_state.cur_view: &mut ...` replacement should not change the semantics of anything (and I feel like CI would've caught it if it did)

bfish713 (approved):

PR #3759, opened 2024-10-15
[Networking] Add a hook for application messages

bfish713:
### This PR: 

This adds a way for the application layer to send arbitrary messages to other participants via the HotShot networking layer.  It's not perfect as it'll do double de/serialization and all messages are fire and forget within hotshot.  
### Key places to review: 

Changes to `SystemContextHandle`
Test added to the newtorking tests

### DIFF
--- a/crates/task-impls/src/network.rs
+++ b/crates/task-impls/src/network.rs
@@ -577,8 +577,12 @@ impl<
                         .await
                 }
                 TransmitType::DaCommitteeBroadcast => {
-                    net.da_broadcast_message(serialized_message, da_committee, broadcast_delay)
-                        .await
+                    net.da_broadcast_message(
+                        serialized_message,
+                        da_committee.iter().cloned().collect(),
+                        broadcast_delay,
+                    )
+                    .await
                 }
             };
 
--- a/crates/task-impls/src/network.rs
+++ b/crates/task-impls/src/network.rs
@@ -46,6 +46,9 @@ pub struct NetworkMessageTaskState<TYPES: NodeType> {
 
     /// Sender to send external events this task generates to the event stream
     pub external_event_stream: Sender<Event<TYPES>>,
+
+    /// This nodes public key
+    pub public_key: TYPES::SignatureKey,
 }
 
 impl<TYPES: NodeType> NetworkMessageTaskState<TYPES> {
@@ -160,11 +163,14 @@ impl<TYPES: NodeType> NetworkMessageTaskState<TYPES> {
 
             // Handle external messages
             MessageKind::External(data) => {
+                if sender == self.public_key {
+                    return;
+                }
                 // Send the external message to the external event stream so it can be processed
                 broadcast_event(
                     Event {
                         view_number: TYPES::Time::new(1),
-                        event: EventType::ExternalMessageReceived(data),
+                        event: EventType::ExternalMessageReceived { sender, data },
                     },
                     &self.external_event_stream,
                 )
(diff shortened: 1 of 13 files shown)


crates/types/src/message.rs
```diff
@@ -117,6 +122,16 @@ pub enum MessageKind<TYPES: NodeType> {
...
 }
 
+/// List of keys to send a message to, or broadcast to all known keys
+pub enum RecipientList<K: SignatureKey> {
+    /// Broadcast to all
+    Broadcast,
+    /// Send a message directly to a key
+    Direct(K),
+    /// Send a message directly to many keys
+    Many(BTreeSet<K>),
```
ss-es, line 132: is there a reason we don't just use `Vec` here? I guess this comes from the type in `da_broadcast_message`, but it seems a bit weird to have `BTreeSet` in that too

crates/testing/src/helpers.rs
```diff
@@ -67,18 +67,46 @@ pub async fn build_system_handle<
...
 
     let initializer = HotShotInitializer::<TYPES>::from_genesis::<V>(TestInstanceState::new(
-        launcher.metadata.async_delay_config,
+        launcher.metadata.async_delay_config.clone(),
     ))
     .await
     .unwrap();
 
+    // See whether or not we should be DA
+    let is_da = node_id < config.da_staked_committee_size as u64;
```
ss-es, line 104: is it possible to take this from `config.known_da_nodes` instead?

I feel like it would be better to have `config.known_da_nodes` initialized to the first `da_staked_committee_size` nodes, and take directly from that here (otherwise we have da nodes getting set independently in two different places, which might lead to surprising behaviour down the line)

ss-es (commented):
I think the main change looked good to me, just had some comments

crates/types/src/message.rs
bfish713: It's a bit weird I agree, I had to change this from Vec.  I think using a Set kind of makes sense because we want to ensure no duplicate keys but idk why we'd use the Btree version, we don't need them ordered

crates/testing/src/helpers.rs
bfish713: Yeah we should do this. However the code is like this all over the test suites, I'll create an issue for this one and follow up since it might be a bigger change.

crates/task-impls/src/network.rs
```diff
@@ -160,11 +163,14 @@ impl<TYPES: NodeType> NetworkMessageTaskState<TYPES> {
 
             // Handle external messages
             MessageKind::External(data) => {
+                if sender == self.public_key {
```
jparr721, line 166: Does this essentially prevent loops with processing messages that we sent?

jparr721 (approved):
This all looks good, but I feel like I don't understand the motivation. Why not just use the external network? Is this basically allowing the application (i.e. sequencer) to talk *back* to us?

crates/types/src/message.rs
ss-es: Oh, the non-duplicates makes sense -- I didn't think of that!

ss-es (approved):
looks good to me

bfish713:
> This all looks good, but I feel like I don't understand the motivation. Why not just use the external network? Is this basically allowing the application (i.e. sequencer) to talk _back_ to us?

What is the external network in this case?

crates/task-impls/src/network.rs
bfish713: It's basically so we don't broadcast messages to ourselves. I feel like it's unexpected behavior from the applications perspective to get it's own message

PR #3771, opened 2024-10-18
Remove unused config stuff

bfish713:
Closes #<ISSUE_NUMBER>


crates/hotshot/src/traits/networking/libp2p_network.rs
```diff
@@ -664,11 +636,6 @@ impl<K: SignatureKey + 'static> Libp2pNetwork<K> {
                 // Subscribe to the QC topic
                 handle.subscribe(QC_TOPIC.to_string()).await.unwrap();
 
-                // Only subscribe to DA events if we are DA
-                if is_da {
-                    handle.subscribe("DA".to_string()).await.unwrap();
-                }
-
```
ss-es, line 671: Is the intention that everyone is always receiving all DA broadcasts/gossip?

ss-es (approved):
looks good to me, just one question

ss-es (changes requested):
oh, wait -- it looks like the PR broke `tests_3` somehow

feel free to dismiss

bfish713:
Figured out test 3, basically the byzantine test for bad proposals is not great in general.  The network task effectively blocks us from sending the multiple bad proposals out.  The reason this PR broke it is because I removed round start delay which changed the timing such that the byzantine node blocks it's own proposal because it thinks it's a duplicate action.  I fixed this by adding more nodes so that we have enough honest leaders to complete 50 other views in time.  I feel like the byzantine leader timing out is totally a reasonable expectation but lmk if you think otherwise

ss-es (approved):

PR #3788, opened 2024-10-23
Remove slow operations from critical path

bfish713:
Closes #<ISSUE_NUMBER>

### This PR: 
 Moves 2 things off the critical path

- Storing the proposal
- Fetching a previous proposal

For storing the proposal we just do this at the end of the voting procedure before sending a vote.  This only blocks sending a vote
For fetching past proposal there was already logic to do this in both the voting and proposing code path after the other dependencies are fulfilled.  We still spawn a task to do the fetch so we can vote on future proposals referencing that view, this is critical for cases where all nodes restart.

I want to make a larger fix where `handle_quorum_proposal_recv` is moved to a separate task I'll follow up with that.  In the meantime this should significantly improve the Decaf situation

### DIFF
--- a/crates/task-impls/src/quorum_proposal_recv/handlers.rs
+++ b/crates/task-impls/src/quorum_proposal_recv/handlers.rs
@@ -213,7 +213,7 @@ pub(crate) async fn handle_quorum_proposal_recv<
 
     let parent = match parent_leaf {
         Some(leaf) => {
-            if let (Some(state), _) = consensus_reader.state_and_delta(leaf.view_number()) {
+            if let (Some(state), _) = consensus_read.state_and_delta(leaf.view_number()) {
                 Some((leaf, Arc::clone(&state)))
             } else {
                 bail!("Parent state not found! Consensus internally inconsistent");
@@ -222,7 +222,7 @@ pub(crate) async fn handle_quorum_proposal_recv<
         None => None,
     };
 
-    if justify_qc.view_number() > consensus_reader.high_qc().view_number {
+    if justify_qc.view_number() > consensus_read.high_qc().view_number {
         if let Err(e) = task_state
             .storage
             .write()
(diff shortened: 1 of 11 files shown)


crates/task-impls/src/quorum_proposal_recv/handlers.rs
```diff
@@ -104,6 +106,35 @@ async fn validate_proposal_liveness<TYPES: NodeType, I: NodeImplementation<TYPES
...
+            event_sender,
+            event_receiver,
+            membership,
+            consensus,
+            sender_public_key,
+            sender_private_key,
+            &lock,
+        )
+        .await;
+    });
```
rob-maron, line 135: Does this get included with the cancellation logic somehow? Do we want it to be?

rob-maron: Sure

bfish713: yeah bounded at 500ms, and doesn't do anything heavy after sending the request

ss-es: would it be bad to add a `async_timeout()` on the entire `fetch_proposal` call? I guess the question is whether it's okay for `fetch_proposal` to abort partway through execution

I agree that it's unlikely `fetch_proposal` will ever get stuck outside the part where we already have an `async_timeout`, but it makes me a bit uncomfortable to drop the handle of tasks that need both a read lock and a write lock (at the very least, I think this might obscure a deadlock if we run into one)

ss-es (commented):
overall looks straightforward to me, just one question about spawning the `fetch_proposal`

ss-es (approved):

crates/task-impls/src/quorum_proposal_recv/handlers.rs
```diff
@@ -176,16 +206,14 @@ pub(crate) async fn handle_quorum_proposal_recv<
...
-            &task_state.upgrade_lock,
-        )
-        .await
-        .ok(),
-    };
-    let consensus_reader = task_state.consensus.read().await;
+            task_state.upgrade_lock.clone(),
+        );
+    }
+    let consensus_read = task_state.consensus.read().await;
```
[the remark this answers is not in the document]
bfish713: I'll fix, are there others I messed up?

PR #3797, opened 2024-10-24
[Combined Net] Remove optimistic VID

rob-maron:
## This PR:
- Removes optimistic VID calculation when the primary network is down
- Reduces lock contention during VID calculation

Subsequent PRs will address the switching logic now that block times are == 8s due to the builder

### DIFF
--- a/crates/types/src/consensus.rs
+++ b/crates/types/src/consensus.rs
@@ -825,9 +825,7 @@ impl<TYPES: NodeType> Consensus<TYPES> {
         epoch: TYPES::Epoch,
     ) -> Option<()> {
         let txns = Arc::clone(consensus.read().await.saved_payloads().get(&view)?);
-        let vid =
-            VidDisperse::calculate_vid_disperse(txns, &membership, view, epoch, None)
-                .await;
+        let vid = VidDisperse::calculate_vid_disperse(txns, &membership, view, epoch, None).await;
         let shares = VidDisperseShare::from_vid_disperse(vid);
         let mut consensus = consensus.write().await;
         for share in shares {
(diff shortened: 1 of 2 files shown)


crates/types/src/consensus.rs
```diff
@@ -824,13 +824,10 @@ impl<TYPES: NodeType> Consensus<TYPES> {
...
-        let consensus = consensus.upgradable_read().await;
-        let txns = consensus.saved_payloads().get(&view)?;
-        let vid =
-            VidDisperse::calculate_vid_disperse(Arc::clone(txns), &membership, view, epoch, None)
-                .await;
+        let txns = Arc::clone(consensus.read().await.saved_payloads().get(&view)?);
+        let vid = VidDisperse::calculate_vid_disperse(txns, &membership, view, epoch, None).await;
         let shares = VidDisperseShare::from_vid_disperse(vid);
-        let mut consensus = ConsensusUpgradableReadLockGuard::upgrade(consensus).await;
+        let mut consensus = consensus.write().await;
```
jbearer, line 830: Before, with the upgradable read, this whole thing used to be atomic, in that consensus state could not change in between us reading the transactions and doing this update. Logically it looks like it doesn't matter, nothing that happens to consensus state is going to affect this update we're doing, since we're just populating potential missing state but not making any effectful changes. But I just want to double check this is intentional

jbearer (approved):

crates/types/src/consensus.rs
rob-maron: Hey yeah - I took a look too - it was not immediately apparent. @bfish713 ?

bfish713: Yeah it's completely fine for it to be non atomic like you both expect

PR #3800, opened 2024-10-24
Add a trait for calculation of exact encoded transaction

dailinsubjam:
Closes https://github.com/EspressoSystems/marketplace-builder-core/issues/169.

### This PR: 
- Added a new trait for calculation of exact encoded transaction submitted to builder

### DIFF
--- a/crates/types/src/traits/block_contents.rs
+++ b/crates/types/src/traits/block_contents.rs
@@ -43,6 +43,12 @@ pub trait Transaction:
 {
 }
 
+/// Abstraction over the minimum block size for transaction submitted to the builder
+pub trait BuilderTransaction {
+    /// The function to estimate transaction size
+    fn minimum_block_size(&self) -> u64;
+}
+
 /// Abstraction over the full contents of a block
 ///
 /// This trait encapsulates the behaviors that the transactions of a block must have in order to be
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -175,9 +175,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.90"
+version = "1.0.91"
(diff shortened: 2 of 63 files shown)


shenkeyao (approved):
LGTM. It would be nice to get approval from consensus as well.

crates/types/src/traits/block_contents.rs
```diff
@@ -43,6 +43,13 @@ pub trait Transaction:
 {
 }
 
+/// Abstraction over the minimum block size for transaction submitted to the builder
+pub trait BuilderTransaction {
+    /// The function to estimate transaction size
+    /// It takes in the transaction itself plus additional_length_overhead which includes transaction table header, transaction table entry etc.
+    fn minimum_block_size(&self, additional_length_overhead: u64) -> u64;
```
QuentinI, line 50: Pasting from Zulip:
> Why are we passing additional_length_overhead as a parameter to the trait method? It depends exclusively on the concrete Transaction type implementing the trait, not on the caller. In fact, caller has no idea what the overhead is without knowing concrete NodeTypes. The whole ((NS_ID_BYTE_LEN + NS_OFFSET_BYTE_LEN) + NUM_TXS_BYTE_LEN + TX_OFFSET_BYTE_LEN) formula applies only to SeqTypes, because that's how the sequencer's BlockHeader works; for TestTypes, for example, the overhead is just 0, because TestTypes::BlockHeader is empty

QuentinI (changes requested):

crates/types/src/traits/block_contents.rs
dailinsubjam: Removed the parameter `additional_length_overhead`

shenkeyao (approved):

QuentinI (approved):

Issue #3726, opened 2024-10-03
[EPOCH TRANSITION] - `Membership` trait has an additional method to fetch a stake table for a given epoch


lukaszrzasik:
### What is this task and why do we need to work on it?

This is required to implement a variable stake table. Stake table can change while transitioning from one epoch to the next. Looking at it from a different angle: there are multiple stake tables, one per epoch.

We need an additional method that fetches a new stake table for the given epoch. This method should be stubbed in HotShot. The real stake table is to be fetched from a contract in L1. This issue should have a corresponding issue in sequencer project in which the method is implemented.

Epoch transition details can be found here: https://www.notion.so/espressosys/Epoch-transition-without-extra-types-and-with-last-votes-broadcast-1112431b68e98092a47aea71b41bec8c

### What are the acceptance criteria to close this issue?

The `Membership` trait has an additional method that fetches a stake table for the given epoch. This method should be stubbed in HotShot.

tbro:
https://github.com/EspressoSystems/HotShot/pull/3801

PR #3801, opened 2024-10-24
Add `Error` type to `Membership` trait

ss-es:
### This PR: 
We are making the implementation of the `Membership` trait for proof-of-stake the responsibility of the sequencer.

This PR allows the `Membership` trait to be implemented with a custom error type (via `lookup_leader`), rather than forcing the sequencer to use `anyhow` or our own internal error type.

### This PR does not: 
The goal was to make a minimal change here (i.e. not have to touch any code in `task-impls`), so I added a new method with a different name rather than changing `leader`

### Key places to review: 
Is there a better naming convention?


tbro (approved):
LGTM

tbro:
https://github.com/EspressoSystems/HotShot/issues/3726

PR #3806, opened 2024-10-25
Lr/epoch end

### LINKED ISSUE #3729: [EPOCH TRANSITION] - Form an eQC at the end of an epoch

An Extended Quorum Certificate should be formed at the end of the epoch. These changes do not require a quorum of nodes in the next epoch. This can be omitted and implemented in another change. These changes should focus on behaviour of nodes in the current epoch, i.e. forming three consecutive certificates in three consecutive views for the last block in the epoch.

Details of the algorithm can be found here: https://www.notion.so/espressosys/Node-perspective-stories-1122431b68e98045a944ff4f148b1101

_No response_

_No response_

The nodes in the current epoch can successfully form an Extended Quorum Certificate.

_No response_

lukaszrzasik:
Closes #3729

### This PR: 

* Adds a method to `Consensus` to check whether the current high QC is for the last block in an epoch
* Adds a method to `Consensus` to check whether the current high QC is an Extended QC. The check is done by leaves traversal. Starting from the leaf referenced in the high QC, three leaves need to reference the same last block in the epoch and the leaves need to be from three consecutive views.
* Changes proposing logic: if the high QC is for the last block in an epoch and the high QC is not an eQC, propose the same last block again.
* Changes voting logic: if the proposed and parent blocks are the same, disregard VID and DAC because we checked it already when previously voting for the same block.

### This PR does not: 

* Change the transactions logic. It means the blocks are still acquired from the builder even if they are not proposed. This should be optimised.
* Increment the epochs.
* Change Quorum Certificate to include additional signature from the nodes in the next epoch.
* Change the stake table.

### DIFF
--- a/crates/task-impls/src/quorum_proposal/handlers.rs
+++ b/crates/task-impls/src/quorum_proposal/handlers.rs
@@ -170,7 +170,18 @@ impl<TYPES: NodeType, V: Versions> ProposalDependencyHandle<TYPES, V> {
 
         let builder_commitment = commitment_and_metadata.builder_commitment.clone();
         let metadata = commitment_and_metadata.metadata.clone();
-        let get_legacy_block_header_fn = async {
+
+        let block_header = if version >= V::Epochs::VERSION
+            && self.consensus.read().await.is_high_qc_forming_eqc()
+        {
+            tracing::info!("Reached end of epoch. Proposing the same block again to form an eQC.");
+            let block_header = parent_leaf.block_header().clone();
+            tracing::debug!(
+                "Proposing block no. {} to form the eQC.",
+                block_header.block_number()
+            );
+            block_header
+        } else if version < V::Marketplace::VERSION {
             TYPES::BlockHeader::new_legacy(
                 state.as_ref(),
                 self.instance_state.as_ref(),
@@ -184,10 +195,8 @@ impl<TYPES: NodeType, V: Versions> ProposalDependencyHandle<TYPES, V> {
             )
             .await
             .wrap()
-            .context(warn!("Failed to construct legacy block header"))
-        };
-
-        let get_marketplace_block_header_fn = async {
+            .context(warn!("Failed to construct legacy block header"))?
+        } else {
             TYPES::BlockHeader::new_marketplace(
                 state.as_ref(),
                 self.instance_state.as_ref(),
@@ -202,23 +211,8 @@ impl<TYPES: NodeType, V: Versions> ProposalDependencyHandle<TYPES, V> {
             )
             .await
             .wrap()
-            .context(warn!("Failed to construct marketplace block header"))
+            .context(warn!("Failed to construct marketplace block header"))?
         };
-
-        let mut block_header = if version < V::Marketplace::VERSION {
-            get_legacy_block_header_fn.await?
-        } else {
-            get_marketplace_block_header_fn.await?
-        };
-
-        if version >= V::Epochs::VERSION && self.consensus.read().await.is_high_qc_forming_eqc() {
-            tracing::info!("Reached end of epoch. Proposing the same block again to form an eQC.");
-            block_header = parent_leaf.block_header().clone();
-            tracing::debug!(
-                "Proposing block no. {} to form the eQC.",
-                block_header.block_number()
-            );
-        }
 
         let proposal = QuorumProposal {
             block_header,


crates/types/src/consensus.rs
```diff
@@ -835,6 +841,76 @@ impl<TYPES: NodeType> Consensus<TYPES> {
...
+        }
+
+        let high_qc_height =
+            if let Some(high_qc_leaf) = self.saved_leaves.get(&high_qc.data.leaf_commit) {
+                high_qc_leaf.height()
+            } else {
+                tracing::warn!("No leaf corresponding to high QC. Consensus inconsistent!");
+                return false;
+            };
+        let locked_height = if let Some(view) = self.validated_state_map.get(&self.locked_view) {
```
ss-es, line 889: It's a bit hard to read with the nested `if let`s

can we write this like

```
let Some(view) = ... else { };
let Some(leaf_commit) = ... else {};
...
```

crates/task-impls/src/quorum_vote/mod.rs
```diff
@@ -307,6 +357,10 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES> + 'static, V: Versions> Handl
                         return;
                     }
                     leaf = Some(proposed_leaf);
+                    // Proposed and parent block are the same. We don't care about VID and DAC.
+                    if is_same_block_proposed {
+                        break;
+                    }
```
ss-es, line 363: is there any danger we get e.g. a `DaCertificateValidated` causing us to cancel the quorum vote task? (it looks like no just from quickly looking at the code but wanted to check)

crates/types/src/consensus.rs
```diff
@@ -835,6 +843,86 @@ impl<TYPES: NodeType> Consensus<TYPES> {
...
+                    *leaf.view_number()
+                );
+
+                if leaf.view_number() == qc_view {
+                    return true;
+                }
+
+                if last_visited_view_number - 1 != leaf.view_number() {
+                    tracing::trace!("The chain is broken. Non consecutive views.");
+                    is_qc_extended = false;
```
ss-es, line 907: does this have any effect since we immediately return?

crates/types/src/consensus.rs
```diff
@@ -835,6 +843,86 @@ impl<TYPES: NodeType> Consensus<TYPES> {
...
+                    return false;
+                }
+                if high_qc_block_number != leaf.height() {
+                    tracing::trace!("The chain is broken. Block numbers do not match.");
+                    is_qc_extended = false;
```
ss-es, line 912: also this

ss-es (commented):
left some comments, I think the version gating comments are the main important ones. otherwise, looks good to me overall

crates/task-impls/src/transactions.rs
```diff
@@ -433,6 +433,25 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> TransactionTask
         None
     }
 
+    /// epochs view change handler
+    #[instrument(skip_all, fields(id = self.id, view_number = *self.cur_view))]
+    pub async fn handle_view_change_epochs(
```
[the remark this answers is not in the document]
ss-es: yeah, I think that makes sense. I think needlessly acquiring a block shouldn't break anything so happy to just do the easier thing for now

crates/task-impls/src/quorum_vote/mod.rs
ss-es: I think I just meant that in the way the `quorum_vote` task is currently structured, I think receiving a `DaCertificateValidated` in line 369 can in principle cause the full task to abort (if I'm reading it correctly?). but I'm not 100% clear on what the code paths here are

but what you said makes sense and answers the question I think: we shouldn't receive a DAC/VID because we would spawn this task just off the justify qc and immediately exit if it's an epoch transition

crates/task-impls/src/quorum_proposal/handlers.rs
```diff
@@ -166,22 +167,28 @@ impl<TYPES: NodeType, V: Versions> ProposalDependencyHandle<TYPES, V> {
...
                 version,
             )
             .await
             .wrap()
-            .context(warn!("Failed to construct legacy block header"))?
-        } else {
+            .context(warn!("Failed to construct legacy block header"))
+        };
+
+        let get_marketplace_block_header_fn = async {
```
ss-es, line 191: nit: I think this could just be called `marketplace_block_header` (similarly `legacy_block_header`)

(I think the `_fn` confused me slightly because it's just a block header future)

crates/task-impls/src/quorum_proposal/handlers.rs
```diff
@@ -196,13 +203,41 @@ impl<TYPES: NodeType, V: Versions> ProposalDependencyHandle<TYPES, V> {
...
+                    block_header.block_number()
+                );
+                block_header
+            } else {
+                return Err(anyhow::anyhow!(
+                    "There is no leaf for the high QC. Consensus inconsistent!"
+                ));
+            }
+        } else {
+            get_marketplace_block_header_fn.await?
```
ss-es, line 234: we can't assume that `V::Epochs::VERSION >= V::Marketplace::VERSION` (I think actually the opposite) -- so I think this needs to be gated too

(I think the earlier logic where we copy the previous block header is fine, because we can expect it to be the same marketplace/legacy block header)

ss-es (commented):
overall looks good, just one actual comment (feel free to ignore the nit)

crates/hotshot/src/types/handle.rs
```diff
@@ -70,6 +70,9 @@ pub struct SystemContextHandle<TYPES: NodeType, I: NodeImplementation<TYPES>, V:
 
     /// Memberships used by consensus
     pub memberships: Arc<Memberships<TYPES>>,
+
+    /// Number of blocks in an epoch, zero means there are no epochs
```
shenkeyao, line 74: Can zero also mean that there's an epoch but no finalized block yet?

crates/task-impls/src/quorum_proposal/handlers.rs
shenkeyao: I feel like it's nice to have `_fn` here so that we know we'll call `.await` to really get the value. No strong opinion though!

crates/types/src/consensus.rs
```diff
@@ -835,6 +843,103 @@ impl<TYPES: NodeType> Consensus<TYPES> {
...
+    /// Returns true if the given qc is for the last block in the epoch
+    pub fn is_qc_for_last_block(&self, cert: &QuorumCertificate<TYPES>) -> bool {
+        let Some(leaf) = self.saved_leaves.get(&cert.data().leaf_commit) else {
+            return false;
+        };
+        let block_height = leaf.height();
+        if block_height == 0 || self.epoch_height == 0 {
+            false
+        } else {
+            block_height % self.epoch_height == 0
```
shenkeyao, line 862: There's a potential division by zero issue here. We should probably add a non-zero check before this line.

crates/types/src/consensus.rs
```diff
@@ -835,6 +843,103 @@ impl<TYPES: NodeType> Consensus<TYPES> {
...
+        let high_qc = self.high_qc();
+        let ret = self.is_qc_extended(high_qc);
+        if ret {
+            tracing::debug!("We have formed an eQC!");
+        };
+        ret
+    }
+
+    /// Returns true if the given qc is an extended Quorum Certificate
+    /// The Extended Quorum Certificate (eQC) is the third Quorum Certificate formed in three
```
shenkeyao, line 879: I thought `eQC` was "epoch QC". 😮 🤣

crates/task-impls/src/quorum_vote/mod.rs
```diff
@@ -307,6 +361,10 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES> + 'static, V: Versions> Handl
                         return;
                     }
                     leaf = Some(proposed_leaf);
+                    // Proposed and parent block are the same. We don't care about VID and DAC.
+                    if is_same_block_proposed {
+                        break;
```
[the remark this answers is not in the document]
bfish713: So here we are in a while loop collecting dependencies, if we break here we possibly won't collect VID or DAC dependencies, then later in line 403 we check we have a vid share, before we vote (same for DAC).  By exiting this loop early we won't get a VID share every time.  

I think the problem is more general though, if at the end of an epoch we don't need a VID share or DAC then this `handle_dep_result` will just never be called.  We will never get the DAC and VID to fullfill the deps and this code will never run.

Are the tests just working because we aren't doing a transition. or because we actually are just resending the DAC or VID?

crates/task-impls/src/quorum_vote/mod.rs
```diff
@@ -285,17 +287,69 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES> + 'static, V: Versions> Handl
...
+                        }
+
+                        tracing::info!("Reached end of epoch. Proposed leaf has the same height and payload as its parent. No need to check the VID and DAC.");
+                        is_same_block_proposed = true;
+
+                        let consensus_reader = self.consensus.read().await;
+
+                        let Some(vid_shares) = consensus_reader
+                            .vid_shares()
+                            .get(&parent_leaf.view_number())
```
bfish713, line 332: Is this always getting the right VID share?  For the first proposal after last block this will of course be correct, but for the next one in the chain where do we set the VID share to equal the original block?

For example for the chain Proposal(bk, v) <- Proposal(bk, v+1) <- proposal(bk, v + 2) when do we set the VID share for v +1 (and v +2) equal to the one from v?

crates/task-impls/src/quorum_vote/mod.rs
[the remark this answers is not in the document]
bfish713: Thanks for the explaining, it makes sense to me know, I didn't realize we still send these out.  I'm now a bit concerned on ordering, It looks like it happens to work because if we get proposal we stop processing, and set everything correct, but process DAC and VID first they have the same "fake" commitment so we don't return early.  Ok for now, but this kind of feels like a house of cards

crates/task-impls/src/quorum_proposal/handlers.rs
ss-es: Looks correct to me now

not sure how expensive `new_marketplace`/`new_legacy` are, but worth noting we are always calling them unnecessarily during epoch transitions here.

we can avoid that by making `let mut block_header = ` a future by removing the `await?`s, then overwriting the future (rather than the block header value) in line 216 with something like `async { Ok(parent_leaf.block_header().clone() }`. Then await the future only on line 224

(but if it's not an expensive call this isn't really important)

crates/types/src/consensus.rs
```diff
@@ -835,6 +843,103 @@ impl<TYPES: NodeType> Consensus<TYPES> {
...
+            if let Some(leaf) = self.saved_leaves.get(&cert.data().leaf_commit) {
+                leaf.block_header().block_number()
+            } else {
+                return false;
+            };
+
+        let mut last_visited_view_number = qc_view;
+        let mut is_qc_extended = true;
+        if let Err(e) =
+            self.visit_leaf_ancestors(qc_view, Inclusive(qc_view - 2), true, |leaf, _, _| {
```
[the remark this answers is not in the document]
bfish713: Yeah that's what was asking,  The code assumes (correctly) that to form an eQC the views are consecutive, I was just checking to see if my understanding matched yours and it does

crates/task-impls/src/quorum_vote/mod.rs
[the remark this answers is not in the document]
bfish713: Yeah it works for now, because of that I agree.  I was just saying it feels like this it could easily get messed up if someone who doesn't realize the VID must be overridden for eQC related proposals, deletes that line of code

bfish713:
This looks good to me now.  I'll let Keyao or Salman accept if their comments are all addressed

ss-es (approved):
looks good to me

shenkeyao (approved):
LGTM!

PR #3807, opened 2024-10-25
3215 dont use high qc for handle_dep_result

pls148:
Closes #<ISSUE_NUMBER>

### This PR: 

Instead of using the high_qc view_number to determine the parent view number for the dependency result, pull the parent_view_number out of the QuorumProposalValidated event.

Originally, we'd also do an explicit wait for ValidatedStateUpdated to fire, but that seems to never actually be necessary. Existing tests all pass with that block removed. The lack of validated_view (meaning the lack of validated_state_map entry for the view number) has been turned into an error instead. We should watch out to make sure that this doesn't happen in the field.

### DIFF
--- a/crates/task-impls/src/helpers.rs
+++ b/crates/task-impls/src/helpers.rs
@@ -371,6 +371,7 @@ pub(crate) async fn parent_leaf_and_state<TYPES: NodeType, V: Versions>(
     private_key: <TYPES::SignatureKey as SignatureKey>::PrivateKey,
     consensus: OuterConsensus<TYPES>,
     upgrade_lock: &UpgradeLock<TYPES, V>,
+    parent_view_number: TYPES::View,
 ) -> Result<(Leaf<TYPES>, Arc<<TYPES as NodeType>::ValidatedState>)> {
     let consensus_reader = consensus.read().await;
     let cur_epoch = consensus_reader.cur_epoch();
@@ -381,7 +382,6 @@ pub(crate) async fn parent_leaf_and_state<TYPES: NodeType, V: Versions>(
             next_proposal_view_number
         )
     );
-    let parent_view_number = consensus_reader.high_qc().view_number();
     let vsm_contains_parent_view = consensus_reader
         .validated_state_map()
         .contains_key(&parent_view_number);
@@ -403,7 +403,7 @@ pub(crate) async fn parent_leaf_and_state<TYPES: NodeType, V: Versions>(
     }
 
     let consensus_reader = consensus.read().await;
-    let parent_view_number = consensus_reader.high_qc().view_number();
+    //let parent_view_number = consensus_reader.high_qc().view_number();
     let parent_view = consensus_reader.validated_state_map().get(&parent_view_number).context(
         debug!("Couldn't find parent view in state map, waiting for replica to see proposal; parent_view_number: {}", *parent_view_number)
     )?;
--- a/crates/task-impls/src/quorum_proposal/handlers.rs
+++ b/crates/task-impls/src/quorum_proposal/handlers.rs
@@ -21,6 +21,7 @@ use hotshot_types::{
     traits::{
         block_contents::BlockHeader, node_implementation::NodeType, signature_key::SignatureKey,
     },
+    vote::HasViewNumber,
 };
 use tracing::instrument;
 use utils::anytrace::*;
@@ -110,6 +111,7 @@ impl<TYPES: NodeType, V: Versions> ProposalDependencyHandle<TYPES, V> {
         view_change_evidence: Option<ViewChangeEvidence<TYPES>>,
         formed_upgrade_certificate: Option<UpgradeCertificate<TYPES>>,
         decided_upgrade_certificate: Arc<RwLock<Option<UpgradeCertificate<TYPES>>>>,
+        parent_view_number: TYPES::View,
     ) -> Result<()> {
         let (parent_leaf, state) = parent_leaf_and_state(
             self.view_number,
@@ -120,6 +122,7 @@ impl<TYPES: NodeType, V: Versions> ProposalDependencyHandle<TYPES, V> {
             self.private_key.clone(),
             OuterConsensus::new(Arc::clone(&self.consensus.inner_consensus)),
             &self.upgrade_lock,
+            parent_view_number,
         )
         .await?;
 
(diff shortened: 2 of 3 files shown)


crates/task-impls/src/quorum_vote/mod.rs
```diff
@@ -328,6 +300,28 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES> + 'static, V: Versions> Handl
...
+
+        let Some(parent_view_number) = parent_view_number else {
+            tracing::error!(
+                "We don't have the parent_view_number for this view {:?}, but we should, because the vote dependencies have completed.",
+                self.view_number
+            );
+            return;
+        };
+
+        let validated_view = self
```
bfish713, line 312: So part of the reason we waited for this view before was because the proposal might still be being processed or fetched.  The reason it all still works ok is because we are probably never behind in tests, and are quick to fetch proposals from others.  I do think there will be a case where returning early here will result in us not voting in a view we otherwise would have.

Further in `update_shared_state` we'll actually call `fetch_proposal` if we don't have the validated state for the view.  Now that code path can't get hit.

bfish713 (changes requested):

crates/task-impls/src/quorum_vote/mod.rs
```diff
@@ -110,31 +111,41 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES> + 'static, V: Versions>
...
-                self.receiver.activate_cloned(),
-                Arc::clone(&self.quorum_membership),
-                OuterConsensus::new(Arc::clone(&self.consensus.inner_consensus)),
-                self.public_key.clone(),
-                self.private_key.clone(),
-                &self.upgrade_lock,
-            )
-            .await
-            .ok(),
+            None => {
```
bfish713, line 114: Can we consolidate the consensus map lookups to this function, So it's easier to see where we are getting all the information.  If we're already looking up the parent here, maybe we don't need to do the `validated_view` thing in the calling function at all

crates/task-impls/src/quorum_vote/mod.rs
```diff
@@ -97,44 +97,66 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES> + 'static, V: Versions>
...
-        let mut maybe_parent = self
-            .consensus
-            .read()
-            .await
+        let mut maybe_parent = consensus_reader
             .saved_leaves()
             .get(&justify_qc.data().leaf_commit)
             .cloned();
+
+        drop(consensus_reader);
```
bfish713, line 121: nit: don't need `consensus_reader` if it's just dropped after on statement

crates/task-impls/src/quorum_proposal/handlers.rs
```diff
@@ -247,52 +243,6 @@ impl<TYPES: NodeType, V: Versions> HandleDepOutput for ProposalDependencyHandle<
 
     #[allow(clippy::no_effect_underscore_binding, clippy::too_many_lines)]
     async fn handle_dep_result(self, res: Self::Output) {
-        let high_qc_view_number = self.consensus.read().await.high_qc().view_number;
```
bfish713, line 250: This isn't really a comment on your code, but in `parent_leaf_and_state` (called from `publish_proposal`) we actually lookup by `high_qc`, I think it works, but I'd much rather see the parent view number be decided by the events we got rather than the high_qc

bfish713 (commented):
Looks good in general, I can approve if you want to make the change to `parent_leaf_and_state` in a new PR

crates/task-impls/src/helpers.rs
```diff
@@ -381,7 +382,7 @@ pub(crate) async fn parent_leaf_and_state<TYPES: NodeType, V: Versions>(
             next_proposal_view_number
         )
     );
-    let parent_view_number = consensus_reader.high_qc().view_number();
```
bfish713, line 384: nit: delete

crates/task-impls/src/quorum_proposal/handlers.rs
```diff
@@ -350,6 +306,11 @@ impl<TYPES: NodeType, V: Versions> HandleDepOutput for ProposalDependencyHandle<
             }
         }
 
+        let parent_view_number = match parent_view_number {
+            Some(view) => view,
+            None => self.consensus.read().await.high_qc().view_number(),
+        };
```
bfish713, line 312: nit: `parent_view_number.unwrap_or(self.consensus...)`

crates/task-impls/src/quorum_vote/handlers.rs
```diff
@@ -172,43 +172,67 @@ pub(crate) async fn update_shared_state<
...
     proposed_leaf: &Leaf<TYPES>,
     vid_share: &Proposal<TYPES, VidDisperseShare<TYPES>>,
+    parent_view_number: Option<TYPES::View>,
 ) -> Result<()> {
     let justify_qc = &proposed_leaf.justify_qc();
 
+    let consensus_reader = consensus.read().await;
+    // Try to find the validated vview within the validasted state map. This will be present
+    // if we have the saved leaf, but if not we'll get it when we fetch_proposal.
+    let mut maybe_validated_view = match parent_view_number {
```
bfish713, line 182: nit: use `.map`

bfish713 (approved):
Looks good except a few nits

crates/task-impls/src/quorum_vote/handlers.rs
```diff
@@ -172,43 +172,67 @@ pub(crate) async fn update_shared_state<
...
         "Proposal's parent missing from storage with commitment: {:?}, proposal view {:?}",
         justify_qc.data.leaf_commit,
         proposed_leaf.view_number(),
     ))?;
-    let consensus_reader = consensus.read().await;
 
-    let (Some(parent_state), _) = consensus_reader.state_and_delta(parent.view_number()) else {
-        bail!("Parent state not found! Consensus internally inconsistent");
+    let Some(validated_view) = maybe_validated_view else {
+        bail!("Don't have a validated view for our view! Consensus internally inconsistent");
```
bfish713, line 230: I think this error is too aggressive, it's possible we just didn't see that view's proposal and failed to fetch it.  Consensus isn't inconsistent in that case.  The error just below is much more serious because it means we stored a view, but not the state update for that view which should never happen

PR #3811, opened 2024-10-28
[Libp2p] Proof of Stake changes

### LINKED ISSUE #3810: [Libp2p] - PoS

Libp2p currently just takes the list of nodes in. We actually want it to be the full stake table so it can interface with PoS.

_No response_

_No response_

- [ ] Tests pass

_No response_

rob-maron:
Closes #3810 

### This PR: 

- Changes Libp2p to use the full stake table instead of the shallow list of nodes
- Updates types accordingly

### This PR does not: 

- Make use of epochs. The epoch it's asking for is zero

### Key places to review: 

- `transport.rs`


ss-es (approved):
looked fine to me, but I'm not that familiar with libp2p

approving because it seems straightforward but someone else should also take a look

PR #3821, opened 2024-10-30
Ignore duplicate txns via LRU cache

bfish713:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/crates/hotshot/src/traits/networking/combined_network.rs
+++ b/crates/hotshot/src/traits/networking/combined_network.rs
@@ -462,11 +462,8 @@ impl<TYPES: NodeType> ConnectedNetwork<TYPES::SignatureKey> for CombinedNetworks
             // Calculate hash of the message
             let message_hash = calculate_hash_of(&message);
 
-            // Check if the hash is in the cache
-            if !self.message_cache.read().contains(&message_hash) {
-                // Add the hash to the cache
-                self.message_cache.write().put(message_hash, ());
-
+            // Check if the hash is in the cache and update the cache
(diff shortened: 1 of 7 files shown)


crates/hotshot/src/traits/networking/combined_network.rs
```diff
@@ -462,11 +462,8 @@ impl<TYPES: NodeType> ConnectedNetwork<TYPES::SignatureKey> for CombinedNetworks
             // Calculate hash of the message
             let message_hash = calculate_hash_of(&message);
 
-            // Check if the hash is in the cache
-            if !self.message_cache.read().contains(&message_hash) {
-                // Add the hash to the cache
-                self.message_cache.write().put(message_hash, ());
-
+            // Check if the hash is in the cache and update the cache
+            if self.message_cache.write().put(message_hash, ()).is_none() {
```
rob-maron, line 466: I haven't profiled this enough to have an opinion on which is faster in most cases, but I'm all for making it more succinct

rob-maron (approved):

crates/hotshot/src/traits/networking/combined_network.rs
bfish713: It's not really for speed, it's so that we update the cache on every duplicate message.  So if a really old message that's still in the cache comes in we make sure it becomes the most recently used and doesn't fall out of cache.

rob-maron: Ah right this does work better with LRU semantics

PR #3834, opened 2024-11-01
[PoS] - Add a legacy builder API endpoint that includes `num_nodes`

### LINKED ISSUE #3833: [PoS] - Add a legacy builder API endpoint that includes `num_nodes`
With PoS, the number of nodes (or total stake) in the network is no longer static. The legacy builder doesn't require the full stake table, just the current `num_nodes` to correctly calculate the `VidCommitment` (we should ensure the marketplace builder doesn't require either)

We should add an API endpoint to request blocks where HotShot additionally provides `num_nodes` to the builder. HotShot will start querying this endpoint after the epoch upgrade, and the legacy builder should serve blocks at both endpoints

shenkeyao:
Closes #3833.

### This PR: 
* Adds a `claim_block_with_num_nodes` endpoint to provide `num_nodes` when requesting a block.
* Adds a trait function with the same name, which will be implemented in the builder repo.


ss-es (approved):
looks good to me!

PR #3845, opened 2024-11-04
[Tech debt] Remove `async-std`

rob-maron:
## This PR:
Removes `async-std` support and `async-compatibility-layer` usage, switches everything to Tokio

As a reminder, this has a couple advantages:
- We only have to support a single executor in HotShot
- It lets us use async primitives unsupported by the compatibility layer
- Development would no longer involve manually setting RUSTFLAGS everywhere
- It lets us migrate off of tide to a more modern Tokio-based framework (we can technically do this now if we rock tokio1, but I think it's best to have everything on a single executor)

In addition to the above, it's clear Tokio is better maintained.


bfish713:
What tech debt do you want to get in before this? Or do you not want to merge this for some other reason?  I'd be very happy to see async-std go

crates/hotshot/src/traits/networking/libp2p_network.rs
```diff
@@ -149,11 +150,11 @@ struct Libp2pNetworkInner<T: NodeType> {
     /// handle to control the network
     handle: Arc<NetworkNodeHandle<T>>,
     /// Message Receiver
-    receiver: UnboundedReceiver<Vec<u8>>,
```
bfish713, line 152: If this is locked because we use multiple receivers can we use async_brodcast instead

bfish713 (approved):

PR #3848, opened 2024-11-05
compile without lockfile

imabdulbasit:
This PR: 
- updates async-lock to v3 for improved async locking.
- sets toolchain to 1.82 in rust-toolchain.toml.
- Run cargo update to refresh the lockfile.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ async-broadcast = "0.7"
 async-compatibility-layer = { version = "1.2.1", default-features = false, features = [
   "logging-utils",
 ] }
-async-lock = "2"
+async-lock = "3"
 async-std = { version = "1", features = ["attributes"] }
 async-trait = "0.1"
 bincode = "1"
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -174,3 +174,6 @@ rust_2018_idioms = { level = "warn", priority = -1 }
 # TODO change to deny
 missing_docs = "warn"
 warnings = "warn"
+unexpected_cfgs = { level = "warn", check-cfg = [
```
sveitser, line 177: Doesn't this level need to be set to "ignore" or something like that?

sveitser: It's "allow"

imabdulbasit: we also need to change this in sequencer

sveitser: Doesn't seem to help

sveitser (approved):

Cargo.toml
sveitser: Actually "warn" is fine we were just not adding it in all the Cargo.toml files that needed it.

Issue #3849, opened 2024-11-05
CI: ensure HotShot builds without lockfile


sveitser:
We should  have a check like what's added in https://github.com/EspressoSystems/espresso-sequencer/pull/2257

To prevent issues like https://github.com/EspressoSystems/espresso-sequencer/issues/2256 going unnoticed.

PR #3850, opened 2024-11-05
CI: compile without lockfile

sveitser:
Closes https://github.com/EspressoSystems/HotShot/issues/3849

Since hotshot is a library I think this check is quite important because (I think) if it fails this means that anyone starting a new project with, or adding hotshot as new dependency to a project will have the build break with likely annoying/difficult to resolve dependency issues.

I was going to use `just` first and noticed the installation in the CI can be simplified. However since the just file does set env vars that are used by the rust-cache action (after the action is initialized and the cache key is computed) I opted for not using just for the new CI job.

CI run of the new job: https://github.com/EspressoSystems/HotShot/actions/runs/11684796595/job/32536807479?pr=3850

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -45,11 +45,7 @@ jobs:
           cache-on-failure: "true"
           save-if: ${{ github.ref == 'refs/heads/main' && matrix.test_suites == 'test-ci-rest' }}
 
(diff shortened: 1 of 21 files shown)


imabdulbasit (approved):
LGTM!

ss-es (approved):

PR #3851, opened 2024-11-05
Lr/eqc voting

### LINKED ISSUE #3730: [EPOCH TRANSITION] - Implement beginning of a new epoch

These changes should focus on behaviour of the nodes in the next epoch, i.e. voting for the last block in the previous epoch, proposing the first block in the new epoch and voting for the first block in the new epoch.

The details of the algorithm can be found here: https://www.notion.so/espressosys/Epoch-transition-without-extra-types-and-with-last-votes-broadcast-1112431b68e98092a47aea71b41bec8c

_No response_

_No response_

The nodes in the next epoch correctly vote for the last block in the previous epoch and for the first block in the new epoch.

_No response_

lukaszrzasik:
Closes #3730

### This PR: 

* Introduces the concept of "epoch safety check". The epoch safety check makes sure that:
* * the proposed block and its justify QC's block are from the same epoch or
* * the justify QC is an Extended Quorum Certificate for the previous epoch.
* Adds logic to broadcast the vote for the last proposal in eQC 3-chain.
* Adds logic to not check whether the node is the leader when receiving a vote for the last proposal in eQC 3-chain. This allows each node to form the eQC on their own.

### DIFF
--- a/crates/types/src/consensus.rs
+++ b/crates/types/src/consensus.rs
@@ -971,43 +971,6 @@ impl<TYPES: NodeType> Consensus<TYPES> {
         }
     }
 
-    /// Returns an `Epoch` for the view following the given view.
-    /// There are three cases:
-    /// 1. The leaf from the given view is not for the last block.
-    ///    The epoch should be calculated based on the next block number.
-    /// 2. The leaf from the given view is for the last block but it is not the last in the 3-chain.
-    ///    The next leaf will be for the same block number.
-    /// 3. The leaf from the given view is for the last block and it is the last in the 3-chain.
-    ///    The next leaf will be for the next block number.
-    /// # Errors
-    /// Returns error when there is no state corresponding to the given view.
-    pub fn get_epoch_for_next_view(&self, view: TYPES::View) -> Result<TYPES::Epoch> {
-        if self.epoch_height == 0 {
-            return Ok(TYPES::Epoch::new(0));
-        }
-        let validated_view = self.validated_state_map.get(&view).context(error!(
-            "Could not find a validated state corresponding to the given view."
-        ))?;
-        let ViewInner::Leaf {
-            leaf: leaf_commit, ..
-        } = validated_view.view_inner
-        else {
-            bail!("Could not find a proposed leaf corresponding to the given view.");
-        };
-        let leaf = self.saved_leaves.get(&leaf_commit).context(error!(
-            "Could not find a leaf corresponding to the given view."
-        ))?;
-        let next_block_number = if self.is_leaf_forming_eqc(leaf_commit) {
-            // Covers the second case.
-            leaf.height()
-        } else {
-            // Covers the first and third case.
-            leaf.height() + 1
-        };
-        let epoch_number = epoch_from_block_number(next_block_number, self.epoch_height);
-        Ok(TYPES::Epoch::new(epoch_number))
-    }
-
     /// Returns true if the `parent_leaf` formed an eQC for the previous epoch to the `proposed_leaf`
     pub fn check_eqc(&self, proposed_leaf: &Leaf<TYPES>, parent_leaf: &Leaf<TYPES>) -> bool {
         if parent_leaf.view_number() == TYPES::View::genesis() {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6513,9 +6513,9 @@ checksum = "388a1df253eca08550bef6c72392cfe7c30914bf41df5269b68cbd6ff8f570a3"
 
 [[package]]
 name = "serde"
-version = "1.0.214"
+version = "1.0.215"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f55c3193aca71c12ad7890f1785d2b73e1b9f63a0bbc353c08ef26fe03fc56b5"
+checksum = "6513c1ad0b11a9376da888e3e0baa0077f1aed55c17f50e7b2397136129fb88f"
 dependencies = [
  "serde_derive",
 ]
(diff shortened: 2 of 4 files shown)


crates/task-impls/src/consensus/handlers.rs
```diff
@@ -40,17 +40,28 @@ pub(crate) async fn handle_quorum_vote_recv<
...
+            task_state
+                .quorum_membership
+                .leader(vote.view_number() + 1, task_state.cur_epoch)?
+                == task_state.public_key,
+            info!(
+                "We are not the leader for view {:?}",
+                vote.view_number() + 1
+            )
+        );
+    }
```
bfish713, line 64: nit: `ensure!(is_vote_leaf_extended || we_are_leader)`

crates/types/src/consensus.rs
```diff
@@ -963,6 +966,25 @@ impl<TYPES: NodeType> Consensus<TYPES> {
...
+        let old_epoch = epoch_from_block_number(parent_leaf.height(), self.epoch_height);
+        let Some(View {
+            view_inner: ViewInner::Leaf {
+                leaf: leaf_commit, ..
+            },
+        }) = self.validated_state_map.get(&parent_leaf.view_number())
+        else {
+            tracing::error!("Couldn't find the provided old leaf in the consensus");
+            return false;
+        };
```
bfish713, line 985: If we have the parent leaf, and need the commit why not just use parent_leaf.commit() instead of looking up in the map (where the leaf presumably already came from)

crates/types/src/consensus.rs
```diff
@@ -963,6 +966,25 @@ impl<TYPES: NodeType> Consensus<TYPES> {
         let epoch_number = epoch_from_block_number(next_block_number, self.epoch_height);
         Ok(TYPES::Epoch::new(epoch_number))
     }
+
+    /// Returns true if the `parent_leaf` formed an eQC for the previous epoch to the `proposed_leaf`
+    pub fn check_eqc(&self, proposed_leaf: &Leaf<TYPES>, parent_leaf: &Leaf<TYPES>) -> bool {
```
bfish713, line 971: This function is checking does `proposed_leaf` contain an eQC as it' justify QC, is that right?

crates/types/src/consensus.rs
```diff
@@ -963,6 +966,25 @@ impl<TYPES: NodeType> Consensus<TYPES> {
...
+        if parent_leaf.height() == 0 {
+            return true;
+        }
```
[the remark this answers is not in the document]
bfish713: maybe just check if proposed_leaf.justify_qc is the gensis qc?

crates/task-impls/src/network.rs
```diff
@@ -380,24 +380,35 @@ impl<
...
+                        Err(e) => {
+                            tracing::warn!(
+                                "Failed to calculate leader for view number {:?}. Error: {:?}",
+                                view_number,
+                                e
+                            );
+                            return None;
+                        }
+                    };
+                    TransmitType::Direct(leader)
```
bfish713, line 403: I feel like this logic is more appropriate in the quorum_vote task, maybe just add the ExtendedVoteSend and handle that event here instead

crates/task-impls/src/helpers.rs
```diff
@@ -502,9 +503,29 @@ pub async fn validate_proposal_safety_and_liveness<
...
     // passes.
 
-    // Liveness check.
     {
         let consensus_reader = validation_info.consensus.read().await;
+        // Epoch safety check:
+        // The proposal is safe if
+        // 1. the proposed block and the justify QC block belong to the same epoch or
+        // 2. the justify QC is the eQC for the previous block
+        let proposal_epoch =
```
bfish713, line 512: Do we also need all this logic in the `validate_proposal_liveness` path, I suppose we can't actually do it there because we are missing the previous proposal but don't we need something to check the epochs?

bfish713 (commented):
I think it makes sense, I need to digest a bit more on how all the logic fits together

crates/types/src/consensus.rs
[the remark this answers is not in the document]
bfish713: I think maybe the only way is if the QC view number is the genesis view `TYPES::View::genesis`

crates/types/src/consensus.rs
bfish713: I was just making sure I was correct, if we already have the parent then it's fine to pass it in, but probably this function makes sense as just starting from the proposed_leaf

crates/task-impls/src/helpers.rs
bfish713: I don't think so, I was just trying to be sure we didn't. I also don't think we can do the check in the liveness path, and we won't vote in that case either

crates/task-impls/src/network.rs
```diff
@@ -399,6 +399,16 @@ impl<
                     TransmitType::Direct(leader),
                 ))
             }
+            HotShotEvent::ExtendedQuorumVoteSend(vote) => {
```
ss-es, line 402: is there a reason we don't want to just use `QuorumVoteSend`?

crates/task-impls/src/quorum_vote/handlers.rs
```diff
@@ -298,6 +298,7 @@ pub(crate) async fn submit_vote<TYPES: NodeType, I: NodeImplementation<TYPES>, V
     storage: Arc<RwLock<I::Storage>>,
     leaf: Leaf<TYPES>,
     vid_share: Proposal<TYPES, VidDisperseShare<TYPES>>,
+    extended_vote: bool,
```
ss-es, line 301: I'm not sure I understand the reason for this either -- are we planning for the `submit_vote` to actually be different for an eqc vs a regular qc? it seems like the only difference is we create a different internal hotshot event (which gets broadcast in exactly the same way)

ss-es (commented):
some minor things that stuck out to me

bfish713 (approved):

PR #3862, opened 2024-11-06
chore(deps): bump `jf-signature` crate to v0.2.0

mrain:
Closes #<ISSUE_NUMBER>

### This PR: 
Bump `jf-signature` crate to v0.2.0 that prevents you from unnoticeably doing something bad with the private keys.
Thus every struct that contains a private signature key type cannot have a derived `serde::Serialize`. So I moved `ValidatorConfig` out of the `HotShotConfig`.

### This PR does not: 

There's still an indirect deps to `jf-signatire` v0.1.0 through PushCDN 0.4.7. It'll be removed after upgrading to PushCDN >0.5.4


ss-es (approved):
looks good to me!

PR #3863, opened 2024-11-06
Add randomized Membership impls for testing

### LINKED ISSUE #3826: Add Membership impl with differing stake table for testing
- [x] Build out basic struct with different randomization implementations
  - [x] Fixed quorum size (no overlap, f overlap, f+1 overlap, 2f overlap, 2f+1 overlap, 3f overlap)
  - [x] Variable quorum size (random between 4-n)
- [x] Figure out how to link into tests

pls148:
Closes #3826 

This branch has been rebased against another PR (https://github.com/EspressoSystems/HotShot/pull/3867) which introduces a bunch of other changes to Memberships

A separate branch pre-rebase has been saved off here, if that helps with reviewing: https://github.com/EspressoSystems/HotShot/tree/ps/test-randomized-committee-premerge

### This PR: 

* Adds additional parser work for cross_tests! to enable template parameters in test types
* Creates two new types of Membership (StableQuorumIterator and RandomOverlapQuorumIterator)
* Adds additional test type test_epoch_success which use StableQuorumIterator and RandomOverlapQuorumIterator

### Key places to review: 

The implementations in hotshot/traits/election/helpers.rs, make sure those are commented well enough to be understood
Review the way I'm passing filter configs into TestTypesRandomizedCommitteeMembers (*FilterConfig)

CI


crates/hotshot/src/traits/election/randomized_committee_members.rs
```diff
@@ -0,0 +1,250 @@
...
+    fn new(
+        eligible_leaders: Vec<PeerConfig<<TYPES as NodeType>::SignatureKey>>,
+        committee_members: Vec<PeerConfig<<TYPES as NodeType>::SignatureKey>>,
+        committee_topic: Topic,
+    ) -> Self {
+        // The only time these two are unequal is in a benchmarking scenario that isn't covered by this test case
+        assert_eq!(
+            eligible_leaders, committee_members,
+            "eligible_leaders should be the same as committee_members!"
+        );
```
rob-maron, line 73: Unless the committees have been unified (across DA/quorum) yet, we expect eligible leaders to be different than the full list of members in the DA committee. This is because we want to enforce that quorum leaders == da leaders for now

crates/hotshot/src/traits/election/randomized_committee_members.rs
```diff
@@ -0,0 +1,250 @@
...
+    /// Get the total number of nodes in the committee
+    fn total_nodes(&self, epoch: <TYPES as NodeType>::Epoch) -> usize {
+        self.make_quorum_filter(epoch).len()
+    }
+
+    /// Get the voting success threshold for the committee
+    fn success_threshold(&self) -> NonZeroU64 {
+        let len = self.stake_table.len() / 2; // Divide by two as we are flipping between odds and evens
+        NonZeroU64::new(((len as u64 * 2) / 3) + 1).unwrap()
+    }
```
rob-maron, line 237: I'm not entirely sure on this - but if this PR introduces stake tables of variable length would we want to update the voting thresholds accordingly?

crates/hotshot/src/traits/networking/memory_network.rs
```diff
@@ -302,22 +302,53 @@ impl<K: SignatureKey + 'static> ConnectedNetwork<K> for MemoryNetwork<K> {
...
+                tracing::error!("Skipping node because not in recipient list: {:?}", &node.0);
+                continue;
+            }
+            // TODO delay/drop etc here
+            let (key, node) = node;
+            trace!(?key, "Sending message to node");
+            if let Some(ref config) = &self.inner.reliability_config {
+                {
+                    let node2 = node.clone();
+                    let fut = config.chaos_send_msg(
```
rob-maron, line 326: Is the `broadcast_message` function copied out to here? Curious - what's the difference in what we're doing between the two?

rob-maron (commented):
Looks good in general, couple of questions

crates/hotshot/src/traits/election/helpers.rs
```diff
@@ -0,0 +1,442 @@
...
+}
+
+/// Create a single u64 seed by merging two u64s. Done this way to allow easy seeding of the number generator
+/// from both a stable SOUND as well as a moving value ROUND (typically, epoch). Shift left by 8 to avoid
+/// scenarios where someone manually stepping seeds would pass over the same space of random numbers across
+/// sequential rounds. Doesn't have to be 8, but has to be large enough that it is unlikely that a given
+/// test run will collide; using 8 means that 256 rounds (epochs) would have to happen inside of a test before
+/// the test starts repeating values from SEED+1.
+fn make_seed(seed: u64, round: u64) -> u64 {
+    seed.wrapping_add(round.wrapping_shl(8))
```
ss-es, line 61: would just doing something like `seed.bitxor(round)` work?

crates/hotshot/src/traits/election/randomized_committee_members.rs
```diff
@@ -0,0 +1,250 @@
...
+        Ok(TYPES::SignatureKey::public_key(&res))
+    }
+
+    /// Get the total number of nodes in the committee
+    fn total_nodes(&self, epoch: <TYPES as NodeType>::Epoch) -> usize {
+        self.make_quorum_filter(epoch).len()
+    }
+
+    /// Get the voting success threshold for the committee
+    fn success_threshold(&self) -> NonZeroU64 {
```
ss-es, line 234: I think these functions need to take `epoch` as an argument, because the success/failure thresholds (as an absolute rather than a percentage) depends on the membership size in that epoch

rather than `stake_table` I think we need to call `stake_table()`

ss-es (commented):
looks great to me!

just left a few comments. most important one is about the `success_threshold`/`failure_threshold` -- unless I'm misunderstanding something, I think that needs to be changed. but that could be a separate PR if you prefer

crates/hotshot/src/traits/election/randomized_committee_members.rs
ss-es: oh actually, nevermind -- I see what you did!

yes I think this should just be a separate PR; it's correct as-is the way you wrote it

crates/hotshot/src/traits/election/randomized_committee_members.rs
ss-es: I think this is a bit weird because `committee_members` here doesn't mean DA (the word `committee` is probably the wrong word here). the DA committee is currently the `committee_members` of `da_membership`, but @tbro is unifying the 2 by eliminating the `Memberships` wrapper struct

I think the only thing `eligible_leaders` is used for is VID benchmarking with a fixed leader

crates/hotshot/src/traits/networking/memory_network.rs
rob-maron: Ah right, I forgot about that stipulation. SGTM, was just curious

crates/types/src/simple_vote.rs
```diff
@@ -24,6 +24,9 @@ use crate::{
     vote::{HasViewNumber, Vote},
 };
 
+/// Marker that data should use the quorum cert type
+pub(crate) trait QuorumMaker {}
```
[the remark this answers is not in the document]
ss-es: from context I think it's meant to be `QuorumMarker` (it looks like it indicates that things need to be signed by the full quorum rather than DA)

crates/testing/tests/tests_1/test_success.rs
```diff
@@ -41,6 +41,25 @@ cross_tests!(
...
+//             completion_task_description: CompletionTaskDescription::TimeBasedCompletionTaskBuilder(
+//                                              TimeBasedCompletionTaskDescription {
+//                                                  duration: Duration::from_secs(60),
+//                                              },
+//                                          ),
+//             ..TestDescription::default()
+//         }
+//     },
+// );
+
```
ss-es, line 62: is this meant to be commented out?

ss-es (approved):
looks good to me, not sure if the test was meant to be commented out or not

PR #3867, opened 2024-11-08
Flatten memberships into a single type

tbro:
Needed for #3726

### This PR: 
Removes `Memberships` struct and updates `Membership` trait to support DA and Quorum w/ a single interface. Essentially, instead of having `qourom_membership` and `committee_membership` (DA) fields in most places we now have a single `membership` field. In some places, where only one of the membership types was previously used, the previous name is preserved. To obtain the correct data, we now have to call the correct method: `membership.leader()` for quorum `membership.da_leader()` for da. 

### This PR does not: 
Shouldn't change behavior. It is only preparation for https://github.com/EspressoSystems/espresso-sequencer/pull/2237

### Key places to review: 
The trait is used extensively and there are changes across many files. 

   * The trait lives in `crates/types/src/traits/election.rs`
   * `SimpleCert` impls in `crates/types/src/simple_certificate.rs`


crates/types/src/traits/election.rs
```diff
@@ -82,15 +120,28 @@ pub trait Membership<TYPES: NodeType>: Clone + Debug + Send + Sync {
...
     /// Returns the number of total nodes in the committee in an epoch `epoch`
     fn total_nodes(&self, epoch: TYPES::Epoch) -> usize;
 
+    /// Returns the number of total DA nodes in the committee in an epoch `epoch`
+    fn da_total_nodes(&self, epoch: TYPES::Epoch) -> usize;
+
     /// Returns the threshold for a specific `Membership` implementation
     fn success_threshold(&self) -> NonZeroU64;
 
+    /// Returns the DA threshold for a specific `Membership` implementation
```
tbro, line 142: link

crates/task-impls/src/network.rs
```diff
@@ -676,10 +674,10 @@ impl<
             kind: message_kind,
         };
         let view_number = message.kind.view_number();
-        let committee_topic = self.quorum_membership.committee_topic();
+        let committee_topic = Topic::Global;
```
tbro, line 677: link

crates/testing/src/helpers.rs
```diff
@@ -145,10 +137,10 @@ pub async fn build_cert<
     V: Versions,
     DATAType: Committable + Clone + Eq + Hash + Serialize + Debug + 'static,
     VOTE: Vote<TYPES, Commitment = DATAType>,
-    CERT: Certificate<TYPES, Voteable = VOTE::Commitment>,
+    CERT: Certificate<TYPES, VOTE::Commitment, Voteable = VOTE::Commitment>,
 >(
     data: DATAType,
-    membership: &TYPES::Membership,
+    da_membership: &TYPES::Membership,
```
ss-es, line 143: I think this should take the stake table + threshold directly (and therefore `build_assembled_sig` should too)

this isn't used in the integration tests like `libp2p::libp2p_network` though, so shouldn't have anything to do with the failure

crates/testing/src/helpers.rs
```diff
@@ -376,18 +367,16 @@ pub async fn build_da_certificate<TYPES: NodeType, V: Versions>(
...
+        vid_commitment(&encoded_transactions, membership.total_nodes(epoch_number));
 
     let da_data = DaData {
         payload_commit: da_payload_commitment,
     };
 
     build_cert::<TYPES, V, DaData, DaVote<TYPES>, DaCertificate<TYPES>>(
         da_data,
-        da_membership,
+        membership,
```
ss-es, line 379: this should pass the da stake table + threshold directly

crates/types/src/simple_certificate.rs
```diff
@@ -145,18 +147,103 @@ impl<TYPES: NodeType, VOTEABLE: Voteable + 'static, THRESHOLD: Threshold<TYPES>>
             _pd: PhantomData,
         }
     }
-    async fn is_valid_cert<MEMBERSHIP: Membership<TYPES>, V: Versions>(
+    async fn is_valid_cert<V: Versions>(
         &self,
+        stake_table: Vec<<TYPES::SignatureKey as SignatureKey>::StakeTableEntry>,
```
ss-es, line 152: not sure if this is immediately fixable (might require epoch numbers in the certs, which is being worked on separately), but I think `is_valid_cert` should actually just take the `Memberships` directly, rather than the stake table + threshold

we always know which stake table/threshold to use for each cert

can you make an issue for this and add it to the board

ss-es (commented):
left some comments, but couldn't find the issue. I'll check a few more things

crates/hotshot/src/traits/election/static_committee.rs
```diff
@@ -19,37 +18,48 @@ use hotshot_types::{
...
 #[derive(Clone, Debug, Eq, PartialEq, Hash)]
-
 /// The static committee election
 pub struct StaticCommittee<T: NodeType> {
     /// The nodes eligible for leadership.
     /// NOTE: This is currently a hack because the DA leader needs to be the quorum
     /// leader but without voting rights.
     eligible_leaders: Vec<<T::SignatureKey as SignatureKey>::StakeTableEntry>,
+    /// The nodes eligible for leadership.
+    da_eligible_leaders: Vec<<T::SignatureKey as SignatureKey>::StakeTableEntry>,
```
rob-maron, line 29: Shouldn't the `da_eligible_leaders` be == `eligible_leaders` since they are the same for now? In that case, `da_leader()` or associated functions should always return as a leader from the selection of all nodes, not just DA

tbro: https://github.com/EspressoSystems/HotShot/pull/3867/commits/36f2f13005b314f9dbfbc2b489a81ec8da9b1933

crates/examples/infra/mod.rs
```diff
@@ -381,22 +381,9 @@ pub trait RunDa<
...
         // Create the quorum membership from all nodes, specifying the committee
         // as the known da nodes
-        let da_membership =
-            <TYPES as NodeType>::Membership::new(all_nodes.clone(), da_nodes, Topic::Da);
-
-        let memberships = Memberships {
-            quorum_membership: quorum_membership.clone(),
-            da_membership,
-        };
+        let memberships = <TYPES as NodeType>::Membership::new(all_nodes.clone(), da_nodes);
```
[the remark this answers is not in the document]
tbro: 178ff184bb28f6209529a95d2d7877b821401f99

crates/examples/infra/mod.rs
```diff
@@ -752,7 +738,7 @@ where
 
         // Create the qurorum membership from the list of known nodes
         let all_nodes = config.config.known_nodes_with_stake.clone();
-        let quorum_membership = TYPES::Membership::new(all_nodes.clone(), all_nodes, Topic::Global);
+        let quorum_membership = TYPES::Membership::new(all_nodes.clone(), all_nodes);
```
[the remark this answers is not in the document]
tbro: d7b9242080ba86418214c9430d5e64b77e0958c4

crates/task-impls/src/request.rs
```diff
@@ -180,15 +180,15 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>> NetworkRequestState<TYPES, I
         let public_key = self.public_key.clone();
 
         // Get the committee members for the view and the leader, if applicable
-        let mut da_committee_for_view = self.da_membership.committee_members(view, epoch);
-        if let Ok(leader) = self.da_membership.leader(view, epoch) {
+        let mut da_committee_for_view = self.da_membership.da_committee_members(view, epoch);
```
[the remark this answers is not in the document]
tbro: c6d30fb7d6c285dbb1fb78a922f82221563e004a

crates/types/src/simple_certificate.rs
tbro: Before this PR that was the logic. We changed it b/c `is_valid_cert` can't be generic of `Membership` b/c we have to call specific functions such `stake_table` or `da_stake_table`. This PR addresses this by passing in the stake table.

crates/testing/src/helpers.rs
tbro: Will we always just pass in DA stake table and success_threshold? If so it isn't clear to me why we need a generic `CERT`.

crates/testing/src/helpers.rs
ss-es: rethinking my comment here, will leave some thoughts on zulip

crates/testing/src/helpers.rs
ss-es: I'm rethinking some of my comments here, but I'll leave some thoughts on zulip

ss-es (approved):
approving, but I think we should follow-up on some of the changes here (left some comments on Zulip)

crates/testing/src/helpers.rs
ss-es, line 143: ignore the previous remark, but I think this should stay

```suggestion
    membership: &TYPES::Membership,
```

ss-es (changes requested):
oops, forgot one thing -- I think `da_membership` is the wrong name here

crates/task-impls/src/quorum_proposal/handlers.rs
```diff
@@ -124,8 +125,8 @@ impl<TYPES: NodeType, V: Versions> ProposalDependencyHandle<TYPES, V> {
             if let HotShotEvent::HighQcRecv(qc, _sender) = event.as_ref() {
                 if qc
                     .is_valid_cert(
-                        self.quorum_membership.as_ref(),
-                        TYPES::Epoch::new(0),
+                        self.quorum_membership.stake_table(TYPES::Epoch::new(0)),
```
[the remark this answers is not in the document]
tbro: https://github.com/EspressoSystems/HotShot/issues/3917#issue-2690102778

crates/types/src/simple_certificate.rs
tbro: https://github.com/EspressoSystems/HotShot/issues/3918#issue-2690133601

crates/testing/src/helpers.rs
tbro: 7aaf224f1fb6bd306101cd2900e0534224255ce6

ss-es (approved):

PR #3874, opened 2024-11-11
add-audit-folder-and-internal-report

0xkato:
Adding an Audit folder, containing the internal review which can also be found in [espresso-audits](https://github.com/EspressoSystems/espresso-audits).


ss-es (approved):
approving, but minor: can the folder be renamed `audits` (lowercase `a`) for consistency with the other folders?

PR #3875, opened 2024-11-11
Fallible tracing subscription

rob-maron:


tbro (approved):
Fixes the issue locally

PR #3877, opened 2024-11-11
Hotstuff 2

bfish713:
Closes #<ISSUE_NUMBER>

### This PR: 

This PR introduces the Hotstuff 2 consensus rules.  Hotstuff 2 is has some notable changes over the original.  These are:

- The decide rule goes from a 3 chain to a 2 chain
- The leader must wait for a QC from the last view (like before) or for a set timeout to collect the highest QC from all other replicas.  The leader proposes with the highest QC it's received
- Nodes send the highest QC they've seen to the next leader when a view finishes.  
- The locked QC becomes the what was before the "high qc" 
    - In other words replicas will not vote for any proposal with a justify QC from a lesser view than any justify QC they previously saw.

### DIFF
--- a/crates/task-impls/src/quorum_proposal/handlers.rs
+++ b/crates/task-impls/src/quorum_proposal/handlers.rs
@@ -90,6 +90,8 @@ pub struct ProposalDependencyHandle<TYPES: NodeType, V: Versions> {
     /// Shared consensus task state
     pub consensus: OuterConsensus<TYPES>,
 
+    /// View timeout from config.
+    pub timeout: u64,
     /// The most recent upgrade certificate this node formed.
     /// Note: this is ONLY for certificates that have been formed internally,
     /// so that we can propose with them.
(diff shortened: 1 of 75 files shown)


crates/types/src/consensus.rs
```diff
@@ -949,7 +949,7 @@ impl<TYPES: NodeType> Consensus<TYPES> {
         let mut is_leaf_extended = true;
         if let Err(e) = self.visit_leaf_ancestors(
             leaf_view,
-            Terminator::Inclusive(leaf_view - 2),
+            Terminator::Inclusive(leaf_view - 1),
```
bfish713, line 952: @lukaszrzasik In HS2 the length of the decide rule is 1 less, so I also reduced it here, are there other places in the epoch code that assume the longer chain length of the old decide rule?

crates/task-impls/src/quorum_proposal/handlers.rs
```diff
@@ -97,9 +104,78 @@ pub struct ProposalDependencyHandle<TYPES: NodeType, V: Versions> {
...
+    async fn wait_for_qc_event(
+        &self,
+        rx: &mut Receiver<Arc<HotShotEvent<TYPES>>>,
+    ) -> Option<QuorumCertificate<TYPES>> {
+        while let Ok(event) = rx.recv_direct().await {
+            if let HotShotEvent::HighQcRecv(qc, _sender) = event.as_ref() {
+                if qc
+                    .is_valid_cert(
+                        self.quorum_membership.as_ref(),
+                        TYPES::Epoch::new(0),
```
bfish713, line 126: @lukaszrzasik Need to get a proper epoch here, I know we can get this from consensus, but need to do this properly like we discussed maybe it's trivial if the epoch is on the QC

crates/types/src/consensus.rs
bfish713: Ok I think I see what's going on, I think this makes things quite confusing though.  I thought with dual certificates we could just keep the decide rule the same.  I would strongly prefer to have no difference between decide rules and help others reach the next epoch simply by seeing proposals from the new epoch. 

For now we can try to adjust this to work with the current code but, in it's current state, it only works when you mismatch original hotstuff and this decide rule.  Basically the current code works because the decide rule we have (not hs-2) is the same as the epoch transition rule.  So if the chain should be one longer with original HS then the changing this code to `-3` shouldn't break things, but I think it does.

Issue #3880, opened 2024-11-12
[DRB] - Benchmark the hash time


shenkeyao:

mrain:
I did a few trials here: https://github.com/mrain/df-sha2-bench

Conclusion so far:
- `sha2` crate already utilizes CPU ISA acceleration, with `asm` feature enabled.
- Current implementation is already the best by now.

mrain:
More on this: an 1 cycle per byte sha2 algorithm ([which can hardly be real](https://bench.cr.yp.to/results-hash.html)) that runs on a 10GHz ASIC can only achieve ~312.5M iterations per second.
While the in my bench above we could get 27M iterations per second on most modern CPUs.

PR #3890, opened 2024-11-14
[DRB] - Add DRB functions

### LINKED ISSUE #3855: [DRB] - Add DRB functions

shenkeyao:
Closes #3855.

### This PR: 
* Creates a new create for dynamic leader election and adds functions for:
  * DRB difficulty level,
  * DRB seed result, and
  * Leader election.

### This PR does not: 
* Update existing code to call these functions.

### DIFF
--- a/crates/hotshot/src/traits/election/dynamic.rs
+++ b/crates/hotshot/src/traits/election/dynamic.rs
@@ -1,0 +1,68 @@
+// Copyright (c) 2021-2024 Espresso Systems (espressosys.com)
+// This file is part of the HotShot repository.
+
+// You should have received a copy of the MIT License
+// along with the HotShot repository. If not, see <https://mit-license.org/>.
+
+use anyhow::{Context, Result};
+use hotshot_types::traits::{node_implementation::NodeType, signature_key::SignatureKey};
+use sha2::{Digest, Sha256};
+
+/// Time a DRB calculation will take, in terms of number of views.
+const DRB_CALCULATION_NUM_VIEW: u64 = 300;
+
+// TODO: Replace `hashes_per_second` parameter with a const once we benchmark the hash time.
+// <https://github.com/EspressoSystems/HotShot/issues/3880>
+/// Difficulty level of the DRB calculation.
+///
+/// Represents the number of times the hash function will be repeatedly called.
+///
+/// # Arguments
+/// * `hashes_per_second` - Number of hashes that can be completed in a second.
+/// * `view_timeout` - Timeout duration for a view, in seconds.
+#[must_use]
+pub fn difficulty_level(hashes_per_second: u64, view_timeout: u64) -> u64 {
+    hashes_per_second * view_timeout * DRB_CALCULATION_NUM_VIEW
+}
+
+// TODO: Replace `hashes_per_second` parameter with a const once we benchmark the hash time.
+// <https://github.com/EspressoSystems/HotShot/issues/3880>
+/// Compute the DRB result for the leader rotation.
+///
+/// This is to be started two epochs in advance.
+///
+/// # Arguments
+/// * `qc_signature` - Aggregated signature in the current quorum proposal.
+/// * `view_timeout` - Timeout duration for a view, in seconds.
+///
+/// # Errors
+/// * If failed to serialize the QC signature.
+pub fn compute_drb<TYPES: NodeType>(
+    qc_signature: &<TYPES::SignatureKey as SignatureKey>::QcType,
+    hashes_per_second: u64,
+    view_timeout: u64,
+) -> Result<usize> {
+    // Hash the QC signature.
+    let mut hash = bincode::serialize(&qc_signature)
+        .with_context(|| "Failed to serialize the QC signature.")?;
+    let difficulty_level = difficulty_level(hashes_per_second, view_timeout);
+    for _iter in 0..difficulty_level {
+        hash = Sha256::digest(hash).to_vec();
+    }
+
+    // Convert the hash to a seed as the DRB result.
+    let mut seed = [0u8; 8];
+    seed.copy_from_slice(&hash[..size_of::<usize>()]);
+    Ok(usize::from_le_bytes(seed))
+}
+
+/// Use the DRB result to get the leader.
+#[must_use]
+pub fn leader<TYPES: NodeType>(
+    stake_table: &[<TYPES::SignatureKey as SignatureKey>::StakeTableEntry],
+    drb: usize,
+) -> TYPES::SignatureKey {
+    let index = drb % stake_table.len();
+    let res = stake_table[index].clone();
+    TYPES::SignatureKey::public_key(&res)
+}
--- a/crates/hotshot/src/traits/election/dynamic.rs
+++ b/crates/hotshot/src/traits/election/dynamic.rs
@@ -6,7 +6,6 @@
 
 use std::hash::{DefaultHasher, Hash, Hasher};
 
-use anyhow::{Context, Result};
 use hotshot_types::traits::{node_implementation::NodeType, signature_key::SignatureKey};
 use sha2::{Digest, Sha256};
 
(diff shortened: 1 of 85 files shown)


crates/hotshot/src/traits/election/dynamic.rs
```diff
@@ -0,0 +1,68 @@
...
+    qc_signature: &<TYPES::SignatureKey as SignatureKey>::QcType,
+    hashes_per_second: u64,
+    view_timeout: u64,
+) -> Result<usize> {
+    // Hash the QC signature.
+    let mut hash = bincode::serialize(&qc_signature)
+        .with_context(|| "Failed to serialize the QC signature.")?;
+    let difficulty_level = difficulty_level(hashes_per_second, view_timeout);
+    for _iter in 0..difficulty_level {
+        hash = Sha256::digest(hash).to_vec();
```
mrain, line 50: we don't have to optimize it now. But there seems to be some redundant memallocs/memcopies here. In the future or in benchmark (#3880 ) we should hope for some tailored implementation, even in assembly.

crates/hotshot/src/traits/election/dynamic.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    // Convert the hash to a seed as the DRB result.
+    let mut seed = [0u8; 8];
+    seed.copy_from_slice(&hash[..size_of::<usize>()]);
+    Ok(usize::from_le_bytes(seed))
+}
+
+/// Use the DRB result to get the leader.
+#[must_use]
+pub fn leader<TYPES: NodeType>(
```
mrain, line 61: This looks like an unweighted version. Should we leave a TODO/PR so that we won't forget it in the future?

crates/hotshot/src/traits/election/dynamic.rs
```diff
@@ -0,0 +1,68 @@
...
+/// * `qc_signature` - Aggregated signature in the current quorum proposal.
+/// * `view_timeout` - Timeout duration for a view, in seconds.
+///
+/// # Errors
+/// * If failed to serialize the QC signature.
+pub fn compute_drb<TYPES: NodeType>(
+    qc_signature: &<TYPES::SignatureKey as SignatureKey>::QcType,
+    hashes_per_second: u64,
+    view_timeout: u64,
+) -> Result<usize> {
```
mrain, line 44: `usize` may not contain sufficient entropy if the total stake weight is large.
I would recommend using any 256 bits type like `[u8; 32]`.

crates/hotshot/src/traits/election/dynamic.rs
```diff
@@ -0,0 +1,68 @@
...
+/// * `hashes_per_second` - Number of hashes that can be completed in a second.
+/// * `view_timeout` - Timeout duration for a view, in seconds.
+#[must_use]
+pub fn difficulty_level(hashes_per_second: u64, view_timeout: u64) -> u64 {
+    hashes_per_second * view_timeout * DRB_CALCULATION_NUM_VIEW
+}
+
+// TODO: Replace `hashes_per_second` parameter with a const once we benchmark the hash time.
+// <https://github.com/EspressoSystems/HotShot/issues/3880>
+/// Compute the DRB result for the leader rotation.
```
mrain, line 30: Should we also document here that it's better to spawn a non-blocking thread for this drb computation?

crates/hotshot/src/traits/election/dynamic.rs
```diff
@@ -0,0 +1,68 @@
...
+/// This is to be started two epochs in advance.
+///
+/// # Arguments
+/// * `qc_signature` - Aggregated signature in the current quorum proposal.
+/// * `view_timeout` - Timeout duration for a view, in seconds.
+///
+/// # Errors
+/// * If failed to serialize the QC signature.
+pub fn compute_drb<TYPES: NodeType>(
+    qc_signature: &<TYPES::SignatureKey as SignatureKey>::QcType,
```
ss-es, line 41: can we make a function that just takes `drb_seed: [u8; 32]`? we can then pass the serialized QC signature when calling it

e.g. similar to https://github.com/EspressoSystems/jellyfish/pull/683/files/de12a3272fe416a655c2b339538c5f43aaf314b3#diff-bb9e8b35f8b9373a119340a627c9c41be713b550265afec50c988c0994016322R52-R58

crates/hotshot/src/traits/election/dynamic.rs
shenkeyao: Added a comment!

crates/hotshot/src/traits/election/dynamic.rs
```diff
@@ -0,0 +1,68 @@
...
+///
+/// Represents the number of times the hash function will be repeatedly called.
+///
+/// # Arguments
+/// * `hashes_per_second` - Number of hashes that can be completed in a second.
+/// * `view_timeout` - Timeout duration for a view, in seconds.
+#[must_use]
+pub fn difficulty_level(hashes_per_second: u64, view_timeout: u64) -> u64 {
+    hashes_per_second * view_timeout * DRB_CALCULATION_NUM_VIEW
+}
```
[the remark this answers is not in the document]
shenkeyao: As discussed on Zulip we don't have all these values for now. I modified this function and added some comments. Hopefully makes sense!

crates/hotshot/src/traits/election/dynamic.rs
shenkeyao: I'm a bit confused by "just takes drb_seed: [u8; 32]" and "pass the serialized QC signature" since the former suggests that the argument of the function is the DRB seed but the latter suggests the QC signature. Does the updated version look good to you? If not, could you clarify a bit more, especially whether you'd like me to _replace_ the `compute_drb_seed` function or _add_ a function for some other feature?

ss-es: sorry, to clarify I mean: can the function take `drb_seed: [u8; 32]` as input *instead of* `qc_signature: &<..>::QcType`

this way we can avoid the `Result<..>` output, and only have to worry about QC serialization when plugging into the function

crates/hotshot/src/traits/election/dynamic.rs
```diff
@@ -0,0 +1,78 @@
...
+    stake_table: &[<TYPES::SignatureKey as SignatureKey>::StakeTableEntry],
+    drb_seed: [u8; 32],
+) -> TYPES::SignatureKey {
+    let mut hasher = DefaultHasher::new();
+    drb_seed.hash(&mut hasher);
+    #[allow(clippy::cast_possible_truncation)]
+    let start_index = (hasher.finish() as usize) % stake_table.len();
+    // TODO: Index with a weighted stake table.
+    // <https://github.com/EspressoSystems/HotShot/issues/3898>
+    let entry = stake_table[start_index + view_number].clone();
```
ss-es, line 76: this will overflow, I think we instead want `view_number.hash(&mut hasher)` before calculating `start_index`

crates/hotshot/src/traits/election/dynamic.rs
shenkeyao: Modified the input and renamed parameters as discussed on Zulip.

crates/hotshot/src/traits/election/dynamic.rs
```diff
@@ -0,0 +1,75 @@
...
+pub fn leader<TYPES: NodeType>(
+    view_number: usize,
+    stake_table: &[<TYPES::SignatureKey as SignatureKey>::StakeTableEntry],
+    drb_result: [u8; 32],
+) -> TYPES::SignatureKey {
+    let mut hasher = DefaultHasher::new();
+    drb_result.hash(&mut hasher);
+    view_number.hash(&mut hasher);
+    #[allow(clippy::cast_possible_truncation)]
+    let index = (hasher.finish() as usize) % stake_table.len();
```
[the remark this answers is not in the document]
shenkeyao: Yeah, I added a TODO comment about switching to the weighted version, but it should be put before this `len()` call. Updated!

mrain (approved):
LGTM!

ss-es (approved):
I think this looks good to me

PR #3911, opened 2024-11-23
[DRB] - Start the DRB calculation two epochs in advance

### LINKED ISSUE #3857: [DRB] - Start the DRB calculation two epochs in advance

shenkeyao:
Closes #3857.

### This PR: 
* Calls the DRB function in the vote task to start the computation for two views in advance.
* Updates the vote task type to include spawned DRB computation tasks.
* Moves and updates the DRB implementation file.

### This PR does not: 
* Handles the use of the DRB result, which will be done in the upcoming PRs.

### Key places to review: 
* All files, but the main changes are in `handlers.rs`.


crates/hotshot/src/tasks/task_state.rs
```diff
@@ -233,6 +233,7 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> CreateTaskState
             network: Arc::clone(&handle.hotshot.network),
             quorum_membership: handle.hotshot.memberships.quorum_membership.clone().into(),
             da_membership: handle.hotshot.memberships.da_membership.clone().into(),
+            drb_computations: BTreeMap::new(),
```
ss-es, line 236: do we need a `BTreeMap` here? I think we only want to run one of these calculations at a time

we can e.g. store the seeds and results in a map and trigger the start of the calculation at `block_height % epoch_height = 1` or something

crates/task-impls/src/quorum_vote/handlers.rs
```diff
@@ -54,6 +58,77 @@ pub(crate) async fn handle_quorum_proposal_validated<
...
+    // * Epoch 1 and 2: Use `[0u8; 32]` as the DRB result since when we first start the computation
+    // in epoch 1, the result is for epoch 3.
+    //
+    // We don't need to handle the special cases explicitly here, because the first proposal with
+    // which we'll start the DRB computation is for epoch 3.
+    if version >= V::Epochs::VERSION {
+        let qc_block_number = proposal.block_header.block_number() - 1;
+
+        // Skip if this is not the expected block.
+        if qc_block_number + 3 == task_state.epoch_height {
```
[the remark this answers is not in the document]
shenkeyao: Somehow I thought `qc_block_number` was the index _in_ the epoch. Good catch!

crates/task-impls/src/quorum_vote/handlers.rs
```diff
@@ -54,6 +58,77 @@ pub(crate) async fn handle_quorum_proposal_validated<
...
+    //
+    // Special cases:
+    // * Epoch 0: No DRB computation since we'll transition to epoch 1 immediately.
+    // * Epoch 1 and 2: Use `[0u8; 32]` as the DRB result since when we first start the computation
+    // in epoch 1, the result is for epoch 3.
+    //
+    // We don't need to handle the special cases explicitly here, because the first proposal with
+    // which we'll start the DRB computation is for epoch 3.
+    if version >= V::Epochs::VERSION {
+        let qc_block_number = proposal.block_header.block_number() - 1;
```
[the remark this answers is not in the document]
shenkeyao: Started a Zulip discussion to follow up on this: https://espresso.zulipchat.com/#narrow/channel/462904-eng-proof-of-stake/topic/epoch.2FPoS.20ugprade/near/484361911.

crates/hotshot/src/tasks/task_state.rs
shenkeyao: > `block_height % epoch_height = 1`

Assuming this is `(qc_block_number + 3) % task_state.epoch_height == 0` instead?

Makes sense not to use `BTreeMap`. I'm going to add a `DrbTasks` struct for this, containing the previous DRB result and the in-progress DRB calculation task, since we may need both of them in each epoch. Hopefully, it's not an overkill!

shenkeyao: On second thought, `BTreeMap` might be more suitable here than a `HashMap` or the new struct I proposed above. We may have at most three things to store here, two DRB results and one in-progress task. I think `BTreeMap` makes garbage collection easier because the three epochs may not be consecutive due to the membership.

Lmk if this doesn't make sense or there's a better design though!

shenkeyao: Fixed. The update also contains some minor changes to replace those `[0; 32]` and `[0; 96]` values with consts.

crates/task-impls/src/quorum_vote/handlers.rs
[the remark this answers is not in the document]
shenkeyao: Fixed. The update also contains some minor changes to replace those `[0; 32]` and `[0; 96]` values with consts.

crates/task-impls/src/quorum_vote/handlers.rs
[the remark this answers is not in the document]
shenkeyao: As discussed on Zulip, moved this part to the end of the function and used the decided QC for calculation.

ss-es (approved):
I think this looks good, but I'll write some thought

crates/hotshot/src/tasks/task_state.rs
shenkeyao: As discussed on Zulip, I'll update the maps in a separate PR. Issue: https://github.com/EspressoSystems/HotShot/issues/3933.

PR #3919, opened 2024-11-25
Handle HighQcSend

lukaszrzasik:
Closes #<ISSUE_NUMBER>

### DIFF
--- a/.github/workflows/typos.yml
+++ b/.github/workflows/typos.yml
@@ -20,4 +20,4 @@ jobs:
         name: Checkout Repository
 
       - name: typos-action
-        uses: crate-ci/typos@v1.27.3
+        uses: crate-ci/typos@v1.28.1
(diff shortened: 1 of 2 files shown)


crates/task-impls/src/network.rs
```diff
@@ -652,6 +652,13 @@ impl<
                     TransmitType::Direct(to),
                 ))
             }
+            HotShotEvent::HighQcSend(quorum_cert, leader, sender) => Some((
+                sender,
+                MessageKind::Consensus(SequencingMessage::General(
+                    GeneralConsensusMessage::HighQc(quorum_cert.to_qc()),
```
ss-es, line 658: can we change `GeneralConsensusMessage::HighQc` to a `QuorumCertificate2`?

it's not deployed so I think we can make the breaking change, and I think we would be gated on the same version (epochs)

ss-es (commented):
just one comment, otherwise looks fine to me

ss-es (approved):

PR #3922, opened 2024-11-26
Lr/double quorum

lukaszrzasik:
Closes #<ISSUE_NUMBER>

### This PR: 

* Modifies the `QuorumProposal` and `Leaf2` types to include the optional next epoch justify QC.
* Adds a new type `NextEpochQuorumData2`. It's a thin wrapper around `QuorumData2`. The additional type makes it possible to distinguish between vote and certificate type aliases.
* Adds a new hotshot event `NextEpochQc2Formed` to signal that a QC for has been formed using votes from the nodes in the next epoch.
* Adds next epoch vote collectors in the consensus task. The collectors collect votes from the nodes that belong to the next epoch. Given enough votes an additional (next epoch) QC is formed.
* Adds logic to save the payload early. The payload is stored as soon as the block is formed and the node is able send the DA proposal. Up until now the payload was stored only after a node received and validated the DA proposal. The payload might be needed earlier to be able to calculate VID shares for the nodes in the next epoch.
* Adds logic to attach the next epoch QC to the proposal if the proposal's normal justify QC is for the last block in the epoch.
* Modifies the proposal validating logic. Adds additional checks:
* * if it's an epoch transition proposal, make sure the next epoch justify QC exists
* * make sure the justify QC and the next epoch justify QC are for the same leaf
* * validate the signatures on the next epoch justify QC
* Modifies the voting logic so that the nodes from the next epoch vote for the last block in the previous epoch.
* Adds logic to send the VID shares for the next epoch nodes when sending a proposal for the last block in the epoch.
* Adds a field in the shared consensus state that stores the highest known next epoch QC. Adds a similar field to the storage.
* Adds an epoch parameter to the threshold methods in the `Membership` trait. This is required because the thresholds depend on the number of nodes in the given epoch.
* Modifies `OverallSafetyTask`. It now uses the total nodes number and threshold directly from the `Membership` object. This is required to properly test the types with the variable stake table.

### This PR does not: 

* Modify the DA logic. It means that the next epoch nodes do not participate in the DAC formation.
* Modify the timeout and view sync logic. These will be adjusted in the next PRs.

### DIFF
--- a/crates/hotshot/src/lib.rs
+++ b/crates/hotshot/src/lib.rs
@@ -493,8 +493,11 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> SystemContext<T
         trace!("Adding transaction to our own queue");
 
         let api = self.clone();
-        let view_number = api.consensus.read().await.cur_view();
-        let epoch = api.consensus.read().await.cur_epoch();
+
+        let consensus_reader = api.consensus.read().await;
+        let view_number = consensus_reader.cur_view();
+        let epoch = consensus_reader.cur_epoch();
+        drop(consensus_reader);
 
         // Wrap up a message
         let message_kind: DataMessage<TYPES> =


crates/types/src/utils.rs
```diff
@@ -262,9 +262,20 @@ pub fn mnemonic<H: Hash>(bytes: H) -> String {
...
     NotInTransition,
 }
+
+/// Returns true if the given block number is the last in the epoch based on the given epoch height.
+#[must_use]
+pub fn is_last_block_in_epoch(block_number: u64, epoch_height: u64) -> bool {
+    if block_number == 0 || epoch_height == 0 {
+        false
+    } else {
+        block_number % epoch_height == 0
```
shenkeyao, line 279: As mentioned on Zulip, I think this should be `block_number + 1`.

shenkeyao: Nvm now I understand that block 0 is a special case.

ss-es (commented):
I think it makes sense overall, just some comments

PR #3941, opened 2024-12-03
#3933 DrbComputations helper for quorum_vote

### LINKED ISSUE #3933: [DRB] - Update the computation map
Zulip discussion: https://espresso.zulipchat.com/#narrow/channel/462904-eng-proof-of-stake/topic/DRB.20integration/near/485669746.

pls148:
Closes #3933

### This PR: 

Moves the logic for DRB computation tasks into a helper struct

### DIFF
--- a/crates/task-impls/src/quorum_vote/drb_computations.rs
+++ b/crates/task-impls/src/quorum_vote/drb_computations.rs
@@ -64,7 +64,7 @@ impl<TYPES: NodeType> DrbComputations<TYPES> {
         }
     }
 
-    /// Stores a seed for a particular epoch for later use by start_new_task()
+    /// Stores a seed for a particular epoch for later use by start_task_if_not_running, called from handle_quorum_proposal_validated_drb_calculation_start
     pub fn store_seed(&mut self, epoch: TYPES::Epoch, drb_seed_input: DrbSeedInput) {
         self.seeds.insert(epoch, drb_seed_input);
     }


crates/task-impls/src/quorum_vote/drb_computations.rs
```diff
@@ -0,0 +1,105 @@
...
+        Self {
+            results: BTreeMap::new(),
+            task: None,
+            seeds: BTreeMap::new(),
+        }
+    }
+
+    /// If a task is currently live AND has finished, join it and save the result.
+    /// Does not block if a live task has not finished.
+    pub async fn try_join_task(&mut self) {
```
shenkeyao, line 34: Is this function used anywhere? I don't think we'll need this, since whenever we overwrite `self.task` we won't use the result of the existing task anymore.

crates/task-impls/src/quorum_vote/drb_computations.rs
```diff
@@ -0,0 +1,105 @@
...
+                }
+            } else {
+                join_handle.abort();
+            }
+            self.task = None;
+        }
+    }
+
+    /// Starts a new task. If a previous task is currently live, will attempt to join it (if finished), or abort it.
+    pub async fn start_new_task(&mut self, epoch: TYPES::Epoch, drb_seed_input: DrbSeedInput) {
```
shenkeyao, line 72: We can get `drb_seed_input` from `self.seeds` so there's no need to have this as an input.

crates/task-impls/src/quorum_vote/drb_computations.rs
```diff
@@ -0,0 +1,105 @@
...
+        self.join_or_abort_task().await;
+        self.seeds.insert(epoch, drb_seed_input);
```
shenkeyao, line 74: We actually want to remove the seed from `self.seeds` after starting a new task using it.

crates/task-impls/src/quorum_vote/handlers.rs
```diff
@@ -205,11 +202,15 @@ pub(crate) async fn handle_quorum_proposal_validated<
...
+
+                    // This will join_or_abort_task before starting the new task
                     task_state
                         .drb_computations
-                        .insert(new_epoch_number, new_drb_task);
+                        .start_new_task(new_epoch_number, drb_seed_input)
+                        .await;
+                } else {
+                    // Only needed if we aren't starting a new task
+                    task_state.drb_computations.join_or_abort_task().await;
```
shenkeyao, line 213: The `else` case is where we aren't on the committee of the next epoch, so we don't have to handle it explicitly.

Is this the only place where `join_or_abort_task` is used? If so we could make it a private function.

crates/task-impls/src/quorum_vote/drb_computations.rs
```diff
@@ -0,0 +1,105 @@
...
+    results: BTreeMap<TYPES::Epoch, DrbResult>,
+
+    /// Currently live computation
+    task: Option<(TYPES::Epoch, JoinHandle<DrbResult>)>,
+
+    /// Stored inputs to computations
+    seeds: BTreeMap<TYPES::Epoch, DrbSeedInput>,
+}
+
+impl<TYPES: NodeType> DrbComputations<TYPES> {
```
shenkeyao, line 21: Looks like we also need a function to insert a seed, which will happen at the beginning of each epoch. Then we can use the seed in the next epoch to start the computation.

crates/task-impls/src/quorum_vote/mod.rs
```diff
@@ -274,7 +277,8 @@ pub struct QuorumVoteTaskState<TYPES: NodeType, I: NodeImplementation<TYPES>, V:
     pub membership: Arc<TYPES::Membership>,
 
     /// Table for the in-progress DRB computation tasks.
-    pub drb_computations: BTreeMap<TYPES::Epoch, JoinHandle<DrbResult>>,
+    //pub drb_computations: BTreeMap<TYPES::Epoch, JoinHandle<DrbResult>>,
```
shenkeyao, line 280: Nit: Can remove this commented-out line.

crates/task-impls/src/quorum_vote/handlers.rs
```diff
@@ -175,26 +173,23 @@ pub(crate) async fn handle_quorum_proposal_validated<
                 .block_number();
 
             // Skip if this is not the expected block.
-            if task_state.epoch_height != 0
-                && (decided_block_number + 3) % task_state.epoch_height == 0
-            {
+            if task_state.epoch_height != 0 && decided_block_number % task_state.epoch_height == 0 {
```
shenkeyao, line 176: This part is inside the `if let Some(decided_view_number) = new_decided_view_number` clause, but we actually want to start the computation no matter if we have decided on a new view, so it should be moved to the beginning of this function.

shenkeyao: Besides, we probably don't have to rely on `decided_block_number % task_state.epoch_height == 0`, since it's possible that we somehow miss the first block. Instead, we can check if `task` has an in-progress computation task for the next epoch, and if not, we start the computation.

crates/task-impls/src/quorum_vote/handlers.rs
```diff
@@ -175,26 +173,23 @@ pub(crate) async fn handle_quorum_proposal_validated<
...
+                    .drb_computations
+                    .garbage_collect(current_epoch_number);
 
                 // Skip if we are not in the committee of the next epoch.
                 if task_state
                     .membership
                     .has_stake(&task_state.public_key, current_epoch_number + 1)
                 {
-                    let new_epoch_number = current_epoch_number + 2;
+                    let new_epoch_number = current_epoch_number + 1;
```
shenkeyao, line 192: We store the seed two epochs in advance, so it's still `+ 2`.

crates/task-impls/src/quorum_vote/handlers.rs
```diff
@@ -175,26 +173,23 @@ pub(crate) async fn handle_quorum_proposal_validated<
                 .block_number();
 
             // Skip if this is not the expected block.
-            if task_state.epoch_height != 0
-                && (decided_block_number + 3) % task_state.epoch_height == 0
```
shenkeyao, line 179: We should keep this condition for the seed-storing process.

crates/task-impls/src/quorum_vote/handlers.rs
```diff
@@ -205,11 +200,23 @@ pub(crate) async fn handle_quorum_proposal_validated<
...
+                if task_state
+                    .membership
+                    .has_stake(&task_state.public_key, current_epoch_number)
+                {
+                    // This will join_or_abort_task before starting the new task
                     task_state
                         .drb_computations
-                        .insert(new_epoch_number, new_drb_task);
+                        .start_new_task(current_epoch_number)
+                        .await;
```
shenkeyao, line 219: Similar to my comment above, this part should be at the start of this function to make it unrelated to the `Decide` status.

crates/task-impls/src/quorum_vote/drb_computations.rs
```diff
@@ -0,0 +1,158 @@
...
+            self.task = None;
+        }
+    }
+
+    /// If a task is currently live AND has finished, join it and save the result.
+    /// If the epoch for the calculation was the same as the provided epoch, return true
+    /// If a task is currently live and NOT finished, abort it UNLESS the task epoch is the same as
+    /// cur_epoch, in which case keep letting it run and return true.
+    /// Return false if a task should be spawned for the given epoch.
+    async fn join_or_abort_task_for_epoch(&mut self, epoch: TYPES::Epoch) -> bool {
```
shenkeyao, line 61: Nit: Probably better to call it `join_or_abort_old_task_for_epoch` or `join_or_abort_old_task` to make it explicit that we'll only abort a task if it's not the same as the given epoch.

crates/task-impls/src/quorum_vote/drb_computations.rs
```diff
@@ -0,0 +1,158 @@
...
+        }
+    }
+
+    /// Stores a seed for a particular epoch for later use by start_new_task()
+    pub fn store_seed(&mut self, epoch: TYPES::Epoch, drb_seed_input: DrbSeedInput) {
+        self.seeds.insert(epoch, drb_seed_input);
+    }
+
+    /// Starts a new task. If a previous task is currently live, will attempt to join it (if finished), or abort it.
+    pub async fn start_new_task(&mut self, epoch: TYPES::Epoch) {
```
shenkeyao, line 94: Looks like we'll always use `start_task_if_not_running` rather than this function, so maybe we can remove it. Same for `join_or_abort_task`.

crates/task-impls/src/quorum_vote/handlers.rs
```diff
@@ -83,6 +81,28 @@ pub(crate) async fn handle_quorum_proposal_validated<
...
     };
 
+    if version >= V::Epochs::VERSION {
+        if let Some(last_leaf_view) = leaf_views.last() {
+            let decided_block_number = last_leaf_view.leaf.block_header().block_number();
+
+            let current_epoch_number = TYPES::Epoch::new(epoch_from_block_number(
+                decided_block_number,
+                task_state.epoch_height,
+            ));
```
shenkeyao, line 91: We can get `current_epoch_number` from `proposal.block_header.block_number()`, so there's no need to use the leaf or decided block info. 

This can also be moved to before the `let LeafChainTraversalOutcome`.

crates/task-impls/src/quorum_vote/drb_computations.rs
```diff
@@ -0,0 +1,126 @@
...
+                join_handle.abort();
+                self.task = None;
+                false
+            }
+        } else {
+            false
+        }
+    }
+
+    /// Stores a seed for a particular epoch for later use by start_new_task()
```
shenkeyao, line 67: Nit: Looks like `start_new_task` should be `handle_quorum_proposal_validated_drb_calculation_seed` now.

shenkeyao (commented):
Just a nit comment.

shenkeyao (approved):

PR #3943, opened 2024-12-04
Fix the return value of builder transaction status API

dailinsubjam:
### This PR: 
- fix the return value of txn stat api

### DIFF
--- a/crates/builder-api/src/v0_1/builder.rs
+++ b/crates/builder-api/src/v0_1/builder.rs
@@ -262,8 +262,7 @@ where
                     .body_auto::<<Types as NodeType>::Transaction, Ver>(Ver::instance())
                     .map_err(Error::TxnUnpack)?;
                 let hash = tx.commit();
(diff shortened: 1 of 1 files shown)


shenkeyao:
Having the same question as @ss-es about the change. :thinking:

dailinsubjam:
@ss-es @shenkeyao Original code will return  `Ok(transaction_commit)`, after the fix it will return `Result(TransactionStatus, BuildError)`. Or am I missing anything?

ss-es:
Oh sorry, I see - yes you're completely right! It used to return `Result<tx.commit(), Error::TxnStat>` but now it returns something like `Result<state.txn_status, Error::TxnStat>`

ignore my comment -- it looks good to me!

shenkeyao:
> @ss-es @shenkeyao Original code will return `Ok(transaction_commit)`, after the fix it will return `Result(TransactionStatus, BuildError)`. Or am I missing anything?

Totally makes sense. 😃

shenkeyao (approved):

PR #3948, opened 2024-12-06
[DRB] - Verify the DRB result from the computation one epoch in advance

### LINKED ISSUE #3858: [DRB] - Verify the DRB result from the computation one epoch in advance

shenkeyao:
Closes #3858.

### This PR: 
- Splits `DrbComputations` into `DrbSeedsAndResults` and `DrbComputation`.
  - Moves the former to the shared `Consensus` but keeps the latter at the same place.
  - Moves most contents in `quorum_vote/drb_computations.rs` to `types/src/drb.rs`, and some to the `quorum_vote/handlers.rs`.
- Updates the proposal task to include the DRB result for the next epoch at the last block.
- Updates the vote task.
  - Third from the last block:
    - Stores the computed result for nodes in epoch `e - 1`.
  - Last block:
    - Stores the received DRB result from the proposal for nodes in epoch `e`.
    - Verifies received result with computed results for nodes in epoch `e - 1`.
- Removes DRB seed and current result fields from `Leaf2` and `Proposal2`. 
- Fixes how the third from last block is defined with a helper function in `utils.rs`.

### This PR does not: 
* Consumes the DRB result for the leader election, which will be done in the next PR.

### Key places to review: 
* `quorum_proposal/handlers.rs`.
* `quorum_vote/handlers.rs`.
* `types/src/drb.rs`.
* (Many other files only contain format changes, probably due to merging both `lr/double-quorum` and `main`.)


crates/task-impls/src/quorum_vote/handlers.rs
```diff
@@ -34,20 +25,175 @@ use hotshot_types::{
...
+/// Store the DRB result from the computation task to the shared `results` table.
+///
+/// Returns the result if it exists.
+async fn store_and_get_computed_drb_result<
     TYPES: NodeType,
     I: NodeImplementation<TYPES>,
     V: Versions,
 >(
+    epoch_number: TYPES::Epoch,
+    task_state: &mut QuorumVoteTaskState<TYPES, I, V>,
```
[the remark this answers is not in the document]
shenkeyao: Looks like we also need the shared `consensus` field of `task_state` in this function to get the `results` table, so I prefer to keep the argument as is. But let me know if you think there's a better way though!

crates/types/src/utils.rs
```diff
@@ -279,3 +279,14 @@ pub fn is_last_block_in_epoch(block_number: u64, epoch_height: u64) -> bool {
         block_number % epoch_height == 0
     }
 }
+
+/// Returns true if the given block number is the third from the last in the epoch based on the
+/// given epoch height.
+#[must_use]
+pub fn is_third_from_last_block_in_epoch(block_number: u64, epoch_height: u64) -> bool {
```
ss-es, line 286: small thing: can we call this `is_epoch_root`? this would align with calling the block the "root" in the membership trait method

shenkeyao: Renamed!

PR #3954, opened 2024-12-10
[Libp2p] Generic DHT persistence

rob-maron:
### This PR: 

Refactors the DHT persistence from being file-backed to being made generic. We want this so the sequencer can use it with SQL storage as well


crates/libp2p-networking/src/network/behaviours/dht/store/persistent.rs
```diff
@@ -164,70 +261,87 @@ impl<R: RecordStore> FileBackedStore<R> {
...
-            .with_context(|| "Failed to serialize records")?;
+        // Spawn a task to save the DHT to the persistent storage
+        let persistent_storage = self.persistent_storage.clone();
+        let record_delta = Arc::clone(&self.record_delta);
+        tokio::spawn(async move {
+            debug!("Saving DHT to persistent storage");
+
+            // Save the DHT to the persistent storage
+            if let Err(e) = persistent_storage
+                .save(serializable_records)
```
ss-es, line 272: can/should this be made to timeout? to allow cleanup if `save` blocks/gets stuck

ss-es (approved):
looks good to me, just one comment

crates/libp2p-networking/src/network/behaviours/dht/store/persistent.rs
```diff
@@ -1,27 +1,116 @@
...
+/// A trait that we use to save and load the DHT to a file on disk
+/// or other storage medium
+#[async_trait]
+pub trait DhtPersistentStorage: Send + Sync + 'static + Clone {
+    /// Save the DHT (as a list of serializable records) to the persistent storage
+    ///
+    /// # Errors
+    /// - If we fail to save the DHT to the persistent storage provider
+    async fn save(&self, _records: Vec<SerializableRecord>) -> anyhow::Result<()> {
+        Ok(())
```
jbearer, line 29: IMO these defaults are more likely to cause us trouble by forgetting to implement something, where they really only save time/make sense for the one `DhtNoPersistence` implementation)

jbearer (approved):

crates/libp2p-networking/src/network/behaviours/dht/store/persistent.rs
rob-maron: Agree, removing defaults

crates/libp2p-networking/src/network/behaviours/dht/store/persistent.rs
rob-maron: Good thinking, doing this now

PR #3971, opened 2024-12-16
Double Voting Prevention

qvkare:
This implements a robust double voting prevention mechanism for the HotShot. The implementation addresses a critical security vulnerability where nodes could potentially cast multiple votes for the same view under certain conditions.

## Problem

In the current implementation, there's a significant security vulnerability in the voting mechanism:

```rust
// Current problematic implementation
if view > self.last_actions.da_vote {
    self.last_actions.da_vote = view;
}
// TODO Add logic to prevent double voting.  For now the simple check if
// the last voted view is less than the view we are trying to vote doesn't work
// because the leader of view n + 1 may propose to the DA (and we would vote)
// before the leader of view n.
```

This implementation fails to prevent double voting because:
1. The leader of view n+1 might propose to DA before the leader of view n
2. Simple view number comparison is insufficient for preventing double votes
3. No persistent tracking of votes per view exists

### 1. New VoteTracker Component

Introduced a new `VoteTracker` struct that provides atomic vote tracking:

```rust
pub struct VoteTracker<TYPES: NodeType> {
    view_votes: HashMap<TYPES::View, HashSet<Arc<TYPES::SignatureKey>>>,
}
```

Key features:
- Thread-safe vote tracking using Arc
- O(1) vote verification
- Automatic cleanup of old view data
- Memory-efficient implementation

### 2. Consensus Integration

Modified the Consensus struct to incorporate vote tracking:

```rust
pub struct Consensus<TYPES: NodeType> {
    // ... existing fields ...
    vote_tracker: VoteTracker<TYPES>,
}
```

Updated vote handling logic:

```rust
pub fn update_action(&mut self, action: HotShotAction, view: TYPES::View) -> bool {
    match action {
        HotShotAction::DaVote => {
            let voter_key = Arc::new(self.public_key().clone());
            if !self.vote_tracker.record_vote(view, voter_key) {
                tracing::warn!("Prevented double voting attempt for view {}", view);
                return false;
            }
            // ... rest of the logic
        }
        // ... other actions
    }
}
```

## Security Implications

1. **Vote Uniqueness**
   - Each node can only vote once per view
   - Secured using node's public key

2. **Byzantine Fault Tolerance**
   - Maintains BFT properties of the protocol
   - Prevents vote splitting attacks
   - Ensures vote finality within views

3. **View Synchronization**
   - Handles out-of-order view proposals correctly
   - Maintains vote consistency across view changes
   - Prevents view-related voting attacks

```mermaid
graph TD
    A[Vote Request] --> B{Check Vote Tracker}
    B -->|Already Voted| C[Reject Vote]
    B -->|First Vote| D[Record Vote]
    D --> E[Update View State]
    E --> F[Cleanup Old Views]
    F --> G[Return Success]
```

### 1. Vote Recording Process

```rust
pub fn record_vote(
    &mut self,
    view: TYPES::View,
    voter_key: Arc<TYPES::SignatureKey>,
) -> bool {
    let votes = self.view_votes.entry(view).or_default();
    if votes.contains(&voter_key) {
        return false;
    }
    votes.insert(voter_key);
    true
}
```

Key aspects:
- Uses efficient HashSet for O(1) lookups
- Maintains vote history per view

### 2. Memory Management

```rust
pub fn cleanup_old_views(&mut self, current_view: TYPES::View) {
    self.view_votes.retain(|view, _| *view >= current_view);
}
```

Features:
- Automatic garbage collection of old view data
- Prevents memory leaks
- Configurable retention policy

### Time Complexity

| Operation | Complexity | Notes |
|-----------|------------|--------|
| Vote Check | O(1) | HashSet lookup |
| Vote Record | O(1) | HashSet insert |
| Cleanup | O(n) | n = number of views |
| Memory Usage | O(v * n) | v = views, n = nodes |

### Unit Tests

```rust
#[test]
fn test_vote_tracking() {
    let mut tracker = VoteTracker::<SimpleVote>::new();
    let view = SimpleVote::View::default();
    let key1 = Arc::new(SimpleVote::SignatureKey::default());
    
    assert!(tracker.record_vote(view, key1.clone()));
    assert!(!tracker.record_vote(view, key1.clone())); // Double vote prevented
}
```


crates/types/src/consensus.rs
```diff
@@ -549,27 +553,39 @@ impl<TYPES: NodeType> Consensus<TYPES> {
...
                     self.last_actions.da_vote = view;
                 }
-                // TODO Add logic to prevent double voting.  For now the simple check if
-                // the last voted view is less than the view we are trying to vote doesn't work
-                // because the leader of view n + 1 may propose to the DA (and we would vote)
-                // before the leader of view n.
-                return true;
+                
+                // Clean up old vote records periodically
+                self.vote_tracker.cleanup_old_views(view);
```
bfish713, line 571: Doesn't this open us up tot he opposite problem would have? Vote in view `n`, then view `n+1` then we can vote for view `n` again?  The old code of course just allows any duplicate vote so there is an improvement with this code.

qvkare: @bfish713 I have implemented a robust system for tracking view sequences. Added a queue mechanism that handles out-of-order votes. This prevents double voting while allowing the network to be resilient. Sorry, I missed the importance of the TODO comment. Comprehensive test coverage of both epoch transitions and network partitioning scenarios is included in this solution.

bfish713 (changes requested):
I think we should discuss the solution to this before adding more details.  I don't think we need this complex of vote tracking inserted at this point.  DA voting is not the same as regular voting so double voting for a view is not really security issue, though not ideal.  Also there is no requirement that votes be sequential.  We can vote for view `n+1` before view `n` for DA, we cannot when voting on a proposal.

The properties we would ideally enforce for voting on DA are:
- Don't vote twice in a view
- Don't vote for old views.  And old view an be considered a view less or equal to the high QC view
- Don't vote for views too far in the future, this can just be a heuristic value to prevent a leader from calculating all it's future views and spamming us with proposals. 

Ideally most of these properties are enforced before we ever get `update_action`

Also can you elaborate on why there are changes to the epoch functions in consensus.rs

PR #3984, opened 2024-12-19
3966 add_epoch_root and sync_l1 on Membership

### LINKED ISSUE #3966: Membership trait changes to support PoS
Reference [meeting notes](https://www.notion.so/espressosys/PoS-Membership-Trait-Discussion-15b2431b68e9803f8c96cbae9fa3a1ad?pvs=4)

- [x] HotShot clones the Membership object when receiving it, wrapping it in `Arc<RwLock<>>`
- [x] Remove requirement for Membership to implement Clone
- [x] Add sync_l1 and add_epoch_root functions, which return an Option<FnOnce(&mut Membership)> or similar. Functions should be called with a read lock held, then upgrade to write lock if a callback is returned. Default implementations should return None. A static helper function can be used to indicate if each function should be called at all.
- Convert the existing Membership methods to async (NOT NEEDED ANY MORE)
- [ ] Audit existing code to avoid Membership calls while holding other locks

pls148:
Closes #3966

### This PR: 

Adds add_epoch_root and sync_l1 to Membership trait, calling them from within quorum_vote.

### DIFF
--- a/crates/task-impls/src/helpers.rs
+++ b/crates/task-impls/src/helpers.rs
@@ -28,7 +28,9 @@ use hotshot_types::{
         signature_key::SignatureKey,
         BlockPayload, ValidatedState,
     },
-    utils::{epoch_from_block_number, is_last_block_in_epoch, Terminator, View, ViewInner},
+    utils::{
+        epoch_from_block_number, is_epoch_root, is_last_block_in_epoch, Terminator, View, ViewInner,
+    },
     vote::{Certificate, HasViewNumber},
 };
 use tokio::time::timeout;
@@ -163,6 +165,56 @@ pub(crate) async fn fetch_proposal<TYPES: NodeType, V: Versions>(
     Ok((leaf, view))
 }
 
+/// Handles calling add_epoch_root and sync_l1 on Membership if necessary.
+async fn decide_from_proposal_add_epoch_root<TYPES: NodeType>(
+    proposal: &QuorumProposal2<TYPES>,
+    leaf_views: &[LeafInfo<TYPES>],
+    epoch_height: u64,
+    membership: &Arc<RwLock<TYPES::Membership>>,
+) {
+    if leaf_views.is_empty() {
+        return;
+    }
+
+    let decided_block_number = leaf_views
+        .last()
+        .unwrap()
+        .leaf
+        .block_header()
+        .block_number();
+
+    // Skip if this is not the expected block.
+    if epoch_height != 0 && is_epoch_root(decided_block_number, epoch_height) {
+        let next_epoch_number =
+            TYPES::Epoch::new(epoch_from_block_number(decided_block_number, epoch_height) + 1);
+
+        let write_callback = {
+            let membership_reader = membership.read().await;
+            membership_reader
+                .add_epoch_root(next_epoch_number, proposal.block_header.clone())
+                .await
+        };
+
+        if let Some(write_callback) = write_callback {
+            let mut membership_writer = membership.write().await;
+            write_callback(&mut *membership_writer);
+        } else {
+            // If we didn't get a write callback out of add_epoch_root, then don't bother locking and calling sync_l1
+            return;
+        }
+
+        let write_callback = {
+            let membership_reader = membership.read().await;
+            membership_reader.sync_l1().await
+        };
+
+        if let Some(write_callback) = write_callback {
+            let mut membership_writer = membership.write().await;
+            write_callback(&mut *membership_writer);
+        }
+    }
+}
+
 /// Helper type to give names and to the output values of the leaf chain traversal operation.
 #[derive(Debug)]
 pub struct LeafChainTraversalOutcome<TYPES: NodeType> {
@@ -202,7 +254,7 @@ impl<TYPES: NodeType + Default> Default for LeafChainTraversalOutcome<TYPES> {
     }
 }
 
-/// calculate the new decided leaf chain based on the rules of hostuff 2
+/// calculate the new decided leaf chain based on the rules of HotStuff 2
 ///
 /// # Panics
 /// Can't actually panic
@@ -211,6 +263,8 @@ pub async fn decide_from_proposal_2<TYPES: NodeType>(
     consensus: OuterConsensus<TYPES>,
     existing_upgrade_cert: Arc<RwLock<Option<UpgradeCertificate<TYPES>>>>,
     public_key: &TYPES::SignatureKey,
+    with_epochs: bool,
+    membership: &Arc<RwLock<TYPES::Membership>>,
 ) -> LeafChainTraversalOutcome<TYPES> {
     let mut res = LeafChainTraversalOutcome::default();
     let consensus_reader = consensus.read().await;
@@ -282,6 +336,14 @@ pub async fn decide_from_proposal_2<TYPES: NodeType>(
         res.included_txns = Some(txns);
     }
 
+    if with_epochs && res.new_decided_view_number.is_some() {
+        let epoch_height = consensus_reader.epoch_height;
+        drop(consensus_reader);
+
+        decide_from_proposal_add_epoch_root(proposal, &res.leaf_views, epoch_height, membership)
+            .await;
+    }
+
     res
 }
 
@@ -317,6 +379,8 @@ pub async fn decide_from_proposal<TYPES: NodeType>(
     consensus: OuterConsensus<TYPES>,
     existing_upgrade_cert: Arc<RwLock<Option<UpgradeCertificate<TYPES>>>>,
     public_key: &TYPES::SignatureKey,
+    with_epochs: bool,
+    membership: &Arc<RwLock<TYPES::Membership>>,
 ) -> LeafChainTraversalOutcome<TYPES> {
     let consensus_reader = consensus.read().await;
     let existing_upgrade_cert_reader = existing_upgrade_cert.read().await;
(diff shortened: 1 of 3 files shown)


crates/types/src/traits/election.rs
```diff
@@ -125,4 +125,32 @@ pub trait Membership<TYPES: NodeType>: Debug + Send + Sync {
 
     /// Returns the threshold required to upgrade the network protocol
     fn upgrade_threshold(&self, epoch: TYPES::Epoch) -> NonZeroU64;
+
+    #[allow(clippy::type_complexity)]
+    /// Handles notifications that a new epoch root has been created
+    /// Is called under a read lock to the Membership. Return a callback
+    /// with Some to have that callback invoked under a write lock.
+    fn add_epoch_root(
```
sveitser, line 133: Didn't we say we need this to be async?

Currently I imagine we would want to do external requests here or even if we de-couple it we may want to take out some async locks or read from channels.

imabdulbasit: I think this can be non async but sync_l1 should be async?

crates/task-impls/src/quorum_vote/handlers.rs
```diff
@@ -135,6 +135,52 @@ async fn handle_quorum_proposal_validated_drb_calculation_seed<
...
+) {
+    let decided_block_number = leaf_views
+        .last()
+        .unwrap()
+        .leaf
+        .block_header()
+        .block_number();
+
+    // Skip if this is not the expected block.
+    if task_state.epoch_height != 0 && (decided_block_number + 3) % task_state.epoch_height == 0 {
```
sveitser, line 156: Currently what would happen if a node misses the block when the update is triggered? Would it never have `add_epoch_root` called for that epoch? 

I'm wondering if we the Membership trait should have a method like `is_epoch_initialized(&self, epoch) -> bool` so that if the node is not online for the beginning of the epoch we can still initialize the membership for the epoch.

Or can we handle this well enough with catchup only in the confirmation layer? What do you think @imabdulbasit ?

crates/types/src/traits/election.rs
sveitser: I don't really want to use `sync_l1` unless it can't be avoided. It shouldn't leak into consensus that we are using the L1 if we can avoid it.

crates/task-impls/src/quorum_vote/handlers.rs
imabdulbasit: I think doing a catchup is better, but I am wondering how we can do that now with Hotshot managing the locks

crates/types/src/traits/election.rs
imabdulbasit: If we have only one function, then yes, it should be async due to calls to L1.

crates/task-impls/src/quorum_vote/handlers.rs
sveitser: For our existing catchup it's hotshot that triggers the catchup (for example by calling `validate_and_apply_header`) and the confirmation layer that performs the catchup. Maybe it should be the same here: hotshot needs to call something to trigger catchup for the membership and the sequencer performs the catchup. 

If we were to do it analoguous for membership we would have to make all the reading functions async and &mut self which seems terrible. I think we can think of something better.

sveitser: For example hotshot could call `add_epoch_root` (probably with another name) for every view then the sequencer could decide if it needs to do catchup for this epoch (and in most cases do (almost) nothing and return `None` as write_callback). This way we ensure the catchup is actually triggered.

I think our overall design is a bit sub-optimal because it's an invariant that the membership is static for an epoch but that invariant isn't really reflected in the code. I think eventually we should enforce this in hotshot and the confirmation layer should somehow provide a constructor for a "EpochMembership" type that takes an `epoch` as input that is called by hotshot where needed. @ss-es mentioned this would be difficult to do in hotshot but I'm not sure pushing the complexity elsewhere is actually better.

imabdulbasit: yeah but then hotshot has to decide when to call it? I think it would calling it for every view would be too much maybe. It makes sense to do catch up when the membership get functions() do not return any data?

imabdulbasit: maybe we do catchup at the node startup and pass the Committee so hotshot does not need to do anything?

sveitser: I think catchup at startup might not work if a node misses some views for some other reason, for example due to a temporary network outage.

sveitser: > yeah but then hotshot has to decide when to call it? I think it would calling it for every view would be too much maybe. 

If it doesn't do anything except for once per epoch then what's the problem with calling it every view?

crates/task-impls/src/quorum_vote/handlers.rs
```diff
@@ -135,6 +135,52 @@ async fn handle_quorum_proposal_validated_drb_calculation_seed<
...
+        let next_epoch_number = TYPES::Epoch::new(
+            epoch_from_block_number(decided_block_number, task_state.epoch_height) + 1,
+        );
+
+        let membership_reader = task_state.membership.read().await;
```
tbro, line 161: Possibly personal preference, but I think it is more idiomatic to prefer introducing a scoped block over explicit `drop`s. Something like:

```rust
let write_callback = {
    let membership_reader = task_state.membership.read().await;
    membership_reader.add_epoch_root(next_epoch_number, proposal.block_header.clone())
}

```

crates/task-impls/src/quorum_vote/handlers.rs
tbro: > I'm wondering if  the Membership trait should have a method like `is_epoch_initialized(&self, epoch) -> bool` so that if the node is not online for the beginning of the epoch we can still initialize the membership for the epoch. But then we also have to dig out the correct header from history which I think might be cumbersome.

Why do we need this? When node comes online it should go through same routine as all nodes, which includes calling `add_epoch_root` for the block being proposed/validated.

sveitser: I might be reading it wrongly but I think the current implementation in this PR only calls the function if `decided_block_number + 3) % task_state.epoch_height == 0` so if it joins at a later point and the decided block has moved on it won't call it. Maybe I'm misunderstanding.

crates/types/src/traits/election.rs
```diff
@@ -125,4 +127,32 @@ pub trait Membership<TYPES: NodeType>: Debug + Send + Sync {
...
+        None
+    }
+
+    #[must_use]
+    /// Used to determine if sync_l1 should be called after add_epoch_root.
+    /// Returning false avoids a redundant read lock for cases where sync_l1
+    /// is a no-op.
+    fn uses_sync_l1() -> bool {
+        false
+    }
```
ss-es, line 149: is this actually needed with the current signature for `sync_l1`?

ss-es: actually, I feel like we wouldn't want to do this even if we could

I think we'd (eventually) always try to take this lock in deployment, so this would only help us in tests. but I don't think we want our tests to have different locking behavior

I might be misunderstanding though

crates/task-impls/src/helpers.rs
```diff
@@ -163,6 +163,55 @@ pub(crate) async fn fetch_proposal<TYPES: NodeType, V: Versions>(
...
+        let write_callback = {
+            let membership_reader = membership.read().await;
+            membership_reader
+                .add_epoch_root(next_epoch_number, proposal.block_header.clone())
+                .await
+        };
+
+        if let Some(write_callback) = write_callback {
+            let mut membership_writer = membership.write().await;
+            write_callback(&mut *membership_writer);
```
ss-es, line 198: can't we just call `membership_reader.sync_l1().await`  here after the `write_callback`?

crates/task-impls/src/helpers.rs
```diff
@@ -163,6 +163,55 @@ pub(crate) async fn fetch_proposal<TYPES: NodeType, V: Versions>(
...
+        }
+
+        if TYPES::Membership::uses_sync_l1() {
+            let write_callback = {
+                let membership_reader = membership.read().await;
+                membership_reader.sync_l1().await
+            };
```
ss-es, line 205: rather than checking `uses_sync_l1()` we can just check whether `add_epoch_root` actually did something, and if so immediately apply the callback and the sync

ss-es (commented):
left some comments, though I'm not necessarily sure in any of them

crates/task-impls/src/helpers.rs
```diff
@@ -163,6 +163,56 @@ pub(crate) async fn fetch_proposal<TYPES: NodeType, V: Versions>(
...
+        let write_callback = {
+            let membership_reader = membership.read().await;
+            membership_reader.sync_l1().await
+        };
+
+        if let Some(write_callback) = write_callback {
+            let mut membership_writer = membership.write().await;
+            write_callback(&mut *membership_writer);
+        }
+    }
```
ss-es, line 213: I think we might want this in a `tokio::spawn`, so we don't block on the `sync_l1` call but I'm not 100% sure what the right approach is here (e.g. I don't know if doing that adds deadlock risks). probably also fine to leave as-is and deal with it if problems come up

ss-es (approved):
overall looks good to me, though I think it's worth merging main in and using `is_epoch_root`

PR #4005, opened 2025-01-07
Lr/epochs more tests

lukaszrzasik:
Closes #<ISSUE_NUMBER>

### This PR: 

* adds more epoch tests
* fixes issues found during testing
* * implements `vids_cloned` for `TestStorage` used to retrieve stored VIDs after node restart
* * adds a check to `collect_garbage` in `Consensus` to make sure that the new anchor view is greater than the old one
* * modifies `parent_leaf_info` in `Consensus` so that it returns `LeafInfo` with VID set to None if VID is not available instead returning None altogether.


ss-es (approved):
looks good to me

PR #4018, opened 2025-01-09
feat: adding metadata input for `calculate_vid_disperse`

### LINKED ISSUE #3973: [VID Upgrade] expose Namespace table in HotShot so VID can take it as input

mrain:
Closes #3973 

### This PR: 

Use `BlockPayload` instead of encoded bytes as input for `calculate_vid_disperse`.

The metadata in `BlockPayload` seems to be for the namespace table: https://github.com/EspressoSystems/espresso-sequencer/blob/fa44122b66b49ca070eab64e59965290f9c0eecf/sequencer/src/block/full_payload/payload.rs#L141
Now VID can take it as input for efficient nsproof generation/verification.

### DIFF
--- a/crates/hotshot/src/lib.rs
+++ b/crates/hotshot/src/lib.rs
@@ -61,7 +61,6 @@ use hotshot_types::{
         signature_key::SignatureKey,
         states::ValidatedState,
         storage::Storage,
-        EncodeBytes,
     },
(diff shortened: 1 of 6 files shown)


ss-es (approved):
looks good to me!

big fan of the [u8] -> TYPES::BlockPayload change

PR #4019, opened 2025-01-09
[DRB] - Store DRB seed and epoch root with consistent block info

### LINKED ISSUE #3998: [DRB] - Store DRB seed with consistent block and QC
We store the DRB seed in `store_drb_seed_and_result` when the third from the last block in an epoch is decided, but the seed corresponds to the current proposal, which is likely 3 views newer than the decided block.

We should determine which QC to use as the seed and its relationship between the decided block.

Links:
* Where we store the seed: https://github.com/EspressoSystems/HotShot/pull/3948/files#diff-8211699150b493f8c0c96fed42c92693ff189962da799c6cfefc8e0c66e8ae8bR462.
* Zulip discussion: https://espresso.zulipchat.com/#narrow/channel/462904-eng-proof-of-stake/topic/DRB.20integration/near/488720346.

### LINKED ISSUE #4008: Store epoch root with consistent block and block header
Similar to https://github.com/EspressoSystems/HotShot/issues/3998, in `decide_from_proposal_add_epoch_root`, the decided block isn't the same as the proposed block. We should use the former to make the logic consistent.

Besides, rename the funciton to `decide_epoch_root`.

shenkeyao:
Closes #3998. Closes #4008.

### This PR: 
* Moves `store_received_drb_result` inside `store_drb_seed_and_result` to handle storing the seed, the received result, and the computed result together.
* Fixes the logic in `store_drb_seed_and_result` to use the decided block rather than the newly proposed block for storing the DRB seed.
* Removes the condition to block voting if failed to store the received result.
* Rename `decide_from_proposal_add_epoch_root` to `decide_epoch_root`.
* Fixes the logic in `decide_epoch_root` for adding epoch root.

### DIFF
--- a/crates/task-impls/src/helpers.rs
+++ b/crates/task-impls/src/helpers.rs
@@ -166,22 +166,12 @@ pub(crate) async fn fetch_proposal<TYPES: NodeType, V: Versions>(
 }
 
 /// Handles calling add_epoch_root and sync_l1 on Membership if necessary.
-async fn decide_from_proposal_add_epoch_root<TYPES: NodeType>(
-    proposal: &QuorumProposal2<TYPES>,
-    leaf_views: &[LeafInfo<TYPES>],
+async fn decide_epoch_root<TYPES: NodeType>(
+    decided_leaf: &Leaf2<TYPES>,
     epoch_height: u64,
     membership: &Arc<RwLock<TYPES::Membership>>,
 ) {
-    if leaf_views.is_empty() {
-        return;
-    }
-
-    let decided_block_number = leaf_views
-        .last()
-        .unwrap()
-        .leaf
-        .block_header()
-        .block_number();
+    let decided_block_number = decided_leaf.block_header().block_number();
 
     // Skip if this is not the expected block.
     if epoch_height != 0 && is_epoch_root(decided_block_number, epoch_height) {
@@ -191,7 +181,7 @@ async fn decide_from_proposal_add_epoch_root<TYPES: NodeType>(
         let write_callback = {
             let membership_reader = membership.read().await;
             membership_reader
-                .add_epoch_root(next_epoch_number, proposal.block_header.clone())
+                .add_epoch_root(next_epoch_number, decided_leaf.block_header().clone())
                 .await
         };
 
@@ -257,7 +247,8 @@ impl<TYPES: NodeType + Default> Default for LeafChainTraversalOutcome<TYPES> {
 /// calculate the new decided leaf chain based on the rules of HotStuff 2
 ///
 /// # Panics
-/// Can't actually panic
+/// If the leaf chain contains no decided leaf while reaching a decided view, which should be
+/// impossible.
 pub async fn decide_from_proposal_2<TYPES: NodeType>(
     proposal: &QuorumProposal2<TYPES>,
     consensus: OuterConsensus<TYPES>,
@@ -340,8 +331,14 @@ pub async fn decide_from_proposal_2<TYPES: NodeType>(
         let epoch_height = consensus_reader.epoch_height;
         drop(consensus_reader);
 
-        decide_from_proposal_add_epoch_root(proposal, &res.leaf_views, epoch_height, membership)
-            .await;
+        // `leaf_views.last()` is never none if we've reached a new decide, so this is safe to
+        // unwrap.
+        decide_epoch_root(
+            &res.leaf_views.last().unwrap().leaf,
+            epoch_height,
+            membership,
+        )
+        .await;
     }
 
     res
@@ -374,6 +371,10 @@ pub async fn decide_from_proposal_2<TYPES: NodeType>(
 ///
 /// Upon receipt then of a proposal for view 9, assuming it is valid, this entire process will repeat, and
 /// the anchor view will be set to view 6, with the locked view as view 7.
+///
+/// # Panics
+/// If the leaf chain contains no decided leaf while reaching a decided view, which should be
+/// impossible.
 pub async fn decide_from_proposal<TYPES: NodeType>(
     proposal: &QuorumProposal2<TYPES>,
     consensus: OuterConsensus<TYPES>,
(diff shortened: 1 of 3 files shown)


crates/task-impls/src/helpers.rs
```diff
@@ -489,8 +490,14 @@ pub async fn decide_from_proposal<TYPES: NodeType>(
         let epoch_height = consensus_reader.epoch_height;
         drop(consensus_reader);
 
-        decide_from_proposal_add_epoch_root(proposal, &res.leaf_views, epoch_height, membership)
-            .await;
+        // `leaf_views.last()` is never none if we've reached a new decide, so this is safe to
+        // unwrap.
+        decide_epoch_root(
+            &res.leaf_views.last().unwrap().leaf,
```
ss-es, line 496: I'm still uncomfortable with the unwrapping; could we not do `if let Ok(decided_leaf) = res.leaf_views.last() { decide_epoch_root(decided_leaf, ...) }`?

crates/task-impls/src/quorum_vote/handlers.rs
```diff
@@ -266,72 +212,97 @@ async fn start_drb_task<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versio
...
-        .leaf
-        .block_header()
-        .block_number();
+    let decided_block_number = decided_leaf.block_header().block_number();
 
-    // Skip if this is not the expected block.
-    if task_state.epoch_height != 0 && is_epoch_root(decided_block_number, task_state.epoch_height)
-    {
-        // Cancel old DRB computation tasks.
+    if task_state.epoch_height != 0 {
```
ss-es, line 238: can this be an e.g. `ensure!(task_state.epoch_height !=0, debug!("Epoch height is 0, skipping DRB storage")`?

crates/task-impls/src/quorum_vote/handlers.rs
```diff
@@ -266,72 +212,97 @@ async fn start_drb_task<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versio
...
             task_state.epoch_height,
         ));
 
-        let mut consensus_writer = task_state.consensus.write().await;
-        consensus_writer
-            .drb_seeds_and_results
-            .garbage_collect(current_epoch_number);
-        drop(consensus_writer);
+        // Skip storing the seed and computed result if this is not the epoch root.
+        if is_epoch_root(decided_block_number, task_state.epoch_height) {
```
ss-es, line 245: can this also be an `ensure!()` to avoid the nested `if`s?

crates/task-impls/src/quorum_vote/handlers.rs
```diff
@@ -266,72 +212,97 @@ async fn start_drb_task<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versio
...
-                .consensus
-                .write()
+            // Skip storing the seed if we are not in the committee of the next epoch.
+            if task_state
+                .membership
+                .read()
                 .await
-                .drb_seeds_and_results
-                .store_seed(new_epoch_number, drb_seed_input);
+                .has_stake(&task_state.public_key, current_epoch_number + 1)
```
ss-es, line 262: I'm not sure we really need this check. I feel like if we decide on a leaf that is an epoch root we can just always store the seed, whether or not we're participating in the next epoch. there's not really any harm in doing it

crates/task-impls/src/helpers.rs
```diff
@@ -338,8 +329,14 @@ pub async fn decide_from_proposal_2<TYPES: NodeType>(
...
```
ss-es, line 335: same comment on unwrapping

I think this one is also the reason behind the panics in the test

crates/task-impls/src/helpers.rs
shenkeyao: Updated as suggested! Replaced the panic with `warn`.

crates/task-impls/src/quorum_vote/handlers.rs
shenkeyao: Similar to the reply above, I don't think we should return an error if this isn't an epoch root, because it may be the last block instead which will be handled later in this function. Does this make sense?

crates/task-impls/src/quorum_vote/handlers.rs
shenkeyao: Agree, updated!

ss-es (approved):
I think this looks good to me, just a few minor comments

crates/task-impls/src/helpers.rs
```diff
@@ -329,14 +329,13 @@ pub async fn decide_from_proposal_2<TYPES: NodeType>(
...
-            epoch_height,
-            membership,
-        )
-        .await;
+        if let Some(decided_leaf_info) = res.leaf_views.last() {
+            decide_epoch_root(&decided_leaf_info.leaf, epoch_height, membership).await;
+        } else {
+            tracing::warn!(
+                "No decided leaf while a view has been decided, which should be impossible."
+            );
```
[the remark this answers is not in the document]
shenkeyao: Changed this to an `info` log and removed "which should be impossible", to still log something but not indicate an error.

crates/task-impls/src/helpers.rs
```diff
@@ -490,14 +489,13 @@ pub async fn decide_from_proposal<TYPES: NodeType>(
...
-            &res.leaf_views.last().unwrap().leaf,
-            epoch_height,
-            membership,
-        )
-        .await;
+        if let Some(decided_leaf_info) = res.leaf_views.last() {
+            decide_epoch_root(&decided_leaf_info.leaf, epoch_height, membership).await;
+        } else {
+            tracing::warn!(
+                "No decided leaf while a view has been decided, which should be impossible."
```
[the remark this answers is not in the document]
shenkeyao: Same change as above.

crates/task-impls/src/quorum_vote/handlers.rs
```diff
@@ -233,76 +233,68 @@ async fn store_drb_seed_and_result<TYPES: NodeType, I: NodeImplementation<TYPES>
...
+    else if is_last_block_in_epoch(decided_block_number, task_state.epoch_height) {
+        // Skip if we are not in the committee of the next epoch.
+        if !task_state
+            .membership
+            .read()
+            .await
+            .has_stake(&task_state.public_key, current_epoch_number + 1)
+        {
+            return Ok(());
         }
```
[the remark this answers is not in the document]
shenkeyao: Removed!

ss-es (approved):

PR #4034, opened 2025-01-16
Remove `external` message support

rob-maron:
### This PR: 

Removes support for sending `external` messages. `external` messages were those that layers above HotShot (like the sequencer) could send and receive, since they did not have direct access to the network

This was originally introduced for the node validator service, but that has since gone a different route (and this is not used at all). Additionally, for sequencer-level catchup we are now implementing a better way for the sequencer to be able to communicate with HotShot networks.

### This PR does not: 

Make any breaking changes


crates/types/src/message.rs
```diff
@@ -123,8 +123,6 @@ pub enum MessageKind<TYPES: NodeType> {
     Consensus(SequencingMessage<TYPES>),
     /// Messages relating to sharing data between nodes
     Data(DataMessage<TYPES>),
-    /// A (still serialized) message to be passed through to external listeners
-    External(Vec<u8>),
```
ss-es, line 127: I think this shouldn't change the serialization of `MessageKind::Consensus` and `MessageKind::Data`, but it might be worth trying to run the `message_compat_tests` from the sequencer to be sure

ss-es (approved):
looks fine to me

crates/types/src/message.rs
rob-maron: Will do before merging!

PR #4067, opened 2025-01-21
Add `root_block_in_epoch`

ss-es:
### This PR: 
Adds `root_block_in_epoch`, which returns the block number of the epoch root in the given epoch.


bfish713:
Where will we actually use this? for catchup we'll be looking up the root for Epoch `e`

bfish713 (approved):

PR #4070, opened 2025-01-22
Epoch-enabled catchup tests

lukaszrzasik:
Closes #<ISSUE_NUMBER>

### This PR: 

* Adds more epoch-enabled catchup tests
* Stores last actioned epoch in `TestStorage`
* Adds new trait `EpochMessage` and implement it for `MessageKind`. This is useful to extract an epoch from a message to record it.
* Implements an `epoch_number` method for `SequencingMessage` to easily extract an associated epoch using `HasEpoch` trait of the message's underlying data.
* Implements `HasEpoch` trait for the "old" types (types prior to epoch upgrade). The implementation is trivial, returns `None`.
* Unfortunately `HasEpoch` trait does not make sense for the `QuorumProposal2` and `QuorumProposalWrapper` types. These types do not have an epoch field. Instead, the epoch needs to be calculated. Because of that the `epoch` method of the `HasEpoch` trait panics for these types.
* Minor refactor of `OverallSafetyTask` which includes:
* * the `Decide`, `Error` and `ReplicaViewTimeout` sent by nodes not belonging to the relevant epoch are ignored
* * all leaves sent in the leaf chain attached to the `Decide` event are taken into account, instead of only the leaf with the highest view. This fixes successful views count.
* Increases `REQUEST_TIMEOUT` to 2000 ms.

### This PR does not: 

~~Add the `test_staggered_restart_with_epochs` test. The test is currently commented out. The problem is the test always fails in CI but it never fails for me locally. It makes debugging it much more time consuming. I will continue doing that but I'd like to have the PR merged in the meantime.~~
After increasing `REQUEST_TIMEOUT` the test passes.

### Key places to review: 

~~`HasEpoch` implementions for  `QuorumProposal2` and `QuorumProposalWrapper` in the `crates/types/src/data.rs` file. Are the panics fine there?~~
The epoch field has been reintroduced to the QuorumProposal2. HasEpoch trait has been implemented for QuorumProposal2 and QuorumProposalWrapper. There are no panics any longer.
`crates/testing/src/overall_safety_task.rs` file

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -165,11 +165,12 @@ dependencies = [
 
 [[package]]
 name = "anstyle-wincon"
-version = "3.0.6"
+version = "3.0.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2109dbce0e72be3ec00bed26e6a7479ca384ad226efdd66db8fa2e3a38c83125"
+checksum = "ca3534e77181a9cc07539ad51f2141fe32f6c3ffd4df76db8ad92346b003ae4e"
 dependencies = [
  "anstyle",
+ "once_cell",
  "windows-sys 0.59.0",
 ]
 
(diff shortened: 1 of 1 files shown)


ss-es (approved):
looks good to me!

PR #4081, opened 2025-01-24
Replace `overall_safety_task`

ss-es:


crates/testing/src/consistency_task.rs
```diff
@@ -186,15 +193,26 @@ fn sanitize_view_map<TYPES: NodeType>(
...
+                        format!("{acc}\n\nNode {node} sent us leaf:\n\n{leaf:?}")
+                    })
             )
         );
 
         if let Some(leaf) = node_leaves.first() {
             result.insert(*view, leaf.1.clone());
         }
+
+        for (parent, child) in result.values().zip(result.values().skip(1)) {
```
bfish713, line 210: nit: isn't this just `leaf_pairs` defined above

crates/testing/src/consistency_task.rs
```diff
@@ -263,28 +378,51 @@ impl<TYPES: NodeType<BlockHeader = TestBlockHeader>, V: Versions> TestTaskState
 
     /// Handles an event from one of multiple receivers.
     async fn handle_event(&mut self, (message, id): (Self::Event, usize)) -> Result<()> {
+        {
+            let mut timeout_task = spawn_timeout_task(
```
bfish713, line 382: Does a timeout event here cancel `timeout_task`?

bfish713: Shouldn't we move this into the the if block just below, that way we are testing that we can get decides within the timeout (the event we care about)

crates/testing/src/consistency_task.rs
```diff
@@ -253,6 +282,92 @@ impl<TYPES: NodeType<BlockHeader = TestBlockHeader>, V: Versions> ConsistencyTas
...
+
+        let inverted_map = invert_network_map::<TYPES, V>(&sanitized_network_map).await?;
+
+        if inverted_map.len() >= self.safety_properties.num_successful_views {
+            Ok(TestProgress::Finished)
+        } else {
+            Ok(TestProgress::Incomplete)
+        }
+    }
+    pub async fn check_view_success(&self) -> Result<()> {
```
bfish713, line 321: This checking that we didn't decide a view that we think should fail? can't we just check this super simply by looking at the `leaf.view` rather than getting it back out of our map

bfish713 (commented):
I didn't thoroughly check the tests.  Main comment is about how we spawn/cancel the timeout in the consistency_task

crates/testing/src/consistency_task.rs
ss-es: yeah timeouts (any event) cancels it, I did this intentionally at first to avoid having to manually increase the timeout on tests with a string of timeouts (e.g. staggered restart)

but now I think I agree with you that it should be moved. I'll make the change

crates/testing/src/consistency_task.rs
ss-es: I don't see a `leaf_pairs` in this scope, though I noticed the check was duplicated in a git merge (fixed)

PR #4092, opened 2025-01-28
Disable CI test logs

ss-es:
Disables CI test logs; discussion in #hotshot on Zulip


bfish713:
Can we instead go to error and downgrade the error logs that get spammed.  I think it's just: https://github.com/EspressoSystems/HotShot/blob/69bb9e4fd87bf25e964aefa8c97e05bdf0cfdc77/crates/hotshot/src/tasks/mod.rs#L164

PR #4097, opened 2025-01-30
#4082 add mutable accessors to QuorumProposalWrapper

pls148:
Closes EspressoSystems/espresso-sequencer#2595 

### This PR: 

* Adds mutation methods on QuorumProposalWrapper
* Addresses a missed comment from EspressoSystems/HotShot#4094

### DIFF
--- a/crates/types/src/data.rs
+++ b/crates/types/src/data.rs
@@ -560,6 +560,19 @@ impl<TYPES: NodeType> ViewChangeEvidence2<TYPES> {
     }
 }
 
+impl<TYPES: NodeType> Committable for ViewChangeEvidence2<TYPES> {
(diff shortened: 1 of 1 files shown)


jparr721 (commented):
Curious, why bother adding mutable and immutable accessors when we could just make all the fields pub? Seems like this is just adding indirection.

bfish713 (approved):

bfish713 (changes requested):
Yeah I feel like all these fields do not need mutable access

ss-es (approved):
I think it's not clear we need all of these, but no objection from me

I think `QuorumProposalWrapper` is really a HotShot-internal type, and it makes sense to me to leave the fields private (or `pub(crate)` at most)

PR #4106, opened 2025-02-04
Adapt VID request / response to epochs

### LINKED ISSUE #4107: VID Request Response
Adapt VID Request Response so that it properly works with the epoch transition mechanism.

lukaszrzasik:
Closes #4107

### This PR: 

* Adapts VID Request Response to the epoch transtion:
* * The VID request is sent only when the node belongs to the current epoch or the node belongs to the next epoch and the proposal is for the last block.
* * The responding node checks whether the requester belongs to the current or next epoch and calculates the VID share accordingly if needed.
* The Request Task checks whether we have the vid share for the required view and public key. This eliminates subtle errors. So far we only checked the view which sometimes caused the request to be prematurely cancelled.
* Changes the `direct_message` logic in PushCDN slightly. PushCDN network uses a loop back mechanism to send a message to itself. This mechanism did not work when the network had been disabled by tests. This was causing an issue where the proposer never received its VID share. It didn't receive it via the network and it never requested it because it already exists in the VID map.
* Adds an epoch-enabled PushCDN crash test.

### Key places to review: 

VID request and response files.
DA file.

### DIFF
--- a/crates/task-impls/src/transactions.rs
+++ b/crates/task-impls/src/transactions.rs
@@ -185,7 +185,6 @@ impl<TYPES: NodeType, I: NodeImplementation<TYPES>, V: Versions> TransactionTask
             fee,
         }) = block
         {
-            tracing::error!("lrzasik: sending BlockRecv, view {:?}, epoch {:?}", block_view, block_epoch);
             broadcast_event(
                 Arc::new(HotShotEvent::BlockRecv(PackedBundle::new(
                     block_payload.encode(),


crates/testing/tests/tests_6/test_epochs.rs
```diff
@@ -557,3 +556,49 @@ cross_tests!(
       metadata
     },
 );
+
+// A run where the CDN crashes part-way through, epochs enabled.
+cross_tests!(
```
bfish713, line 561: This fails without your changes right?

crates/task-impls/src/response.rs
```diff
@@ -84,14 +84,22 @@ impl<TYPES: NodeType, V: Versions> NetworkResponseState<TYPES, V> {
...
-                                || !valid_signature::<TYPES>(request, sender)
-                            {
+                            if !valid_signature::<TYPES>(request, sender) {
                                 continue;
                             }
-                            if let Some(proposal) =
-                                self.get_or_calc_vid_share(request.view, sender).await
+                            if let Some(proposal) = self
+                                .get_or_calc_vid_share(request.view, target_epoch, sender)
+                                .await
```
bfish713, line 102: Hmm just thinking through this, do we need more changes to disperse the shares to each epoch or is that already done.  Basically for the transition block there is 2 sets of vid shares for both the outgoing and upcoming epoch.

Maybe that's already completed?

bfish713 (approved):
Logic looks good, maybe there is a follow up

